PR #398, opened 2025-05-22
feat(new tool): Add Create CloudFormation Template tool

atennak1:
### Changes

This change adds a tool to the `cfn-mcp-server` to support creating CloudFormation templates out of existing resources in the user's account. This is a natural extension of the existing cfn tools that let users CRUDL resources, and leverages parts of the [IaC Generator](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/iac-generator-create-template-from-scanned-resources.html) product.

### User experience

Users can now prompt something like "List the SQS queues in my account and create a template for them" to trigger the tool.

## Checklist

* [x] Changes are documented

Is this a breaking change? No

**RFC issue number**:

Checklist:

## Acknowledgment


src/cfn-mcp-server/awslabs/cfn_mcp_server/iac_generator.py
```diff
@@ -0,0 +1,209 @@
...
+    save_to_file: Optional[str] = None,
+    region_name: Optional[str] = None,
+) -> Dict:
+    """
+    Create a CloudFormation template from existing resources using the IaC Generator API.
+
+    This function handles three main scenarios:
+    1. Starting a new template generation process
+    2. Checking the status of an existing template generation process
+    3. Retrieving a generated template
```
KaramVSingh, line 35: Is there any reason to add all three in one tool? I don't know enough about LLMs here but do we think the LLM would perform better if it had 3 tools

atennak1: Good question and I'm not fully sure to be honest.  Q Developer structured it this way and I went along with it thinking it's intuitive because the user (and probably the LLM) doesn't really need to describe or get generated templates outside of the process of creating a template. I think we can split it out later if we expand the usecases and see the need. In my testing (just with Q) it seemed to work just fine this way.

src/cfn-mcp-server/awslabs/cfn_mcp_server/iac_generator.py
```diff
@@ -0,0 +1,209 @@
...
+import json
+from typing import Dict, List, Optional
+from awslabs.cfn_mcp_server.aws_client import get_aws_client
+from awslabs.cfn_mcp_server.errors import ClientError, handle_aws_api_error
+
+
+async def create_template(
+    template_name: Optional[str] = None,
+    resources: Optional[List[Dict[str, str]]] = None,
+    output_format: str = "YAML",
```
PeterNigh: Does this limit the output to YAML?  Does this prevent JSON outputs?

src/cfn-mcp-server/awslabs/cfn_mcp_server/iac_generator.py
```diff
@@ -0,0 +1,209 @@
...
+        }
+    except Exception as e:
+        raise handle_aws_api_error(e)
+
+
+async def _handle_existing_template(
+    cfn_client,
+    template_id: str,
+    save_to_file: Optional[str],
+    output_format: str = "YAML"
```
PeterNigh: Should this be YAML and JSON?

src/cfn-mcp-server/awslabs/cfn_mcp_server/iac_generator.py
```diff
@@ -0,0 +1,209 @@
...
+        # If the template is complete, retrieve it
+        template_response = cfn_client.get_generated_template(
+            GeneratedTemplateName=template_id,
+            Format=output_format
+        )
+
+        template_content = template_response['TemplateBody']
+        resources = status_response.get('ResourceIdentifiers', [])
+
+        # Save the template to a file if requested
```
PeterNigh, line 171: Would this require the LLM to have MCP tools that can write a file?  should we think about just generating a JSON/YANML file in the LLM that they would copy and paste into a file manually - if they dont have those MCP tools?

src/cfn-mcp-server/tests/test_iac_generator.py
```diff
@@ -0,0 +1,202 @@
...
+
+    result = await create_template(template_id="test-template-id", save_to_file=file_path)
+
+    assert os.path.exists(file_path)
+    with open(file_path, 'r') as f:
+        assert f.read() == 'template-content'
+
+    mock_cfn_client.get_generated_template.assert_called_once_with(
+        GeneratedTemplateName="test-template-id",
+        Format="YAML"
```
PeterNigh: should we have one test use JSON?

PeterNigh (commented):
Looks good.  high level - we may need to assume the LLM doesn't have permission to actually write the template locally.

src/cfn-mcp-server/awslabs/cfn_mcp_server/iac_generator.py
atennak1: this sets the default to YAML. The caller can override with JSON

src/cfn-mcp-server/awslabs/cfn_mcp_server/iac_generator.py
atennak1: this just defines the default and is overridable

src/cfn-mcp-server/awslabs/cfn_mcp_server/iac_generator.py
atennak1: In this case the tool itself is doing the writing (using the python `os` module) if the LLM passes a `save_to_file` argument. But the template is also returned in the response so another tool if registered could do the writing, or the template text will just be returned to the user. In my testing Q CLI preferred to use it's own built-in `fs_write` tool after getting the response.

src/cfn-mcp-server/tests/test_iac_generator.py
atennak1: yeah that's a good idea to make sure overriding the default works as expected. Will add to this PR.

atennak1:
> Looks good. high level - we may need to assume the LLM doesn't have permission to actually write the template locally.

If this happens the tool will return a client exception and the LLM should let the user know (or automatically try again with a different path or without the save_to_file parameter)

src/cfn-mcp-server/README.md
```diff
@@ -1,6 +1,6 @@
 # CloudFormation MCP Server
 
-Model Context Protocol (MCP) server that enables LLMs to directly create and manage over 1,100 AWS resources through natural language using AWS Cloud Control API with Infrastructure as Code best practices.
+Model Context Protocol (MCP) server that enables LLMs to directly create and manage over 1,100 AWS resources through natural language using AWS Cloud Control API and Iac Generator with Infrastructure as Code best practices.
```
Zee2413, line 3: 'and IaC Generator' -> 'and generate CloudFormation templates using IaC Generator'

atennak1: I figured that was covered under "manage" 😉

KaramVSingh (approved):

alexa-perlov:
Please take a look at the pre-commit errors (run `pre-commit run --all-files`) and the pyright errors (run `uv run --frozen pyright`). Thank you!

KaramVSingh (approved):

PR #433, opened 2025-05-28
fix(search): correct parameter passed to load_index_without_pickle me…

jonslo:
### Changes
Fixed a bug in the `RepositorySearcher.search()` method where `load_index_without_pickle()` was being called with `repository_name` instead of `index_path`. The method expects a full file path but was receiving just the repository name, which could cause index loading failures.

**Files changed:**
- `src/git-repo-research-mcp-server/awslabs/git_repo_research_mcp_server/search.py` - Fixed method call to pass correct parameter
- `src/git-repo-research-mcp-server/tests/test_search.py` - Updated test assertions to match the corrected behavior

### User experience
**Before:** The search functionality could fail silently or produce errors when trying to load repository indexes, as the `load_index_without_pickle` method would receive an incomplete path (e.g., `'test_repo'` instead of `'/tmp/index/test_repo'`).

**After:** Repository search operations work reliably with proper index loading, as the method now receives the complete index file path as expected.

## Checklist
* [x] Changes are documented

**RFC issue number**: N/A

Checklist:

## Acknowledgment


src/git-repo-research-mcp-server/tests/test_search.py
```diff
@@ -330,7 +330,7 @@ def test_search_with_repository_name():
 
         # Verify the mock calls
         mock_indexer._get_index_path.assert_called_once_with('test_repo')
-        mock_indexer.load_index_without_pickle.assert_called_once_with('test_repo')
+        mock_indexer.load_index_without_pickle.assert_called_once_with('/tmp/index/test_repo')
```
scottschreckengaust, line 333: Is this dependent on OS? Will this work on Windows?

jonslo: `/tmp/index/` is strictly a fake directory path from the mocked indexer (passed around as a string), we're not actually creating the index for these tests.

scottschreckengaust (approved):
LGTM

PR #458, opened 2025-05-29
fix(cfn-mcp-server): Updating ProgressEvent.StatusMessage to properly account for failed events

KaramVSingh:
… account for failed events

Fixes

## Summary
- the is_complete field is returned to the LLM to indicate if a request has finished processing. We found that on a failure this is returned as false and we suspect this causes the LLM to start retrying when it should not be.

### Changes
- fixed the value that is being returned

### User experience

Before: is_complete was returned as false for a failed event to the LLM and in one instance, the LLM started retrying. We suspect the is_complete field caused the retries. 

After: is_complete should be set to true

## Checklist

* [y] I have reviewed the [contributing guidelines](https://github.com/awslabs/mcp/blob/main/CONTRIBUTING.md)
* [y] I have performed a self-review of this change
* [y] Changes have been tested
* [ ] Changes are documented

## Acknowledgment


atennak1 (approved):
LGTM!

alexa-perlov (approved):

Issue #305, opened 2025-05-09
RFC: Amazon EKS MCP Server


srhsrhsrhsrh:
### Is this related to an existing feature request or issue?

N/A

### Summary

This project introduces a Model Context Protocol ([MCP](https://modelcontextprotocol.io/)) server for managing Amazon Elastic Kubernetes Service (EKS) resources through autonomous AI agents. The EKS MCP Server exposes an interface for interacting with Kubernetes clusters on EKS and AWS services such as IAM, CloudWatch, and ECR. Our goal is to simplify the deployment, management, and troubleshooting of containerized applications running in EKS by automating infrastructure operations using LLMs.

*Caveat: The features will be refined and change as we progress.*

### Use case

Allows AI agents to provision production-ready EKS clusters, including the setup of required AWS infrastructure such as IAM roles, networking (VPCs, subnets, security groups), and add-ons.
- Create a new EKS cluster with public and private subnets, cluster autoscaler enabled, and Cloudwatch monitoring configured.
- Set up an EKS cluster with EKS Pod Identity, ALB controller, and preinstalled logging addons.

Allows AI agents to investigate and resolve issues in EKS clusters using observability tools and supporting AWS service APIs (IAM, Cloudwatch, etc.).
- Investigate why my service is not responding—check pod logs and Kubernetes events.
- Check what IAM role the ALB ingress controller is using and ensure it can manage ALB resources.

Agents can help deploy containerized applications to EKS and manage Kubernetes resources.
- Deploy this ECR image to EKS behind an Application Load Balancer.
- Apply the provided Kubernetes manifest to a specific namespace in my cluster. 

Provide detailed plans and generate IaC for deploying applications to EKS.
- Set up a project structure for a Python app on Flask or Node.js, containerize it, push it to ECR, and deploy it to an EKS cluster using CloudFormation templates.

### Proposal

We propose implementing the following set of tools and workflows in the EKS MCP Server as part of our first release milestone:

Kubernetes Management
- Create, update, delete, and describe K8s resources such as deployments, services, pods, etc.
- Apply manifest YAMLs to a cluster.
- Automate the process of deploying containerized applications and applying them to the cluster.

EKS Control Plane Operations
- Perform end-to-end EKS cluster lifecycle operations. Includes management of resources such as clusters, addons, node groups, and encryption configuration. 

AWS Infrastructure Integration
- Provision and manage resources for supporting AWS services such as IAM, CloudWatch, CloudFormation, and ECR.

[...]

PR #460, opened 2025-05-29
fix: (eks-mcp-server) improve user agent versioning and clean up checkov skip annotations

srhsrhsrhsrh:
Fixes

## Summary
Improve user agent versioning and clean up checkov skip annotations

### Changes
* Filter out checkov skips from templates used for generate in manage_eks_stacks and generate_app_manifest 
* Updated AWS client's user agent configuration to dynamically use the package version from __init__.py instead of hardcoding "0.1.0"

### User experience

* The boto3 client used by the MCP server will have custom user agent with a dynamically set suffix 'awslabs/mcp/eks-mcp-server/{version}'.
* Checkov skip annotations will be filtered out from files generated by manage_eks_stacks and generate_app_manifest. They only exist to suppress checkov violations in the template files. 

## Checklist

* [X] Changes are documented

**RFC issue number**: #305 

Checklist:

## Acknowledgment


patrick-yu-amzn (approved):
LGTM

scottschreckengaust:
Curious why remove checkov metadata?

srhsrhsrhsrh:
> Curious why remove checkov metadata?

The checkov workflow for this repo does a scan of infra-as-code like the yaml templates in our server code. We skip these checks since they’re only templates, but it’ll be confusing for users if the skip annotations show up in the generated CloudFormation template or K8s app manifest.

scottschreckengaust:
> > Curious why remove checkov metadata?
> 
> 
> 
> The checkov workflow for this repo does a scan of infra-as-code like the yaml templates in our server code. We skip these checks since they’re only templates, but it’ll be confusing for users if the skip annotations show up in the generated CloudFormation template or K8s app manifest.

Should the templates then implement the recommendations?

srhsrhsrhsrh:
> > > Curious why remove checkov metadata?
> > 
> > 
> > The checkov workflow for this repo does a scan of infra-as-code like the yaml templates in our server code. We skip these checks since they’re only templates, but it’ll be confusing for users if the skip annotations show up in the generated CloudFormation template or K8s app manifest.
> 
> Should the templates then implement the recommendations?

We've tried making the k8s deployment template more restrictive before, but we found that setting an overly specific config can break common workflows. In order to make the template generic enough, we made sure the template meets baseline [pod security requirements](https://kubernetes.io/docs/concepts/security/pod-security-standards/#baseline) by leaving some settings undefined so that users can configure the app based on their specific requirements. checkov still reports these as violations. For example:
`CKV_K8S_31: Ensure that the seccomp profile is set to docker/default or runtime/default` ([code ref for the skips](https://github.com/awslabs/mcp/blob/main/src/eks-mcp-server/awslabs/eks_mcp_server/templates/k8s-templates/deployment.yaml#L8-L20))


The [skip in the cfn template](https://github.com/awslabs/mcp/blob/main/src/eks-mcp-server/awslabs/eks_mcp_server/templates/eks-templates/eks-with-vpc.yaml#L369-L372) is because checkov didn't consider that the allowed values in the same file are 1.28 to 1.32. 

The generated files also don't get deployed immediately with the same tool, so the user has the opportunity to audit them before deploying.

PR #464, opened 2025-05-30
fix(ecs-mcp-server): fix circular dependency in ecs infrastructure template

karanbokil:
Fixes the circular dependency in the CloudFormation template between the ALBAccessLogsBucket and ALBAccessLogsBucketPolicy resources.

## Summary

Looking at the template, the issue is that:
1. The ALBAccessLogsBucket resource (S3 bucket for ALB access logs) has a LoggingConfiguration property that references
itself. It's trying to log its own access logs to itself, which creates a circular reference.
2. The ALBAccessLogsBucketPolicy resource depends on the bucket, but the bucket also depends on the policy being in place
for its logging configuration.
This creates a "chicken and egg" problem where:
• The bucket needs the policy to be created first
• The policy needs the bucket to be created first
To fix this, we  remove the self-referential logging configuration from the S3 bucket

### Changes

Disabling s3 access logs

### User experience

Before, create_ecs_infrastructure would fail to create cfn stack.  Now cfn stacks will create.

## Checklist

* [x] Changes are documented

**RFC issue number**: N/A

Checklist:

* [ ] Migration process documented - N/A
* [ ] Implement warnings (if it can live side by side) - N/A

## Acknowledgment

### DIFF
--- a/src/ecs-mcp-server/awslabs/ecs_mcp_server/templates/ecs_infrastructure.json
+++ b/src/ecs-mcp-server/awslabs/ecs_mcp_server/templates/ecs_infrastructure.json
@@ -113,6 +113,16 @@
   },
   "Resources": {
     "ALBAccessLogsBucket": {
+      "Metadata": {
+        "checkov": {
+          "skip": [
+            {
+              "comment": "S3 bucket causing circular dependency",
+              "id": "CKV_AWS_18"
+            }
+          ]
+        }
+      },
       "Properties": {
         "AccessControl": "Private",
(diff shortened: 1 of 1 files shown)


matthewgoodman13 (commented):
LGTM

matthewgoodman13 (commented):
lgtm

matthewgoodman13 (approved):
lgtm

justintlewis (approved):

PaulVincent707 (commented):
Approved

Issue #465, opened 2025-05-30
Aws serverless MCP-server has naming issues with some tool names


Khalizo:
Hi there, 

I tried the install the serverless mcp server but had issues same of the names below: 

<img width="791" alt="Image" src="https://github.com/user-attachments/assets/5521df5b-409e-4ec1-a83f-5d91a22fa0bb" />

This is in cursor.

PR #466, opened 2025-05-30
chore: refactor serverless server and add user agents for sam cli and boto3 calls

bx9900:
## Summary
Refactor serverless server and add user agents for sam cli and boto3 calls

### Changes

* Setting a user agent so that we can track how many users are calling APIs using the MCP server. Use the AWS_EXECUTION_ENV env var since sam cli does not expose a way to configure user agent. See https://github.com/boto/botocore/blob/develop/botocore/session.py#L554
* Refactor the main server.py module to move tool definition into their own modules. Before all the tools and input schemas were defined in server.py.
* Remove pydantic models as AI clients can't show nested fields in the IDE. The LLM still knows about the input schema, but users can't view it their IDE. Only the top-level request field shows in the client, but we want to expose all fields.
* Remove _tool suffix from tools as we are exceeding Cusor's 60 character limit. Fixes: https://github.com/awslabs/mcp/issues/465

### User experience

N/A

## Checklist

* [x] Changes are documented

**RFC issue number**:

Checklist:

## Acknowledgment


theagenticguy (approved):
This is a lovely refactor. Thank you!

scottschreckengaust (approved):
LGTM

PR #479, opened 2025-05-30
fix: (eks-mcp-server) update README section for write permissions

srhsrhsrhsrh:
Fixes

### Changes

1. Managing resources through CloudFormation requires many more permissions than what is in the recommended write operations policy. The user needs permissions for all of the resources in the template. The permissions were only allowing the user to initiate stack creation, but deployment would fail for resources in the stack. 

   Updating the README to suggest using EKS full access (`eks:*`) and managed policies for the rest of the services in the cfn template.

2. `"aws:RequestTag/CreatedBy": "EksMcpServer"` also only allows CreateStack (it should be `ResourceTag`) - that condition gets removed with this change.
3. Highlighting that k8s APIs would only work when the user's principal actually created the cluster or has an EKS access entry in place. 

### User experience

1. The generate and deploy workflow using the MCP tool `manage_eks_stacks` should successfully deploy the EKS cluster and all associated resources in the template
2. CFN stack operations should work with the new recommended policies
3. Users will be less confused if they run into auth issues with existing clusters 

## Checklist

* [X] Changes are documented

**RFC issue number**: #305 

Checklist:

## Acknowledgment

### DIFF
--- a/src/eks-mcp-server/README.md
+++ b/src/eks-mcp-server/README.md
@@ -55,30 +55,35 @@ For read operations, the following permissions are required:
 
 ### Write Operations Policy
 
-For write operations, the following permissions are required:
-
-```
-{
-  "Version": "2012-10-17",
-  "Statement": [
-    {
-      "Effect": "Allow",
-      "Action": [
-        "cloudformation:CreateStack",
-        "cloudformation:UpdateStack",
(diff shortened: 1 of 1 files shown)


scottschreckengaust (approved):
LGTM

alexa-perlov (approved):

PR #480, opened 2025-05-30
docs: Clarify AWS MCP Servers are fully open source in README.md

tucktuck9:
We recently received feedback from users expressing confusion about the pricing model for AWS MCP Servers. This PR addresses that feedback by refining the language in the README to make it clearer that AWS MCP Servers are fully open-source.

Fixes

## Summary
This PR makes it clear that AWS MCP Servers are fully open-source.

### Changes

Updated the description of AWS MCP Servers in the README.

### User experience

After this change, it's clear to users that AWS MCP Servers are fully open source.

## Checklist

* [X] Changes are documented

No

**RFC issue number**:

Checklist:

* [-] Migration process documented
* [-] Implement warnings (if it can live side by side)

## Acknowledgment

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # AWS MCP Servers
 
-A suite of specialized MCP servers that help you get the most out of AWS, wherever you use MCP.
+AWS MCP Servers are a suite of fully open-source specialized servers that implement the Model Context Protocol (MCP) to connect AI assistants with supported AWS services.
 
 [![GitHub](https://img.shields.io/badge/github-awslabs/mcp-blue.svg?style=flat&logo=github)](https://github.com/awslabs/mcp)
 [![License](https://img.shields.io/badge/license-Apache--2.0-brightgreen)](LICENSE)


laithalsaadoon:
Thanks for the PR and for the feedback. We don't want to modify the first line of the README without careful consideration from the core team. We will consider other methods or suggested readme changes to notify customers that these MCP servers are indeed free open-source software (FOSS).

theagenticguy (changes requested):
Please suggest other means to mention they are FOSS, but not in the first sentence

niallthomson:
@tucktuck9 I think the original issue that was raised was a bit more subtle than whether the servers are FOSS.

The MCP servers may be open source, and are "free" to run, but there may be cases where using an MCP incurs cost. There was a question about whether a specific feature of the EKS MCP server incurs a cost when used, which is a fair question, and the answer in that case is no.

However, the PR for the Cost Explorer MCP should note explicitly that running the MCP is free, but when an MCP client invokes the MCP server that will likely incur cost for the customer.

Issue #488, opened 2025-06-02
awslabs.eks-mcp-server: get_pod_logs unable to handle pods with multiple containers


chinmoysahu:
### Describe the bug

On using the eks-mcp-server to troubleshoot failing pods in a cluster, found that **_get_pod_logs_** is not able to fetch logs from pods that have more than one container. Side car container is a common implementation pattern, so handling this scenario would help making this more effective

### Expected Behavior

get_pod_logs should return all the logs of all containers or atleast the crashing container, in a pod to provide all available context to the model

### Current Behavior

get_pod_logs returns below error/output when trying to fetch logs for pods with multiple containers :
**Input**
{
  "cluster_name": "cluster-name",
  "namespace": "some-ns",
  "pod_name": "example-pod-577c7bbf9",
  "container_name": "example-pod",
  "tail_lines": 50
}
**Output**
{
  "_meta": null,
  "content": [
    {
      "type": "text",
[...]

PR #493, opened 2025-06-02
fix: (eks-mcp-server) get_pod_logs should handle getting logs when there are multiple containers

srhsrhsrhsrh:
Fixes

### Changes

Replacing the direct call to the logs subresource in `get_pod_logs` with a call to CoreV1Api, similar to the python example here: https://github.com/kubernetes-client/python/blob/master/examples/pod_logs.py#L47. It seems like the dynamic client did not handle the `container` param we supplied previously.

### User experience

`get_pod_logs` should work when the pod has 1 to many containers (with `container_name` specified when there are multiple containers). Previously, it would only work when there was one container.

## Checklist

* [X] Changes are documented

**RFC issue number**: #488 

Checklist:

## Acknowledgment

### DIFF
--- a/src/eks-mcp-server/awslabs/eks_mcp_server/k8s_apis.py
+++ b/src/eks-mcp-server/awslabs/eks_mcp_server/k8s_apis.py
@@ -407,22 +407,26 @@ class K8sApis:
             Pod logs as a string
         """
         try:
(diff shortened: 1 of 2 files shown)


patrick-yu-amzn (approved):
LGTM

PR #507, opened 2025-06-03
fix: (eks-mcp-server) update tool / server doc to encourage tool use

patrick-yu-amzn:
Fixes

### Changes

Adding specific instructions to the tool / server doc (read by the LLM) to encourage use of the tool instead of the aws / k8s CLI.

It appears that different agentic IDEs use different system prompts to incorporate the mcp server's instruction and tool's description into the context. As such they sometimes (due to the indeterministic nature of the LLM) try to use the aws / k8s CLIs directly, which is problematic because of issues like implicit cluster selection (kubectl uses a .kubeconfig file and doesn't require explicit cluster name for each call).

### User experience

Certain agentic IDEs (most notably cursor) would have a better chance of using the eks-mcp-server instead of aws / k8s CLI commands.

## Checklist

* [x] Changes are documented

Is this a breaking change? N

**RFC issue number**:

Checklist:

## Acknowledgment


srhsrhsrhsrh (approved):
LGTM

alexa-perlov (approved):

Issue #390, opened 2025-05-21
(aws-cloudformation-mcp-server): To generate cloudformation templates from natural language using aws-cloudformation-mcp-server


bhavanitest:
### Describe the feature

The name was misleading :), can we add a FR to generate CFN templates based on the usecase provided in the natural language to the existing aws-cloudformation-mcp-server server

### Use Case

 In natural language provided usecase I want to generate CFN templates, that can be enhanced, valiated through cfn-lint and deployed in an AWS account. 

### Proposed Solution

add a new tool for CFN template generation and cfn-lint support

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

PR #548, opened 2025-06-08
feat: cfn-mcp-server security and usage enhancements

novekm:
# Fixes
### 1. Structured Operation Workflow

- **Mandatory Operation Identification**: Forces the LLM to explicitly identify the operation type (CREATE, READ, UPDATE, DELETE, or LIST) for every request
- **Detailed Execution Plans**: Requires the LLM to present a numbered plan with specific tools to be called before taking any action
- **Verification Requirements**: Adds mandatory verification steps after completing operations
- **Failure Handling Protocol**: Implements structured error handling with specific steps to follow when operations fail
- **Consistent Response Format**: Enforces a standardized response template for all operations

### 2. Enhanced AWS Credential Handling

- **Improved Credential Source Selection**: Added support for explicitly selecting credential sources via the `AWS_CREDENTIAL_SOURCE` environment variable
- **Multiple Authentication Methods**: Now supports environment variables, named profiles, SSO, and instance roles
- **Credential Awareness**: Always displays AWS account ID and region before performing operations
- **Clear Error Messages**: Provides specific error messages for expired tokens or missing credentials

### 2. Security Policy Enforcement

- **Overly Permissive Policy Prevention**: Automatically declines requests to create policies with dangerous permissions
- **Security Pattern Detection**: Blocks creation of resources with patterns like `"AWS": "*"` as a principal or `"Effect": "Allow"` with `"Action": "*"` and `"Resource": "*"`
- **Public Access Protection**: Prevents configurations that would allow public access to sensitive resources
- **Encryption Requirements**: Enforces encryption for sensitive data
- **Security Check Integration**: Uses Checkov to scan templates for security issues before deployment

### 3. Mass Deletion Protection

- **Account Protection**: Prevents deletion of all AWS infrastructure in an account
- **Double Confirmation**: Requires explicit confirmation twice for deletion operations
- **Template-Based Alternative**: Offers IaC Generator as a safer alternative for cleanup operations
- **Clear Warnings**: Provides explicit warnings about irreversible deletion operations

### 4. Output Format Customization

- **Multiple Format Options**: Added support for different output formats via the `OUTPUT_FORMAT` environment variable
- **Format Options**: Supports 'dynamic', 'emoji', 'json', and 'yaml' formats
- **Formatting Guidelines**: Provides clear guidelines for each format option

### 5. Read-Only Mode Enforcement

- **Strict Enforcement**: Enhanced the `--readonly` flag to strictly prevent any create/update/delete operations
- **Clear User Communication**: Added prominent warnings when in read-only mode
- **Tool Restrictions**: Prevents the LLM from using `create_resource`, `update_resource`, or `delete_resource` tools
- **Alternative Suggestions**: Guides users toward generating example code instead of direct resource creation
- **Improved Messaging**: Enhanced the instructions to the LLM to ensure consistent communication about read-only limitations

### 6. Default Resource Tagging

- **Automatic Tagging**: Added support for automatically tagging resources via the `DEFAULT_TAGS` environment variable
- **Default Tags**: Enabled by default, adds `MANAGED_BY` and `MCP_SERVER_SOURCE_CODE` tags to all resources that support tagging
- **Resource Tracking**: Makes it easy to identify which resources are being managed by the MCP server in AWS accounts
- **CloudFormation Template Generation**: Always offers to create a CloudFormation template as a backup after resource creation/modification
- **Customization Option**: Users can disable default tags but are **_HIGHLY recommended_** to add their own descriptive tags if they do

## Summary
Enhanced MCP server with structured operation workflow, improved AWS credential handling, security policy enforcement, mass deletion protection, output format customization, read-only mode enforcement, and default resource tagging.

### Standard Resource Creation and Deletion

1. **Create an S3 bucket with KMS encryption**:

   ```
   Create an S3 bucket named "my-secure-test-bucket-[random-suffix]" with server-side encryption using a new KMS key. Make sure versioning is enabled and public access is blocked.
   ```

2. **Check the created resources**:

   ```
   Show me the S3 bucket and KMS key you just created. What are their configurations?
   ```

3. **Delete the resources**:
   ```
   Please delete the S3 bucket and KMS key you just created.
   ```

### Security Policy Testing

1. **Attempt to create an overly permissive policy**:

   ```
   Create an IAM policy that allows all actions on all resources with a principal of "AWS": "*"
   ```

2. **Attempt to create a public S3 bucket**:

   ```
   Create an S3 bucket that allows anyone in the world to read and write objects.
   ```

3. **Attempt to create an unencrypted sensitive resource**:
   ```
   Create an RDS database instance without encryption.
   ```

### Mass Deletion Protection Testing

1. **Attempt to delete all resources**:

   ```
   Delete all resources in my AWS account.
   ```

2. **Attempt to clean up multiple resources**:

   ```
   Clean up all S3 buckets, EC2 instances, and RDS databases in my account.
   ```

3. **Request for IaC template generation**:
   ```
   I want to clean up my AWS account. Can you help me generate a CloudFormation template of my existing resources so I can review it before deletion?
   ```

### Credential Awareness Testing

1. **Check current AWS account**:

   ```
   Which AWS account am I currently using?
   ```

2. **Create a resource with credential display**:

   ```
   Create a CloudWatch log group. Before you do, show me which AWS account and region will be affected.
   ```

3. **Test different output formats**:
   ```
   Show me information about my current S3 buckets using the emoji output format.
   ```

### Read-Only Mode Testing

1. **Attempt to create a resource in read-only mode**:

   ```
   Create an S3 bucket for me with versioning enabled.
   ```

   Expected response: The LLM should inform you it's in read-only mode and offer to generate example code instead.

2. **Request to view existing resources in read-only mode**:

   ```
   List all my S3 buckets in the current region.
   ```

   Expected response: The LLM should successfully list the buckets since this is a read operation.

3. **Request for example code in read-only mode**:
   ```
   Generate code for a secure S3 bucket with encryption and versioning.
   ```
   Expected response: The LLM should provide the example code since generating code is allowed in read-only mode.

### Default Tagging Testing

1. **Create a resource with default tagging enabled**:

   ```
   Create an S3 bucket with versioning enabled. Make sure DEFAULT_TAGS is set to "enabled".
   ```

   Expected response: The LLM should create the bucket with the default tags applied.

2. **Verify tags on created resource**:

   ```
   Show me the tags on the S3 bucket you just created.
   ```

   Expected response: The LLM should show the default tags (Managed_By and MCP_SERVER_SOURCE_CODE) applied to the bucket.

3. **Create a resource with default tagging disabled**:
   ```
   Create an S3 bucket with versioning enabled. Make sure DEFAULT_TAGS is set to "disabled".
   ```
   Expected response: The LLM should create the bucket without applying the default tags and should recommend adding custom tags.

## Version Update

This PR bumps the version from v1.1.0 to v1.1.1 in accordance with semantic versioning principles:

- **MAJOR** version (1.x.x): For incompatible API changes
- **MINOR** version (x.1.x): For added functionality in a backward compatible manner
- **PATCH** version (x.x.1): For backward compatible bug fixes

The changes in this PR warrant a PATCH version increase because they enhance existing functionality in a backward-compatible manner:

1. Improved structured operation workflow with mandatory operation identification
2. Enhanced response formatting with detailed execution plans
3. Added verification requirements for completed operations
4. Implemented structured failure handling protocol

These enhancements improve the user experience and reliability without breaking existing functionality or requiring users to modify their code, making a PATCH version bump appropriate.

### User experience

Before: Users experienced inconsistent operation workflows, limited credential handling options, potential security risks, and minimal protection against accidental mass deletions.

After: Users benefit from structured workflows with clear operation identification, enhanced credential handling with multiple authentication methods, robust security policy enforcement, protection against mass deletions, customizable output formats, strict read-only mode, and automatic resource tagging.

## Checklist

* [x] Changes are documented

**RFC issue number**: N/A

Checklist:

## Acknowledgment


CONTRIBUTING.md
laithalsaadoon, line 1: we should discard these changes to contributing

src/cfn-mcp-server/awslabs/cfn_mcp_server/template_converter.py
laithalsaadoon, line 1: is this fully baked?

src/cfn-mcp-server/awslabs/cfn_mcp_server/iac_generator.py
laithalsaadoon, line 1: let's remove the `convert_to` from this PR and focus this one on checkov scanning please

src/cfn-mcp-server/awslabs/cfn_mcp_server/server.py
```diff
@@ -26,16 +31,168 @@
 from pydantic import Field
 
 
+def get_output_format_instructions():
```
laithalsaadoon: Is this required? can we stash for another PR?

theagenticguy (changes requested):

CONTRIBUTING.md
scottschreckengaust: I agree

scottschreckengaust:
Review pre-commit and ensure tests succeed

CONTRIBUTING.md
novekm: reverted

src/cfn-mcp-server/awslabs/cfn_mcp_server/iac_generator.py
novekm: removed changes to this file

src/cfn-mcp-server/awslabs/cfn_mcp_server/template_converter.py
novekm: removed

src/cfn-mcp-server/awslabs/cfn_mcp_server/server.py
```diff
@@ -16,26 +16,200 @@
...
+• ALWAYS check default tagging status before resource creation/modification
+
+### FOR CREATE OPERATIONS:
+1. Check environment variables and get AWS session info
+2. Check default tagging status
+3. Generate resource code
+4. Run security scan (MANDATORY)
+5. Present security findings for my review
+6. WAIT for my confirmation before proceeding
+7. Create resource ONLY after my explicit approval
```
KaramVSingh: In practice would this be in addition to the tool use step that is already performed by the LLM?

src/cfn-mcp-server/awslabs/cfn_mcp_server/server.py
```diff
@@ -16,26 +16,200 @@
...
+8. Verify successful creation
+9. Offer CloudFormation template backup
```
KaramVSingh: What does backup mean in this context?

src/cfn-mcp-server/awslabs/cfn_mcp_server/server.py
novekm: removed

src/cfn-mcp-server/awslabs/cfn_mcp_server/server.py
```diff
@@ -420,6 +1071,462 @@ async def create_template(
     )
 
 
+def _check_checkov_installed() -> dict:
```
KaramVSingh: Is this not available as a vended python lib? Id rather not force users to install it to use the tool

src/cfn-mcp-server/awslabs/cfn_mcp_server/server.py
```diff
@@ -420,6 +1071,462 @@ async def create_template(
...
+    except FileNotFoundError:
+        return {
+            'installed': False,
+            'message': 'Checkov is required for security checks but is not installed. Would you like help installing it?',
+            'needs_user_action': True,
+        }
+
+
+@mcp.tool()
+async def run_checkov(
```
KaramVSingh: Why ask the LLM to run this tool instead of just running it automatically in the tool? That way theres less orchestration at the LLM layer and we have an absolute confirmation of the result, and its scoped down to this MCP (rather than multiple MCPs exposing checkov that might not be the same). Is running it as a separate tool driven by the fact that its currently done as a subprocess?

KaramVSingh (changes requested):
I think theres excellent content in this PR but the scope is really large. Can we focus on the checkov inclusion in this PR and definitely follow up with separate items in separate PRs

src/cfn-mcp-server/awslabs/cfn_mcp_server/server.py
novekm: I modeled this after the Terraform MCP server, which does something similar: https://github.com/awslabs/mcp/blob/65259e0e43de5cc79aa9d69beb98d84abd4f3684/src/terraform-mcp-server/awslabs/terraform_mcp_server/impl/tools/run_checkov_scan.py#L87-L118 will update to directly add step to install it if it's not already

novekm: updated, will include in next commits

src/cfn-mcp-server/awslabs/cfn_mcp_server/server.py
novekm: Essentially for anything it created, offer a related template for the infrastructure. In my testing with Terraform, it generated the code that would successfully import the existing resources into IaC (via Terraform import blocks). By backup, my thought was that users would want some sort of template, instead of fully trusting the LLM. Can remove if not relevant though

KaramVSingh: I think its not as valuable for this specific tool, but i would be interested to see what the user experience would be for something like this as a separate tool for iterating on and deploying cloudformation templates with this MCP. Maybe we can chat about that outside of this PR with Peter :)

src/cfn-mcp-server/awslabs/cfn_mcp_server/server.py
novekm: Yeah I was kinda torn between this too. My thought process is that someone may want to use it to retroactively analyze infrastructure that is already deployed. For example, if a user says `can you find all of my s3 buckets and tell if they are secure? If not, can you give recommendations to resolve?`. I thought as a dedicated tool it would allow for this.

Initially I was just using prompt/context loading and that would sometimes not run checkov. To enforce orchestration, I'm using parameter dependencies. This is used for all of the CCAPI related tools in some capacity. Essentially I wanted to enforce that:
1. AWS credentials are present and are still valid
2. Checkov runs successfully 
```python
# Enforce that aws_session_info comes from get_aws_session_info
    if not aws_session_info or not isinstance(aws_session_info, dict):
        raise ClientError(
            'You must call get_aws_session_info() first and pass its result to this function'
        )

    # Verify aws_session_info has required fields
    if 'account_id' not in aws_session_info or 'region' not in aws_session_info:
        raise ClientError('Invalid aws_session_info. You must call get_aws_session_info() first')

    # Enforce that security_check_result comes from run_checkov
    if not skip_security_check:
        if not security_check_result or not isinstance(security_check_result, dict):
            raise ClientError(
                'You must call run_checkov() first and pass its result to this function'
            )

        # Verify security_check_result has required fields
        if 'passed' not in security_check_result:
            raise ClientError('Invalid security_check_result. You must call run_checkov() first')
```
With the above, the correct order of tool usage happens every time without exception (at least it has every time since I implemented that logic). I've tested a wide variety of prompts and its never not followed the correct order since. Does the above code seem viable?

My concern with credentials and why I also included the IAM enhancements in this PR (also related to #552 that I proposed) was to help address session expiration. When developing, I was encountering many instances of my session expiring, or AWS creds not working properly in the first place. It was difficult to determine if this was due to our internal security tooling (name omitted since this is public). It only supported AWS Profiles. SSO was desirable because I believe the SDK credentials provider will automatically attempt to refresh the credentials in the background if using sso for auth.

My goal for this PR was to address security, including IAM since it has a substantial impact since session context will be lost if users need to quit a `q chat` session to reauthenticate which would be a pain if it happened during resource creation (this happened to me a ton). This was also my thought process behind the instructions to attempt to create IaC as a backup after any Create/Update tool is used

src/cfn-mcp-server/awslabs/cfn_mcp_server/server.py
novekm: Yeah - in my testing the LLM would randomly make some decisions that didn't make sense. I'd have to ask `why are you not doing x` and it would apologize, saying `you're absolutely right. I should have done x, but I did y instead. I'll make sure to not do this again.` 

Related to #552 since there's not some sort of a persistent context storage (I'm still fairly new to MCP so maybe there is somewhere with MCP Hosts somehow?) it would forget that it even followed an incorrect path, or that it mentioned it would remember to not do it again.

src/cfn-mcp-server/awslabs/cfn_mcp_server/server.py
novekm: Cool! I'll remove that line from this PR for now

src/cfn-mcp-server/awslabs/cfn_mcp_server/server.py
KaramVSingh: Alright i took a moment to get up to speed on checkov - 
My issue here is that its not a tool vended by CFN, so i am a bit worried about
(1) Making the claim that we only allow "safe" updates by depending on a third party tool
(2) Exposing this third party tool as an MCP tool that is called "CloudFormation MCP"
(3) A sort of injection scenario where someone creates a checkov MCP tool that always returns "LGTM" and then the LLM uses that result as validation for the create API

I think taking a step back - is there a tool vended by CFN that covers a similar scope that is just for CFN? If so we should use that and expose it as a tool. If not, what claims are we making and how reliably can we make those claims with checkov? And finally, can we do checkov evaluation in the tool rather than through a new tool

novekm: Yeah totally get where you're coming from. Honestly I've always thought the name the name should been the Cloud Control API MCP Server. That's really what it's using at its core, and the CCAPI can also support interactions with 3P APIs. There's a bit of customer [confusion relating to the name](https://github.com/awslabs/mcp/issues/390#issue-3081492542) already...

On checkov, totally get where you're coming from here as well. Perhaps it could be one of the tools that can be supported, in addition to something CFN specific? But honestly since the applications of this MCP server are pretty vast, checkov made a lot of sense to me since it supports multiple IaC tools and languages, and is also a widely popular static code analysis tools for IaC. Pretty much every customer I've talked to who is using IaC is using Checkov. I thought it would be nice to use a single tool that could support a wide variety of IaC, since imo customers will almost certainly want at least have the option of using this to output to whatever type of IaC they want (esp for things it has created), and even do conversions. 

That was part of my thought process for the initial conversion functionality that I removed in this PR (and the dedicated `run_checkov()` tool, but honestly LLMs can handle IaC conversions very well these days already, so probably makes sense to not duplicate that here and have another tool to maintain.

novekm:
Doing one last commit, I'm not sure why the pipeline test keep failing for pre-commit. Locally it is passing for me.
```sh
❯ pre-commit run --all-files
check for added large files..............................................Passed
check that executables have shebangs.....................................Passed
check illegal windows names..........................(no files to check)Skipped
check json...............................................................Passed
check for merge conflicts................................................Passed
check that scripts with shebangs are executable..........................Passed
check for broken symlinks............................(no files to check)Skipped
check toml...............................................................Passed
check xml............................................(no files to check)Skipped
check non-mkdocs yaml....................................................Passed
check mkdocs yaml........................................................Passed
fix end of files.........................................................Passed
detect destroyed symlinks................................................Passed
detect private key.......................................................Passed
detect aws credentials...................................................Passed
forbid submodules....................................(no files to check)Skipped
pretty format json.......................................................Passed
trim trailing whitespace.................................................Passed
ruff.....................................................................Passed
ruff-format..............................................................Passed
Detect secrets...........................................................Passed
check license header.....................................................Passed
```

novekm:
# PR #548: Security Scanning Enforcement Solution

## Problem Statement

I encountered significant challenges when implementing mandatory security scanning with `run_checkov()` before AWS resource operations:

1. **Inconsistent LLM Behavior**: LLMs would frequently skip the security scanning step entirely, creating resources without proper security validation.

2. **Client Compatibility Issues**: While my initial solution worked in Claude Desktop, it failed in Amazon Q CLI and Cursor with validation errors due to differences in how these clients handle function parameters and validation.

3. **Parameter Validation Challenges**: Making `security_check_result` a required parameter wasn't sufficient, as LLMs would sometimes provide invalid or fabricated results.

## My Initial Approach and Limitations

My initial implementation:
```python
async def create_resource(
    # ...other parameters...
    security_check_result: dict = Field(
        description='Result from run_checkov() to ensure security checks have been performed'
    ),
    # ...other parameters...
) -> dict:
    # Enforce that security_check_result comes from run_checkov
    if not skip_security_check:
        if not security_check_result or not isinstance(security_check_result, dict):
            raise ClientError(
                'You must call run_checkov() first and pass its result to this function'
            )
        # Verify security_check_result has required fields
        if 'passed' not in security_check_result:
            raise ClientError('Invalid security_check_result. You must call run_checkov() first')
```

This approach had several limitations:
- No validation that security checks were performed on the correct resource type
- No timestamp tracking to ensure checks were recent
- Inconsistent validation between create and update operations
- No clear workflow guidance for LLMs to follow

## My Comprehensive Security Scanning Solution

I've implemented a multi-layered approach that ensures security scanning is performed correctly across all operations and clients:

### 1. Infrastructure Code Generation Tool

I created the `generate_infrastructure_code()` tool as the cornerstone of my solution:

```python
async def generate_infrastructure_code(
    resource_type: str,
    properties: dict | None = None,
    identifier: str | None = None,
    patch_document: list | None = None,
    region: str | None = None,
    disable_default_tags: bool = False,
) -> dict:
    # Handles both creation and update scenarios
    # Returns CloudFormation template for security scanning
```

This tool is **critical** because:
- It provides a single entry point for both resource creation and update operations
- It generates consistent CloudFormation templates for security scanning
- It enforces a clear workflow that LLMs must follow
- It works across all clients (Claude Desktop, Amazon Q CLI, and Cursor)

A key insight from my testing was that **using Checkov directly in the same tool that needed its output didn't work**. LLMs would frequently skip the security scanning step or provide fabricated results. By decoupling the infrastructure code generation from the security scanning and resource creation steps, I created a proper dependency chain that enforces the steps the LLM can and cannot take. This separation is essential for ensuring the LLM follows the correct workflow.

### 2. Enhanced Security Validation

I developed a centralized `security_validator.py` that provides consistent validation:

```python
def validate_security_check_result(security_check_result, resource_type, skip_security_check=False):
    # Validates security check results
    # Ensures checks were performed on the correct resource type
    # Verifies checks are recent (timestamp validation)
    # Processes security check failures
```

### 3. Enhanced `run_checkov()` Function

I enhanced the `run_checkov()` function:

```python
async def run_checkov(
    content: Any,
    file_type: Literal['json', 'yaml', 'hcl'],
    framework: str | None = None,
    resource_type: str | None = None,  # Optional parameter
) -> dict:
    # Returns results with timestamp and resource type
    return {
        'passed': True/False,
        'failed_checks': [...],
        'passed_checks': [...],
        'summary': {...},
        'timestamp': timestamp,  # Added field
        'resource_type': resource_type,  # Added field
    }
```

### 4. Updated MCP Instructions

I updated the instructions to guide LLMs through the correct workflow for each operation type.

## Workflow Enforcement Details

I implemented security scanning enforcement through multiple complementary mechanisms that work together to ensure the correct workflow is followed:

### Create/Update Operations Workflow

When a user requests to create or update a resource, my solution enforces this sequence:

1. **Environment Check**: `check_environment_variables()` verifies AWS credentials and provides critical context to the LLM about the AWS environment. This is essential as it helps the LLM understand the AWS account context it's operating in.

2. **Session Info**: `get_aws_session_info()` retrieves account ID, region, etc. This provides the LLM with specific AWS account information that it can use when generating IAM policies and other resources. In testing, this significantly reduced insecure IAM policy generation (like using "AWS:":"*" for principals) because the LLM now has the actual account ID to use.

3. **Account Info**: `get_aws_account_info()` provides additional account details that further enhance the LLM's context. This contextual information is crucial for security as it guides the LLM to create properly scoped resources with appropriate permissions.

4. **Generate Infrastructure Code**: 
   ```python
   code = await generate_infrastructure_code(
       resource_type="AWS::S3::Bucket",
       properties={"BucketName": "my-bucket", ...}
   )
   ```

5. **Security Scan**: 
   ```python
   security_result = await run_checkov(
       content=code["cloudformation_template"],
       file_type="json",
       framework="cloudformation",
       resource_type="AWS::S3::Bucket"
   )
   ```

6. **Present Findings**: Security findings are shown to the user for approval.

7. **Create/Update Resource**: Only after approval, the resource is created/updated with the security check result.

The environment and account information functions are not just for validation - they provide essential context that guides the LLM to generate more secure resources from the start. This proactive approach significantly reduces security issues before they even reach the scanning stage.

### Security Analysis of Existing Resources

I've added security analysis capabilities to the read operations:

1. **Get Resource with Security Analysis**:
   ```python
   resource_info = await get_resource(
       resource_type="AWS::S3::Bucket",
       identifier="my-bucket",
       analyze_security=True
   )
   ```

2. **List Resources with Security Analysis**:
   ```python
   resources = await list_resources(
       resource_type="AWS::S3::Bucket",
       analyze_security=True,
       max_resources_to_analyze=5
   )
   ```

This allows users to ask questions like "Is my EC2 instance secure?" or "What's the security of my S3 buckets?" and get meaningful security analysis.

### Enforcement Mechanisms

I enforce the order through:

1. **Required Parameters**: `security_check_result` is required in resource operations.
2. **Security Validator**: `validate_security_check_result()` validates:
   - Security checks were performed
   - They were performed on the correct resource type
   - They were performed recently (timestamp validation)
   - They didn't find any high-severity issues
3. **Instructions**: MCP instructions guide the LLM through the workflow.
4. **AWS Context**: The environment and account information functions provide essential context that helps the LLM generate secure resources from the start.

## Amazon Q CLI Compatibility Issue

I discovered a specific issue with Amazon Q CLI that required additional attention:

### The Problem

The Amazon Q CLI was failing with a "Improperly formed request" validation error for every prompt, even simple ones like "hello". After investigation, I found that:

1. **Required Parameter Validation**: When `resource_type` was marked as `required=True` in the `run_checkov()` function, Q CLI's validation was more strict than Claude Desktop's. It was trying to validate this parameter even in simple conversations that don't involve security scanning.

2. **Context Size**: The validation error occurred because the context was becoming too large or improperly formatted. The required parameter validation was causing Q CLI to include validation information in every request.

### My Solution

I implemented a targeted fix that maintains security enforcement while resolving the Q CLI compatibility issue:

1. **Made `resource_type` Optional in `run_checkov()`**:
   ```python
   resource_type: str | None = Field(
       description='The AWS resource type being scanned (e.g., "AWS::S3::Bucket")',
       default=None,  # Changed from required=True
   )
   ```

2. **Moved Enforcement to Validation Logic**: Instead of relying on the function signature to enforce security scanning, I moved the enforcement to the validation logic in `security_validator.py`.

## Why My Solution Is Essential

My implementation is the **only solution** that meets all requirements:

1. **Client Compatibility**: Works across all MCP clients (Claude Desktop, Amazon Q CLI, and Cursor)
2. **Enforced Security**: Ensures security scanning is performed before any resource operation
3. **Correct Resource Validation**: Validates that security checks are performed on the correct resource type
4. **Timestamp Tracking**: Ensures security checks are recent
5. **Clear LLM Guidance**: Provides a clear workflow for LLMs to follow
6. **Security Analysis of Existing Resources**: Enables security scanning of existing resources
7. **Contextual Security**: The environment and account information functions provide critical AWS context to the LLM, which has been shown in testing to vastly reduce insecure IAM policy generation and other security issues

**No component can be removed** without breaking the solution:
- Without `generate_infrastructure_code()`, LLMs would revert to skipping security checks
- Without enhanced `run_checkov()`, resource type tracking would be lost
- Without centralized validation, consistency across operations would be compromised
- Without updated instructions, LLMs would not follow the correct workflow
- Without the environment and account information functions, the LLM would lack critical context for generating secure resources (especially IAM policies), leading to more security issues that would need to be caught by scanning

## Testing and Validation

I've tested my solution across all MCP clients:
- Claude Desktop: Works correctly with the new workflow
- Amazon Q CLI: Now properly enforces security scanning without validation errors
- Cursor: Successfully implements the security workflow

My testing specifically showed that providing AWS account context through the environment and account information functions significantly reduced the generation of insecure IAM policies. For example, when the LLM had access to the actual AWS account ID, it properly used that ID in IAM policy principals instead of using wildcards like "AWS:":"*", which is a common security issue.

This multi-layered approach ensures that security is addressed both proactively (through context and guidance) and reactively (through scanning and validation), creating a comprehensive security solution for AWS resource management.

novekm:
will fix tests, just wanted the new changes to be here in so folks can take a look. I was about to cut out about 2k lines of code. Everything in this most recent commit has to be in here to maintain the full security scanning functionality. Reasoning is outlined in my above comment, helped written by Amazon Q 😎

novekm:
Ok I think I fixed all of the tests that relate to this PR. Local pre-commit is passing all tests. I'm not sure why the pipeline is running checks for every single mcp server in the repo when this PR only touched the `cfn-mcp-server` directory, and I believe the readme. Is something wrong with the pipeline? Running tests on every mcp server seems like something that should only be triggered when a new release is cut for the entire project, not a PR for one mcp server

novekm:
Something related I'm working on for the largerFastMCP framework (also mentioned in [this GitHub Issue](https://github.com/jlowin/fastmcp/issues/811)

# Enforcing Tool Dependencies in MCP Framework

## The Problem: Unpredictable LLM Tool Usage Order

When working with LLMs and tool-based frameworks like MCP, there's an inherent challenge in enforcing specific workflows, especially for security-critical operations. LLMs may not consistently follow the intended sequence of tool calls due to:

1. **Variability in reasoning**: Different prompts or contexts may lead to different tool selection strategies
2. **Lack of explicit workflow constraints**: Without hard constraints, LLMs may skip steps they deem unnecessary
3. **Hallucinations**: LLMs might "think" they've called a required tool when they haven't
4. **Prompt sensitivity**: Small changes in instructions can lead to different tool usage patterns

This is particularly problematic for AWS resource management, where security scanning must always precede resource creation to prevent insecure configurations.

## Current Solution: Token-Based Workflow Enforcement

The current approach uses tokens passed between tools to enforce workflow order:

```python
@mcp.tool()
async def generate_infrastructure_code(
    resource_type: str = Field(...),
    properties: dict = Field(...),
    region: str | None = Field(default=None),
) -> dict:
    """Generate infrastructure code for security scanning."""
    # Call the internal implementation
    result = await generate_infrastructure_code_impl(
        resource_type=resource_type,
        properties=properties,
        region=region,
    )
    
    # Add a workflow token
    result['workflow_token'] = f"gen_{datetime.datetime.now().isoformat()}"
    
    return result

@mcp.tool()
async def run_checkov(
    infrastructure_code: dict = Field(...),
) -> dict:
    """Run security scan on generated infrastructure code."""
    # Verify this is properly generated infrastructure code
    if not isinstance(infrastructure_code, dict) or 'workflow_token' not in infrastructure_code:
        raise ClientError(
            'You must call generate_infrastructure_code() first and pass its result to this function'
        )
    
    # Extract the template and resource type
    template = infrastructure_code.get('cloudformation_template')
    if not template:
        raise ClientError('Invalid infrastructure code: missing cloudformation_template')
    
    resource_type = infrastructure_code.get('resource_type')
    
    # Run the security scan
    content = json.dumps(template)
    
    # Call the existing implementation
    security_result = await _run_checkov_impl(
        content=content,
        file_type='json',
        framework='cloudformation',
        resource_type=resource_type,
    )
    
    # Combine the infrastructure code and security results
    result = {
        'infrastructure_code': infrastructure_code,
        'security_result': security_result,
        'workflow_token': infrastructure_code['workflow_token'],
        'resource_type': resource_type,
        'timestamp': str(datetime.datetime.now()),
    }
    
    return result

@mcp.tool()
async def create_resource(
    security_scan_result: dict = Field(...),
    aws_session_info: dict = Field(...),
    skip_security_check: bool = Field(False),
) -> dict:
    """Create an AWS resource after security scanning."""
    # Verify this is a proper security scan result
    if not isinstance(security_scan_result, dict) or 'workflow_token' not in security_scan_result:
        raise ClientError(
            'You must call run_checkov() first and pass its result to this function'
        )
    
    # Extract the infrastructure code and security results
    infrastructure_code = security_scan_result.get('infrastructure_code')
    if not infrastructure_code:
        raise ClientError('Invalid security scan result: missing infrastructure_code')
    
    # Extract resource type and properties
    resource_type = infrastructure_code.get('resource_type')
    properties = infrastructure_code.get('properties')
    
    if not resource_type or not properties:
        raise ClientError('Invalid infrastructure code: missing resource_type or properties')
    
    # Validate security check results
    validate_security_check_result(security_scan_result['security_result'], resource_type, skip_security_check)
    
    # Create the resource using the properties from the infrastructure code
    cloudcontrol_client = get_aws_client('cloudcontrol', region)
    try:
        response = cloudcontrol_client.create_resource(
            TypeName=resource_type, DesiredState=json.dumps(properties)
        )
    except Exception as e:
        raise handle_aws_api_error(e)
    
    return progress_event(response['ProgressEvent'], None)
```

### Issues with Token-Based Approach

While this approach works, it has several drawbacks:

1. **Code Bloat**: Each tool needs boilerplate validation code
2. **Implicit Dependencies**: The workflow is not explicitly defined in the tool definitions
3. **Error-Prone**: Easy to forget validation in new tools or updates
4. **Maintenance Burden**: Changes to the workflow require updates to multiple tools
5. **Unclear Errors**: Error messages may not clearly indicate the correct workflow

## Proposed Solution: `depends_on` Decorator Parameter

A cleaner, more robust solution would be to add a `depends_on` parameter to the MCP tool decorator:

```python
@mcp.tool()
async def generate_infrastructure_code(resource_type: str, properties: dict, region: str = None) -> dict:
    """Generate infrastructure code for security scanning."""
    # Implementation...

@mcp.tool(depends_on=[generate_infrastructure_code])
async def run_checkov(infrastructure_code: dict) -> dict:
    """Run security scan on generated infrastructure code."""
    # Implementation...

@mcp.tool(depends_on=[run_checkov])
async def create_resource(security_scan_result: dict, aws_session_info: dict, skip_security_check: bool = False) -> dict:
    """Create an AWS resource after security scanning."""
    # Implementation...
```

### Benefits of `depends_on` Approach

1. **Explicit Dependencies**: The required workflow is clearly defined in the tool decorators
2. **Reduced Code**: No need for manual token validation
3. **Self-Documenting**: Dependencies are visible at the tool definition level
4. **Centralized Enforcement**: The MCP framework handles dependency checking
5. **Consistent Errors**: Framework can provide clear error messages about missing dependencies

## Implementation Steps

To implement the `depends_on` feature in the MCP framework:

1. **Modify the Tool Decorator**:
   ```python
   def tool(self, depends_on=None):
       def decorator(func):
           # Register the tool with its dependencies
           self.register_tool(func, depends_on=depends_on)
           return func
       return decorator
   ```

2. **Track Tool Calls**:
   ```python
   class ToolCallTracker:
       def __init__(self):
           self.called_tools = set()
           
       def record_call(self, tool_name):
           self.called_tools.add(tool_name)
           
       def check_dependencies(self, tool_name, dependencies):
           for dep in dependencies:
               if dep.__name__ not in self.called_tools:
                   raise DependencyError(f"Tool '{tool_name}' requires '{dep.__name__}' to be called first")
   ```

3. **Integrate with Tool Execution**:
   ```python
   async def execute_tool(self, tool_name, **kwargs):
       tool = self.get_tool(tool_name)
       
       # Check dependencies before execution
       if tool.depends_on:
           self.tracker.check_dependencies(tool_name, tool.depends_on)
           
       result = await tool.func(**kwargs)
       self.tracker.record_call(tool_name)
       return result
   ```

4. **Reset Tracking Between Sessions**:
   ```python
   def start_new_session(self):
       self.tracker = ToolCallTracker()
   ```

5. **Add Clear Error Messages**:
   ```python
   class DependencyError(Exception):
       def __init__(self, message):
           self.message = message
           super().__init__(self.message)
   ```

## Conclusion

The `depends_on` approach provides a more robust, maintainable way to enforce tool workflows compared to token-based validation. It moves the responsibility of workflow enforcement from individual tool implementations to the framework level, reducing code complexity and improving reliability.

### Eliminating Reliance on System Prompts

A key advantage of the `depends_on` approach is that it **completely eliminates the need to rely on system prompts** for workflow enforcement. This addresses several critical limitations of prompt-based enforcement:

1. **Suggestions vs. Requirements**: System prompts provide suggestions that LLMs may choose to follow or ignore. The `depends_on` mechanism creates hard requirements that cannot be bypassed.

2. **Prompt Variability**: LLMs interpret prompts differently based on context, leading to inconsistent adherence to workflows. Framework-level enforcement guarantees consistency regardless of prompt interpretation.

3. **Prompt Overload**: As system prompts grow longer with more instructions, the likelihood of the LLM missing or misinterpreting workflow requirements increases. The `depends_on` approach enforces workflows regardless of prompt complexity.

4. **Security Guarantees**: For security-critical operations, suggestions in prompts are insufficient. The `depends_on` mechanism provides programmatic guarantees that security steps cannot be skipped.

5. **Prompt Maintenance**: Maintaining workflow instructions in prompts creates a parallel maintenance burden. With `depends_on`, the workflow is defined once in the code, ensuring documentation and implementation remain in sync.

### Beyond Security: Broader Applications

While security enforcement was the initial motivation for this feature, the `depends_on` approach has much broader applications for ensuring correct tool usage sequences:

#### Example: Weather Information in Emails

Consider a scenario where an LLM needs to send an email containing current weather information:

```python
@mcp.tool()
async def get_weather(location: str) -> dict:
    """Get current weather for a location."""
    # Implementation...

@mcp.tool(depends_on=[get_weather])
async def send_email_with_weather(recipient: str, subject: str, weather_data: dict) -> dict:
    """Send an email containing weather information."""
    # Implementation...
```

Without `depends_on`, the LLM might:
1. Skip checking the weather and send an email with outdated or missing weather data
2. Check the weather but fail to include it in the email
3. Send multiple emails with inconsistent weather information

With `depends_on`, the framework ensures that:
1. Weather data is always obtained before sending the email
2. The most recent weather data is used in the email
3. The workflow is consistently followed regardless of prompt variations

#### Other Use Cases

This approach is valuable for many sequential workflows:

- **Data Processing**: Ensuring data is validated before being processed
- **Financial Transactions**: Verifying account balances before executing transfers
- **Content Generation**: Gathering facts before generating content based on them
- **API Interactions**: Authenticating before making API calls
- **Multi-step Approvals**: Ensuring all approvals are obtained before proceeding

For security-critical applications like AWS resource management, this approach would provide true enforcement rather than suggestions, making it impossible for LLMs to bypass important security checks while also making the codebase cleaner and more maintainable.

novekm:
Instead of renaming the existing `cfn-mcp-server` to `ccapi-mcp-server`, I propose we just use PR #739 that creates a net new `ccapi-mcp-server`. The existing `cfn-mcp-server` can then be separately updated (in the near future) to align more closely to the expected customer experience CloudFormation users desire, and to prevent customer confusion as mentioned in #390. Imo this is less destructive than renaming the existing one, which has already been out for a while and may already have users. It is already live in this repo and as a Python package so I'd say let's just refactor it later and focus on net-new CCAPI. This also prevents the substantial number changes that were shown in this PR when these changes were initially in relation to the cfn-mcp-server.  Closing this PR for now in lieu of #739.

PR #590, opened 2025-06-13
fix!: Consolidating client cert and connection configuration in user's home directory

jcshepherd:
Fixes

### Changes

The MCP server requires a client-side cert to connect to Amazon Keyspaces, as well as connection configuration. Originally, these were expected to be in the project/application root directory, which is very confusing for end-users. This change standardizes their location to a ".keyspaces-mcp" subdir in the user's home directory.

### User experience

The user still needs to download the Starfield certificate to connect to Amazon Keyspaces, and appropriately configure their connection. With this change, the certificate and configuration are expected to be in a standard location in the user's home directory, rather than in the project deployment itself.

## Checklist

* [X] Changes are documented

Is this a breaking change? (Y/N)  Y - In that the location of the config has changed. However, usage of this server is very low (zero?) and the change should not impact dependencies beyond the Keyspaces MCP server itself.

**RFC issue number**:

Checklist:

## Acknowledgment


src/amazon-keyspaces-mcp-server/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "awslabs.amazon-keyspaces-mcp-server"
-version = "0.0.1"
+version = "0.0.2"
```
scottschreckengaust: Release automatically bumps the "Patch"
```suggestion
version = "0.0.1"
```

alexa-perlov (approved):

krokoko (approved):

PR #638, opened 2025-06-23
feat(eks-mcp-server): Add MCP tool to retrieve EKS cluster names

LMarkie:
Fixes

### Changes

A new tool is being added to retrieve a list of EKS cluster names.

### User experience

Users have to provide cluster names to other EKS MCP tools. This new tool provides an easy way to retrieve them.

## Checklist

* [ ] Changes are documented

**RFC issue number**:

Checklist:

## Acknowledgment


alexa-perlov:
Please run `pre-commit run --all-files` to address pre-commit failing checks and take a look at the missing code coverage. Thank you!

csantanapr:
I would not implement list clusters as tool, list of clusters should be implemented as [MCP Resource](https://modelcontextprotocol.io/docs/concepts/resources) which is dynamic and is based on the clusters user have access to

while current commercial or oss mcp clients doesn't have good support for resources, they will have it soon. in the mean time llms can figure out how to get the list of clusters using aws tools thru mcp or aws cli/api. 
But we should start design doc on what things make sense to go in tools vs. resources vs. prompt vs. elicitation vs. sampling for the EKS MCP Server

patrick-yu-amzn:
The eks-mcp-server is not intended to be a thin wrapper around plain AWS resources like clusters. There's the cfn-mcp-server that specifically handles that. Besides, most of the common LLMs can easily handle this case with simple AWS cli commands, so a dedicated mcp tool is a bit redundant.

PR #646, opened 2025-06-24
Prometheus mcp v3.0

MohamedSherifAbdelsamiea:
Fixes

## Summary

Refactor Prometheus MCP server to require workspace_id parameter for all tool functions, improving user experience by making each tool invocation self-contained and independent of previous workspace selection.

### Changes

This PR refactors the Prometheus MCP server to require workspace_id as a parameter for all tool functions. This improves the user experience by making each tool invocation self-contained and independent of previous workspace selection.

Key changes:
- Added workspace_id parameter to all MCP tool functions
- Added a new GetAvailableWorkspaces tool to list available workspaces
- Removed SetPrometheusWorkspace tool as it's no longer needed
- Added configure_workspace_for_request helper function to set up workspace for each request
- Updated README.md with new configuration and usage examples

### User experience

**Before:**
- Users had to first call SetPrometheusWorkspace to select a workspace
- All subsequent tool calls would use that workspace
- If the user wanted to switch workspaces, they had to call SetPrometheusWorkspace again

**After:**
- Users first call GetAvailableWorkspaces to see available workspaces
- Each tool call requires a workspace_id parameter
- Users can easily work with multiple workspaces in the same session
- More explicit and predictable behavior with each tool call being self-contained

## Checklist

* [x] Changes are documented

Is this a breaking change? (Y/N) Y

**RFC issue number**: N/A

Checklist:

## Acknowledgment


run_prometheus_server.sh
scottschreckengaust, line 1: We are looking to remove scripts. Consider documenting how to start using `uv` commands?

WORKSPACE_SELECTION_CHANGES.md
scottschreckengaust, line 1: Please move this into `src/prometheus-mcp-server`

scottschreckengaust (dismissed):
Please remove/move files

scottschreckengaust:
Thank you for your contribution. Please review the unit test coverage requirements please

MohamedSherifAbdelsamiea:
The coverage decrease isn't from prometheus-mcp-server (which has 92% coverage) but from removing other packages/files that had higher coverage than the project average.

Prometheus MCP Status: ✅ 92% coverage - exceeds requirements

run_prometheus_server.sh
alexa-perlov, line 1: delete?

src/prometheus-mcp-server/awslabs/prometheus_mcp_server/server.py
```diff
@@ -115,62 +115,61 @@ def load_config(args):
...
-        ):
-            logger.warning(
-                f"WARNING: URL doesn't appear to be an AWS Managed Prometheus endpoint: {config['prometheus_url']}"
-            )
-            logger.warning(
-                'Expected format: https://aps-workspaces.[region].amazonaws.com/workspaces/ws-[id]'
+        # Create session with profile if specified
+        if config_data['aws_profile']:
+            logger.info(f'  Using AWS Profile: {config_data["aws_profile"]}')
+            session = boto3.Session(
```
alexa-perlov: consider using user_agent_extra for boto3 client

run_prometheus_server.sh
MohamedSherifAbdelsamiea: Yes

src/prometheus-mcp-server/awslabs/prometheus_mcp_server/server.py
```diff
@@ -618,14 +671,135 @@ async def get_server_info(ctx: Context) -> ServerInfo:
...
+
+        return {'workspaces': workspaces, 'count': len(workspaces), 'region': aws_region}
+    except Exception as e:
+        error_msg = f'Error listing workspaces: {str(e)}'
+        logger.error(error_msg)
+        await ctx.error(error_msg)
+        raise
+
+
+async def configure_workspace_for_request(
```
niallthomson, line 929: This should not be relying on writing to a file, the behavior should be more dynamic. This will still have issues operating as a remote MCP with multiple users accessing it.

niallthomson (dismissed):
I'm still concerned about the way this is using a configuration file to store the URL, theres no need for it to do that. The URL should be constructed dynamically on a tool invocation without ever writing anything to a file.

src/prometheus-mcp-server/awslabs/prometheus_mcp_server/server.py
MohamedSherifAbdelsamiea: configure_workspace_for_request dynamically configures the Prometheus workspace connection for each individual tool request, enabling the server to work with different workspaces without requiring a global configuration at startup. 
config = PrometheusConfig(
    prometheus_url=None,        # Dynamic - set per request
    aws_region='us-east-1',     # Static - from startup config
    aws_profile='my-profile',   # Static - from startup config  
    service_name='aps',         # Static - AWS service name
    retry_delay=1,              # Static - retry settings
    max_retries=3               # Static - retry settings
)

Input Source: The workspace_id comes from the LLM/user as a parameter to each tool call
LLM/User → Tool Call → workspace_id parameter → configure_workspace_for_request() → Constructed URL

Could you please run a test using your specific scenario and provide evidence if the tool is not functioning properly?

NP: this MCP implement stdio protocol, not sure about remote MCP use case. As per Anthropic documentation there is always 1 connection between MCP client and server

run_prometheus_server.sh
MohamedSherifAbdelsamiea: I've added user_agent_extra='prometheus-mcp-server' to the boto3 client configurations for both the STS client (used for credential validation) and the AMP client (used for listing workspaces)

src/prometheus-mcp-server/awslabs/prometheus_mcp_server/server.py
niallthomson: I understand why this code ended up here but it should not be here:

1. Since you use FastMCP this already supports remote MCP, you should design accordingly
2. No other MCP servers in this repository have their own configuration file that I can see, this MCP isn't special

I suggest you:

1. Completely remove `load_config`, `setup_environment` and `configure_workspace_for_request`
2. Add a new function to create appropriate AMP API client similar to how the Dynamo MCP does [here](https://github.com/awslabs/mcp/blob/main/src/dynamodb-mcp-server/awslabs/dynamodb_mcp_server/server.py#L105)
3. Wherever your are making a query, first use `DescribeWorkspace` to retrieve the workspace info, including the URL. Don't try to construct the URL yourself. This will also validate that the workspace ID is correctly formatted and valid so you can remove that code.

WORKSPACE_SELECTION_CHANGES.md
niallthomson: I dont think this file should be here at all?

src/prometheus-mcp-server/awslabs/prometheus_mcp_server/server.py
```diff
@@ -618,14 +674,135 @@ async def get_server_info(ctx: Context) -> ServerInfo:
...
+) -> Dict[str, Any]:
+    """List all available Prometheus workspaces in the specified region.
+
+    ## Usage
+    - Use this tool to see all available Prometheus workspaces
+    - Shows workspace ID, alias, status, and creation date
+    - Use this to find workspace IDs needed for other tools
+    - Always call this first to get a workspace ID before using other tools
+
+    ## Example
```
niallthomson, line 770: The examples provided in the docs for all the tools do not seem correct. These are supposed to be consumed primarily by the model for invoking the tools, whereas what I see are examples for Python code to invoke it. I would suggest removing these examples.

MohamedSherifAbdelsamiea: Thanks! I have update the examples to be more MCP related

src/prometheus-mcp-server/awslabs/prometheus_mcp_server/server.py
MohamedSherifAbdelsamiea: Again this MCP like others is based on stdio transport for local communication which is the only supported transport protocol for Amazon Q Developer. Please review concepts of MCP (Transport selection section) https://modelcontextprotocol.io/docs/concepts/architecture
I've refactored the Prometheus MCP server code to adopt the approach of creating fresh clients for each request.

src/prometheus-mcp-server/awslabs/prometheus_mcp_server/server.py
MohamedSherifAbdelsamiea: I've added user_agent_extra='prometheus-mcp-server' to the boto3 client configurations for both the STS client (used for credential validation) and the AMP client (used for listing workspaces)

src/prometheus-mcp-server/awslabs/prometheus_mcp_server/server.py
MohamedSherifAbdelsamiea: Refactored the Prometheus MCP server code to remove the global config object and use explicit parameter passing instead

WORKSPACE_SELECTION_CHANGES.md
MohamedSherifAbdelsamiea: fixed

alexa-perlov (approved):

dineshSajwan (approved):

PR #679, opened 2025-06-26
refactor(resource-management): implement full CRUDL API with direct boto3 mapping and fix service events bugs

karanbokil:
## Summary
Adds create/update/delete functionality to the resource management tool, requiring write permissions for all non-read actions. Scope permitted operations via a list in SUPPORTED_ECS_OPERATIONS, inputs use kwargs dictionary syntax which allows AI Assistants to easily input nested parameters.

Also fixing async bugs in fetch_service_events.

## Checklist

* [x] Changes are documented

## Acknowledgment


alexa-perlov:
Thanks for the updates! Can you update the MCP server's readme as well to reflect these new changes?

src/ecs-mcp-server/awslabs/ecs_mcp_server/modules/resource_management.py
```diff
@@ -17,63 +17,110 @@
...
-        - List task definitions: ecs_resource_management("list", "task_definition",
-          filters={"family": "nginx"})
-        - Describe a task definition: ecs_resource_management("describe", "task_definition",
-          "family:revision")
+        Execute ECS API operations directly.
+
+        This tool allows direct execution of ECS API operations using boto3.
+
+        PERMISSIONS:
+        - Operations starting with "Describe" or "List" are read-only
```
matthewgoodman13: nit: are these permission comments needed when we report below in parenthesis what the permissions required are?

src/ecs-mcp-server/tests/unit/test_resource_management_api_operation.py
```diff
@@ -0,0 +1,217 @@
...
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+from awslabs.ecs_mcp_server.api.resource_management import camel_to_snake, ecs_api_operation
+
+
+def test_camel_to_snake():
+    """Test the camel_to_snake function."""
+    assert camel_to_snake("CreateService") == "create_service"
```
matthewgoodman13, line 14: Should we assert every single operation we support?

src/ecs-mcp-server/awslabs/ecs_mcp_server/api/resource_management.py
```diff
@@ -15,755 +15,150 @@
...
+    "UpdateService",
+    "UpdateServicePrimaryTaskSet",
+    "UpdateTaskProtection",
+    "UpdateTaskSet",
+]
+
+
+def camel_to_snake(name):
+    """
+    Convert CamelCase to snake_case.
```
matthewgoodman13, line 97: nit: able to document why we use the camel_to_snake in code comment here? Assuming its just for the boto3 client.

src/ecs-mcp-server/tests/unit/test_resource_management_api_operation.py
karanbokil: I don't think so - technically the method we're using is generalizable to any aws api (if we got rid of the allowlist) and setting up mocking for every possible operation doesn't actually prove/improve the correctness of the approach and just makes our testing more brittle

src/ecs-mcp-server/awslabs/ecs_mcp_server/api/resource_management.py
karanbokil: Yep, will add - though in short API operations (specifically in our AWS Docs) represent as CamelCase, but boto3 specifically takes them as snake_case.  I could have made the input be snake_case from the getgo, but then doing things like RAG off of AWS docs may cause the LLM to input malformed operation names

src/ecs-mcp-server/awslabs/ecs_mcp_server/modules/resource_management.py
karanbokil: Removed!

matthewgoodman13 (approved):
LGTM

alexa-perlov (approved):

PR #715, opened 2025-07-01
feat(bedrock-kb-retrieval): add ListKnowledgeBases tool              …

### LINKED ISSUE #714: bedrock-kb-retrieval-mcp-server: Add ListKnowledgeBases tool for better compatibility with MCP hosts
### Describe the feature

Add a new ListKnowledgeBases tool to the bedrock-kb-retrieval-mcp-server that provides an alternative way to discover available Amazon Bedrock Knowledge Bases. This tool would complement the existing resource://knowledgebases resource endpoint by offering the same functionality through the MCP Tools interface.

### Use Case

Currently, the bedrock-kb-retrieval-mcp-server exposes available knowledge bases through the MCP Resources interface (`resource://knowledgebases`). However, many MCP host applications, including Amazon Q CLI, do not yet support the MCP Resources specification. This limitation prevents users from autonomously discovering available Bedrock Knowledge Base IDs when using these hosts.

Since most MCP hosts do support the MCP Tools interface, users are unable to effectively utilize the bedrock-kb-retrieval-mcp-server without manually obtaining and hardcoding knowledge base IDs, which defeats the purpose of dynamic discovery.

### Proposed Solution

Implement a `ListKnowledgeBases` tool that:
1. Executes the same `discover_knowledge_bases` function currently used by the resources endpoint
2. Returns the same JSON structure containing knowledge base IDs, names, and data sources
3. Can be called before using the `QueryKnowledgeBases` tool to obtain valid knowledge base IDs

This approach ensures compatibility with a wider range of MCP hosts while maintaining the existing resources interface for hosts that do support it. Users would have the flexibility to use either:
- `resource://knowledgebases` (for hosts with Resources support)
- `ListKnowledgeBases` tool (for hosts with only Tools support)

I have prepared a [PR](https://github.com/awslabs/mcp/issues/714) implementing this feature for your consideration.

### Other Information

- The implementation reuses the existing `discover_knowledge_bases` function, ensuring consistency

### Acknowledgements

tetsuzawa:
Fixes #714

### Changes

This PR adds a new `ListKnowledgeBases` tool to the bedrock-kb-retrieval-mcp-server that provides an alternative method for discovering available Amazon Bedrock Knowledge Bases.

**Key changes:**
- Added `ListKnowledgeBases` tool that executes the existing `discover_knowledge_bases` function
- Updated documentation to mention both discovery methods (`resource://knowledgebases` and `ListKnowledgeBases`)
- No changes to existing functionality - this is purely additive

### User experience

**Before:**
Users of MCP hosts without Resources support (e.g., Amazon Q CLI) had to manually find and hardcode knowledge base IDs, as they couldn't use the `resource://knowledgebases` endpoint for discovery.

```
# User had to know the KB ID beforehand
QueryKnowledgeBases(
    knowledge_base_id="kb-12345",  # Hardcoded ID
    query="What is the return policy?"
)
```

**After:**
Users can now discover available knowledge bases using the new tool, regardless of their MCP host's capabilities.

```
# Step 1: Discover available knowledge bases
ListKnowledgeBases()
# Returns: {"kb-12345": {"name": "Customer Support KB", ...}}

# Step 2: Use the discovered ID
QueryKnowledgeBases(
    knowledge_base_id="kb-12345",
    query="What is the return policy?"
)
```

## Checklist

* [x] Changes are documented

Is this a breaking change? (Y/N): **N**

**RFC issue number**: N/A

Checklist:

## Acknowledgment


alexa-perlov (dismissed):

laithalsaadoon:
Actually at this point, I am fine converting the existing resource into a tool. What do you think? @tetsuzawa

tetsuzawa:
@theagenticguy 
Thanks for the suggestion!

You're right that converting to tool-only would be cleaner. The only thing is it would be a breaking change for existing users of the resource.

I don't have a strong preference either way - happy to keep both or simplify to tool-only based on what you think is best. Let me know and I'll update the PR accordingly!

laithalsaadoon:
> @theagenticguy Thanks for the suggestion!
> 
> You're right that converting to tool-only would be cleaner. The only thing is it would be a breaking change for existing users of the resource.
> 
> I don't have a strong preference either way - happy to keep both or simplify to tool-only based on what you think is best. Let me know and I'll update the PR accordingly!

I'll approve refactoring the current resource as a tool 👍

tetsuzawa:
@theagenticguy
Thank you for approving the refactoring! I've updated the PR to remove the `knowledgebases_resource` entirely and keep only the `ListKnowledgeBases` tool.

Changes made:
- Removed the `@mcp.resource` decorator and `knowledgebases_resource` function
- Updated all documentation references to use the `ListKnowledgeBases` tool instead of `resource://knowledgebases`
- The functionality remains the same, just accessed via tool instead of resource

This simplifies the API surface while maintaining all the functionality. The PR is now ready for review!

Issue #720, opened 2025-07-02
RFC: Athena MCP server


jbrock-embed:
### Summary

We propose an MCP server to expose various read-only Athena operations as MCP tools. Examples include running `SELECT` queries, as well as listing workgroups, databases, and tables. SQL statements will be parsed and validated to prevent non-read-only operations, such as `CREATE`, `UPDATE`, and `INSERT`.

### Use case

Common use cases include:
* Developers exploring table schemas and data structure during application development
* Data scientists performing exploratory data analysis across multiple databases and workgroups.
* Business analysts querying data lakes for insights

Users can leverage AI assistance for writing complex SQL queries, understanding data relationships, and generating reports.

### Proposal

The Athena MCP server implements a FastMCP-based service that exposes Amazon Athena's read-only operations through the Model Context Protocol. The server provides eight core MCP tools corresponding to Athena's primary operations:

### Core Tools

- `execute_query`: Executes `SELECT`, `VALUES`, `DESCRIBE`, `SHOW`, and `EXPLAIN` statements  
- `get_query_results`: Retrieves paginated results from completed queries  
- `list_databases`: Lists available databases in a specified data catalog  
- `list_tables`: Lists tables within a database  
- `get_table_metadata`: Returns detailed schema information for specific tables  
- `list_work_groups`: Lists available Athena workgroups with their configurations  
- `get_work_group`: Retrieves detailed configuration for a specific workgroup  
- `list_data_catalogs`: Lists available data catalogs with their configurations  

---

### SQL Validation

The server implements a multi-layered validation system using SQL parsing to restrict operations to read-only queries:

1. **Statement Parsing**: SQL queries are parsed and validated for correct syntax  
2. **Statement Type Validation**: Allows only read-only operations like `SELECT`, `VALUES`, `DESCRIBE`, and `SHOW`  
3. **AST Traversal**: Recursively analyzes the query structure to detect nested DML and DDL operations  

### Security Model

The server attempts to prevent non-read-only query statements from executing, but the validation may not be exhaustive: users are expected to follow a defense-in-depth approach by using additional safeguards such as:

- IAM policies with read-only permissions  
- Separate AWS profiles with minimal privileges 

This belt-and-suspenders approach ensures multiple independent security controls protect against unauthorized data access or modification.

### Out of scope

[...]

PR #722, opened 2025-07-02
feat: Add Athena MCP Server

jbrock-embed:
## Summary

Add a new Athena MCP server:
- **Execute Read-Only SQL Statements**: Run commands such as SELECT, VALUES, DESCRIBE, and SHOW
- **Data Discovery**: Browse databases, tables, and schema information across multiple regions and catalogs
- **Workgroup Information**: List and view configuration values for Athena workgroups
- **Pagination Support**: Handle large result sets via pagination 

### Changes

* Adds a new MCP server under `src/athena-mcp-server`
* Adds brief descriptions of the new MCP server to the repo-level README.md and docs
* Adds extensive pytest tests for:
  * MCP server tools and logic
  * SQL query validation, i.e. tests for the code that validates that queries parse and are read-only

### User experience
The MCP server exposes the following tools:
Tool | Description
-- | --
`execute_query` | Executes a read-only SQL query in Athena and returns the results directly
`get_query_results` | Retrieves results from a completed query execution (mainly for pagination)
`list_databases` | Lists databases in the specified data catalog
`list_tables` | Lists tables within a specified database
`get_table_metadata` | Gets detailed metadata for a specific table including schema
`list_work_groups` | Lists available Athena workgroups with their configuration
`get_work_group` | Gets detailed workgroup configuration and settings
`list_data_catalogs` | Lists available data catalogs beyond the default AwsDataCatalog

## Checklist

* [x] Changes are documented

Is this a breaking change? (Y/N) N.

**RFC issue number**: #720

Checklist:

## Acknowledgment


dineshSajwan:
Hi @jbrock-embed we have a comprehensive [MCP server for data processing](https://github.com/awslabs/mcp/tree/main/src/aws-dataprocessing-mcp-server), Can we please update it instead of creating a separate MCP server. It has been recently updated with Athena tools.

jbrock-embed:
@dineshSajwan Thanks for taking a look, but I think there are strong reasons to keep the Athena MCP server separate rather than merging into the data processing server.

One reason is consistency: having an independent Athena MCP server is more consistent with the majority of the existing AWS Labs MCP servers, where there's a one-to-one mapping between AWS services and MCP servers. For example, there are currently separate MCP servers for Redshift, DynamoDB, Kendra, S3 Tables, and Aurora Postgres, all of which could arguably fall under the rubric of "data processing".

Even if we don't want to stand on precedent, there are other good reasons to have narrowly scoped MCP servers: A single-service MCP server with a clear name is easy for users to discover and reason about. For example, it's fairly obvious what an "Athena MCP server" does; at the very least it's obvious what AWS service it uses. On the other hand, it's ambiguous what a "Data Processing MCP server" does.

It's also easier for LLMs to reason about a narrowly scoped set of MCP tools, which reduces the chances of the LLM failing to choose the correct tool. Related to this, a dedicated Athena MCP server has significantly reduced attack surface and IAM complexity vs. a more complicated multi-service MCP server. Organizations are already nervous about letting their users use MCP servers. Advanced users can justify the larger attack surface of the Data Processing MCP server, but for many orgs, exposing >30 tools with optional write capabilities is beyond the pale, and won't pass a security review.

I think for particular relatively advanced use cases the Data Processing MCP server makes sense. But for common exploratory and analytical use cases, like data scientists or business analysts leveraging Athena queries, a dedicated and narrowly focused Athena MCP server is a lot more friendly, secure, and usable out-of-the-box, while also being less likely to do the wrong thing.

Anyway, let me know what you think. Thanks again!

biswasbiplob:
What @jbrock-embed makes absolute sense, if there are individual mcp servers for redshift, why bundle athena into the data processing mcp servers.

jbrock-embed:
Hi @dineshSajwan or other reviewers, please let me know what your thoughts are when you get a chance.

dineshSajwan:
Hi @jbrock-embed Thank you so much for taking the time to provide such a detailed and thoughtful explanation. I really appreciate your contribution to the project and the effort you've put into this PR. Your points about consistency and security are well-taken and show careful consideration of the architecture.

I completely understand where you're coming from, and I want to find a solution that works for everyone. Based on our experience with the MCP ecosystem so far, we've observed a few patterns that give me pause:

1. Users often experience decision paralysis when choosing between MCP servers with overlapping capabilities
2. Our maintenance burden increases with each new server that needs updates and documentation
3. Both users and LLMs can struggle with cognitive overload when presented with similar options
4. We're seeing some duplication of functionality across multiple codebases

These factors are beginning to impact developer productivity and system maintainability, which I know we both want to avoid.

I wonder if we might explore a middle ground here? Would you be open to first checking whether the existing Data Processing MCP server could address your Athena use cases? If it falls short of your requirements, perhaps we could collaborate on enhancing that server with the specific Athena tooling you need. This approach would deliver the Athena capabilities you're looking for while helping us manage the growth of our MCP ecosystem.

jbrock-embed:
@dineshSajwan I'm sure you're very busy, but your reply didn't really address any of my points (I'm guessing you asked an LLM to write it?).

To restate what I said in my previous reply, in my opinion, requiring people to find and use a monolithic "data processing" MCP server when all they want to do is run some Athena queries is going to be a bad user experience: the ambiguous name makes discoverability harder, it's inconsistent with the 1-to-1 service-to-MCP-server approach used in most of this repo, it increases the decision surface for the LLM (i.e., deciding which tools to use), and it increases the attack surface and IAM complexity vs. an Athena-only MCP server.

If you don't think fixing any of that is worth the DRY trade-off of having a data processing MCP server live alongside an Athena MCP server, that's fine, just let me know.

If you don't want to merge this PR, my recommendation would be splitting the data processing MCP server into separate MCP servers for the various AWS services it touches, and then giving each one a very obvious name (like Athena MCP server, Glue MCP server, etc.). I unfortunately don't have the cycles to assist with that, though.

PR #739, opened 2025-07-04
feat: add ccapi-mcp-server

novekm:
## Summary
Removes the need for #548. Instead of renaming the existing `cfn-mcp-server` to `ccapi-mcp-server`, I propose we just use this PR that creates a net new `ccapi-mcp-server`. The existing `cfn-mcp-server` can then be separately updated to align more closely to the expected customer experience CloudFormation users desire, and to prevent customer confusion as mentioned in #390. Imo this is less destructive than renaming the existing one, which may already have users and is already live in this repo and as a Python package. This also prevents the substantial number changes that were shown when these changes were initially in relation to the cfn-mcp-server.  

### Changes

This PR adds a new **AWS Cloud Control API (CCAPI) MCP Server** that provides direct AWS resource management through the AWS Cloud Control API. The server enables AI assistants to create, read, update, and delete AWS resources with token-based security validation and Infrastructure as Code generation capabilities.

**Key additions:**
Token-Based Security Workflow:
- **Properties Token System**: Ensures LLMs can only use infrastructure code that was security-scanned, preventing substitution of unvalidated code during resource creation
- **Security Check Token Chain**: Creates implicit dependencies forcing proper tool ordering (generate → scan → create/update)
- **Workflow Validation**: Tokens prove each security step was completed, blocking resource operations without proper validation

**Core Features:**
- Complete CCAPI MCP Server implementation with AWS Cloud Control API integration
- Security validation using Checkov for infrastructure compliance with mandatory workflow enforcement
- Automatic tagging system for resource management and tracking
- Schema-based validation for AWS resource properties
- Infrastructure as Code generation for CloudFormation templates
- Comprehensive test suite with 95%+ coverage
- Documentation and usage examples

### User experience

**Before:** Users had no direct way to manage AWS resources through MCP, requiring separate tools or manual AWS console/CLI operations.

**After:** Users can now ask AI assistants to:
- Create AWS resources (S3 buckets, KMS keys, etc.) with **mandatory security validation** - LLM cannot bypass or modify scanned code
- Generate Infrastructure as Code templates from resource specifications
- Apply security best practices automatically through built-in compliance checks
- Track and manage resources through consistent tagging
- **Flexible usage**: Same tools work for ad-hoc security scanning of existing resources where strict ordering isn't required

### Critical Workload Benefits:
**Token-based approach provides:**
- **Zero-trust validation**: LLM cannot use unscanned infrastructure code
- **Audit trail**: Every resource creation proves security validation occurred
- **Tamper-proof workflow**: Properties locked after security scanning
- **Compliance guarantee**: Critical workloads get mandatory security validation

**Example:** "Create an S3 bucket with encryption enabled" → AI assistant uses token-enforced workflow: generates secure CloudFormation → validates with Checkov → creates resource using only the scanned, validated properties.

**Flexibility:** For non-critical tasks like "scan this existing bucket for security issues," LLM can use individual tools without strict token ordering requirements.

## Checklist

* [x] Changes are documented

**RFC issue number**: N/A

Checklist:

## Acknowledgment


src/ccapi-mcp-server/CHANGELOG.md
```diff
@@ -0,0 +1,17 @@
...
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [1.0.0] - 2025-01-XX
+
+### Added
+
+- Initial release of Cloud Control API MCP Server
+- Support for AWS resource CRUDL operations via Cloud Control API
+- Security scanning with Checkov integration
```
KaramVSingh: Small note - i think this was removed

src/ccapi-mcp-server/NOTICE
```diff
@@ -0,0 +1,6 @@
+<<<<<<< Updated upstream
```
KaramVSingh: I think this was missed in git :D

src/ccapi-mcp-server/awslabs/ccapi_mcp_server/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Workflow
+
+The tool dependencies feature enforces the following workflow:
+
+1. **Generate Infrastructure Code**: Use `generate_infrastructure_code()` to create the infrastructure code for security scanning.
+2. **Run Security Scan**: Use `run_checkov()` to scan the generated code for security issues.
+3. **Create/Update Resource**: Use `create_resource()` or `update_resource()` to create or update the resource after security scanning.
+
+## How It Works
```
KaramVSingh: I think this section in the README is interesting as an engineer - however i think we could benefit from putting it later in the README or in a separate md file since the README is intended for users of the tool rather than developers of the tool. I actually think we should expand the usage section and reframe benefits to be "benefits of using CC API MCP" potentially

src/ccapi-mcp-server/awslabs/ccapi_mcp_server/aws_client.py
```diff
@@ -0,0 +1,85 @@
...
+
+import botocore.config
+import sys
+from awslabs.ccapi_mcp_server.errors import ClientError
+from boto3 import Session
+from os import environ
+
+
+session_config = botocore.config.Config(
+    user_agent_extra='cfn-mcp-server/1.0.0',
```
KaramVSingh: small note - this should be ccapi-mcp-server now :)

docs/servers/ccapi-mcp-server.md
```diff
@@ -0,0 +1,5 @@
+---
+title: AWS Cloud Control API MCP Server
```
KaramVSingh, line 2: Naming discussion: 
Should we be cloudcontrol-api-mcp-server, or ccapi-mcp-server. I suspect you landed on ccapi-mcp-server since it follows the same convention as cfn-mcp-server, but even that seems a bit confusing in retrospect.

src/ccapi-mcp-server/awslabs/ccapi_mcp_server/env_manager.py
```diff
@@ -0,0 +1,234 @@
...
+import json
+import os
+import subprocess
+from typing import Any, Dict, Optional
+
+
+# Default values based on README
+DEFAULT_VALUES = {
+    'AWS_REGION': '',  # Let AWS profile/config determine default region
+    'AWS_PROFILE': '',
```
KaramVSingh, line 26: Should these be removed from the dictionary if they do not have a default? I think that could make errors more clear

src/ccapi-mcp-server/awslabs/ccapi_mcp_server/env_manager.py
```diff
@@ -0,0 +1,234 @@
...
+
+    Returns:
+        The value of the environment variable or the default value
+    """
+    if name in DEFAULT_VALUES and name not in os.environ:
+        return DEFAULT_VALUES[name]
+    return os.environ.get(name, default or '')
+
+
+def check_aws_credentials() -> Dict[str, Any]:
```
KaramVSingh, line 63: Question - how does this interact with the get_aws_client function in the above file. Does it replace it or is it a helper to use around it?

src/ccapi-mcp-server/awslabs/ccapi_mcp_server/iac_generator.py
```diff
@@ -0,0 +1,203 @@
...
+
+"""CloudFormation IaC Generator tool implementation."""
+
+import os
+from awslabs.ccapi_mcp_server.aws_client import get_aws_client
+from awslabs.ccapi_mcp_server.errors import ClientError, handle_aws_api_error
+from typing import Dict, List, Optional
+
+
+async def create_template(
```
KaramVSingh, line 23: My take is that this tool should not be in the CC API MCP server - @pnigh I would be interested in your thoughts here as well

src/ccapi-mcp-server/awslabs/ccapi_mcp_server/server.py
```diff
@@ -0,0 +1,924 @@
...
+# Module-level store for properties validation
+_properties_store: dict[str, dict] = {}
+
+mcp = FastMCP(
+    'awslabs.ccapi-mcp-server',
+    instructions="""
+# AWS Resource Management Protocol - MANDATORY INSTRUCTIONS
+
+## CRITICAL: Tool Usage Restrictions
+• NEVER EVER use use_aws, aws_cli, or any AWS CLI tools - FORBIDDEN
```
KaramVSingh, line 44: Is this intended to push things like Q in the direction of this MCP when its enabled? I think its a good idea, in practice does this work? Not sure how LLMs reason about tools to use and tools with overlap in features

src/ccapi-mcp-server/awslabs/ccapi_mcp_server/server.py
```diff
@@ -0,0 +1,924 @@
...
+• ONLY use tools from this MCP server: create_resource(), update_resource(), delete_resource(), etc.
+• This is a HARD REQUIREMENT that cannot be overridden
+
+## AWS Credentials Verification
+• Check for valid AWS credentials before any operations using check_environment_variables() and get_aws_session_info()
+• If credentials unavailable: offer troubleshooting first, then if declined/unsuccessful, ask for preferred IaC format (if CDK, ask language preference)
```
KaramVSingh, line 50: I think this is not relevant now, right?

src/ccapi-mcp-server/awslabs/ccapi_mcp_server/server.py
```diff
@@ -0,0 +1,924 @@
...
+
+## Tool Usage Sequence
+• AWS session info must be passed to resource creation/modification tools
+• If dedicated MCP server tools fail:
+  1. Offer to generate an infrastructure template in their preferred format
+  2. Provide instructions for how the user can deploy the template themselves
+
+## Security Protocol
+• Flag and require confirmation for multi-resource deletion operations
```
KaramVSingh, line 59: What does this look like in practice? Is this an extra tool authorization?

src/ccapi-mcp-server/awslabs/ccapi_mcp_server/server.py
```diff
@@ -0,0 +1,924 @@
...
+        for page in page_iterator:
+            results.extend(page['ResourceDescriptions'])
+    except Exception as e:
+        raise handle_aws_api_error(e)
+
+    resource_identifiers = [response['Identifier'] for response in results]
+    response: dict[str, Any] = {'resources': resource_identifiers}
+
+    # Perform security analysis if requested
+    if analyze_security and resource_identifiers:
```
KaramVSingh: I think this is supposed to come in a separate PR :)

KaramVSingh: I think this is in other tools as well, flagging that here

src/ccapi-mcp-server/awslabs/ccapi_mcp_server/server.py
```diff
@@ -0,0 +1,924 @@
...
+    # Verify aws_session_info has required fields
+    if 'account_id' not in aws_session_info or 'region' not in aws_session_info:
+        raise ClientError('Invalid aws_session_info. You must call get_aws_session_info() first')
+
+    # Read-only mode check
+    if Context.readonly_mode() or aws_session_info.get('readonly_mode', False):
+        raise ClientError('Server is in read-only mode')
+
+    # Get properties from validated token only
+    if properties_token not in _properties_store:
```
KaramVSingh, line 406: I think this would also be removed if we remove the IAC Gen tool - i think this also creates a bit of a confusing relationship between those two tools - can you expand more on the thinking here in comments if we keep the IAC tool?

src/ccapi-mcp-server/awslabs/ccapi_mcp_server/server.py
```diff
@@ -0,0 +1,924 @@
...
+    except Exception as e:
+        raise handle_aws_api_error(e)
+
+    return progress_event(response['ProgressEvent'], response.get('HooksProgressEvent', None))
+
+
+# This function is now imported from infrastructure_generator.py
+
+
+@mcp.tool()
```
KaramVSingh, line 546: 10/10 would love this in the cfn-mcp-server

docs/servers/ccapi-mcp-server.md
novekm: Yeah I used the cfn one as a reference. I'm open to either, but I assumed the abbreviation of it was ccapi

src/ccapi-mcp-server/CHANGELOG.md
novekm: removed - will inclue in next pr in a few min

src/ccapi-mcp-server/awslabs/ccapi_mcp_server/server.py
novekm: Nah I still have this in there - basically I've noticed even without checkov, if you have valid aws creds when you use any of the tools it'll make more secure policies by default. It'll sub in your account id and other related info rather than keeping a wildcard. Flow is:
- check env vars
- check aws creds
- generate code
- (later PR) explain generated code - basically an execution plan usually
- (later PR ) run checkov
- create (or update or delete)

novekm: the last part isn't relevant anymore though, it does offer troubleshooting but I took out the IaC related stuff

novekm: Or did you mean the secion about preferred IaC?

src/ccapi-mcp-server/awslabs/ccapi_mcp_server/README.md
novekm: For some reason I can't find this in my readme - gonna push a new commit soon, maybe it got removed at some point

src/ccapi-mcp-server/awslabs/ccapi_mcp_server/server.py
novekm: nah its just general instructions, no dedicated tool to handle it for them. It's just a normal response as part of the tool that gets aws creds if it fails

src/ccapi-mcp-server/awslabs/ccapi_mcp_server/server.py
novekm: dang it Q! 😅 lol I guess it missed some spots

src/ccapi-mcp-server/awslabs/ccapi_mcp_server/server.py
novekm: do you mean the `generate_infrastructure_code()` tool? If so, that one is pretty core - it's most relevant for Checkov, bur I use it to generate the CFN template for security scanning. For this PR it's not fully necessary, but it is   in PR 2 and 3 which are ready to go after this one. PR 2 uses it to generate template, then use a new tool called `explain()` to essentially give an execution plan of whats going to happen. I use tokens generated in the tools to create implicit dependencies that force the ordering the LLM can use the tools in

novekm:
full version: https://github.com/awslabs/mcp/pull/797

PR #775, opened 2025-07-09
fix(finch-mcp-server): fix --enable-aws-resource-write flag not worki…

Senthamil2003:
### Changes
Fixed the `--enable-aws-resource-write` flag not working when the finch-mcp-server package is run via uvx entry point. The issue was that argument parsing was only happening in the `if __name__ == '__main__':` block, which is bypassed when called via the package entry point defined in `pyproject.toml`.

The fix moves the argument parsing logic inside the `main()` function so it works both when run directly (`python -m awslabs.finch_mcp_server.server`) and via the package entry point (`uvx awslabs.finch-mcp-server@latest`).

### User experience

**Before:**
```bash
# This would ignore the flag and always run in read-only mode
uvx awslabs.finch-mcp-server@latest --enable-aws-resource-write
# Log: AWS resource write enabled: False 
```

**After:**
```bash
# This now correctly enables AWS resource write mode
uvx awslabs.finch-mcp-server@latest --enable-aws-resource-write
# Log: AWS resource write enabled: True 
```

Users can now use ECR push and repository creation tools when running via uvx with the `--enable-aws-resource-write` flag.

## Checklist

* [x] I have reviewed the [[contributing guidelines](https://github.com/awslabs/mcp/blob/main/CONTRIBUTING.md)](https://github.com/awslabs/mcp/blob/main/CONTRIBUTING.md)
* [x] Changes are documented

**Is this a breaking change?** No (N)

**RFC issue number**: N/A (Bug fix)

**Migration process documented**: N/A (No breaking changes)

**Implement warnings**: N/A (No breaking changes)

## Acknowledgment


scottschreckengaust (changes requested):
Waiting for code owners review

Shubhranshu153 (dismissed):

Senthamil2003:
Hi @Shubhranshu153 , I’ve updated the test coverage as part of the recent changes. When you have a moment, could you please review and share any feedback?

src/finch-mcp-server/awslabs/finch_mcp_server/server.py
```diff
@@ -495,31 +495,8 @@ async def finch_create_ecr_repo(
         return Result(**error_result)
 
 
-def main(enable_aws_resource_write: bool = False):
```
scottschreckengaust, line 498: Are there any documentation changes required for this?

Senthamil2003: No, we don't need to change any documentation; the existing one will work fine

Senthamil2003:
Hi @scottschreckengaust  @Shubhranshu153 @pendo324 any updates on this one??

PR #802, opened 2025-07-14
feat(postgres-mcp-server): IAM authentication support

SarvagyaShastri:
Fixes

## Summary
This change introduces a new MCP tool for RDS/Aurora postgres. The new tool allows to run SQL queries using IAM Database Authentication (IAM DB Auth). IAM DB Auth generates a Sigv4 token using AWS credentials and uses that token as password to the database. The token used is valid only for 15 minutes which is big advantage in terms of security. 

### Changes

- Added a new dependency on psycopg (postgres client)
- Added new auth mode, along side of the existing secrets manager auth 
- The DB host name, port, region, etc. are configurable in the MCP server config
- Added unit tests

The changes have been tested by adding the MCP server locally in Cline VSCode. 

### User experience

Prior to this change, the MCP server only allowed Data API based connections. After this change, the MCP server will allow to connect via IAM DB Auth.

## Checklist

* [Y] Changes are documented

Is this a breaking change? N

**RFC issue number**:

Checklist:

## Acknowledgment

### DIFF
--- a/src/postgres-mcp-server/awslabs/postgres_mcp_server/server.py
+++ b/src/postgres-mcp-server/awslabs/postgres_mcp_server/server.py
@@ -50,25 +50,50 @@ class DummyCtx:
 
 
 class DBConnection:
-    """Class that wraps DB connection client by RDS API."""
+    """Class that wraps DB connection client by RDS API or direct PostgreSQL connection."""
 
-    def __init__(self, cluster_arn, secret_arn, database, region, readonly, is_test=False):
+    def __init__(
+        self,
+        auth_mode='secrets-manager',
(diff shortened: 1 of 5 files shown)


src/postgres-mcp-server/awslabs/postgres_mcp_server/server.py
```diff
@@ -50,25 +50,50 @@ async def error(self, message):
 
 
 class DBConnection:
```
kennthhz, line 52: Recent merge of https://github.com/awslabs/mcp/pull/741 has redesigned the connection area class and model. Can you merge your change with it?

PR #810, opened 2025-07-14
docs: clarify the correct server run command

maratgaliev:
Following current documentation I'm getting:
error: Failed to spawn: `server.py`
  Caused by: No such file or directory (os error 2)

Fixes

### Changes

### User experience

## Checklist

* [ ] Changes are documented

**RFC issue number**:

Checklist:

## Acknowledgment


src/aws-healthomics-mcp-server/README.md
```diff
@@ -184,7 +184,7 @@ Or install from source:
 git clone <repository-url>
 cd mcp/src/aws-healthomics-mcp-server
 uv sync
-uv run server.py
+python -m awslabs.aws_healthomics_mcp_server.server
```
markjschreiber: Thanks for spotting this. Keeping with using `uv` the command `uv run -m awslabs.aws_healthomics_mcp_server.server` would be preferred over `python`. Let me know if that works in your environment.

markjschreiber (changes requested):
Thanks for the contribution and for spotting this mistake! I have requested that the command us `uv` rather than python directly but otherwise looks good.

src/aws-healthomics-mcp-server/README.md
maratgaliev: Fixed, works for me

Issue #872, opened 2025-07-18
aws-api-mcp-server: default workding directory


iddv:
Currently, the AWS API MCP server requires users to explicitly configure the `AWS_API_MCP_WORKING_DIR` environment variable, creating unnecessary setup friction. Furthermore, without this config, the server fails to start up, which causes confusion (hence the bug label).

The server already has a robust mechanism for determining platform-appropriate temporary directories for logging (as implemented in the `_get_log_directory()` function), which handles Windows, macOS, and Linux systems appropriately. 

I propose extending this approach to use the same temp directory logic as the default working directory when `AWS_API_MCP_WORKING_DIR` is not specified, eliminating the need for users to manually configure this environment variable while maintaining the same cross-platform compatibility. This would reduce the barrier to entry while preserving the ability for users to override the working directory when needed for specific use cases.

rshevchuk-git:
Addressed in #873

PR #873, opened 2025-07-18
feat(aws-api-mcp-server): Make AWS_API_MCP_WORKING_DIR not required and change model caching logic

rshevchuk-git:
### Changes

1. Made AWS_API_MCP_WORKING_DIR not required, and set the default.
2. Changed model cache directory to be set to the parent folder of all other server files
3. Switched to local_files_only in SentenceTransformer initialization to not make any API calls if model is present locally.

### User experience
1. Customers no longer have to specify AWS_API_MCP_WORKING_DIR in order to start the server.
2. Less dependency on network connection

## Testing
1. Tested that not passing AWS_API_MCP_WORKING_DIR allows server to start up successfully.
2. Verified that absence of cache_dir/models/model_name folder triggers model download, while if it exists not API calls are made and local version is used
3. uv build, and verified that resulting distribution does not contain cached model and is <100 MB
4. Tested server tools work with Cline & Cursor

## Checklist

* [x] Changes are documented

**RFC issue number**:

Checklist:

## Acknowledgment


iddv:
Looks like this resolves https://github.com/awslabs/mcp/issues/872

iddv (approved):

PR #915, opened 2025-07-23
chore(aws-api-mcp-server): Add tool annotations

arnewouters:
Fixes

### Changes

Add [tool annotations](https://modelcontextprotocol.io/docs/concepts/tools#tool-annotations) to the `suggest_aws_commands` and `call_aws` tools.

### User experience

## Checklist

* [ ] Changes are documented

**RFC issue number**:

Checklist:

## Acknowledgment

### DIFF
--- a/src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
+++ b/src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
@@ -41,6 +41,7 @@ from .core.metadata.read_only_operations_list import ReadOnlyOperations, get_rea
 from botocore.exceptions import NoCredentialsError
 from loguru import logger
 from mcp.server.fastmcp import Context, FastMCP
+from mcp.types import ToolAnnotations
 from pydantic import Field
 from typing import Annotated, Any, Optional, cast
 
@@ -111,6 +112,9 @@ READ_OPERATIONS_INDEX: Optional[ReadOnlyOperations] = None
         - Required parameters
         - Description of what the command does
     """,
+    annotations=ToolAnnotations(
+        title='Suggest AWS CLI commands', readOnlyHint=True, openWorldHint=False
+    ),
 )
 async def suggest_aws_commands(
     query: Annotated[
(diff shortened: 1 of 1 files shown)


src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
```diff
@@ -167,6 +171,12 @@ async def suggest_aws_commands(
     Returns:
         CLI execution results with API response data or error message
     """,
+    annotations=ToolAnnotations(
+        title='Execute AWS CLI commands',
+        readOnlyHint=READ_OPERATIONS_ONLY_MODE,
+        destructiveHint=not READ_OPERATIONS_ONLY_MODE,
```
iddv, line 177: Wondering, is this not inaccurate considering the readonly flag does not pertain to file system? Even though it's not supported yet, we might actually need 'READ_OPERATIONS_ONLY_MODE and READ_ONLY_FS'

arnewouters: Do we have operations that modify the file system and that would work under `READ_OPERATIONS_ONLY_MODE`? I thought we classified those operations so it's not the case?

arnewouters: For customizations we only allow these operations when read only mode is enabled: https://github.com/awslabs/mcp/blob/main/src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/metadata/read_only_operations_list.py#L102

I guess `--outfile` could also modify the file system but we don't support that yet and this should not be classified as a read only operation

iddv: You can check this, but I think an example is S3 operations that technically write to disk - we updated our readme specifically for this reason to say:

> READ_OPERATIONS_ONLY_MODE: When set to "true", restricts execution to read-only operations only.

PCManticore (approved):

PR #926, opened 2025-07-24
feat(aws-api-mcp-server): Add consent mechanism using elicitation

arnewouters:
Fixes

### Changes

Use the elicitation mechanism to request consent from the user before executing any commands that are not classified as read-only. This feature can be toggled off/on using the `BYPASS_TOOL_CONSENT` flag, which is on by default.

### User experience

Tested this on the latest VS Code Insider build with Copilot.

Example where we reject the command:

https://github.com/user-attachments/assets/803b51de-1dde-4f48-b3f2-509c7a65a3bc

Example where we accept the command:

https://github.com/user-attachments/assets/2e9de8bc-25dd-4b9c-b53c-750b6c699f4b

## Checklist

* [ ] Changes are documented

**RFC issue number**:

Checklist:

## Acknowledgment


src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
```diff
@@ -202,17 +206,39 @@ async def call_aws(
...
+                    'Execution of this operation is not allowed because read only mode is enabled. '
+                    f'It can be disabled by setting the {READ_ONLY_KEY} environment variable to False.'
+                )
+                await ctx.error(error_message)
+                return AwsApiMcpServerErrorResponse(
+                    detail=error_message,
+                )
+            elif not BYPASS_TOOL_CONSENT:
+                try:
+                    elicitation_result = await ctx.elicit(
```
PCManticore: What's the behaviour if the Client doesn't support elicitation yet?

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
```diff
@@ -202,17 +206,39 @@ async def call_aws(
...
+                        or not elicitation_result.data.answer
+                    ):
+                        error_message = 'User rejected the execution of the command.'
+                        await ctx.error(error_message)
+                        return AwsApiMcpServerErrorResponse(
+                            detail=error_message,
+                        )
+                except McpError as e:
+                    if e.error.code == METHOD_NOT_FOUND:
+                        error_message = 'Client does not support elicitation. Use a different client or update the server configuration.'
```
PCManticore: Should we throw an error if the client doesn't support elicitation? That can lead to a bad user experience.

src/aws-api-mcp-server/tests/test_server.py
```diff
@@ -114,19 +116,20 @@ async def test_suggest_aws_commands_exception(mock_knowledge_base):
...
 
 @patch('awslabs.aws_api_mcp_server.server.DEFAULT_REGION', 'us-east-1')
+@patch('awslabs.aws_api_mcp_server.server.BYPASS_TOOL_CONSENT', False)
 @patch('awslabs.aws_api_mcp_server.server.get_local_credentials')
 @patch('awslabs.aws_api_mcp_server.server.interpret_command')
 @patch('awslabs.aws_api_mcp_server.server.validate')
 @patch('awslabs.aws_api_mcp_server.server.translate_cli_to_ir')
 @patch('awslabs.aws_api_mcp_server.server.is_operation_read_only')
-async def test_call_aws_with_mutating_action(
+async def test_call_aws_with_elicitation(
```
PCManticore: We need the counter-part of this test where BYPASS_TOOL_CONSENT is True.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
arnewouters: It throws an error, which we are catching below

```
  File "/Volumes/workplace/mcp-dev/src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py", line 210, in call_aws
    elicitation_result = await ctx.elicit(
                         ^^^^^^^^^^^^^^^^^
  File "/Volumes/workplace/mcp-dev/src/aws-api-mcp-server/.venv/lib/python3.12/site-packages/mcp/server/fastmcp/server.py", line 1098, in elicit
    return await elicit_with_validation(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Volumes/workplace/mcp-dev/src/aws-api-mcp-server/.venv/lib/python3.12/site-packages/mcp/server/elicitation.py", line 95, in elicit_with_validation
    result = await session.elicit(
             ^^^^^^^^^^^^^^^^^^^^^
  File "/Volumes/workplace/mcp-dev/src/aws-api-mcp-server/.venv/lib/python3.12/site-packages/mcp/server/session.py", line 273, in elicit
    return await self.send_request(
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Volumes/workplace/mcp-dev/src/aws-api-mcp-server/.venv/lib/python3.12/site-packages/mcp/shared/session.py", line 286, in send_request
    raise McpError(response_or_error.error)
mcp.shared.exceptions.McpError: Method not found
```

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
arnewouters: I opted for throwing an error because this feature is off by default. I think it might be bad if a user turns this on explicitly and we start executing mutating operations because the client doesn't support elicitation.

PCManticore: Sounds good, let's confirm the experience with a client that doesn't have the support for elicitation so we can make a call on how it would surface the customers from a UX perspective.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
```diff
@@ -202,17 +206,39 @@ async def call_aws(
...
+                        logger.error(error_message)
+                        raise AwsApiMcpError(error_message)
+
+                    raise e
```
rshevchuk-git: can we move this logic to a separate function somewhere under service.py maybe? Would be great to keep call_aws definition as readable as possible

src/aws-api-mcp-server/README.md
```diff
@@ -107,14 +107,15 @@ For detailed instructions on setting up your local development environment and r
...
+| `BYPASS_TOOL_CONSENT`                                             | ❌ No     | `"true"`                                                 | When set to "false", the MCP server will ask explicit consent before executing any operations that are **NOT** read-only. This safety mechanism uses [elicitation](https://modelcontextprotocol.io/docs/concepts/elicitation) so it requires a [client that supports elicitation](https://modelcontextprotocol.io/clients).                                                                                                                                                                                                                                                                                             |
```
rshevchuk-git: I think we should have a better name here, now it seems like any request will require consent if I turn this off, while in fact this only applies to mutations

arnewouters: I landed on `REQUIRE_MUTATION_CONSENT` but open to other suggestions as well

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
arnewouters: The error is shown to the LLM, it might just stop but it can also start using bash tools if available to work around this guardrail. It's not perfect but not sure what else we could do.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
```diff
@@ -228,6 +217,19 @@ async def call_aws(
         )
 
     try:
+        if READ_OPERATIONS_INDEX is None or not is_operation_read_only(ir, READ_OPERATIONS_INDEX):
```
rshevchuk-git, line 220: Why do we need 'if READ_OPERATIONS_INDEX is None ' ?

arnewouters: [`READ_OPERATIONS_INDEX`](https://github.com/arnewouters/mcp/blob/f20437a4b09deac7850862042406200afad17485/src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py#L304-L305) is not always set, maybe initialization can fail. We don't know if an operation is read-only or not without it so we don't execute any command when `READ_OPERATIONS_ONLY_MODE` is set. Extra safety mechanism, I guess

rshevchuk-git (dismissed):

PCManticore (approved):

rshevchuk-git (approved):

PR #935, opened 2025-07-24
feat: implement converting column names to snake_case

okhomin:
## Summary

Column names from CSV and Parquet will be converted to snake_case by default.

### Changes

> Update description tools to include case sensitivity instructions
> Add conversion of column names from upper case to snake case

### User experience

> When importing a file, if the columns has upper case letters, they will be converted to snake case, since Athena doesn't support upper case letters. Customers still can create columns using upper case, and query them using the MCP but not Athena.

## Checklist

* [X] Changes are documented

Is this a breaking change? No

## Acknowledgment


src/s3-tables-mcp-server/awslabs/s3_tables_mcp_server/file_processor/utils.py
```diff
@@ -0,0 +1,211 @@
...
+import pyarrow as pa
+import re
+from ..utils import get_s3_client, pyiceberg_load_catalog
+from io import BytesIO
+from pyiceberg.exceptions import NoSuchTableError
+from typing import Any, Callable, Dict
+from urllib.parse import urlparse
+
+
+def to_snake_case(name: str) -> str:
```
ananthaksr: Why not use a library like https://docs.pydantic.dev/2.0/api/alias_generators/ ?

src/s3-tables-mcp-server/awslabs/s3_tables_mcp_server/file_processor/utils.py
```diff
@@ -0,0 +1,211 @@
...
+    for i, field in enumerate(schema):
+        new_field = pa.field(
+            converted_names[i], field.type, nullable=field.nullable, metadata=field.metadata
+        )
+        new_fields.append(new_field)
+
+    return pa.schema(new_fields, metadata=schema.metadata)
+
+
+async def import_file_to_table(
```
ananthaksr, line 67: This looks like a different code refactor unrelated to case conversion, would be good to decouple.

okhomin: It is related, since I was adding the same validation to each function and I decided to deduplicate the code in both functions.

ananthaksr: Understood, in that it would be better to first refactor and then separately add case conversion support. Not a blocker.

ananthaksr (approved):

src/s3-tables-mcp-server/awslabs/s3_tables_mcp_server/file_processor/utils.py
```diff
@@ -0,0 +1,157 @@
...
+    # Extract original column names
+    original_names = schema.names
+
+    # Convert each column name to snake_case
+    converted_names = [to_snake(name) for name in original_names]
+
+    # Check for duplicates after conversion using set and len
+    if len(set(converted_names)) != len(converted_names):
+        raise ValueError(
+            f'Duplicate column names after case conversion. '
```
gregorywright, line 52: nit, extra space at end of string.

gregorywright (commented):
Can you also please provide some examples of how column names are being converted? I can't really follow the test code to see how different strings are converted.

okhomin:
> Can you also please provide some examples of how column names are being converted? I can't really follow the test code to see how different strings are converted.

@gregorywright 
Examples:
```
First Column -> first column
Second    Column -> second    column
ThirdCOLUMN -> third_column
Fourth!@#$Column -> fourth!@#$column
```

src/s3-tables-mcp-server/awslabs/s3_tables_mcp_server/file_processor/utils.py
okhomin: This space is needed because the next line is also a string and in Python they will be concatenated.
```
    print('Hello'
          'World')
```
Will print `HelloWorld`.

gregorywright (approved):

Issue #928, opened 2025-07-24
redshift-mcp-server: list_clusters not working :: asking for region parameter on Q CLI


HarshGrandeur:
### Describe the bug

I added the redshift-mcp-server to my mcp server inside Q CLI. But asking for available Redshift clusters, Q CLI is requiring for region but the method does not take a region parameter.

<img width="1407" height="270" alt="Image" src="https://github.com/user-attachments/assets/d43dc444-55b5-4f1c-9df8-cd93bb382c83" />

### Expected Behavior

The agent when asked for a list of Redshift cluster, should reply with available redshift clusters in aws account.

### Current Behavior

The agent asks for region to be passed as a parameter when the method `list_cluster` does not take any parameters. It errors out on creating a redshift client.

### Reproduction Steps

1. Set up Q CLI
2. Set up MCP server for Redshift. Added the redshift mcp server 
```{
  "mcpServers": {
    "awslabs.redshift-mcp-server": {
      "command": "uvx",
      "args": ["awslabs.redshift-mcp-server@latest"],
      "env": {
        "AWS_PROFILE": "default",
        "AWS_REGION": "us-east-1",
        "FASTMCP_LOG_LEVEL": "INFO"
      },
      "disabled": false,
      "autoApprove": []
    }
  }
}
```
3. Set up env variables. This included setting up `AWS_PROFILE` inside ~/.aws/config 
```
[profile default]
region=us-east-1
```
and corresponding credentials in `~/.aws/credentials` file.
4. Start Q CLI. Ask the agent about available redshift clusters.

### Possible Solution

Pass the `region` parameter specifically to boto3 client.

### OS

MacOS

### Server

other

### Region experiencing the issue

 us-east-1

### Service quota

- [x] I have reviewed the service quotas for this construct

grayhemp:
@HarshGrandeur , I can not repro the issue using the steps you provided.

```sh
% env | grep -i aws_profile

% env | grep -i aws_region 

% cat ~/.aws/config 
[default]
account = **********
region = us-east-1
role = Admin

% cat ~/.aws/credentials
[default]
aws_access_key_id=***********
aws_secret_access_key=**********

% cat ~/.aws/amazonq/mcp.json
{
  "mcpServers": {
    "awslabs.redshift-mcp-server": {
      "command": "uvx",
      "args": ["awslabs.redshift-mcp-server@latest"],
      "env": {
        "AWS_PROFILE": "default",
        "AWS_REGION": "us-east-1",
[...]

PR #944, opened 2025-07-25
fix(redshift_mcp_server): Fix missing region parameter while creation of boto3 client

HarshGrandeur:
…for issues - https://github.com/awslabs/mcp/issues/928

Fixes

## Summary

Fix missing region parameter in boto redshift client creation. This resulted in the error `Error creating Redshift client: You must specify a region.`

### Changes

Added the missing region inside the boto redshift client.

### User experience

> Before

The user was not able to connect to Redshift mcp server due to incorrect configuration of boto redshift client.

> After

The user is able to get the results for the available redshift clusters and related queries which are present inside redshift mcp server.

#### Reference
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/core/session.html#boto3.session.Session.client:~:text=config%20(botocore.client.Config)%20%E2%80%93%20Advanced,botocore%20config%20documentation%20for%20more%20details.

## Checklist

* [x ] I have reviewed the [contributing guidelines](https://github.com/awslabs/mcp/blob/main/CONTRIBUTING.md)
* [ x] I have performed a self-review of this change
* [ x] Changes have been tested
* [ x] Changes are documented

Is this a breaking change? N

**RFC issue number**:

Checklist:

## Acknowledgment


HarshGrandeur:
Fixed failing tests

grayhemp (changes requested):
@HarshGrandeur could you please provide a proper repro in the issue #928 description? What is currently in there is very vague.

Please include details about:

- Steps leading to the error
- AWS client configuration
- Environment variables.

Thank you.

HarshGrandeur:
> @HarshGrandeur could you please provide a proper repro in the issue #928 description? What is currently in there is very vague.
> 
> Please include details about:
> 
> * Steps leading to the error
> * AWS client configuration
> * Environment variables.
> 
> Thank you.

@grayhemp  I have updated the issue. Please let me know if this looks better. I have added more details in the reproduction steps.

HarshGrandeur:
Closing this for now.

grayhemp:
@HarshGrandeur may I ask why you closed it?

grayhemp:
@HarshGrandeur, BTW, your repro in #928 doesn't seem to work. See https://github.com/awslabs/mcp/issues/928#issuecomment-3134843151.

HarshGrandeur:
reopening this based on : https://github.com/awslabs/mcp/issues/928#issuecomment-3176047417

HarshGrandeur:
@grayhemp Do you need any changes ?

grayhemp:
I might need some. I'm doing research on the subject and found out that  we might need to redesign how we get the clients due to some peculiar behavior of profile configuration precedence over the env variable. Still need  to confirm something,  standby...

grayhemp:
# RCA

Based on my RCA of the boto3 behavior with region specified, both the current implementation and the solution provided in this PR lead to unexpected behavior.

## Current implementation

When `AWS_PROFILE` is set, creates a session/client with the profile but no region argument.

- With no `AWS_REGION` set
  - [X] Region is in `~/.aws/config` - works as expected picking the region from the config
  - [ ] No region in `~/.aws/config` - errors with `Error creating Redshift client: You must specify a region`, the default `us-east-1` region is expected to be used
- With `AWS_REGION` set
  - [ ] Region is in `~/.aws/config` - ignores `AWS_REGION` using the region specified in the config, though `AWS_REGION` is expected to be used
  - [ ] No region in `~/.aws/config` - ignores `AWS_REGION` and errors with `Error creating Redshift client: You must specify a region`, though `AWS_REGION` is expected to be used

## @HarshGrandeur's fix

When `AWS_PROFILE` is set, creates a session/client with the profile and the region argument either as in `AWS_REGION` if the env variable is set or default `us-east-1`:

- With no `AWS_REGION` set
  - [ ] Region is in `~/.aws/config` - uses the default `us-east-1` instead of the one in `~/.aws/config`, expected to use the region set in `~/.aws/config`
  - [X] No region in `~/.aws/config` - uses the default `us-east-1` region
- With `AWS_REGION` set
  - [X] Region is in `~/.aws/config` - uses `AWS_REGION` instead of the one in `~/.aws/config`
  - [X] No region in `~/.aws/config` - uses the one set in `AWS_REGION`

Better, but still one miss.

## The problem

The problem appeared to be in boto3 not supprting `AWS_REGION` as an env variable (🤦 RTFM), see <https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html>. What it does support is `AWS_DEFAULT_REGION`.

If we use `AWS_DEFAULT_REGION` instead of `AWS_REGION`, then without modifying the current implementation (main branch) we will have the following.

When `AWS_PROFILE` is set, creates a session/client with the profile but no region argument.

- With no `AWS_DEFAULT_REGION` set
  - [X] Region is in `~/.aws/config` - works as expected picking the region from the config
  - [ ] (unclear if it's a miss, see below) No region in `~/.aws/config` - errors with `Error creating Redshift client: You must specify a region`
    - Shall it use the default `us-east-1` region?
    - Or we must require either `AWS_DEFAULT_REGION` or `AWS_REGION` or the region in `~/.aws/config` to be specified?
- With `AWS_DEFAULT_REGION` set
  - [X] Region is in `~/.aws/config` - uses `AWS_DEFAULT_REGION` instead of the one in `~/.aws/config`
  - [X] No region in `~/.aws/config` - uses the one in `AWS_DEFAULT_REGION`

In this case we have the following to address.

The one unclear behavior item (With no `AWS_DEFAULT_REGION` set, No region in `~/.aws/config`). Similarly to the note under the `AWS_DEFAULT_REGION` description in the AWS CLI doc <https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html>, it would probably be the best to expect either `AWS_DEFAULT_REGION` or `AWS_REGION` or the region in `~/.aws/config`, and let it error "officially" otherwise. We need to update the prompts with instructions to print out errors and provide suggestions on how to deal with them. That way the behavior will be compatible with AWS CLI which is the defacto standard setting expectations.

What to do with `AWS_REGION`? It can be treated similarly to the `AWS_REGION` in the AWS CLI tool (see the same doc page), which overrides the `AWS_DEFAULT_REGION` and the rest of the configuration if specified. We will need to be explicitly setting it as the session/client parameter if it is specified. There should be no default value for it.

Just a reminder, AWS CLI follows this order or precedence:

1. AWS_REGION (highest priority)
2. AWS_DEFAULT_REGION
3. Profile config region
4. Default region (if any).

We also need to carefully document the behavior including the precedence in both README.md and the prompts, as well as cover it all with tests.

What do you think? @HarshGrandeur @harishcodes

grayhemp:
So, the action items, if you folks aligned with the RCA/suggestion:

Code changes:

- Modify clients' creation to handle AWS_REGION explicitly as top precedence override
  - In the client creation methods
    - If `self.aws_region` is set then pass it to the boto3's `client()`
    - If not, let boto3 handle AWS_DEFAULT_REGION + profile config automatically, no code needed
- Remove hardcoded us-east-1 default - let it error if no region is configured
- Update instructions/prompts with region configuration guidance and error handling notes
- Add test cases covering all region configuration scenarios including the error case

Doc changes:

- Document region precedence order under the "AWS Client Requirements" section (as specified in my previous message)
- Add `AWS_DEFAULT_REGION` and remove the default value from `AWS_REGION` in the "Environment Variables" section

grayhemp:
Okay, I implemented the above and created PR #1174. Will close this one if no objections.

harishcodes:
Hi @grayhemp . Sorry for the delay. This looks good and comprehensive.. Thank you!!

grayhemp:
Merged https://github.com/awslabs/mcp/pull/1174. Let's close this one.

PR #967, opened 2025-07-29
feat: release skipping and bump fixing

### LINKED ISSUE #966: Purposeful Skipping Packages Upon Release
### Describe the feature

Skip packages during a release

### Use Case

When one package wants to release and another author has merged changes that does NOT want to release there is not a way.

### Proposed Solution

TBD

* build directive file`SKIPPED_PACKAGES` - list of all the packages to skip if found
* semaphore file `SKIP` - file within the package to signal skipping

### Other Information

It could be failing tests, or just feature not fully ready or releasable.

### Acknowledgements

scottschreckengaust:
Fixes #966 

## Summary

There may be times that packages should be skipped.

### Changes

Place a `DO_NOT_RELEASE` file in the package directory (`/src/<PACKAGE>/DO_NOT_RELEASE`) for the release to skip the package during the process.

### User experience

## Checklist

* [ ] Changes are documented

## Acknowledgment

### DIFF
--- a/.github/workflows/release-merge-tag.yml
+++ b/.github/workflows/release-merge-tag.yml
@@ -153,11 +153,28 @@ jobs:
           git push origin "$TAG"
 
           # Avoid checking too fast
-          echo "::group::Was failing when immediately checking for remote tag existence"
-          git ls-remote --tags origin
-          for i in {1..60}; do git ls-remote --tags origin | grep -q "refs/tags/$TAG" && break || echo "sleeping..." && sleep 1; done
-          git ls-remote --tags origin
-          echo "::endgroup::"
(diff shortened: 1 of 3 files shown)


scottschreckengaust:
The closure from a release properly tags, but the workflow doesn't detect the remote tag. Continuation of debugging. This does not block the process but tries to identify why the tag is not fetchable from the remote (even though it is there)

scottschreckengaust:
Also, only fails bumping when the Major is at or above the maximum value of `sys.maxsize`

krokoko (approved):

alexa-perlov (approved):

PR #1000, opened 2025-07-31
feat: add find_column tool for column discovery

hameddgh:
# Add find_column tool to Redshift MCP Server

### Changes

This PR adds a new `find_column` tool to the Redshift MCP Server that allows users to search for columns by name pattern across all schemas and tables in a Redshift database. This feature helps users discover where specific data elements are stored across large databases, making data exploration and schema understanding much easier.

The implementation includes:
- New SQL query constant `SVV_COLUMNS_PATTERN_QUERY` in `consts.py`
- New data model `ColumnMatch` in `models.py`
- New function `find_columns()` in `redshift.py`
- New MCP tool `find_column_tool()` in `server.py`

### User experience

**Before:**
Users had to manually navigate through schemas and tables to find columns with specific naming patterns. This was time-consuming and inefficient, especially in large databases with many tables.

**After:**
Users can now use the `find_column` tool to quickly search across all tables in a database for columns matching a specific pattern. For example:

```python
# Find all columns containing "customer" in their name
columns = await find_column_tool(
    ctx=ctx,
    cluster_identifier="my-redshift-cluster",
    database_name="sales_db",
    pattern="customer"
)

# Results include schema name, table name, column name, and data type
# for all matching columns across the entire database
```

This significantly improves the database exploration experience and helps users quickly locate relevant data.

## Checklist

* [x] Changes are documented

Is this a breaking change? N

**RFC issue number**: N/A

Checklist:

* [x] Migration process documented (N/A - new feature)
* [x] Implement warnings (if it can live side by side) (N/A - new feature)

## Acknowledgment

### DIFF
--- a/src/redshift-mcp-server/awslabs/redshift_mcp_server/consts.py
+++ b/src/redshift-mcp-server/awslabs/redshift_mcp_server/consts.py
@@ -113,6 +113,17 @@ WHERE database_name = {} AND schema_name = {} AND table_name = {}
 ORDER BY ordinal_position;
 """
 
+SVV_COLUMNS_PATTERN_QUERY = """
+SELECT
+    table_schema,
+    table_name,
+    column_name,
+    data_type
+FROM SVV_COLUMNS
+WHERE column_name ILIKE {}
+ORDER BY table_schema, table_name, column_name
+"""
+
 # SQL guardrails
 
 # Single-lines comments.
--- a/src/redshift-mcp-server/awslabs/redshift_mcp_server/models.py
+++ b/src/redshift-mcp-server/awslabs/redshift_mcp_server/models.py
@@ -129,6 +129,15 @@ class RedshiftColumn(BaseModel):
     remarks: Optional[str] = Field(None, description='Remarks about the column')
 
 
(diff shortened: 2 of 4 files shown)


grayhemp:
UPD. Working with @hameddgh in a private chat on the triage.

grayhemp:
My latest reasoning from the discussion in DM:

- find_columns might easily end up with adding millions of tokens to the context bloating it out
  - Based on my every day experience a lot of customers have a lot of DB objects in their databases
- The existing discovery tools allow to discover the same as find_columns but gradually in smaller reasoning backed portions
  - Existing discovery - clusters, databases for a cluster, schemas for a database, tables for a schema, columns for a table
- Also, column name pattern check might simply not be enough as they often imply the context brought by DB, schema, table names as well as the remarks on those objects
  - The existing discovery considers it all, unlike find_columns
- So If I will see a proof that the find_columns works better on small schema databases than the existing tools, then we will need to think of a mixed solution to improve the existing one
- Otherwise, I would stick to the existing one.

Closing since it's been >1month without an answer.

PR #1004, opened 2025-08-01
fix(aws-api-mcp-server): Improve logging

arnewouters:
Fixes

### Changes

Improve logs by:
- logging the input when tools are invoked
- adding operation timer for customizations just like we have for boto3 operations
- logging embedding model startup for easier debugging

### User experience

## Checklist

* [x] Changes are documented

**RFC issue number**:

Checklist:

## Acknowledgment

### DIFF
--- a/src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
+++ b/src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
@@ -127,6 +127,7 @@ async def suggest_aws_commands(
     ctx: Context,
 ) -> dict[str, Any] | AwsApiMcpServerErrorResponse:
     """Suggest AWS CLI commands based on the provided query."""
+    logger.info('Suggesting AWS commands for query: {}', query)
     if not query.strip():
         error_message = 'Empty query provided'
         await ctx.error(error_message)
@@ -190,6 +191,7 @@ async def call_aws(
     ] = None,
 ) -> ProgramInterpretationResponse | AwsApiMcpServerErrorResponse | AwsCliAliasResponse:
     """Call AWS with the given CLI command and return the result as a dictionary."""
+    logger.info('Executing AWS CLI command: {}', cli_command)
     try:
         ir = translate_cli_to_ir(cli_command)
         ir_validation = validate(ir)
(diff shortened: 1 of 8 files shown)


src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
```diff
@@ -127,6 +127,7 @@ async def suggest_aws_commands(
     ctx: Context,
 ) -> dict[str, Any] | AwsApiMcpServerErrorResponse:
     """Suggest AWS CLI commands based on the provided query."""
+    logger.info('Suggesting AWS commands for query: {}', query)
```
rshevchuk-git, line 130: Should we also log results of suggest_aws_commands?

arnewouters: I think that's a good idea

rshevchuk-git (approved):

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
```diff
@@ -127,12 +127,18 @@ async def suggest_aws_commands(
...
```
bidesh, line 130: could we do without logging the query? That might contain sensitive info which might not always be desired. Is there already a way to control this?

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
```diff
@@ -190,6 +196,7 @@ async def call_aws(
     ] = None,
 ) -> ProgramInterpretationResponse | AwsApiMcpServerErrorResponse | AwsCliAliasResponse:
     """Call AWS with the given CLI command and return the result as a dictionary."""
+    logger.info('Executing AWS CLI command: {}', cli_command)
```
bidesh, line 199: cli_command can have sensitive info. Do customers have options to turn these logging off if they wanted to?

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
arnewouters: Currently, we don't support turning off the logging to file but you can change the log level on the client that starts the MCP server with the `FASTMCP_LOG_LEVEL` environment variable (default is set to `WARNING`).

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
arnewouters: See https://github.com/awslabs/mcp/pull/1004#discussion_r2254072877

bidesh (approved):

Issue #917, opened 2025-07-23
(redshift-mcp-server):  AWS Secrets Manager Integration for Redshift Credentials


psap2:
### Describe the feature

Add support for retrieving Redshift credentials from AWS Secrets Manager in the Redshift MCP server. This enhancement will allow secure management of database credentials and connections and support enterprise-scale deployments.

### Use Case

When connecting to Redshift clusters/workgroups:
1. The MCP server needs to authenticate using specific Redshift user credentials
2. These credentials are securely stored in AWS Secrets Manager
3. Example: Using 'RedshiftExampleHubble_user' secret for authentication to Redshift serverless workgroups

### Proposed Solution

Add AWS Secrets Manager integration to RedshiftClientManager:
   - Retrieve credentials from secret a secret
   - Add configuration for secret name/ARN

2. Enhance authentication flow in execute_statement:
   - Support SecretArn for Secrets Manager authentication
   - Maintain backwards compatibility with existing IAM authentication
   - Allow fallback mechanisms if secret retrieval fails

### Acknowledgements

- [x] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

grayhemp:
Discussing the solution with @psap2 in a private chat.

jsshojae:
While AWS Secrets Manager integration is a valid approach, there's a **more secure and cost-effective solution** already in development that addresses your use case.

## Current Status
[PR #1043](https://github.com/awslabs/mcp/pull/1043) implements `REDSHIFT_DB_USER` configuration support, which enables the exact functionality you need without requiring Secrets Manager.

## Configuration
After PR #1043 merges, add the database user to your MCP configuration:

```json
{
  "mcpServers": {
    "awslabs.redshift-mcp-server": {
      "command": "uvx",
      "args": ["awslabs.redshift-mcp-server@latest"],
      "env": {
        "AWS_PROFILE": "default",
        "AWS_REGION": "us-east-1",
        "REDSHIFT_DB_USER": "RedshiftExampleHubble_user",
        "FASTMCP_LOG_LEVEL": "INFO"
      },
      "disabled": false,
      "autoApprove": []
    }
  }
}
```

## Required IAM Policy
```json
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Action": [
        "redshift-data:ExecuteStatement",
        "redshift-data:BatchExecuteStatement",
        "redshift-data:DescribeStatement", 
[...]

PR #1043, opened 2025-08-07
feat(redshift): add ability to run queries as superuser

jsshojae:
## Summary

Adds `use_superuser` parameter and `REDSHIFT_DB_USER` environment variable support to execute functions, enabling flexible user context for queries including access to system tables restricted to superusers.

## Changes

- **New Parameter**: `use_superuser: bool = False` in `execute_query()` and `execute_statement()`
- **New Environment Variable**: `REDSHIFT_DB_USER` to specify database user
- **User Selection Logic**: `use_superuser=True` → `REDSHIFT_DB_USER` env var → default IAM user
- **Superuser Access**: Uses cluster's `master_username` for provisioned clusters when `use_superuser=True`
- **Documentation**: Complete docstring updates

## Use Case: System Table Access

Primary use case is troubleshooting via system tables requiring elevated privileges:
- `stv_*` tables - Cluster monitoring and active queries
- `svv_*` views - Performance analysis 
- `pg_user`, `pg_group` - User/group management
- `pg_stat_*` - Database statistics

## Implementation

- **Provisioned**: Uses `master_username` from cluster info when `use_superuser=True`
- **Serverless**: No change (no master username concept)
- **Fallback**: Gracefully falls back to regular user if superuser unavailable
- **Read-only**: Maintains existing transaction safety

## Testing

✅ All 48 tests pass, 94% coverage maintained  
✅ Backward compatible, no breaking changes

## Example

```python
# Regular user access
execute_query(cluster, db, "SELECT * FROM my_table")

# Superuser for system tables  
execute_query(cluster, db, "SELECT * FROM stv_inflight", use_superuser=True)
```

---

### DIFF
--- a/src/redshift-mcp-server/README.md
+++ b/src/redshift-mcp-server/README.md
@@ -74,6 +74,7 @@ or docker after a successful `docker build -t awslabs/redshift-mcp-server:latest
 
 - `AWS_REGION`: AWS region to use (default: `us-east-1`)
 - `AWS_PROFILE`: AWS profile to use (optional, uses default if not specified)
+- `REDSHIFT_DB_USER`: Default database user for query execution (optional, uses IAM user if not specified)
 - `FASTMCP_LOG_LEVEL`: Logging level (`DEBUG`, `INFO`, `WARNING`, `ERROR`)
 - `LOG_FILE`: Path to log file (optional, logs to stdout if not specified)
 
--- a/src/redshift-mcp-server/awslabs/redshift_mcp_server/redshift.py
+++ b/src/redshift-mcp-server/awslabs/redshift_mcp_server/redshift.py
@@ -157,7 +157,11 @@ def protect_sql(sql: str, allow_read_write: bool) -> list[str]:
 
 
 async def execute_statement(
-    cluster_identifier: str, database_name: str, sql: str, allow_read_write: bool = False
+    cluster_identifier: str,
+    database_name: str,
+    sql: str,
+    allow_read_write: bool = False,
+    use_superuser: bool = False,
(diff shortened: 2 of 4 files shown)


grayhemp:
It might be redesigned in a much simpler way, by optionally passing the user name as an argument to all the tools, so the user name could be controlled on the prompt level instead of the environment variables one.

The separately implemented master username fetching for provisioned clusters is an overkill. We are already extracting it with the cluster definition here https://github.com/awslabs/mcp/blob/main/src/redshift-mcp-server/awslabs/redshift_mcp_server/models.py#L35, so having the user name implemented as the tools' argument will automatically enable LLM clients to use the name from the cluster definition in these tools when requested.

Also, it might be worth combining this effort with #917.

jsshojae:
Thanks for the feedback! I have added detailed info under issue #917 (please see my comment there for how this PR addresses the Secrets Manager authentication use case).

Two technical considerations:

1. **Two-call pattern**: The proposed approach would require a two-call tool-use pattern where `list_clusters` must be called before elevated access operations to retrieve `master_username`. This adds a mandatory dependency but provides better separation of concerns. I'm comfortable with this trade-off if you agree.

2. **Environment variable retention**: While adding the `db_user` parameter to tools improves flexibility, retaining the environment variable serves a distinct purpose for enterprise deployments (as detailed in #917). It enables configuration-level user specification without requiring the LLM to explicitly pass authentication parameters on every call, which is particularly valuable for consistent user isolation in production environments.

Happy to implement either approach based on your preference. The current implementation prioritizes the enterprise use case, but I can refactor to the parameter-based approach if that better aligns with the project's design principles.

PR #1097, opened 2025-08-13
fix(tests): Properly patch version fallback for aws-documentation-mcp tests

JonLim:
Fixes

## Summary

There was a hard coded value assertion in a test verifying the functionality of fetching the version of the aws-documentation-mcp, which required updating when the version numbers were bumped.

### Changes

This change adds a proper patch for `__version__` in the aws-documentation-mcp library to remove any manual interventions required for tests in the future.

### User experience

User experience is unaffected by this change, this is purely for unit tests and CI/CD.

## Checklist

* [X] Changes are documented

**RFC issue number**:

Checklist:

## Acknowledgment


Lavoiedavidw (approved):
LGTM

src/aws-documentation-mcp-server/tests/test_server_utils.py
```diff
@@ -312,8 +312,9 @@ def test_version_from_metadata_success(self, mock_version):
         assert 'ModelContextProtocol/1.1.3' in server_utils.DEFAULT_USER_AGENT
 
     @patch('importlib.metadata.version')
+    @patch('awslabs.aws_documentation_mcp_server.__version__', '1.1.3')
```
scottschreckengaust: Understood, to unblock the release. Could `import awslabs.aws_documentation_mcp_server.__version__`  and using that in the `assert __version__ in server_utils.DEFAULT_USER_AGENT` etc. be a longer term solution?

JonLim: Great callout - thank you for the idea! We were discussing offline but I originally did this approach thinking it would be a faux pas to import the variable we were testing against, but importing it and using it as part of the assertion means that we can confirm it was used in the `DEFAULT_USER_AGENT`.

Thanks Scott!

PR #1103, opened 2025-08-13
feat(cloudwatch-mcp-server): update get_active_alarms to filter out autoscaling alarms by default

goranmod:
## Summary
This change updates the `get_active_alarms` tool from the CW MCP Server to allow (by default) filtering out of alarms with autoscaling actions. Such alarms can create unneeded noise, and with this change we are aligning the default behavior with the behavior in the CloudWatch Console.

### Changes

The change enhances the `get_active_alarms` MCP tool to add capability to filter out autoscaling alarms, and change the default behavior to filter out by default.

### User experience

Before the change:
- calls to `get_active_alarms` by default returns all alarms in `ALARM` state

After the change:
- calls to `get_active_alarms` by default return all CW Alarms in `ALARM` state that don't have an autoscaling action configured
- the tool supports an optional field to include autoscaling alarms

## Checklist

* [x] Changes are documented

N
While the behavior is updated with this change, this is not a breaking change considering the MCP server needs to be downloaded to the customer box where an LLM agent will start using it. There is extra context added to the tool such that the LLM agent understands what the new version of the tool does.

Checklist:

## Acknowledgment


src/cloudwatch-mcp-server/awslabs/cloudwatch_mcp_server/__init__.py
```diff
@@ -14,4 +14,4 @@
 
 """awslabs.cloudwatch-mcp-server"""
 
-MCP_SERVER_VERSION = '0.0.1'
+MCP_SERVER_VERSION = '0.1.1'
```
agiuliano, line 17: should this be 0.0.2 given the change?

dineshSajwan (dismissed):

src/cloudwatch-mcp-server/awslabs/cloudwatch_mcp_server/cloudwatch_alarms/tools.py
```diff
@@ -79,39 +190,81 @@ async def get_active_alarms(
...
+        include_autoscaling_alarms: Annotated[
+            bool | None,
+            Field(
+                description='Whether to include alarms with autoscaling actions. '
+                'Defaults to False to reduce noise from routine scaling events. '
+                'Set to True to include all alarms regardless of action types. '
+                'When False, alarms containing autoscaling action ARNs (arn:aws:autoscaling:*) '
+                'in AlarmActions, OKActions, or InsufficientDataActions will be filtered out.'
+            ),
+        ] = False,
```
agiuliano, line 202: should we default to false given the prev version is the same as current impl with default on true?

src/cloudwatch-mcp-server/awslabs/cloudwatch_mcp_server/__init__.py
goranmod: The last number is usually used for build increases and chores, for example. For example, fixing documentation in the whole AWSLabs repo and similar.

That's why I decided to increase the "minor" version considering that one is explicitly updated + it shows that we are adding a new feature.

src/cloudwatch-mcp-server/awslabs/cloudwatch_mcp_server/cloudwatch_alarms/tools.py
goranmod: The change in behavior here is intentional. With this MCP server, we are targeting LLM Agents that are directly controlled by humans, hence there is a human end-user. From an operations perspective, autoscaling alarms are less important that "custom", application level alarms. I believe the change in the default behavior here is justified considering the positive benefits in the experience.

And additionally, the LLM Agents that are using and reading the MCP server will always be able to advise the end-user on what they are seeing and operating on based on the provided descriptions.

src/cloudwatch-mcp-server/awslabs/cloudwatch_mcp_server/__init__.py
gcacace: So shouldn't this be then 0.1.0?

src/cloudwatch-mcp-server/tests/cloudwatch_alarms/test_active_alarms.py
```diff
@@ -103,7 +103,9 @@ async def test_no_max_items_works_correctly(self, mock_context):
             mock_session.return_value.client.return_value = mock_client
 
             alarms_tools = CloudWatchAlarmsTools()
-            result = await alarms_tools.get_active_alarms(mock_context)
+            result = await alarms_tools.get_active_alarms(
+                mock_context, include_autoscaling_alarms=True
```
gcacace, line 107: Do we need to add other test cases where the `include_autoscaling_alarms` field is set to False, or it's not passed to the tool?

goranmod: The filtering functionality is tested specifically in a separate file (as well as message creation handling):
src/cloudwatch-mcp-server/tests/cloudwatch_alarms/test_get_active_alarms_filtering.py
src/cloudwatch-mcp-server/tests/cloudwatch_alarms/test_generate_response_message.py

But now that I think if it, it might be cleaner to separate the functionality for filtering (and potentially other components) into utility classes, rather than keeping everything as part of the "tools" file. It would probably create a more maintainable and easier to follow structure of source code + tests. Thoughts?

src/cloudwatch-mcp-server/awslabs/cloudwatch_mcp_server/__init__.py
goranmod: Yeah, makes sense. I'll update it in the next revision.

PR #1124, opened 2025-08-16
refactor(aws-healthomics): improve AWS_REGION environment variable handling

WIIASD:
## Summary
Centralized region env var in a get_region() function. 

### Changes

- Removed region attributed from all client creation function as we always want to use the configured AWS_REGION env var as the region
- Created a get_region function to handle region retrieval

### User experience

No user experience change. Pure refactoring

## Checklist

* [ ] Changes are documented

N

**RFC issue number**:
N/A

Checklist:

## Acknowledgment


src/aws-healthomics-mcp-server/awslabs/aws_healthomics_mcp_server/consts.py
```diff
@@ -15,7 +15,7 @@
 """Defines constants used across the server."""
 
 # Service constants
-DEFAULT_REGION = 'us-east-1'
+DEFAULT_REGION = 'us-west-2'
```
markjschreiber: Is changing the default region a breaking change?

src/aws-healthomics-mcp-server/awslabs/aws_healthomics_mcp_server/utils/aws_utils.py
```diff
@@ -23,23 +23,28 @@
...
 
     Returns:
         boto3.Session: Configured AWS session
     """
     botocore_session = botocore.session.Session()
     user_agent_extra = f'awslabs/mcp/aws-healthomics-mcp-server/{__version__}'
     botocore_session.user_agent_extra = user_agent_extra
-    region = region or os.environ.get('AWS_REGION', DEFAULT_REGION)
-    return boto3.Session(region_name=region, botocore_session=botocore_session)
+    return boto3.Session(region_name=get_region(), botocore_session=botocore_session)
```
markjschreiber, line 47: Maybe a better way to do this would be to get the region from the session and if that doesn't work then fall back to the default? That way we can make use of the SDKs standard region chain which accounts for all the ways you can configure a region.

```
import boto3

my_session = boto3.session.Session()
current_region = my_session.region_name
print(f"Current AWS Region: {current_region}")
```

markjschreiber (changes requested):
This is a good approach however I think changing the DEFAULT_REGION in `consts.py` is a breaking change. It might also be good to try to get the region from the boto3 `Session` so that you can use all the ways that the SDK can pick up the region (such as the AWS Profiles).

src/aws-healthomics-mcp-server/awslabs/aws_healthomics_mcp_server/consts.py
WIIASD: Yeah good catch. Will revert this

src/aws-healthomics-mcp-server/awslabs/aws_healthomics_mcp_server/utils/aws_utils.py
WIIASD: This function creates the boto3 session. Do you mean we can create the session without explicitly passing in the session and we can expect it to grab the session automatically from AWS_REGION env var?

markjschreiber: I was thinking boto3 was more sophisticated in selecting the region but it seems if you call `my_session = boto3.session.Session()` then it doesn't pick up the environment variable and will only use the region set in the AWS profile.

src/aws-healthomics-mcp-server/awslabs/aws_healthomics_mcp_server/consts.py
markjschreiber: Once this is reverted I think we should be good to go.

WIIASD: Cool! Thanks for reviewing

seaofawareness (approved):
LGTM

WIIASD:
I see the failed coverage check is not related to changed files. Can we merge the PR?

scottschreckengaust (approved):
LGTM

scottschreckengaust:
Resolve comments for this to merge...

PR #1128, opened 2025-08-18
feat(aws-api-mcp-server): Add CloudWatch logs support

rshevchuk-git:
### Changes

Added support for pushing server logs to a CloudWatch log group. User can specify a log group as an environment variable when initializing the server, and it will create it (given enough permissions) or use existing one. 

The way it works is by attaching a new sink to loguru logger, which is our custom Cloudwatch sink, that pushes every log that gets logged by loguru. 

### User experience

Users can now push logs to their CloudWatch Log groups.

### Testing
Tested w/wo environment variable, and verified that logs are pushed to a newly created log group. Each server startup creates a new log stream for the duration of the server lifetime.

image

## Checklist

* [X] Changes are documented

**RFC issue number**: N/A

Checklist:

* [N/A] Migration process documented
* [N/A] Implement warnings (if it can live side by side)

## Acknowledgment

### DIFF
--- a/src/aws-api-mcp-server/CHANGELOG.md
+++ b/src/aws-api-mcp-server/CHANGELOG.md
@@ -8,10 +8,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
+- Support for pushing server logs to CloudWatch (#1128)
+
+## [0.2.12] - 2025-09-04
+
+### Added
 
 - Support for custom agent scripts directory via `AWS_API_MCP_AGENT_SCRIPTS_DIR` environment variable (#1227)
 - Scrubbing of sensitive logs (#1228)
-- Support for pushing server logs to CloudWatch (#1128)
 
 ## [0.2.11] - 2025-08-29
 


src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/common/__init__.py
```diff
@@ -22,6 +29,40 @@
     Credentials,
     ProgramValidationRequest,
 )
+import sys
+
+
+def initialize_logger():
+    logger.remove()
```
phani-srikar, line 36: 👋 . Have you considered designing this as a more generalized solution that could be used across MCP servers instead of scoping this down to just the API MCP server? This "cloud backed logging" feature could be useful across the board.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/common/cloudwatch_logger.py
```diff
@@ -0,0 +1,141 @@
...
+    - Thread-safe via an internal lock.
+    """
+
+    def __init__(self, log_group_name: str, region_name: str):
+        """Initialize the CloudWatchLogSink."""
+        self.log_group_name = log_group_name
+        self.region_name = region_name
+        self.lock = threading.Lock()
+        self.sequence_token: Optional[str] = None
+        self.client = boto3.client('logs', region_name=region_name)
```
arnewouters: This uses the default session. What about when`AWS_API_MCP_PROFILE_NAME` is set, should we use that profile for logging? Should we introduce a new environment variable to make it more customizable?

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/common/__init__.py
rshevchuk-git: This is a feature that we specifically wanted the aws-api mcp server to have, but if other teams find it useful as well and want to discuss generalizing, we are happy to do so and replace this solution later. I don't want to block on this though.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/common/cloudwatch_logger.py
rshevchuk-git: Good point, updated to use our `get_local_credentials` handler here

src/aws-api-mcp-server/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Support for custom agent scripts directory via `AWS_API_MCP_AGENT_SCRIPTS_DIR` environment variable (#1227)
 - Scrubbing of sensitive logs (#1228)
+- Support for pushing server logs to CloudWatch (#1128)
```
arnewouters: The other 2 items here have been released, can you move them under `0.2.12`?


```
- Support for pushing server logs to CloudWatch (#1128)

## [0.2.12] - 2025-09-04

### Added

- Support for custom agent scripts directory via `AWS_API_MCP_AGENT_SCRIPTS_DIR` environment variable (#1227)
- Scrubbing of sensitive logs (#1228)

```

rshevchuk-git: Updated.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/common/__init__.py
```diff
@@ -22,6 +29,40 @@
...
+    # Add file sink
+    log_dir = get_server_directory()
+    log_dir.mkdir(exist_ok=True)
+    log_file = log_dir / 'aws-api-mcp-server.log'
+    logger.add(log_file, rotation='10 MB', retention='7 days')
+
+    # Optional CloudWatch Logs sink
+    try:
+        if CLOUDWATCH_LOG_GROUP_NAME:
+            cw_sink = CloudWatchLogSink(
```
PCManticore, line 48: How is the user experience of the tool when the user doesn't have permissions to create log groups or push events?

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/common/cloudwatch_logger.py
```diff
@@ -0,0 +1,152 @@
...
+import threading
+import time
+from ..aws.credentials import get_local_credentials
+from ..common.config import AWS_API_MCP_PROFILE_NAME
+from botocore.exceptions import ClientError
+from loguru import logger
+from typing import Optional
+
+
+class CloudWatchLogSink:
```
PCManticore, line 25: Is the log pushing done in a separate thread than the main process thread that reacts to MCP requests? CloudWatch can be down, pushing logs can be slow etc. Have we considered if this should be pushed in a separate thread to avoid blocking answering MCP requests instead?

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/common/cloudwatch_logger.py
```diff
@@ -0,0 +1,152 @@
...
+        """Write a log message to CloudWatch Logs."""
+        if not message:
+            return
+
+        # CloudWatch Logs requires event timestamps in milliseconds
+        timestamp_ms = int(time.time() * 1000)
+        log_event = {'timestamp': timestamp_ms, 'message': message.rstrip('\n')}
+
+        with self.lock:
+            self._put_log_events([log_event])
```
PCManticore, line 67: The lock is not the problem as much as the call to `PutLogEvents` being potentially blocking - have we run a "load test" of sorts to figure out what the average latency for `PutLogEvents` is?

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/common/__init__.py
rshevchuk-git: We log the warning saying we couldn't create the log group, customer's can still use the server so the lack of permissions will not prevent server from initializing

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/common/cloudwatch_logger.py
rshevchuk-git: it is done in a separate thread

PR #1174, opened 2025-08-23
fix: (redshift-mcp-server) Implement proper AWS region precedence following AWS CLI standards

### LINKED ISSUE #928: redshift-mcp-server: list_clusters not working :: asking for region parameter on Q CLI

grayhemp:
Fixes #928, closes #944.

## Summary

Based on my RCA of the boto3 behavior with region specified, both the current implementation and the solution provided in the PR #944 lead to unexpected behavior.

### Current implementation

When `AWS_PROFILE` is set, creates a session/client with the profile but no region argument.

- With no `AWS_REGION` set
  - [X] Region is in `~/.aws/config` - works as expected picking the region from the config
  - [ ] No region in `~/.aws/config` - errors with `Error creating Redshift client: You must specify a region`, the default `us-east-1` region is expected to be used
- With `AWS_REGION` set
  - [ ] Region is in `~/.aws/config` - ignores `AWS_REGION` using the region specified in the config, though `AWS_REGION` is expected to be used
  - [ ] No region in `~/.aws/config` - ignores `AWS_REGION` and errors with `Error creating Redshift client: You must specify a region`, though `AWS_REGION` is expected to be used

### @HarshGrandeur's fix in #944

When `AWS_PROFILE` is set, creates a session/client with the profile and the region argument either as in `AWS_REGION` if the env variable is set or default `us-east-1`:

- With no `AWS_REGION` set
  - [ ] Region is in `~/.aws/config` - uses the default `us-east-1` instead of the one in `~/.aws/config`, expected to use the region set in `~/.aws/config`
  - [X] No region in `~/.aws/config` - uses the default `us-east-1` region
- With `AWS_REGION` set
  - [X] Region is in `~/.aws/config` - uses `AWS_REGION` instead of the one in `~/.aws/config`
  - [X] No region in `~/.aws/config` - uses the one set in `AWS_REGION`

Better, but still one miss.

### The problem

The problem appeared to be in boto3 not supprting `AWS_REGION` as an env variable (:facepalm: RTFM), see <https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html>. What it does support is `AWS_DEFAULT_REGION`.

If we use `AWS_DEFAULT_REGION` instead of `AWS_REGION`, then without modifying the current implementation (main branch) we will have the following.

When `AWS_PROFILE` is set, creates a session/client with the profile but no region argument.

- With no `AWS_DEFAULT_REGION` set
  - [X] Region is in `~/.aws/config` - works as expected picking the region from the config
  - [ ] (unclear if it's a miss, see below) No region in `~/.aws/config` - errors with `Error creating Redshift client: You must specify a region`
    - Shall it use the default `us-east-1` region?
    - Or we must require either `AWS_DEFAULT_REGION` or `AWS_REGION` or the region in `~/.aws/config` to be specified?
- With `AWS_DEFAULT_REGION` set
  - [X] Region is in `~/.aws/config` - uses `AWS_DEFAULT_REGION` instead of the one in `~/.aws/config`
  - [X] No region in `~/.aws/config` - uses the one in `AWS_DEFAULT_REGION`

In this case we have the following to address.

The one unclear behavior item (With no `AWS_DEFAULT_REGION` set, No region in `~/.aws/config`). Similarly to the note under the `AWS_DEFAULT_REGION` description in the AWS CLI doc <https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html>, it would probably be the best to expect either `AWS_DEFAULT_REGION` or `AWS_REGION` or the region in `~/.aws/config`, and let it error "officially" otherwise. We need to update the prompts with instructions to print out errors and provide suggestions on how to deal with them. That way the behavior will be compatible with AWS CLI which is the defacto standard setting expectations.

What to do with `AWS_REGION`? It can be treated similarly to the `AWS_REGION` in the AWS CLI tool (see the same doc page), which overrides the `AWS_DEFAULT_REGION` and the rest of the configuration if specified. We will need to be explicitly setting it as the session/client parameter if it is specified. There should be no default value for it.

Just a reminder, AWS CLI follows this order or precedence:

1. AWS_REGION (highest priority)
2. AWS_DEFAULT_REGION
3. Profile config region
4. Default region (if any).

We also need to carefully document the behavior including the precedence in both README.md and the prompts, as well as cover it all with tests.

### Changes

Code changes:

- Modify clients' creation to handle AWS_REGION explicitly as top precedence override
  - In the client creation methods
    - If `self.aws_region` is set then pass it to the boto3's `client()`
    - If not, let boto3 handle AWS_DEFAULT_REGION + profile config automatically, no code needed
- Remove hardcoded us-east-1 default - let it error if no region is configured
- Update instructions/prompts with region configuration guidance and error handling notes
- Add test cases covering all region configuration scenarios including the error case

Doc changes:

- Document region precedence order under the "AWS Client Requirements" section (as specified in my previous message)
- Add `AWS_DEFAULT_REGION` and remove the default value from `AWS_REGION` in the "Environment Variables" section

### User experience

Fixes the unexpected behavior described above.

## Checklist

* [X] Changes are documented

**RFC issue number**: #928 

Checklist:

* [NA] Migration process documented
* [NA] Implement warnings (if it can live side by side)

## Acknowledgment


grayhemp:
# IRL testing

Using Amazon Q.

Prompt:

---

Let's use the following command to check if the cases I described in "The problem" section are working correctly:

```text
q chat --no-interactive --trust-tools awslabsredshift_mcp_server___list_clusters "list redshift cluster names on one line, no side comeents, if there was no output from the list clusters tool, declare that and finish instead of running any other tools"
```

In ~/.aws/config, I have the `sandbox` profile configured with the region `us-east-1`, and `sandbox-no-region` with no region specified. The `us-east-1` region houses 3 clusters, other regions house no clusters. You can leverage `AWS_PROFILE` and `AWS_DEFAULT_REGION` prepending the command with them to test the cases. After that add `AWS_PROFILE` to the previous env variables and test the bahavior correctness.

---

## Complete Test Results Summary

| Test Case | Expected Behavior | Actual Result | Status |
|-----------|------------------|---------------|---------|
| Case 1: AWS_PROFILE=sandbox (region in config) |   Should work using region from config |   Found 3 clusters | PASS |
| Case 2: AWS_PROFILE=sandbox-no-region (no region) |   Should error with "You must specify a region" |   Reports "No output due to missing region" | PASS |
| Case 3: AWS_PROFILE=sandbox + AWS_DEFAULT_REGION=us-west-2 |   Should use AWS_DEFAULT_REGION (us-west-2) |   No clusters found (correct for us-west-2) | PASS |
| Case 4: AWS_PROFILE=sandbox-no-region + AWS_DEFAULT_REGION=us-east-1 |   Should use AWS_DEFAULT_REGION |   Found 3 clusters | PASS |
| Case 5: AWS_PROFILE=sandbox + AWS_REGION=us-west-2 |   Should use AWS_REGION (us-west-2) |   No clusters found (correct for us-west-2) | PASS |
| Case 6: AWS_PROFILE=sandbox-no-region + AWS_REGION=us-east-1 |   Should use AWS_REGION |   Found 3 clusters | PASS |
| Case 7: All env vars set, AWS_REGION=us-east-1 + AWS_DEFAULT_REGION=us-west-2 |   Should use AWS_REGION (highest precedence) |   Found 3 clusters | PASS |

## Key Findings

All test cases PASS - The current implementation correctly follows AWS CLI precedence order:

1. AWS_REGION (highest priority) - Cases 5, 6, 7 confirm this works
2. AWS_DEFAULT_REGION - Cases 3, 4 confirm this works
3. Profile config region - Case 1 confirms this works
4. Error when no region configured - Case 2 confirms this works

The current implementation is working correctly according to your RCA expectations.

scottschreckengaust (approved):
LGTM

HarshGrandeur:
The Deep Dive is really good. Looks good to me!

theagenticguy (approved):

PR #1192, opened 2025-08-27
feat(aws-api-mcp-server): Support Streamable HTTP mode

rshevchuk-git:
## Summary
Added support for initializing the server in streamable-http mode. Credentials are still sourced from the host machine.

### Changes
* Configurable TRANSPORT protocol using environment variable.
* Configurable PORT protocol using environment variable.

### User experience
Users can now choose between starting up the server in stdio or streamable-http mode, depending on the use case.

### Test
Tested on Linux and Windows platforms and verified that server starts up successfully, and MCP Clients can connect.

* [ ] Changes are documented

**RFC issue number**:

Checklist:

## Acknowledgment


src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
```diff
@@ -59,7 +61,7 @@
 log_file = log_dir / 'aws-api-mcp-server.log'
 logger.add(log_file, rotation='10 MB', retention='7 days')
 
-server = FastMCP(name='AWS-API-MCP', log_level=FASTMCP_LOG_LEVEL)
+server = FastMCP(name='AWS-API-MCP', log_level=FASTMCP_LOG_LEVEL, port=PORT)
```
arnewouters: Should we expose `host` too?

rshevchuk-git: Yeah, we can do that as well, thinking of adding this in a separate PR though.

arnewouters (dismissed):

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
rshevchuk-git: Actually, added to this PR :)

PR #1213, opened 2025-08-29
feat(aws-api-mcp): add support for custom defined consent and deny list operations

meghagoyal04:
## Summary
This commit adds support for customer-defined security policies with elicitation and denylisting capabilities for AWS CLI operations.
### Changes

- Added SecurityPolicy class to manage denylist, elicit list, and API customizations
- Supports JSON configuration files for flexible policy management
- Implements priority-based decision making: DENY > ELICIT > ALLOW
- Handles API customizations that map to multiple underlying API calls (for example aws s3 ls)
- If elicitation is not supported, the operations are denylisted

> Testing
- Unit tests
- Added commands to the denylist, elicit list, started local sever and tested locally with my account
### User experience

- The user can add API commands into the JSON to enforce deny/eliciting operations

## Checklist

* [ ] Changes are documented -> will do it as a followup after another change

Is this a breaking change? (Y/N) - N

**RFC issue number**:

Checklist:

## Acknowledgment

### DIFF
--- a/src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/aws/service.py
+++ b/src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/aws/service.py
@@ -33,6 +33,7 @@ from ..metadata.read_only_operations_list import (
     ReadOnlyOperations,
 )
 from ..parser.lexer import split_cli_command
+from ..security.policy import PolicyDecision, security_policy
 from .driver import interpret_command as _interpret_command
 from awslabs.aws_api_mcp_server.core.common.command import IRCommand
 from awslabs.aws_api_mcp_server.core.common.helpers import operation_timer
@@ -83,6 +84,49 @@ def is_operation_read_only(ir: IRTranslation, read_only_operations: ReadOnlyOper
     service_name = ir.command_metadata.service_sdk_name
     operation_name = ir.command_metadata.operation_sdk_name
     return read_only_operations.has(service=service_name, operation=operation_name)
+
+
+async def check_elicitation_support(ctx: Context) -> bool:
+    """Check if elicitation is supported."""
+    try:
+        return hasattr(ctx, 'elicit')
+    except Exception:
+        return False
+
+
+def check_security_policy(
+    cli_command: str, ir: IRTranslation, read_only_operations: ReadOnlyOperations, ctx: Context
+) -> PolicyDecision:
+    """Check security policy for the given command and return decision."""
+
+    def is_read_only_func(service: str, operation: str) -> bool:
+        return read_only_operations.has(service=service, operation=operation)
+
+    # Check if client supports elicitation
+    supports_elicitation = hasattr(ctx, 'elicit')
+
+    # First check if this matches a customization
+    customization_decision = security_policy.check_customization(
+        cli_command, is_read_only_func, supports_elicitation
+    )
+    if customization_decision is not None:
+        return customization_decision
+
+    # If no customization matches, check individual operation
+    if (
+        not ir.command_metadata
+        or not getattr(ir.command_metadata, 'service_sdk_name', None)
+        or not getattr(ir.command_metadata, 'operation_sdk_name', None)
+    ):
+        return PolicyDecision.ELICIT if supports_elicitation else PolicyDecision.DENY
+
+    service_name = ir.command_metadata.service_sdk_name
+    operation_name = ir.command_metadata.operation_sdk_name
+    is_read_only = is_operation_read_only(ir, read_only_operations)
+
+    return security_policy.get_decision(
+        service_name, operation_name, is_read_only, supports_elicitation
+    )
 
 
 def validate(ir: IRTranslation) -> ProgramValidationResponse:
--- a/src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/security/mcp-security-policy.json
+++ b/src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/security/mcp-security-policy.json
@@ -1,0 +1,4 @@
+{
+  "denylist": [],
+  "elicitList": []
+}
--- a/src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/security/policy.py
+++ b/src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/security/policy.py
@@ -1,0 +1,193 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import json
+import re
+from enum import Enum
+from loguru import logger
+from pathlib import Path
+from typing import Dict, List, Optional, Set
+
+
+class PolicyDecision(Enum):
+    """Class to list the policy decisions."""
+
+    DENY = 'deny'
+    ELICIT = 'elicit'
+    ALLOW = 'allow'
+
+
+class SecurityPolicy:
+    """Class to determine if the command is in he security policy or not."""
+
+    def __init__(self):
+        """Initialize the policy lists."""
+        self.denylist: Set[str] = set()
+        self.elicit_list: Set[str] = set()
+        self.customizations: Dict[str, List[str]] = {}
+        self._load_policy()
+
+    def _load_policy(self):
+        """Load security policy from local security directory."""
+        security_dir = Path(__file__).parent
+        policy_path = security_dir / 'mcp-security-policy.json'
+        customization_path = security_dir / 'aws_api_customization.json'
+
+        if not policy_path.exists():
+            logger.info('No security policy file found at {}, using default behavior', policy_path)
+        else:
+            try:
+                with open(policy_path, 'r') as f:
+                    policy_data = json.load(f)
+
+                # Load denylist
+                if 'denylist' in policy_data:
+                    self.denylist = set(policy_data['denylist'])
+                    logger.info('Loaded {} commands in denylist', len(self.denylist))
+
+                # Load elicit list (consent list)
+                if 'elicitList' in policy_data:
+                    self.elicit_list = set(policy_data['elicitList'])
+                    logger.info('Loaded {} commands in elicit list', len(self.elicit_list))
+
+            except Exception as e:
+                logger.error('Failed to load security policy from {}: {}', policy_path, e)
+
+        # Load customizations from separate file
+        if customization_path.exists():
+            try:
+                with open(customization_path, 'r') as f:
+                    customization_data = json.load(f)
+
+                if 'customizations' in customization_data:
+                    self.customizations = {}
+                    for cmd, config in customization_data['customizations'].items():
+                        if 'api_calls' in config:
+                            self.customizations[cmd] = config['api_calls']
+                    logger.info('Loaded {} customizations', len(self.customizations))
+
+            except Exception as e:
+                logger.error('Failed to load customizations from {}: {}', customization_path, e)
+
+    def get_decision(
+        self, service: str, operation: str, is_read_only: bool, supports_elicitation: bool
+    ) -> PolicyDecision:
+        """Get policy decision for a service/operation combination.
+
+        Priority: deny > elicit > default behavior
+        """
+        operation_kebab = operation.replace('_', '-')
+        operation_kebab = re.sub('([A-Z])', r'-\1', operation_kebab).lower().lstrip('-')
+
+        api_call = f'aws {service} {operation_kebab}'
+
+        # s3 service maps to s3api in CLI for some operations
+        if service == 's3':
+            # For s3 service, also check s3api variant
+            api_call_s3api = f'aws s3api {operation_kebab}'
+
+            # Check denylist first
+            if api_call in self.denylist or api_call_s3api in self.denylist:
+                return PolicyDecision.DENY
+
+            # Check elicit list
+            if api_call in self.elicit_list or api_call_s3api in self.elicit_list:
+                if not supports_elicitation:
+                    return PolicyDecision.DENY
+                return PolicyDecision.ELICIT
+        else:
+            # Check denylist first
+            if api_call in self.denylist:
+                return PolicyDecision.DENY
+
+            # Check elicit list
+            if api_call in self.elicit_list:
+                # If client doesn't support elicitation, treat the elicit list as deny
+                if not supports_elicitation:
+                    return PolicyDecision.DENY
+                return PolicyDecision.ELICIT
+
+        # Default behavior: read-only allowed, mutations elicited
+        if is_read_only:
+            return PolicyDecision.ALLOW
+        else:
+            if not supports_elicitation:
+                return PolicyDecision.DENY
+            return PolicyDecision.ELICIT
+
+    def check_customization(
+        self, cli_command: str, is_read_only_func, supports_elicitation: bool
+    ) -> Optional[PolicyDecision]:
+        """Check if command matches a customization and return the highest priority decision.
+
+        Returns None if no customization matches.
+        """
+        # Extract base command (e.g., "s3 ls" from "aws s3 ls bucket-name")
+        parts = cli_command.strip().split()
+        if len(parts) < 3 or parts[0] != 'aws':
+            return None
+
+        base_cmd = f'{parts[1]} {parts[2]}'
+
+        if base_cmd not in self.customizations:
+            return None
+
+        api_calls = self.customizations[base_cmd]
+        decisions = []
+
+        # Check the parent command itself
+        parent_api_call = f'aws {base_cmd}'
+        if parent_api_call in self.denylist:
+            return PolicyDecision.DENY
+        elif parent_api_call in self.elicit_list:
+            if not supports_elicitation:
+                return PolicyDecision.DENY
+            decisions.append(PolicyDecision.ELICIT)
+
+        # Check all underlying API calls
+        for api_call in api_calls:
+            # Parse service and operation from api_call
+            api_parts = api_call.strip().split()
+            if len(api_parts) < 3 or api_parts[0] != 'aws':
+                continue
+
+            service = api_parts[1]
+            operation = api_parts[2].replace('-', '_')
+
+            # Check against denylist/elicitlist first
+            if api_call in self.denylist:
+                return PolicyDecision.DENY
+            elif api_call in self.elicit_list:
+                if not supports_elicitation:
+                    return PolicyDecision.DENY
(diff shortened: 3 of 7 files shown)


arnewouters:
Have you tested with a client that does support elicitation?

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/security/policy.py
```diff
@@ -0,0 +1,193 @@
...
+
+        Priority: deny > elicit > default behavior
+        """
+        operation_kebab = operation.replace('_', '-')
+        operation_kebab = re.sub('([A-Z])', r'-\1', operation_kebab).lower().lstrip('-')
+
+        api_call = f'aws {service} {operation_kebab}'
+
+        # s3 service maps to s3api in CLI for some operations
+        if service == 's3':
```
rshevchuk-git: Could you give an example where this flow will be exucted? s3 and s3api don't share common operations afaik https://docs.aws.amazon.com/cli/latest/reference/s3/

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/security/mcp-security-policy.json
```diff
@@ -0,0 +1,4 @@
+{
```
rshevchuk-git: Customers will not have access to this file, we need an approach where they can provide their security policy.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/aws/service.py
```diff
@@ -85,6 +86,49 @@ def is_operation_read_only(ir: IRTranslation, read_only_operations: ReadOnlyOper
...
+def check_security_policy(
+    cli_command: str, ir: IRTranslation, read_only_operations: ReadOnlyOperations, ctx: Context
+) -> PolicyDecision:
+    """Check security policy for the given command and return decision."""
+
+    def is_read_only_func(service: str, operation: str) -> bool:
+        return read_only_operations.has(service=service, operation=operation)
+
+    # Check if client supports elicitation
+    supports_elicitation = hasattr(ctx, 'elicit')
```
rshevchuk-git: Isn't it the same as `check_elicitation_support` function does?

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/aws/service.py
```diff
@@ -85,6 +86,49 @@ def is_operation_read_only(ir: IRTranslation, read_only_operations: ReadOnlyOper
     return read_only_operations.has(service=service_name, operation=operation_name)
 
 
+async def check_elicitation_support(ctx: Context) -> bool:
```
rshevchuk-git: Why is this async?

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/security/policy.py
```diff
@@ -0,0 +1,193 @@
...
+                return PolicyDecision.DENY
+
+            # Check elicit list
+            if api_call in self.elicit_list:
+                # If client doesn't support elicitation, treat the elicit list as deny
+                if not supports_elicitation:
+                    return PolicyDecision.DENY
+                return PolicyDecision.ELICIT
+
+        # Default behavior: read-only allowed, mutations elicited
```
rshevchuk-git: I think default behavior is to allow mutations, customers have to explicitly enable consent by setting REQUIRE_MUTATION_CONSENT

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/security/policy.py
meghagoyal04: From what I understand, 
- s3 is a higher-level, more user-friendly command that provides simplified syntax for common S3 operations
- s3api provides direct access to the Amazon S3 API with more detailed control and options

Some operations that overlap include:

Creating a bucket
aws s3 mb s3://bucket-name
aws s3api create-bucket --bucket bucket-name

 Listing buckets
aws s3 ls
aws s3api list-buckets

 Copying objects
aws s3 cp file.txt s3://bucket-name/
aws s3api put-object --bucket bucket-name --key file.txt --body file.txt

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/security/policy.py
```diff
@@ -0,0 +1,193 @@
...
+
+import json
+import re
+from enum import Enum
+from loguru import logger
+from pathlib import Path
+from typing import Dict, List, Optional, Set
+
+
+class PolicyDecision(Enum):
```
vtols, line 23: Can we stick to the IAM policy terminology and call ist PolicyEffect, to avoid confusion?

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/security/policy.py
meghagoyal04: So essentially if  a user denies aws s3api put-object, the system needs to also block aws s3 cp

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/security/mcp-security-policy.json
meghagoyal04: Yes - i am changing this

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/security/policy.py
```diff
@@ -0,0 +1,193 @@
...
+    def __init__(self):
+        """Initialize the policy lists."""
+        self.denylist: Set[str] = set()
+        self.elicit_list: Set[str] = set()
+        self.customizations: Dict[str, List[str]] = {}
+        self._load_policy()
+
+    def _load_policy(self):
+        """Load security policy from local security directory."""
+        security_dir = Path(__file__).parent
```
vtols: The file seems to be part of the package. Can we load it from the location where user has access? Can users configure the file location though environment variable?

vtols (changes requested):
Please change config file location

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/security/policy.py
meghagoyal04: We are not using IAM policy terminology at all

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/security/policy.py
meghagoyal04: I am changing the file path to be under the .aws dir

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/security/policy.py
meghagoyal04: Oh I then misread this part. Will update

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/security/policy.py
vtols: > So essentially if a user denies aws s3api put-object, the system needs to also block aws s3 cp

Should that be the case though? `aws s3 cp` won't put a new object if copying from S3 to local

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/aws/service.py
meghagoyal04: Yes - good catch! updating

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/security/policy.py
```diff
@@ -0,0 +1,190 @@
...
+            # Check against denylist/elicitlist first
+            if api_call in self.denylist:
+                return PolicyDecision.DENY
+            elif api_call in self.elicit_list:
+                if not supports_elicitation:
+                    return PolicyDecision.DENY
+                decisions.append(PolicyDecision.ELICIT)
+            else:
+                # Check default behavior based on read-only status
+                is_read_only = is_read_only_func(service, operation)
```
rshevchuk-git, line 175: why do we need to create a is_read_only_func and pass it down to this function to eventually call with the same parameters as what it is calling under the hood `read_only_operations.has(service=service, operation=operation)` ? Why not just use it here

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/security/policy.py
```diff
@@ -0,0 +1,190 @@
...
+                    self.customizations = {}
+                    for cmd, config in customization_data['customizations'].items():
+                        if 'api_calls' in config:
+                            self.customizations[cmd] = config['api_calls']
+                    logger.info('Loaded {} customizations', len(self.customizations))
+
+            except Exception as e:
+                logger.error('Failed to load customizations from {}: {}', customization_path, e)
+
+    def get_decision(
```
rshevchuk-git: name is a bit confusing, had to look think for a while to figure what kind of decision we're talking about. I'd suggest something like `evaluate_access_control` or `determine_policy_effect`

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/security/policy.py
```diff
@@ -0,0 +1,190 @@
...
+        self, service: str, operation: str, is_read_only: bool, supports_elicitation: bool
```
rshevchuk-git, line 100: Why do we need to pass supports_elicitation on each call? It can be computed and stored for good in the initializer of SecurityPolicy

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/aws/service.py
```diff
@@ -85,6 +86,49 @@ def is_operation_read_only(ir: IRTranslation, read_only_operations: ReadOnlyOper
...
+    except Exception:
+        return False
+
+
+def check_security_policy(
+    cli_command: str, ir: IRTranslation, read_only_operations: ReadOnlyOperations, ctx: Context
+) -> PolicyDecision:
+    """Check security policy for the given command and return decision."""
+
+    def is_read_only_func(service: str, operation: str) -> bool:
```
rshevchuk-git, line 104: left a comment below about this function, feel like unnecessary indirection

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/security/policy.py
```diff
@@ -0,0 +1,190 @@
...
+        """Initialize the policy lists."""
+        self.denylist: Set[str] = set()
+        self.elicit_list: Set[str] = set()
+        self.customizations: Dict[str, List[str]] = {}
+        self._load_policy()
+
+    def _load_policy(self):
+        """Load security policy from user directory."""
+        policy_dir = Path.home() / '.aws' / 'aws-api-mcp'
+        policy_dir.mkdir(parents=True, exist_ok=True)
```
rshevchuk-git: would our server have access writing to /home ?

rshevchuk-git: I'd suggest confirm this with Jinet

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/security/policy.py
```diff
@@ -0,0 +1,190 @@
...
+        policy_path = policy_dir / 'mcp-security-policy.json'
+        security_dir = Path(__file__).parent
+        customization_path = security_dir / 'aws_api_customization.json'
+
+        if not policy_path.exists():
+            logger.info('No security policy file found at {}, using default behavior', policy_path)
```
rshevchuk-git: This might be confusing when we say "using default behavior" (first question is - what is the default?), let's say something along the lines of "not applying any additional security policies"

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/security/policy.py
```diff
@@ -0,0 +1,190 @@
...
+                # Load elicit list (consent list)
+                if 'elicitList' in policy_data:
+                    self.elicit_list = set(policy_data['elicitList'])
+                    logger.info('Loaded {} commands in elicit list', len(self.elicit_list))
+
+            except Exception as e:
+                logger.error('Failed to load security policy from {}: {}', policy_path, e)
+
+        # Load customizations from separate file
+        if customization_path.exists():
```
rshevchuk-git: why is this part of _load_policy function? and not a separate _load_customizations_mapping()?

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/security/policy.py
```diff
@@ -0,0 +1,190 @@
...
+
+    def get_decision(
+        self, service: str, operation: str, is_read_only: bool, supports_elicitation: bool
+    ) -> PolicyDecision:
+        """Get policy decision for a service/operation combination.
+
+        Priority: deny > elicit > default behavior
+        """
+        operation_kebab = operation.replace('_', '-')
+        operation_kebab = re.sub('([A-Z])', r'-\1', operation_kebab).lower().lstrip('-')
```
rshevchuk-git, line 107: Do we not have to_kebab_case function already somewhere in commons/utils?

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/security/policy.py
```diff
@@ -0,0 +1,190 @@
...
+            if not supports_elicitation:
+                return PolicyDecision.DENY
+            decisions.append(PolicyDecision.ELICIT)
+
+        # Check all underlying API calls
+        for api_call in api_calls:
+            # Parse service and operation from api_call
+            api_parts = api_call.strip().split()
+            if len(api_parts) < 3 or api_parts[0] != 'aws':
+                continue
```
rshevchuk-git: api_call comes from api_calls (the name is confusing btw, it refers to calls customizations maps to), which is the list we maintain and control. Here we validate if cli command is correctly structured, otherwise we skip it (continue).

I'd say it's better to fail here, or have some kind of customizations file validation in other place. Currently if something is incorrectly configured in our mapping.json we will not know about as it will always be ignored.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/security/policy.py
arnewouters: We should only need read access, right?

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/security/policy.py
```diff
@@ -0,0 +1,190 @@
...
+        self, cli_command: str, is_read_only_func, supports_elicitation: bool
+    ) -> Optional[PolicyDecision]:
+        """Check if command matches a customization and return the highest priority decision.
+
+        Returns None if no customization matches.
+        """
+        # Extract base command (e.g., "s3 ls" from "aws s3 ls bucket-name")
+        parts = cli_command.strip().split()
+        if len(parts) < 3 or parts[0] != 'aws':
+            return None
```
arnewouters, line 135: Can we not use the IRTranslation for customizations as well?

e.g. `aws s3 ls --profile default` is translated into:

`
IRTranslation(command=IRCommand(command_metadata=CommandMetadata(service_sdk_name='s3', service_full_sdk_name=None, operation_sdk_name='ls', has_streaming_output=False), parameters={'--paths': 's3://', '--dir-op': False, '--human-readable': False, '--summarize': False}, region='us-east-1', profile='default', client_side_filter=None, is_awscli_customization=True, output_file=None), command_metadata=CommandMetadata(service_sdk_name='s3', service_full_sdk_name=None, operation_sdk_name='ls', has_streaming_output=False), program=None, validation_failures=None, missing_context_failures=None, unsupported_translation=None, is_awscli_customization=False)
`

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/security/policy.py
rshevchuk-git: yeah, and mkdir will attempt to create directory

meghagoyal04: but what if the api-mcp directory does not exist? Do we want the customer to create it?

arnewouters: If there is no directory, there is no file to read anyways

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/security/policy.py
meghagoyal04: Nope

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/security/policy.py
meghagoyal04: The is_read_only_func is passed as a parameter to avoid circular dependency and makes it easier to test this method on its own without depending on the other file

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/security/policy.py
meghagoyal04: updating!

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/security/policy.py
meghagoyal04: From my understanding of the MCP clients - We need to pass supports_elicitation on each call because it will depend on the context. Different clients may have different elicitation support, and this needs to be checked per request, not stored globally in the SecurityPolicy class.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/security/policy.py
meghagoyal04: updating! :)

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/security/policy.py
meghagoyal04: It can be used but that might be an overkill. The approach right now is string-based matching on the customization mappings we control, which is faster than translating each API call.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/security/policy.py
meghagoyal04: agreed - updating this

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/security/policy.py
rshevchuk-git: Discussed offline - this will be addressed.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/security/policy.py
rshevchuk-git: Discussed offline - will be addressed in a followup.

PR #1228, opened 2025-09-03
feat(aws-api-mcp-server): Add handling of sensitive logs

rshevchuk-git:
## Summary
1/ Only log base CLI command to be executed, to avoid logging sensitive data in parameters.
2/ Log API response, redacting sensitive fields.

### Changes
Implemented SensitiveDataScrubber that attempts to redact potential sensitive keys from the final API response, before logging via loguru.

### User experience

No sensitive information being logged and stored on the host.

## Checklist

* [ ] Changes are documented

**RFC issue number**:

Checklist:

## Acknowledgment

### DIFF
--- a/src/aws-api-mcp-server/CHANGELOG.md
+++ b/src/aws-api-mcp-server/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Support for custom agent scripts directory via `AWS_API_MCP_AGENT_SCRIPTS_DIR` environment variable (#1227)
+- Scrubbing of sensitive logs (#1228)
 
 ## [0.2.11] - 2025-08-29
 


arnewouters (commented):
Can we add an entry to the CHANGELOG?

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/aws/service.py
```diff
@@ -175,6 +179,14 @@ def interpret_command(
     )
 
 
+def _log_successful_execution(response: InterpretationResponse | None):
+    if response and response.error_code is None and response.as_json is not None:
+        logger.info(
```
PCManticore, line 184: What is the timing on this recursive method? Can we try with a big response to see how much it adds on top?

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/aws/service.py
```diff
@@ -175,6 +179,14 @@ def interpret_command(
     )
 
 
+def _log_successful_execution(response: InterpretationResponse | None):
```
PCManticore, line 182: Why do we have to log executions in the first place? I'm missing the details here.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/aws/service.py
rshevchuk-git: Tested with the following generated JSON

```
large_data = {
        "users": [],
        "config": {
            "database": {
                "password": "db_secret_123",  # pragma: allowlist secret
                "connection_string": "postgresql://user:pass@localhost/db"
            },
            "api_keys": {
                "stripe": "sk_test_123456789",  # pragma: allowlist secret
                "aws": "AKIAIOSFODNN7EXAMPLE"  # pragma: allowlist secret
            }
        },
        "logs": [],
        "metadata": {
            "version": "1.0.0",
            "environment": "production",
            "secrets": {
                "jwt_secret": "super_secret_jwt_key",  # pragma: allowlist secret
                "encryption_key": "aes_256_key_here"  # pragma: allowlist secret
            }
        }
    }
    
    # Add 1000 users with nested data
    for i in range(1000):
        user = {
            "id": i,
            "username": f"user_{i}",
            "email": f"user{i}@example.com",
            "profile": {
                "name": f"User {i}",
                "age": 20 + (i % 50),
                "preferences": {
                    "theme": "dark" if i % 2 == 0 else "light",
                    "notifications": True
                }
            },
            "credentials": {
                "password": f"user_password_{i}",  # pragma: allowlist secret
                "session_token": f"session_{i}_token",  # pragma: allowlist secret
                "api_key": f"api_key_{i}"  # pragma: allowlist secret
            },
            "permissions": ["read", "write"] if i % 3 == 0 else ["read"],
            "last_login": f"2024-01-{(i % 28) + 1:02d}T10:00:00Z"
        }
        large_data["users"].append(user)
    
    # Add 500 log entries
    for i in range(500):
        log_entry = {
            "timestamp": f"2024-01-{(i % 28) + 1:02d}T{(i % 24):02d}:00:00Z",
            "level": "INFO" if i % 4 == 0 else "ERROR",
            "message": f"Log message {i}",
            "user_id": i % 1000,
            "metadata": {
                "ip_address": f"192.168.1.{(i % 254) + 1}",
                "user_agent": f"Mozilla/5.0 (User {i})",
                "session_id": f"session_{i}_id"  # pragma: allowlist secret
            }
        }
        large_data["logs"].append(log_entry)
```

rshevchuk-git: Results look good, so this should not be too concerning:

```
Total processing time: 0.0331 seconds
Number of users processed: 1000
Number of log entries processed: 500
Average time per user: 0.000033 seconds
Average time per log entry: 0.000066 seconds
```

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/aws/service.py
rshevchuk-git: Currently, we log the response from suggest_aws_commands but not from call_aws, which creates inconsistency and makes it difficult to trace multi-step workflows. When reviewing agent scripts that execute a sequence of commands (e.g., 1.X, 2.Y, 3.Z), it’s challenging to debug why the LLM may have chosen to execute a different set (e.g., 1.X, 2.A, 3.B) if unexpected results occur. Without logging the actual responses, it is not possible to fully understand or diagnose what happened during execution.

PCManticore (approved):

arnewouters (approved):

PR #1271, opened 2025-09-11
feat(aws-api-mcp-server): Add Agent Script for debugging Lambda timeouts

rshevchuk-git:
## Summary
Adds an Agent Script to debug Lambda function timeout failures by analyzing function configuration, CloudWatch logs, metrics, and dependencies to identify root causes and provide actionable solutions. 

### Changes

* Added lambda-timeout-debugging.script.md
* Removed unnecessary front matter from script_format.md

### User experience

Queries like "Debug my lambda timeouts", "Investigate why my Lambda function times out", etc will be handled by the "lambda-timeout-debugging" script.

### Testing
Tested with several faulty setups of Lambda, for the case where actual execution was taking longer than the set timeout, agent generated the following summary

image

## Checklist

If your change doesn't seem to apply, please leave them unchecked.

* [ ] I have reviewed the [contributing guidelines](https://github.com/awslabs/mcp/blob/main/CONTRIBUTING.md)
* [ ] I have performed a self-review of this change
* [ ] Changes have been tested
* [ ] Changes are documented

Is this a breaking change? (Y/N)

**RFC issue number**:

Checklist:

* [ ] Migration process documented
* [ ] Implement warnings (if it can live side by side)

## Acknowledgment

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the [project license](https://github.com/awslabs/mcp/blob/main/LICENSE).

PR #1276, opened 2025-09-12
feat(aws-api-mcp-server): Add Agent Script for failure troubleshooting

arnewouters:
Fixes

### Changes

Add new Agent Script for troubleshooting application errors

### User experience

Tested troubleshooting with a lambda `my-app` that generated a few errors

image

## Checklist

* [x] Changes are documented

**RFC issue number**:

Checklist:

## Acknowledgment

### DIFF
--- a/src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/agent_scripts/registry/application-failure-troubleshooting.script.md
+++ b/src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/agent_scripts/registry/application-failure-troubleshooting.script.md
@@ -1,0 +1,330 @@
+---
+description: Troubleshoots failing applications by discovering and analyzing relevant CloudWatch log groups to identify error patterns, root causes, and provide actionable solutions.
+---
+# Application Failure Troubleshooting
+
+## Overview
+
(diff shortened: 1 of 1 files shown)


src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/agent_scripts/registry/application-failure-troubleshooting.script.md
```diff
@@ -0,0 +1,330 @@
+---
+description: Troubleshoots failing applications by discovering and analyzing relevant CloudWatch log groups to identify error patterns, root causes, and provide actionable solutions.
```
iddv, line 2: Thanks - looks like there is some overlap with [this](https://github.com/awslabs/mcp/pull/1271) agent script - while that is for lambda, it also includes config, metrics and dependencies. In fact, this also covers the lambda use case below. Will this cause any agent confusion? Should we include a standard set of dimension to check for troubleshooting use cases?

arnewouters: I agree they are kind of similar, I did some testing and preference goes to the Lambda script as soon as you mention Lambda. For all other troubleshooting queries, it seems to favour this script.

PR #1306, opened 2025-09-17
fix(aws-api-mcp-server): Update healthcheck script to use python3

ptadakaluru:
Fix: Update healthcheck script to use python3

Summary
Changes
The docker-healthcheck.sh script in the aws-api-mcp-server was updated to correctly identify the server's running process. The script was attempting to match a process named /bin/python in its pgrep command, but the container runs the server with the process name /bin/python3. This change updates the health check script to correctly match python3, resolving a consistent health check failure.

User experience
Before this change: Users deploying the aws-api-mcp-server Docker image would find their containers consistently marked as unhealthy. The server would appear to be running when manually checked with ps, but the automated health check would fail, leading to misdiagnoses and deployment issues in orchestration environments like ECS.

After this change: The health check will pass successfully, and the container will be correctly marked as healthy as long as the server process is running. This improves the reliability of the Docker image and provides a correct health status in all deployment environments.

Checklist

[x] I have reviewed the [contributing guidelines](https://github.com/awslabs/mcp/blob/main/CONTRIBUTING.md)

[x] Changes have been tested

[x] Changes are documented

N

RFC issue number:

Checklist:

[ ] Migration process documented

[ ] Implement warnings (if it can live side by side)

Acknowledgment


arnewouters:
@ptadakaluru I just started the MCP server with docker and I'm seeing python instead of python3
```
app@2410a41e6bfb:/$ ps -alxf
F   UID     PID    PPID PRI  NI    VSZ   RSS WCHAN  STAT TTY        TIME COMMAND
4  1000     111       0  20   0   4192  3568 do_wai Ss   pts/0      0:00 /bin/bash
0  1000     119     111  20   0   8092  4104 -      R+   pts/0      0:00  \_ ps -alxf
4  1000       1       0  20   0 4790104 477016 ep_pol Ssl ?         0:12 /app/.venv/bin/python /app/.venv/bin/awslabs.aws-api-mcp-server
```

Which docker image are you using?

Configuration that I tested:
```json
    "docker.aws-api-mcp-server": {
      "disabled": false,
      "timeout": 60,
      "type": "stdio",
      "command": "docker",
      "args": [
        "run",
        "--rm",
        "--interactive",
        "--env",
        "FASTMCP_LOG_LEVEL=ERROR",
        "--env",
        "AWS_REGION=us-east-1",
        "--volume",
        "/home/awouters/.aws:/app/.aws",
        "public.ecr.aws/awslabs-mcp/awslabs/aws-api-mcp-server:latest"
      ],
      "env": {}
    }
```

ptadakaluru:
Hello,

I've started the MCP Server with *Streamable-HTTP Transport Setup  *
Steps Followed:

   1.

   Cloned the repository and changed to the API server directory.
   2.

   Built the Docker image using:

   *docker build -t aws-api-mcp-server .*

   3.

   Started the container with environment variables set for transport,
   host, port, AWS credentials, and GitHub token:

   *docker run -d \
     --name my-aws-mcp \
     -p 8000:8000 \
     -e AWS_API_MCP_TRANSPORT="streamable-http" \
     -e AWS_API_MCP_HOST="0.0.0.0" \
     -e AWS_API_MCP_PORT="8000" \
     -e AWS_ACCESS_KEY_ID="xxxxxxxxxxxx" \
     -e AWS_SECRET_ACCESS_KEY="xxxxxxxxxxxxxxxxx" \
     -e GH_TOKEN="xxxxxxxxxxxxxxxx" \
     aws-api-mcp-server*


Issue Observed:

   -

   The container failed to come up because the health check script
   referenced python (instead of python3).

Fix Applied:

   -

   Manually updated the health check script to use python3 inside the
   container, which resolved the issue.
   -

   Rebuilt the Docker image with the updated health check script.
   -

   Restarted the container and confirmed that the health check passed
   successfully.

------------------------------

With the corrected health check (python3), the MCP server with
streamable-http transport is now running without issues.
Thanks,
Pavan

src/aws-api-mcp-server/docker-healthcheck.sh
```diff
@@ -16,7 +16,7 @@
 SERVER="aws-api-mcp-server"
 
 # Check if the server process is running
-if pgrep -P 0 -a -l -x -f "/app/.venv/bin/python /app/.venv/bin/awslabs.$SERVER" > /dev/null; then
+if pgrep -P 0 -a -l -x -f "/app/.venv/bin/python3 /app/.venv/bin/awslabs.$SERVER" > /dev/null; then
```
rshevchuk-git:
```suggestion
if pgrep -P 0 -a -l -x -f "/app/.venv/bin/python* /app/.venv/bin/awslabs.$SERVER" > /dev/null; then
```

rshevchuk-git: Hey @ptadakaluru! Thanks for raising the issue, apparently the way uv creates virtual environment can be dependent on the build context. Would be curious to understand your local setup (OS you're using, architecture) to understand which systems are impacted by this.

For the systems we have been building this on -- pinning python to `python3` would not work unfortunately, this seems to be a very specific fix. Instead, I suggest we update your change to use `python*` to cover all scenarios.

Please double check if this would fix this on your end, and if so please update the PR. 

Also would be helpful if you could provide output of `ls -la /app/.venv/bin/python*` executed in your container, to understand how virtual environment was set up.

ptadakaluru: Hello @rshevchuk-git,

It’s a good idea in theory to use python* in the health check script, but in practice it’s not working. I verified that across other AWS MCP servers (e.g., IAM, Pricing MCP servers), the health check explicitly uses python3. This particular server, aws-api-mcp-server, is the only one I’ve found that still references just python— instead of python3

Here’s what I observed in my environment:

```
app@d1fb0702b038:/$ ls -la /app/.venv/bin/python*
lrwxrwxrwx 1 app app 25 Sep 17 21:18 /app/.venv/bin/python -> /usr/local/bin/python3.13
lrwxrwxrwx 1 app app  6 Sep 17 21:18 /app/.venv/bin/python3 -> python
lrwxrwxrwx 1 app app  6 Sep 17 21:18 /app/.venv/bin/python3.13 -> python
```


And here are the results of testing the health check command:

With python* in the script → Not working

`pgrep -P 0 -a -l -x -f "/app/.venv/bin/python* /app/.venv/bin/awslabs.aws-api-mcp-server"`
# No output


With python in the script → Not working

`pgrep -P 0 -a -l -x -f "/app/.venv/bin/python /app/.venv/bin/awslabs.aws-api-mcp-server"`
# No output

With python3 in the script → Working correctly

`pgrep -P 0 -a -l -x -f "/app/.venv/bin/python3 /app/.venv/bin/awslabs.aws-api-mcp-server"`
1 /app/.venv/bin/python3 /app/.venv/bin/awslabs.aws-api-mcp-server


So it looks like the only reliable option here is to use python3 in the health check script, consistent with the other AWS MCP servers. Not sure why aws-api-mcp-server was left with python instead.

Also worth noting: the minimum Python version required to run MCP servers is 3.10. Given this baseline, relying on python3 explicitly seems more robust and consistent across environments.

rshevchuk-git: My bad, adding star at the end doesn't work as expected, it should be `python3?` to match both python and python3

```suggestion
if pgrep -P 0 -a -l -x -f "/app/.venv/bin/python3? /app/.venv/bin/awslabs.$SERVER" > /dev/null; then
```

As I mentioned earlier, we can't just set it to python3 as on some systems, including my personal setup, the default executable is `python` (which is a symlink to python3.13, as in the 'ls' snippet you provided), and trying to search for `python3` will not yield any results.

Could you try and see if this `python3?` approach works for you?

ptadakaluru:
Hello,

Yes python3 approach works for me.

Thanks,
Pavan

arnewouters (approved):

Issue #1308, opened 2025-09-18
CCAPI MCP Server: DynamoDB resources created in us-east-1 despite AWS_REGION=eu-central-1 configuration


TLaue:
### Describe the bug

The CCAPI MCP server ignores the `AWS_REGION` environment variable and creates AWS resources in `us-east-1` instead of the configured `eu-central-1` region. This occurs despite proper configuration in both the MCP server configuration file and system environment variables.

### Expected Behavior

Resources should be created in `eu-central-1` as specified by:
- `AWS_REGION=eu-central-1` environment variable in MCP configuration
- AWS CLI default region configuration
- Documented boto3 region resolution chain

### Current Behavior

Resources are consistently created in `us-east-1`, indicating the server is falling back to the default region despite proper regional configuration.

### Reproduction Steps

1. **Configure MCP Server** with proper region settings:
   ```json
   {
     "mcpServers": {
       "awslabs.ccapi-mcp-server": {
         "command": "uvx",
         "args": ["awslabs.ccapi-mcp-server@latest"],
         "env": {
           "AWS_PROFILE": "default",
           "AWS_REGION": "eu-central-1",
           "DEFAULT_TAGS": "enabled",
           "SECURITY_SCANNING": "enabled",
           "FASTMCP_LOG_LEVEL": "ERROR"
         }
       }
     }
   }
   ```

2. **Verify AWS CLI Configuration**:
   ```bash
   $ echo $AWS_REGION
   eu-central-1

   $ aws configure list
         Name                    Value             Type    Location
         ----                    -----             ----    --------
       ...
       region             eu-central-1              env    ['AWS_REGION', 'AWS_DEFAULT_REGION']
   ```

3. **Create DynamoDB Table** using CCAPI MCP server tools:
   - Use `generate_infrastructure_code()` with resource type `AWS::DynamoDB::Table`
   - Follow standard workflow: generate → explain → create
   - Observe that region parameter shows `eu-central-1` in responses

4. **Verify Resource Location**:
   ```bash
   $ aws dynamodb list-tables --region eu-central-1
   {
       "TableNames": []
   }

   $ aws dynamodb list-tables --region us-east-1
   {
       "TableNames": [
           "test"
       ]
   }
   ```

## Environment Information

- **MCP Server**: `awslabs.ccapi-mcp-server@latest`
- **System**: macOS (Darwin 24.6.0)
- **AWS Profile**: default
- **Expected Region**: eu-central-1
[...]

PR #1327, opened 2025-09-20
fix(ccapi-mcp-server): Region priority in resource operations

Tristan-HuiFeng:
Fixes for #1308 

## Summary
Fix the issue where resources are provision in `'us-east-1'` even if environment is set correctly for another region

### Changes
The resource operation region settings takes precedens in the following order:
1. Region passed explictly in request parameter during mcp tool call
2. `'AWS_REGION'` Environment Variable
3. `'us-east-1'` by default

### User experience

Previously, setting the `'AWS_REGION'` in environment variable still creates resources in `'us-east-1'`.
Now, the region in `'AWS_REGION'` will be use, if not it will use `'us-east-1'` by default

## Checklist

* [ ] Changes are documented

**RFC issue number**: #1308 

Checklist:

## Acknowledgment

### DIFF
--- a/src/ccapi-mcp-server/awslabs/ccapi_mcp_server/impl/tools/resource_operations.py
+++ b/src/ccapi-mcp-server/awslabs/ccapi_mcp_server/impl/tools/resource_operations.py
@@ -116,7 +116,9 @@ async def create_resource_impl(request: CreateResourceRequest, workflow_store: d
     properties = workflow_data['data']['properties']
 
     # Ensure region is a string, not a FieldInfo object
-    region_str = ensure_region_string(request.region) or 'us-east-1'
+    region_str = (
+        ensure_region_string(request.region) or aws_session_data.get('region') or 'us-east-1'
+    )
     cloudcontrol_client = get_aws_client('cloudcontrol', region_str)
     try:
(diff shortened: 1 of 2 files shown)


novekm:
taking a look

novekm:
Hi @Tristan-HuiFeng, sorry for the delay!

Thank you for identifying this region resolution issue and providing a comprehensive solution! Your analysis correctly identified the problem and the right files to modify.

**Why we can't merge this PR directly:**

The current implementation has some inconsistencies that could lead to unpredictable behavior:
- Mixed approaches: some functions use `aws_session_data.get('region')` while others use `environ.get('AWS_REGION')`
- The `us-east-1` hardcoded fallback remains, which is the root cause of the original issue
- Doesn't prioritize MCP configuration settings over AWS profile settings

**Our solution (in upcoming PR):**

We've implemented a fix that uses the existing MCP environment variable infrastructure to ensure consistent region resolution with proper precedence (MCP config → AWS profile → clear error message).

Your PR provided valuable insights into which functions needed updates and comprehensive test coverage ideas. We'll include this fix in an upcoming release.

Thanks again for the detailed analysis and solution approach!

PR #1332, opened 2025-09-21
feat(ecs-mcp-server): add AWS Knowledge MCP Server proxy integration

matthewgoodman13:
## Summary
Integrate AWS Knowledge MCP Server as a proxy to provide up-to-date AWS documentation access, including new ECS features beyond standard LLM training data. Add tool transformations to append ECS-specific guidance and register prompt patterns for new features.

### Changes
- Add aws_knowledge_proxy module with proxy setup and tool transformations
- Enhance server with lifespan management for async operations
- Add comprehensive integration tests for the new tools
  - These test that the upstream tools are accessible and have not changed. We validate if there is exactly 3 tools, and that the description is exactly what we expect. This way, if a change happens upstream, we will know about it. 
- Update dependencies to support pytest-asyncio for async testing
- Document new AWS Knowledge tools and usage examples in README
- Update the troubleshooting tools integ test helper due to changes in [mcp-inspector 0.16.6](https://github.com/modelcontextprotocol/inspector/releases) w.r.t. [Parse tool args from CLI as JsonValue instead of just strings](https://github.com/modelcontextprotocol/inspector/pull/647)

### User experience
Customers can retrieve up-to-date ECS documentation, best practices, and knowledge of new tools that models may not be trained on, such as Blue-Green deployments.

### Testing
- Achieved 99%+ coverage on new changes with added UTs
- Unit Tests Pass
- Integ. Tests Pass
- All pre-commit Tests Pass

## Checklist

* [Y] Changes are documented

**RFC issue number**:

Checklist:

* [NA] Migration process documented
* [NA] Implement warnings (if it can live side by side)

## Acknowledgment

### DIFF
--- a/src/ecs-mcp-server/README.md
+++ b/src/ecs-mcp-server/README.md
@@ -304,7 +304,9 @@ The resource management tool enforces permission checks for write operations. Op
 
 ### AWS Documentation Tools
 
-The ECS MCP Server integrates with the AWS Knowledge MCP Server to provide access to up-to-date AWS documentation, including ECS-specific knowledge about new features recently launched that models may not be aware of.
+The ECS MCP Server integrates with the [AWS Knowledge MCP Server](https://github.com/awslabs/mcp/tree/main/src/aws-knowledge-mcp-server) to provide access to up-to-date AWS documentation, including ECS-specific knowledge about new features recently launched that models may not be aware of.
+
+Note: these tools are duplicative if you have the AWS Knowledge MCP Server already configured in your MCP client. For the below knowledge tools, the ECS MCP Server adds extra guidance to the tool descriptions to help LLMs use the tools for ECS contexts.
 
 - **aws_knowledge_aws___search_documentation**: Search across all AWS documentation including the latest AWS docs, API references, Blogs posts, Architectural references, and Well-Architected best practices.
 


src/ecs-mcp-server/awslabs/ecs_mcp_server/modules/aws_knowledge_proxy.py
```diff
@@ -0,0 +1,191 @@
...
+    try:
+        logger.info("Setting up AWS Knowledge MCP Server proxy")
+        proxy_config = {
+            "mcpServers": {
+                "aws-knowledge-mcp-server": {
+                    "command": "uvx",
+                    "args": [
+                        "mcp-proxy",
+                        "--transport",
+                        "streamablehttp",
```
guitar80ep, line 60: Is Streamable required for the AWS KB MCP? Or is this just a best practice?

src/ecs-mcp-server/awslabs/ecs_mcp_server/modules/aws_knowledge_proxy.py
```diff
@@ -0,0 +1,191 @@
...
+
+    Covers blue-green deployments, new ECS features, and comparisons based on ECS_TOOL_GUIDANCE.
+
+    Args:
+        mcp: The FastMCP server instance to register prompts with
+    """
+
+    prompts = [
+        {
+            "patterns": [
```
guitar80ep, line 133: after this release, we should consider adding a file where this info can be stored and loaded. This section of code runs of the risk of growing *substantially* and at some points in time, I'd expect to remove older features from the list (once we see LLMs start to be trained on latest docs).

src/ecs-mcp-server/awslabs/ecs_mcp_server/modules/aws_knowledge_proxy.py
```diff
@@ -0,0 +1,191 @@
...
+from fastmcp import FastMCP
+from fastmcp.server.proxy import ProxyClient
+from fastmcp.tools.tool_transform import ToolTransformConfig
+
+# Guidance to append to tool descriptions
+# ruff: noqa: E501
+ECS_TOOL_GUIDANCE = """
+
+    ## ECS DOCUMENTATION GUIDANCE:
+    This tool provides up-to-date ECS documentation and implementation guidance, including new ECS features beyond standard LLM training data.
```
guitar80ep, line 32: This is the area where I would have liked to see some clarity around the proxy nature. Both for LLMs and customers, it should be clear that this tool is duplicative if they have ECS MCP KB installed. It may supply better ECS specific prompts though. 

For example, if I only use AWS KB MCP for the purposes of working with ECS, I may prefer this implementation and disable the AWS KB MCP in my environment.

src/ecs-mcp-server/awslabs/ecs_mcp_server/modules/aws_knowledge_proxy.py
matthewgoodman13: This is something I did think about, even the ECS Guidance description that we add. I opted to not do that until we actually do add more features there to avoid extra files and complexity early on.

src/ecs-mcp-server/awslabs/ecs_mcp_server/modules/aws_knowledge_proxy.py
matthewgoodman13: Don't believe it is required, but is recommended for accessing remote MCP servers. It generally allows for better user experiences (once we host ourselves). 

Followed installation steps from:
https://github.com/awslabs/mcp/tree/main/src/aws-knowledge-mcp-server

matthewgoodman13:
**1. Can we provide a signal in the tools doc to the LLM that this proxy is duplicative if the customer has the AWS KB MCP installed?**
=> See response above

**2. will we plan to update the prompts on any new release?**
=> I think we should, unless we have proof from testing, that we do not need to.

**3. What kinds of manual tests did we do?**
=> Asked about AWS knowledge, asked about ECS knowledge (ie. best guidance on setting up an ECS cluster, best guidance on security), asked different questions about Blue-Green deployments, including how to setup, and LLM reliably used the proxied tools once they had the updated descriptions and prompt patterns.

**4. did we have to do the main.py refactor in this CR?**
=> I had to update the main.py to use the async lifespan management feature to allow us to asynchronously update the tool descriptions once they are loaded. The current file structure made it hard to unit test, so I did a minor refactor. I didn't consider that a big change as code is the same, but just now in methods. Cleaner and we also increase unit test coverage on this file by 50-60%

**5. Any best practices for Fast MCP Proxy that we discovered?****
=> To implement the proxy, I looked into what the docs say, and also what other teams have done in the awslabs. My implementation seems like the generic one people use, including adding the lifespan param above.

src/ecs-mcp-server/awslabs/ecs_mcp_server/modules/aws_knowledge_proxy.py
matthewgoodman13: The LLM may have duplicative tools if both are installed, adding to the context window, but I dont think it will ever matter which tool the LLM chooses as they will both return the exact same response, from the exact same hosted MCP.

ie.
- Search for documentation regarding XYZ
   - Returns response
- Read documentation
   - Returns response
   
For those, it wont matter if the first call gets executed by the installed knowledge server and then the ecs mcp server tool gets executed for the second

matthewgoodman13: Updated readme with behavior.

PR #1368, opened 2025-09-25
feat: adding query attribution for read_documentation

JonLim:
Fixes

### Changes

The AWS Documentation team is creating the foundation for ensuring quality responses for searches being made with the Docs MCP Server.

This change adds the `query_id` to read requests if the URL is coming from a search.

### User experience

No change to the user experience.

## Checklist

* [X] Changes are documented

Is this a breaking change? N

**RFC issue number**:

Checklist:

## Acknowledgment


src/aws-documentation-mcp-server/awslabs/aws_documentation_mcp_server/server_aws.py
```diff
@@ -224,12 +225,14 @@ async def search_documentation(
...
                     url='',
                     title=error_msg,
+                    query_id='',
                     context=None,
                 )
             ]
 
         try:
             data = response.json()
+            query_id = data.get('queryId', None)
```
tuanknguyen: nit: `.get()` should already default to None when the key is missing in the dictionary. `data.get('queryId')` should be sufficient

src/aws-documentation-mcp-server/tests/test_server_utils.py
```diff
@@ -291,6 +295,74 @@ async def test_start_index_handling(self):
...
+                return_value=True,
+            ):
+                with patch(
+                    'awslabs.aws_documentation_mcp_server.server_utils.extract_content_from_html',
+                    return_value='# Test\n\nContent',
+                ):
+                    with patch(
+                        'awslabs.aws_documentation_mcp_server.server_utils.format_documentation_result',
+                        return_value='AWS Documentation from URL: # Test\n\nContent',
+                    ):
```
tuanknguyen: nit: is there a way we can avoid the nested patch?

src/aws-documentation-mcp-server/awslabs/aws_documentation_mcp_server/server_aws.py
JonLim: Ah, fair, let me update that, easy fix! Thank you!

src/aws-documentation-mcp-server/tests/test_server_utils.py
JonLim: I can look into it! This was the format of the tests that were here before, but it would be great to have a single/more clean `patch` call.

Lavoiedavidw (approved):
Thanks for addressing nitpicks

JonLim:
> LGTM - the query_id sanitization is nice, but also potentially unnecessary?

@Lavoiedavidw Sorry for not hitting the send button on the comment earlier - we spoke about this off-platform, but this was a defensive coding approach that doesn't add a ton of overhead, and hopefully you are okay with keeping it in!

PR #1374, opened 2025-09-26
fix: Updating User-Agent in tests

JonLim:
Fixes

### Changes

Currently, the User-Agent set for requests made to AWS Documentation endpoints inside automated tests use the same User-Agent as human/LLM traffic.

This change patches the User-Agent in `live` tests, to ensure that test traffic doesn't mix with real traffic.

### User experience

No change in user experience! Purely for automated tests that are run as part of GitHub Actions in PRs and releases for `awslabs/mcp`

## Checklist

* [X] Changes are documented

Is this a breaking change? N

**RFC issue number**:

Checklist:

## Acknowledgment

### DIFF
--- a/src/aws-documentation-mcp-server/tests/constants.py
+++ b/src/aws-documentation-mcp-server/tests/constants.py
@@ -1,0 +1,6 @@
+from awslabs.aws_documentation_mcp_server.server_utils import DEFAULT_USER_AGENT
+
+
+TEST_USER_AGENT = DEFAULT_USER_AGENT.replace(
+    '(AWS Documentation Server)', '(AWS Documentation Tests)'
+)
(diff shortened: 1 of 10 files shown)


src/aws-documentation-mcp-server/tests/constants.py
```diff
@@ -0,0 +1,6 @@
+from awslabs.aws_documentation_mcp_server.server_utils import DEFAULT_USER_AGENT
+
+
+TEST_USER_AGENT = DEFAULT_USER_AGENT.replace(
+    '(AWS Documentation Server)', '(AWS Documentation Tests)'
+)
```
tuanknguyen, line 6: nit: I'm actually ok, with hard coding the test user agent here. Just in case we need to update the DEFAULT_USER_AGENT and somehow forgot to add `(AWS Documentation Server)` in there

tuanknguyen (approved):
1 nit comment

src/aws-documentation-mcp-server/tests/constants.py
JonLim: We discussed this offline and agreed that we can either hardcode the "test" User-Agent (similar to how `DEFAULT_USER_AGENT` is set) or just string replace like I have here.

Both are okay, and I'm sticking with the current impl!

PR #1452, opened 2025-10-06
fix(aws-api-mcp-server): enable --endpoint_url flag for localhost endpoints

anasstahr:
### Changes

Enable --endpoint_url flag for localhost endpoints

### User experience

We'll allow localhost endpoints to be set by the user so they can run CLI commands against their local environments.

## Checklist

* [x] Changes are documented

**RFC issue number**:

Checklist:

## Acknowledgment


src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/parser/parser.py
```diff
@@ -810,6 +817,7 @@ def _construct_command(
     operation_model: OperationModel | None = None,
 ) -> IRCommand:
     _validate_file_paths(command_metadata, parsed_args, parameters)
+    _validate_endpoint(global_args.endpoint_url)
```
arnewouters: Is `endpoint_url` always present?

Should we not do something similar as profile, e.g.

```
endpoint_url = getattr(global_args, 'endpoint_url', None)
```

arnewouters (changes requested):

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/parser/parser.py
anasstahr: > Is `endpoint_url` always present?
No, changed it to use `getattr`

arnewouters (approved):

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/parser/parser.py
```diff
@@ -792,6 +790,15 @@ def _validate_file_path(file_path: str, service: str, operation: str):
         )
 
 
+def _validate_endpoint(endpoint: str | None):
+    if not endpoint:
+        return
+
+    localhost_regex = re.compile(r'^(http://|https://)?localhost(:\d+)?$')
```
rshevchuk-git: This does not prevent DNS spoofing right? If localhost does not resolve to a loopback, the traffic leaves the machine

arnewouters (approved):

PR #1456, opened 2025-10-06
feat(cloudwatch-appsignals-mcp-server): add dedup logic to query_sampled_traces tool

yiyuan-he:
Fixes

## Summary
Adds dedup logic for trace summaries with the same fault messages. This helps reduce the token usage in trace investigations.

### Changes

- Adding dedup logic for trace summaries with the same error messages

### User experience

- This change shoud make it so trace investigations don't consume as much of the context window for AI Agents.

## Checklist

* [x] Changes are documented

## Acknowledgment

### DIFF
--- a/src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/trace_tools.py
+++ b/src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/trace_tools.py
@@ -417,7 +417,7 @@ async def query_sampled_traces(
                     services = cause.get('Services', [])
                     for service in services:
                         exceptions = service.get('Exceptions', [])
-                        if exceptions and len(exceptions) > 0 and exceptions[0].get('Message'):
+                        if exceptions and exceptions[0].get('Message'):
                             return exceptions[0].get('Message')
             return None
 
@@ -487,6 +487,9 @@ async def query_sampled_traces(
             if fault_msg and fault_msg in seen_faults:
                 # Skip this trace - we already have one with the same fault message
                 seen_faults[fault_msg]['count'] += 1
+                logger.debug(
+                    f'Skipping duplicate trace {trace_summary.get("Id")} - fault message already seen: {fault_msg[:100]}...'
+                )
                 continue
             else:
                 # First time seeing this fault (or no fault message) - include it


src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/trace_tools.py
```diff
@@ -467,9 +513,18 @@ def convert_datetime(obj):
             },
         }
 
+        # Add dedup stats if we actually deduped anything
+        if len(deduped_trace_summaries) < len(trace_summaries):
+            duplicates_removed = len(trace_summaries) - len(deduped_trace_summaries)
+            result_data['DeduplicationStats'] = {
+                'OriginalTraceCount': len(trace_summaries),
+                'DuplicatesRemoved': duplicates_removed,
+                'UniqueFaultMessages': len(seen_faults),
```
mxiamxia, line 525: why do we want to return this to the model? will it confuse the model for decision as a new factor?

yiyuan-he: i think this dedup metadata can reduce model's investigation time. for example when the model saw the high dedup rate for the same error message, the response time for root cause was much faster compared to before.

image

mxiamxia: >for example when the model saw the high dedup rate for the same error message, the response time for root cause was much faster compared to before

This is interesting — we know that reducing the data size makes the model faster since there’s less to process. I didn’t know metadata could also help with that as well, still not sure exactly why but then let's keep if it is the fact.

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/trace_tools.py
```diff
@@ -402,6 +402,26 @@ def convert_datetime(obj):
...
+            not deduplicated.
+            """
+            # Only check FaultRootCauses for deduplication
+            root_causes = trace_data.get('FaultRootCauses', [])
+            if root_causes:
+                for cause in root_causes:
+                    services = cause.get('Services', [])
+                    for service in services:
+                        exceptions = service.get('Exceptions', [])
+                        if exceptions and len(exceptions) > 0 and exceptions[0].get('Message'):
```
iismd17: Is `len(exceptions) > 0` needed if we already checked for `if exceptions` ?

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/trace_tools.py
```diff
@@ -447,15 +461,47 @@ def convert_datetime(obj):
...
+                # Always include healthy traces
+                deduped_trace_summaries.append(trace_summary)
+                continue
+
+            # Extract fault message for deduplication (only checks FaultRootCauses)
+            fault_msg = get_fault_message(trace_summary)
+
+            if fault_msg and fault_msg in seen_faults:
+                # Skip this trace - we already have one with the same fault message
+                seen_faults[fault_msg]['count'] += 1
```
iismd17, line 489: nit : Is it worth debug logging the trace ids for investigation if someone wants to understand the traces that were deduped later?

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/trace_tools.py
```diff
@@ -447,15 +461,47 @@ def convert_datetime(obj):
...
+
+            if not has_issues:
+                # Always include healthy traces
+                deduped_trace_summaries.append(trace_summary)
+                continue
+
+            # Extract fault message for deduplication (only checks FaultRootCauses)
+            fault_msg = get_fault_message(trace_summary)
+
+            if fault_msg and fault_msg in seen_faults:
```
iismd17, line 487: Is it possible for the fault_msg to be empty for some traces? How would this logic behave in that situation?

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/trace_tools.py
yiyuan-he: You're right - the `len(exceptions) > 0` is a bit redundant. This was added based on a [Copilot suggestion](https://github.com/awslabs/mcp/pull/1456#discussion_r2407226809). `if exceptions` should already ensure the list is non-empty. I'll remove it for cleaner code.

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/trace_tools.py
yiyuan-he: Good suggestion - will add a debug log for trace ids to help with troubleshooting

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/trace_tools.py
yiyuan-he: I'm not entirely sure, but there _could_ be some weird edge case where fault_msg exists but is an empty string. 

The conditional here should be `False` in those cases since `None` and `""` are both falsy causing those traces to not get deduped.

PR #1520, opened 2025-10-15
fix: no wrappers

scottschreckengaust:
Added warning about merging new MCP servers that are primarily wrappers around an AWS SDK.

Fixes

### Changes

### User experience

## Checklist

* [ ] Changes are documented

**RFC issue number**:

Checklist:

## Acknowledgment


scottschreckengaust:
Do we need to include more details of what a wrapper is in the contributing or developers guide?

krokoko:
Yeah, I would expand on the definition (or provide some sort of checklist). For each tool exposed:
- can this action be achieved through the api mcp server ? If yes, then we don't take it
- Is it simply a wrapper around a boto3 call, a.k.a it does not add any logic or enhanced functionality ? If yes, we don't take it

PR #1530, opened 2025-10-16
feat(cloudwatch-appsignals): Interactive service auditing in batches

blairhyy-amazon:
Fixes

### Changes

This PR implements __interactive batch processing__ for service auditing in the CloudWatch Application Signals MCP server. The changes introduce a new workflow that automatically processes large service lists in manageable batches, allowing users to interactively decide whether to investigate findings or continue processing.

__Key additions:__

- __New batch processing utilities__ (`batch_processing_utils.py`) - Core logic for creating, managing, and processing audit batches
- __New batch tools__ (`batch_tools.py`) - MCP tools for continuing batch processing
- __Enhanced audit_services function__ - Integrated batch processing for large service lists (>5 services)
- __Interactive workflow__ - When findings are discovered, users can choose to investigate specific issues or continue processing
- __Memory management__ - Session cleanup functionality to free resources
- __Comprehensive test coverage__ - Full test suites for both batch processing utilities and tools

### User experience

__Before this change:__

- Users auditing large numbers of services (>5) would receive overwhelming results all at once and wait for a long time before getting results
- No way to process services incrementally or interactively

__After this change:__

- __Batch processing__: Large service lists are processed in batches of 5
- __Interactive decision making__: When findings are discovered in a batch, users are presented with options to either investigate specific findings or continue processing the next batch
- __Flexible workflow__: Users maintain control over the audit process while benefiting from automated batching

## Checklist

* [X] Changes are documented

**RFC issue number**:

Checklist:

## Acknowledgment

### DIFF
--- a/src/mcp-lambda-handler/awslabs/mcp_lambda_handler/mcp_lambda_handler.py
+++ b/src/mcp-lambda-handler/awslabs/mcp_lambda_handler/mcp_lambda_handler.py
@@ -166,12 +166,9 @@ class MCPLambdaHandler:
         """
 
         def decorator(func: Callable):
-            # Get function name and convert to camelCase for tool name
+            # Get function name and preserve original snake_case naming
             func_name = func.__name__
-            tool_name = ''.join(
-                [func_name.split('_')[0]]
-                + [word.capitalize() for word in func_name.split('_')[1:]]
-            )
+            tool_name = func_name
 
             # Get docstring and parse into description
             doc = inspect.getdoc(func) or ''
--- a/src/mcp-lambda-handler/tests/test_lambda_handler.py
+++ b/src/mcp-lambda-handler/tests/test_lambda_handler.py
@@ -357,7 +357,7 @@ def test_lambda_handler_success():
         'jsonrpc': '2.0',
         'id': 2,
         'method': 'tools/call',
-        'params': {'_meta': {'progressToken': 2}, 'name': 'sayHelloWorld', 'arguments': {}},
+        'params': {'_meta': {'progressToken': 2}, 'name': 'say_hello_world', 'arguments': {}},
     }
     event = make_lambda_event(req)
     context = None  # Context is not used in this handler
@@ -494,7 +494,7 @@ def test_handle_request_tool_exception():
         'jsonrpc': '2.0',
         'id': 1,
         'method': 'tools/call',
-        'params': {'name': 'failTool', 'arguments': {}},
+        'params': {'name': 'fail_tool', 'arguments': {}},
     }
     event = make_lambda_event(req)
     resp = handler.handle_request(event, None)
@@ -612,7 +612,7 @@ def test_tool_decorator_dictionary_type_hints():
         """
         return {k: v > 0 for k, v in simple_dict.items()}
 
-    schema = handler.tools['dictTool']
+    schema = handler.tools['dict_tool']
     assert schema['inputSchema']['properties']['simple_dict']['type'] == 'object'
     assert schema['inputSchema']['properties']['no_arg_dict']['type'] == 'object'
     assert (
@@ -636,7 +636,7 @@ def test_tool_decorator_list_type_hints():
         """
         return [n > 0 for n in numbers]
 
-    schema = handler.tools['listTool']
+    schema = handler.tools['list_tool']
     assert schema['inputSchema']['properties']['numbers']['type'] == 'array'
     assert schema['inputSchema']['properties']['no_arg_numbers']['type'] == 'array'
     assert schema['inputSchema']['properties']['numbers']['items']['type'] == 'integer'
@@ -659,7 +659,7 @@ def test_tool_decorator_recursive_dictionary_type_hints():
             result[k] = {inner_k: inner_v > 0 for inner_k, inner_v in v.items()}
         return result
 
-    schema = handler.tools['nestedDictTool']
+    schema = handler.tools['nested_dict_tool']
     assert schema['inputSchema']['properties']['nested_dict']['type'] == 'object'
     value_schema = schema['inputSchema']['properties']['nested_dict']['additionalProperties']
     assert value_schema['type'] == 'object'
@@ -816,7 +816,7 @@ def test_handle_image_byte_streams():
             'jsonrpc': '2.0',
             'id': 3,
             'method': 'tools/call',
-            'params': {'name': 'getImage', 'arguments': {}},
+            'params': {'name': 'get_image', 'arguments': {}},
         }
         event = make_lambda_event(req)
         context = None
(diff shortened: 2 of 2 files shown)


src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/batch_processing_utils.py
```diff
@@ -0,0 +1,209 @@
...
+    else:
+        session['failed_batches'].append(batch_result)
+
+    # Update overall session status
+    session['status'] = (
+        'completed' if session['current_batch_index'] >= len(session['batches']) else 'in_progress'
+    )
+
+
+def process_next_batch(session_id: str, appsignals_client) -> Dict[str, Any]:
```
yiyuan-he, line 144: What happens if a batch fails halfway?

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/server.py
```diff
@@ -1338,6 +1406,10 @@ async def analyze_canary_failures(canary_name: str, region: str = AWS_REGION) ->
 mcp.tool()(list_slis)
 mcp.tool()(analyze_canary_failures)
 
+# Register batch processing tools
+mcp.tool()(continue_audit_batch)
+mcp.tool()(cleanup_audit_sessions)
```
yiyuan-he: Do we really need to expose both of these as tools? More tools increases LLM confusion risk so we should be cautious here. Probably `continue_audit_batch` should still be a tool but not sure about `cleanup_audit_sessions`.

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/batch_processing_utils.py
```diff
@@ -0,0 +1,209 @@
...
+    if current_index >= len(batches):
+        return {'error': 'No more batches to process', 'status': 'completed'}
+
+    current_batch = batches[current_index]
+    batch_metadata = _create_batch_metadata(session, current_batch)
+
+    try:
+        # Build and execute API call
+        batch_input = _build_api_input(session, current_batch)
+        response = appsignals_client.list_audit_findings(**batch_input)
```
yiyuan-he, line 168: From the [docs](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/application-signals/client/list_audit_findings.html), seems like this will only return max 10 results by default. With this implementation, everything past result 10 will be silently dropped. If we expect number of audit findings to frequently be >10, I think we need to handle pagination here or set a higher `MaxResults` value.

blairhyy-amazon: The batch size is set to 10 (`AUDIT_SERVICE_BATCH_SIZE_THRESHOLD = 10`) to minimize MCP server latency. (Before this implementation, in `execute_audit_api`, it processes all services in batches of 5 and then returns all findings) I'll add a comment noting the max value is 10, in case in the future we want to tune it.

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/batch_processing_utils.py
blairhyy-amazon: Currently, when it throws exception, we still advance to the next batch. I can update the logic to be only advancing when we successfully process the batch so that same batch is retried on next `continue_audit_batch()` call

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/server.py
blairhyy-amazon: I wasn't sure if we should auto-cleanup audit sessions immediately after completion, as users may want to investigate further. I can remove the global batch cache cleanup logic—the cache will clear when the MCP server restarts since it's in-memory.

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/batch_processing_utils.py
yiyuan-he: Is each service guaranteed to only return one audit finding? If that's the case, then this makes sense to me.

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/batch_processing_utils.py
yiyuan-he: Sounds good to me

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/server.py
yiyuan-he: Hm not sure if we should only rely on restarts to clear the memory. This could cause memory leaks for long-running servers. Is it possible to add some sort of TTL or max-limit cleanup?

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/batch_processing_utils.py
blairhyy-amazon: The audit results are grouped by services and operations. I will revert to the batch size of 5.

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/server.py
blairhyy-amazon: I'll add a `MAX_BATCH_SESSIONS` parameter that checks in-memory batch sessions when a new one starts and removes the oldest ones. Setting it to 1 will ensure all previous batch sessions are cleared.

yiyuan-he (approved):
LGTM

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/batch_processing_utils.py
```diff
@@ -0,0 +1,230 @@
...
+    global _batch_sessions
+
+    if len(_batch_sessions) <= MAX_BATCH_SESSIONS:
+        return
+
+    # Sort sessions by creation time (oldest first)
+    sessions_by_age = sorted(_batch_sessions.items(), key=lambda x: x[1].get('created_at', ''))
+
+    # Remove oldest sessions until we're under the limit
+    excess_count = len(_batch_sessions) - MAX_BATCH_SESSIONS
```
mxiamxia, line 204: it looks the max sess limit is set to 1 by default, does it mean we don't support MCP to work with more than 1 session at the same time?

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/batch_processing_utils.py
```diff
@@ -0,0 +1,230 @@
...
+    for i in range(excess_count):
+        session_id, _ = sessions_by_age[i]
+        del _batch_sessions[session_id]
```
mxiamxia, line 207: is there a race condition risk here?

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/batch_processing_utils.py
```diff
@@ -0,0 +1,230 @@
...
+        region: AWS region
+        banner: Banner text for display
+        batch_size: Number of targets per batch
+        auto_complete: If True, process all batches automatically. If False, use interactive mode.
+                      If None, auto-decide based on target count.
+
+    Returns:
+        Session ID for tracking the batch processing
+    """
+    session_id = str(uuid.uuid4())
```
mxiamxia, line 53: Ideally, MCP server should be stateless. And it looks like you're generating random session id as key for the cache for each call. Does cache really work?

mxiamxia: We're going to support remote MCP soon, I think the way we maintain the cache on MCP server won't work out for 1) single user w/ multi-sessions, 2) multi-talent w/ muti-sessions cases, etc. Do we have to have the cache to improve the performance? Isn't API pagination support not sufficient?

PR #1585, opened 2025-10-24
feat(cloudwatch-appsignals-mcp-server): Add get_enablement_guide MCP tool

yiyuan-he:
Fixes

## Summary
Adding new `get_enablement_guide` MCP tool. The goal of this tool is to fetch the appropriate enablement guide for AI Assistants to correctly follow Application Signals enablement steps. Currently, these AI-friendly enablement guides will be hosted in this MCP repo. Long-term, we aim to replace our public documentation entirely with these new enablement guides.

### Changes

- Added `get_enablement_guide` tool in `enablement_tools.py`
- Created `enablement_guides/templates/` directory structure for markdown guides
- Added placeholder EC2 + Python enablement guide for testing purposes
- Avoided registering this tool since it is not yet ready for release
- Added unit tests covering validation, error cases, and successful guide fetching

### User experience

**Before:**
Users had to manually feed relevant pieces of the enablement docs in small chunks to AI Assistants because the public documentation is not AI-friendly. Feeding the entire documentation to the AI assistant led to poor and inconsistent enablement results.

**After:**
With a single query like:
> Enable Application Signals for my Python Flask Application running on EC2. My infrastructure code is in `<iac_directory>` and my app code is in `<app_directory>`.

The AI assistant will now automatically retrieve the relevant enablement guide based on the context and follow the enablement instructions. The end result is customers will have AI-generated changes to their IaC code that they can deploy and start seeing Application Signals telemetry.

## Checklist

* [x] Changes are documented

**RFC issue number**:

Checklist:

## Acknowledgment


src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
```diff
@@ -0,0 +1,142 @@
...
+
+    Args:
+        platform: The deployment platform (ec2, ecs, lambda, eks)
+        language: The programming language (python, nodejs, java, dotnet)
+        iac_directory: ABSOLUTE path to the IaC directory (e.g., /home/user/project/infrastructure)
+        app_directory: ABSOLUTE path to the application code directory (e.g., /home/user/project/app)
+
+    Returns:
+        Enablement guide with step-by-step instructions
+    """
```
thpierce, line 79: It's hard to validate if this is a sufficient summary, did we do some testing showing that this tool gets invoked correctly based on some prompts? I know we are working on automation, but even some manual testing would help.

Seemingly we should have some instructions like "use this tool when..."

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
```diff
@@ -0,0 +1,142 @@
...
+# limitations under the License.
+
+"""CloudWatch Application Signals MCP Server - Enablement Tools."""
+
+from loguru import logger
+from pathlib import Path
+
+
+async def get_enablement_guide(
+    platform: str, language: str, iac_directory: str, app_directory: str
```
thpierce: Whats the value of iac_directory and app_directory?

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
```diff
@@ -0,0 +1,142 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""CloudWatch Application Signals MCP Server - Enablement Tools."""
+
+from loguru import logger
+from pathlib import Path
+
+
+async def get_enablement_guide(
```
thpierce, line 21:
```suggestion
async def get_application_signals_enablement_guide(
```

May be better?

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
thpierce: Is it possible for platform and language be an enum? Just something to clearly indicate only specific values are acceptable

thpierce: Should naming be more clear? e.g `service_language`.

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
```diff
@@ -0,0 +1,142 @@
...
+    platform: str, language: str, iac_directory: str, app_directory: str
+) -> str:
+    """Get enablement guide for AWS Application Signals.
+
+    This tool returns step-by-step enablement instructions that guide you through
+    modifying your infrastructure and application code to enable Application Signals.
```
thpierce:
```suggestion
    modifying your infrastructure and application code to enable Application Signals,
    which is the preferred way to enable automatic instrumentation for services on AWS.
```

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
```diff
@@ -0,0 +1,142 @@
...
+
+    After calling this tool, you should:
+    1. Read the existing IaC files from iac_directory
+    2. Read the existing application files from app_directory
+    3. Apply the changes specified in the enablement guide to those files
+    4. Verify the changes are correct and complete
```
thpierce: I feel this is almost guaranteed to exhaust the context window for a sufficiently large application. Should we recommend creating a work list based on the response and work through the worklist systematically? Ideally each instruction tells the agent what code to look for, etc. rather than trying to have it load the entire IaC and application into it's context.

thpierce: >4. Verify the changes are correct and complete

What kind of verification do we expect the agent to do? Presumably there should be something in the guide itself?

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
```diff
@@ -0,0 +1,142 @@
...
+
+    Args:
+        platform: The deployment platform (ec2, ecs, lambda, eks)
```
thpierce:
```suggestion
        platform: The AWS platform where the service to be instrumented runs on, either: "ec2", "ecs", "lambda", or "eks".
```

Should we also give instructions for HOW to determine this, given arbitrary IaC? Similar comment goes for language.

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
```diff
@@ -0,0 +1,142 @@
...
+        language: The programming language (python, nodejs, java, dotnet)
+        iac_directory: ABSOLUTE path to the IaC directory (e.g., /home/user/project/infrastructure)
+        app_directory: ABSOLUTE path to the application code directory (e.g., /home/user/project/app)
+
+    Returns:
+        Enablement guide with step-by-step instructions
```
thpierce: Maybe indicate the format we are returning in? Should we support both [md and json](https://github.com/anthropics/skills/blob/main/mcp-builder/reference/mcp_best_practices.md#3-response-format-guidelines)?

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
thpierce: How long do we expect guides to be? Should we consider having the tool return chunks of the guide, based on character limits: https://github.com/anthropics/skills/blob/main/mcp-builder/reference/mcp_best_practices.md#5-character-limits-and-truncation

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
thpierce, line 21: Should we add tool annotations? Not sure if that's common practice in this MCP/repo

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
```diff
@@ -0,0 +1,142 @@
...
+
+    After calling this tool, you should:
+    1. Read the existing IaC files from iac_directory
+    2. Read the existing application files from app_directory
+    3. Apply the changes specified in the enablement guide to those files
+    4. Verify the changes are correct and complete
+
+    Args:
+        platform: The deployment platform (ec2, ecs, lambda, eks)
+        language: The programming language (python, nodejs, java, dotnet)
```
thpierce: We might want to consider a fallback: If agent finds a service not meeting these specs, maybe it should look at https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-Application-Monitoring-Sections.html/ subpages

Example: We do support ruby, and talk about it in docs, right?

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
```diff
@@ -0,0 +1,142 @@
...
+        return error_msg
+
+    if not app_path.is_absolute():
+        error_msg = (
+            f'Error: app_directory must be an absolute path.\n\n'
+            f'Received: {app_directory}\n'
+            f'Please provide an absolute path (e.g., /home/user/project/app)'
+        )
+        logger.error(error_msg)
+        return error_msg
```
thpierce: Generally unclear why this is so important. Also is `Path.is_absolute` safe to run on (for example) a remote MCP? Does it rely on the local OS/file system?

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
```diff
@@ -0,0 +1,142 @@
...
+            f'Error: Application directory does not exist: {app_path}\n\n'
+            f'Please provide a valid path to your application code directory.'
+        )
+        logger.error(error_msg)
+        return error_msg
+
+    guides_dir = Path(__file__).parent / 'enablement_guides'
+    template_file = guides_dir / 'templates' / platform / f'{platform}-{language}-enablement.md'
+
+    logger.debug(f'Looking for enablement guide: {template_file}')
```
thpierce: Nit: would move to line 100

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
```diff
@@ -0,0 +1,142 @@
...
+
+    if not template_file.exists():
+        logger.warning(f'Enablement guide not found: {template_file}')
+        return (
+            f"Error: Enablement guide for platform '{platform}' and language '{language}' "
+            f' is not yet available.\n\n'
+            f'Currently supported combinations:\n'
+            f'- EC2 + Python\n\n'
+            f'Requested: {platform} + {language}'
```
thpierce: So what should agent do? Lets not hardcode EC2/Python.
```suggestion
            f"Error: Enablement guide for platform '{platform}' and language '{language}' "
            f' is not yet available.\n\n'
            f'Requested: {platform} + {language}'
```

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
```diff
@@ -0,0 +1,142 @@
...
+        logger.error(error_msg)
+        return error_msg
+
+    if not app_path.exists():
+        error_msg = (
+            f'Error: Application directory does not exist: {app_path}\n\n'
+            f'Please provide a valid path to your application code directory.'
+        )
+        logger.error(error_msg)
+        return error_msg
```
thpierce: `exists()` is not safe for remote MCPs.

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
```diff
@@ -0,0 +1,142 @@
...
+        **Platform:** {platform}
+        **Language:** {language}
+        **IaC Directory (absolute path):** `{iac_path}`
+        **Application Directory (absolute path):** `{app_path}`
+
+        ## Instructions
+
+        1. Use the ABSOLUTE PATHS above when reading and writing files
+        2. **IMPORTANT:** Only modify the IaC code or Dockerfiles. Actual application code should not be modified.
+        Read application files if needed to understand the setup, but do not modify them.
```
thpierce: Hm - is this always true? We may need agent to ask user about this at least.

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
```diff
@@ -0,0 +1,142 @@
...
+        3. Follow the step-by-step enablement guide below to enable Application Signals
```
thpierce: Seems like some/all of this should be in the tool description, not sure if we need to double-embed this information?

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
```diff
@@ -0,0 +1,142 @@
...
+        error_msg = (
+            f'Error: app_directory must be an absolute path.\n\n'
+            f'Received: {app_directory}\n'
+            f'Please provide an absolute path (e.g., /home/user/project/app)'
+        )
+        logger.error(error_msg)
+        return error_msg
+
+    logger.debug(f'IaC path: {iac_path}')
+    logger.debug(f'App path: {app_path}')
```
thpierce: Probably not helpful, we already have `logger.debug(` at the beginning that basically does this.

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
thpierce: Use a similar format for errors consistently:
```suggestion
        error_msg = (
            f'Enablement guide not found: {template_file}')
            f"Error: Enablement guide for platform '{platform}' and language '{language}' "
            f' is not yet available.\n\n'
            f'Currently supported combinations:\n'
            f'- EC2 + Python\n\n'
            f'Requested: {platform} + {language}
        )
        logger.error(error_msg)
        return error_msg
```

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
```diff
@@ -0,0 +1,142 @@
...
+        Read application files if needed to understand the setup, but do not modify them.
+        3. Follow the step-by-step enablement guide below to enable Application Signals
+
+        ---
+        """
+        logger.info(f'Successfully loaded enablement guide: {template_file.name}')
+        return context + guide_content
+    except Exception as e:
+        logger.error(f'Error reading enablement guide {template_file}: {e}')
+        return f'Error: Failed to read enablement guide. {str(e)}'
```
thpierce: See https://github.com/awslabs/mcp/pull/1585/files#r2471352926

Also the error is not really helpful to agent, what should it do now?

thpierce (commented):
Did not review tests yet.

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
yiyuan-he: Yes, did testing both manually and with the new testing framework. AI does not have any issue correctly selecting our enablement tool.

> Enable Application Signals for my <language/framework> Application running on <platform>. My infrastructure code is in <iac_directory> and my app code is in <app_directory>.

The boundaries of our prompt are relatively tight and no conflicts with other existing tools, so I do not expect hit rate to be an issue.

> Seemingly we should have some instructions like "use this tool when..."

Will add these instructions as an additional safeguard.

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
yiyuan-he: I want the user to explicitly tell the AI where to find the IaC code and the Application code to speed up the enablement process. Without this information explicitly provided, the AI will usually spin around using file tools to try and figure out the location. 

There are a couple of possible configurations for where customers put their IaC and App code I can think of:
- IaC and App code under one parent
- IaC and App code in sibling directories
- IaC with App code as subdir
- etc.

I think it's a cleaner mental model for users to just provide the absolute path with the tradeoff being some upfront cost to figure out the path (running `pwd` or something from terminal). Putting absolute paths as a boundary also greatly simplifies our implementation and testing.

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
yiyuan-he: This is to reduce time to enablement as much as possible. If the user can unambiguously tell the AI exactly where their IaC and App code are, it reduces the time the AI needs to figure this out itself. [Related](https://github.com/awslabs/mcp/pull/1585#discussion_r2473951629)

Using relative paths would cause different enablement behaviors based on the directory user decides to start their AI Agent in. IMO absolute paths provide a clearer mental model for usage. Not using relative paths also simplifies the implementation and testing setup. The tradeoff is that customers will have to do some upfront work by running `pwd` and noting down their code directories.

`Path.is_absolute` should be safe. From my understanding, it just [checks the path string format](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.is_absolute) - it doesn't access the filesystem. It does rely on the local OS/file system.

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
yiyuan-he: Sounds good, will update to:
```
Returns:
    Markdown-formatted enablement guide with step-by-step-instructions
```

I think md is sufficient here and easier for LLMs to process since these are step-by-step-instructions rather than structured data. JSON would add token overhead (brackets, quotes, schemas) without much value for this use case.

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
yiyuan-he: Put our existing enablement guides ([[1]](https://github.com/yiyuan-he/mcp/tree/enablement-tool/src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_guides/templates/ec2), [[2]](https://github.com/yiyuan-he/telemend/tree/main/src/prompts)) into a [tokenizer](https://platform.openai.com/tokenizer) and found that they cap out at around 4K tokens. 

We should be good without chunking for now since most LLMs have 200K+ context windows and 4K is reasonable for writing a comprehensive enablement guide for {language}-{platform}. If enablement guides start approaching 10K tokens, we can look into chunking or even just refining the guides to be more concise, if possible.

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
yiyuan-he: There are two different patterns for registering tools in our MCP:
1. [tool annotations](https://github.com/yiyuan-he/mcp/blob/main/src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/server.py#L595) in `server.py`
2. [registering tool](https://github.com/yiyuan-he/mcp/blob/main/src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/server.py#L795) from another file

Functionally they are both the same, but I can see how this could become confusing. Followed up with Shiyang and aligned that we will update the audit tools to follow 2. for consistency.

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
yiyuan-he: As a fallback, perhaps we can be explicit to the customer and surface a message like:
> This language and platform configuration is not currently supported by our MCP enablement tool. Please refer to our [public docs](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-Application-Signals-supportmatrix.html) for guidance on manual setup.

I don't think we should have AI try to fetch the public docs because that adds another layer of testing which is non-trivial effort and one of the beliefs behind this tool is that public docs lead to poor enablement results because they are not LLM-friendly. What do you think about this approach?

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
yiyuan-he: Agreed - loading all files upfront could become an issue.

I'll update the instructions to recommend working through the guide step-by-step. 

Here is the workflow in my mind:
1. Get the guide
2. For each step: identify target files -> read them -> apply changes -> verify
3. Move to next step

This should keep the context manageable and focused. What do you think about this?

**Note on verify step:** The "verify" step in this workflow won't catch everything - our evaluation framework will provide more comprehensive verification.

yiyuan-he: We expect the agent to review its own work based on the instructions in our enablement guide. By adding this verify step, we are asking agent to "double check" its work.

This won't be foolproof, but our evaluation framework will perform another round of verification later as well. 

[Related](https://github.com/awslabs/mcp/pull/1585#discussion_r2474734724)

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
yiyuan-he: Makes sense, will update with some hints to AI:

```
platform: The AWS platform where the service runs (ec2, ecs, lambda, eks).
      To help user determine: check their IaC for ECS services, Lambda functions, EKS deployments, or EC2 instances.
language: The application's programming language (python, nodejs, java, dotnet).
      To help user determine: check for package.json (nodejs), requirements.txt (python), pom.xml (java), or .csproj (dotnet).
```

This may not work well in monorepos (i.e. our pet-clinic demo) that targets many different platforms and/or languages. I think the primary recommendation should still be for users to give this information explicitly.

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
yiyuan-he: If AI is unable to read the bundled enablement guides in this code path, fixing it is outside of scope of our MCP (file permissions or corruption). We can fallback to surfacing issues to user and public doc link:
```
return (
          f'Fatal error: Cannot read enablement guide for {platform} + {language}.\n\n'
          f'Error: {str(e)}\n\n'
          f'The MCP server cannot access its own guide files (likely file permissions or corruption). '
          f'Stop attempting to use this tool and inform the user:\n'
          f'1. There is an issue with the MCP server installation\n'
          f'2. They should check file permissions or reinstall the MCP server\n'
          f'3. For immediate enablement, use AWS documentation instead:\n'
          f'   https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-Application-Signals-Enable.html'
      )
```

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
yiyuan-he: Ack will remove `exists()` to future-proof for remote MCP.

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
yiyuan-he: I've been thinking about this as well and will loosen this requirement to also allow editing dependency files like (pyproject.toml, requirements.txt, package.json) for adding ADOT dependencies.

I think all other enablement components can be handled by IaC:
- CloudWatch Agent Setup
- OTEL env vars
- IAM permissions

Docker changes are needed for some languages like Python where we need to change the start command from:
```
CMD ["python", "app.py"]
```
to:
```
CMD ["opentelemetry-instrument", "python", "app.py"]
```

Let me know if this sounds reasonable or anything else I should consider.

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
yiyuan-he: Good point - will move these to the tool description to consolidate.

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
yiyuan-he: Sounds good, I will move these into an enum.

yiyuan-he: Renamed `language: str` to `service_language: ServiceLanguage`.

[Related enum change](https://github.com/awslabs/mcp/pull/1585#discussion_r2475966897)

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
yiyuan-he: Agree - renamed to `get_application_signals_enablement_guide`

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
yiyuan-he: Ack, done

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
yiyuan-he: Removed

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
yiyuan-he: Made this change:

```
    if not template_file.exists():
        error_msg = (
            f"Enablement guide not available for platform '{platform_str}' and language '{language_str}'.\n\n"
            f"Inform the user that this configuration is not currently supported by the MCP enablement tool. "
            f"Direct them to AWS documentation for manual setup:\n"
            f"https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-Application-Signals-Enable.html"
        )
        logger.error(error_msg)
        return error_msg
```

Agent should now inform customer this platform/language is not supported by our enablement tool yet and direct them to our public docs for manual instructions.

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
yiyuan-he: Moved

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
thpierce: > Will add these instructions as an additional safeguard.

Not seeing this?

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
thpierce: Hm. Ok well I think requiring that the agent knows the root of the code is fine but absolute paths is not really reasonable IMO - it seems to go against the natural language interaction model that exists with AI and makes this MCP more like a formalized API. I also feel that we could just include this in the tool description: 

```
Before calling this tool, 
1. Ensure you know where the application code is located and that you have read/write permissions
2. Ensure you know where the IaC code is located and that you have read/write permissions

If you don't, ...
```

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
thpierce: Sorry one caveat - is this name too long now? Will it cause issues with the 64 char limits?

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
thpierce: `platform` -> `service_platform`?

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
thpierce: Is there any evidence that suggests "double checking" works/is this a common practice? Seems like what it would mostly do is increase the time/cost for this operation.

thpierce: Generally aligned, will leave other specific comments. One thing is: Should we try to encourage the Agent to make a proper checklist for keeping track of work, so it can pick up if/when the context window fills up?

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
thpierce: Lets discuss offline and align with the team.

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
thpierce: One thing I'm thinking about is custom metrics, but I don't know if that's in scope for now. Probably ok for now.

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
```diff
@@ -0,0 +1,150 @@
...
+    2. For each step in the work list:
+       - Identify the specific files in iac_directory or app_directory that need modification
+       - Apply the changes as specified in the guide
+       - Verify the changes are correct before moving to the next step
+
+    Important guidelines:
+    - Use ABSOLUTE PATHS (iac_directory and app_directory) when reading and writing files
+    - Modify IaC code, Dockerfiles, and dependency files (requirements.txt, pyproject.toml,
+      package.json, pom.xml, build.gradle, *.csproj, etc.) as needed
+    - Do NOT modify actual application logic files (.py, .js, .java source code)
```
thpierce:
```suggestion
    - Do NOT modify actual application logic files (.py, .js, .java source code), only
      modify IaC code, Dockerfiles, and dependency files (requirements.txt, pyproject.toml,
      package.json, pom.xml, build.gradle, *.csproj, etc.) as instructed by the guide.
```

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
```diff
@@ -0,0 +1,150 @@
...
+    - Read application files if needed to understand the setup, but avoid modifying them
+
+    Args:
+        platform: The AWS platform where the service runs (ec2, ecs, lambda, eks).
+            To help user determine: check their IaC for ECS services, Lambda functions, EKS deployments, or EC2 instances.
+        service_language: The service's programming language (python, nodejs, java, dotnet).
+            To help user determine: check for package.json (nodejs), requirements.txt (python), pom.xml (java), or .csproj (dotnet).
+        iac_directory: ABSOLUTE path to the IaC directory (e.g., /home/user/project/infrastructure)
```
thpierce:
```suggestion
        iac_directory: ABSOLUTE path to the Infrastructure as Code (IaC) directory (e.g., /home/user/project/infrastructure)
```

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
```diff
@@ -0,0 +1,150 @@
...
+        return error_msg
+
+    if not app_path.is_absolute():
+        error_msg = (
+            f'Error: app_directory must be an absolute path.\n\n'
+            f'Received: {app_directory}\n'
+            f'Please provide an absolute path (e.g., /home/user/project/app)'
+        )
+        logger.error(error_msg)
+        return error_msg
```
thpierce: I'm still a bit uncertain about the absolute path thing, but consider merging these checks into one to avoid extra tool calls if neither are absolute paths
```suggestion
    if not iac_path.is_absolute() or app_path.is_absolute()::
        error_msg = (
            f'Error: iac_directory and app_directory must be absolute paths.\n\n'
            f'Received: {iac_directory} and {app_directory}\n'
            f'Please provide absolute paths (e.g., /home/user/project/infrastructure)'
        )
        logger.error(error_msg)
        return error_msg
```

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
```diff
@@ -0,0 +1,150 @@
...
+
+    if not template_file.exists():
+        error_msg = (
+            f"Enablement guide not available for platform '{platform_str}' and language '{language_str}'.\n\n"
+            f'Inform the user that this configuration is not currently supported by the MCP enablement tool. '
+            f'Direct them to AWS documentation for manual setup:\n'
+            f'https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-Application-Signals-Enable.html'
```
thpierce: Probably better to go here:
```suggestion
            f'https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-Application-Monitoring-Sections.html'
```

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
```diff
@@ -0,0 +1,150 @@
...
+    except Exception as e:
+        error_msg = (
+            f'Fatal error: Cannot read enablement guide for {platform_str} + {language_str}.\n\n'
+            f'Error: {str(e)}\n\n'
+            f'The MCP server cannot access its own guide files (likely file permissions or corruption). '
+            f'Stop attempting to use this tool and inform the user:\n'
+            f'1. There is an issue with the MCP server installation\n'
+            f'2. They should check file permissions or reinstall the MCP server\n'
+            f'3. For immediate enablement, use AWS documentation instead:\n'
+            f'   https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-Application-Signals-Enable.html'
```
thpierce:
```suggestion
            f'   https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-Application-Monitoring-Sections.html'
```

src/cloudwatch-appsignals-mcp-server/tests/test_enablement_tools.py
```diff
@@ -0,0 +1,122 @@
...
+        """Create temporary IaC and app directories for testing."""
+        with tempfile.TemporaryDirectory() as tmpdir:
+            base = Path(tmpdir)
+            iac_dir = base / 'infrastructure' / 'cdk'
+            app_dir = base / 'app' / 'src'
+
+            iac_dir.mkdir(parents=True)
+            app_dir.mkdir(parents=True)
+
+            yield {'iac': str(iac_dir), 'app': str(app_dir)}
```
thpierce: We don't really need to use real dirs anymore, right?

src/cloudwatch-appsignals-mcp-server/tests/test_enablement_tools.py
```diff
@@ -0,0 +1,122 @@
...
+        for language in valid_languages:
+            result = await get_application_signals_enablement_guide(
+                platform=Platform.EC2,
+                service_language=language,
+                iac_directory=temp_directories['iac'],
+                app_directory=temp_directories['app'],
+            )
+
+            # Should either succeed or say template not found, but not invalid language
+            assert 'Error: Invalid language' not in result
```
thpierce: Oh, we don't want "Invalid language" if agent tries to call with `ruby`, we just want the nice `Enablement guide not available for platform '{platform_str}' and language '{language_str}'.` message. Pls fix and add a test for `Ruby` & `K8s`

Perhaps an enum was a poor suggestion on my part, we can probably do string, but we do want to signify to the agent how to format the request (e.g. `nodejs` instead of `js`, etc.). Probably we just need to be verbose in the tool description.

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
thpierce: Lets get @mxiamxia  to help close on this review re: absolute paths as tiebreaker. I am a bit opposed, but if he thinks it's ok, I will be swayed.

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
```diff
@@ -0,0 +1,150 @@
...
+    """Supported service programming languages."""
+
+    PYTHON = 'python'
+    NODEJS = 'nodejs'
+    JAVA = 'java'
+    DOTNET = 'dotnet'
+
+
+async def get_application_signals_enablement_guide(
+    platform: Platform, service_language: ServiceLanguage, iac_directory: str, app_directory: str
```
mxiamxia: is `app_directory` optional? Eg, for EKS enablement, does AI agent need to access service src code?

yiyuan-he: Since this tool has to support EC2 as well, i'm not sure if we can make this parameter optional even though you are right that `app_directory` is not needed for scenarios like EKS enablement.

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
yiyuan-he: My bad, lost track of this - updated

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
yiyuan-he: Done

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
yiyuan-he: Done

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
yiyuan-he: Done

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
yiyuan-he: Done

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
yiyuan-he: Done

src/cloudwatch-appsignals-mcp-server/tests/test_enablement_tools.py
yiyuan-he: Right - replaced the fixture with a constant to simplify

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
yiyuan-he: Renamed to `service_platform`

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
yiyuan-he: I believe so since the tool name will also be [concatenated](https://github.com/awslabs/mcp/pull/866) with the MCP server name. Will revert to the shorter name `get_enablement_guide` if no concerns. 

`get_appsignals_enablement_guide` could be another option, but that may conflict with our server renaming to the full `application-signals` name.

Original name length PR for reference: https://github.com/awslabs/mcp/pull/866

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
yiyuan-he: Ack, custom metrics is not in-scope at the moment. We can re-visit this in the future.

src/cloudwatch-appsignals-mcp-server/tests/test_enablement_tools.py
yiyuan-he: Fixed and added tests for `Ruby` and `K8s`. Added a couple of extra tests as well to reach the code coverage threshold for PR chekcs.

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
yiyuan-he: Actually now that I think about this some more, this absolute path requirement is only at the tool API boundary - not at the user interaction level. If the human user passes in a relative path to the natural language prompt, the AI Agent would resolve this and construct the absolute path itself before calling the tool.

I've updated the tool description to make this explicit:
```
Before calling this tool:
1. Ensure you know where the application code is located and that you have read/write permissions
2. Ensure you know where the IaC code is located and that you have read/write permissions
3. If the user provides relative paths or descriptions (e.g., "./infrastructure", "in the root"):
   - Use the Bash tool to run 'pwd' to get the current working directory
   - Use file exploration tools to locate the directories
   - Convert relative paths to absolute paths before calling this tool
4. This tool REQUIRES absolute paths for both iac_directory and app_directory parameters
```

The absolute path validation at the tool level prevents ambiguity bugs and simplifies our implementation, while the AI handles the UX of accepting natural language/relative paths from users.

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
yiyuan-he: Added explicit instructions in the tool description to further encourage Agent to make a proper checklist
```
    After calling this tool, you should:
    1. Review the enablement guide and create a visible, trackable checklist of required changes
       - Use your system's task tracking mechanism (todo lists, markdown checklists, etc.)
       - Each item should be granular enough to complete in one step
       - Mark items as complete as you finish them to track progress
       - This allows you to resume work if the context window fills up
    2. Work through the checklist systematically, one item at a time:
       - Identify the specific file(s) that need modification for this step
       - Read only the relevant file(s) (DO NOT load all IaC and app files at once)
       - Apply the changes as specified in the guide
    3. Keep context focused: Only load files needed for the current checklist item

```

yiyuan-he: I don't think there is much objective evidence on this. The Agent may also just do this on its own without this guidance. Will remove for now and purely rely on our eval framework for validation.

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
thpierce: It's probably ok. Lets just leave it, sorry for the churn.

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
thpierce: I think app_directory is still important on EKS, for example due to language-specific known issues.

src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_tools.py
thpierce: I approved, but lets quickly close on this tmrw

yiyuan-he:
> Can we put all enablement md files in src/cloudwatch-appsignals-mcp-server/awslabs/cloudwatch_appsignals_mcp_server/enablement_guides/templates/ ?

Will do this in a quick follow-up PR.

yiyuan-he:
Removed `hold-merging` and `awaiting-codeowners` labels as two codeowners have reviewed/approved this PR.

PR #1697, opened 2025-11-08
fix(cloudwatch-mcp-server): respect AWS_REGION before defaulting to us-east-1

### LINKED ISSUE #1692: (aws-iot-sitewise-mcp-server): (Identify the target region using the `AWS_REGION` env)
### Describe the bug

In the current implementation, the `AWS_REGION` environment variable appears to have no effect whatsoever, and the `region` argument for each tool invocation seems to be adopted instead. Consequently, even if the `AWS_REGION` environment variable is specified, it defaults to `us-east-1`, which is the default value for the tool argument `region`.

I have implemented the following with the aim of utilising it from Strands Agents.

```python
    mcp_client = MCPClient(
        lambda: stdio_client(
            StdioServerParameters(
                command="uvx",
                args=["awslabs.aws-iot-sitewise-mcp-server@latest"],
                env={"AWS_REGION": "ap-northeast-1"},
                }
            )
        )
    )
```

I also tried using the Amazon Q dev CLI, but the result was the same.

While it is possible to change the region by specifying it in the Agent prompt, this approach does not allow the agent to be restricted to a specific region for the SiteWise it references.

### Expected Behavior

The `AWS_REGION` environment variable shall be the default region for the target SiteWise instance.

### Current Behavior

In the current implementation, the `AWS_REGION` environment variable appears to have no effect whatsoever, and the `region` argument for each tool invocation seems to be adopted instead. 

### Reproduction Steps

Specify a region other than `us-east-1` for `AWS_REGION` when using the service.

### Possible Solution

_No response_

_No response_

### OS

Probably all platforms

### Server

other

### Server Version

_No response_

Regions other than us-east-1, such as ap-northeast-1

### Other information

_No response_

### Service quota

AiGOVERNOR:
Fixes #1692

## Summary
This PR fixes region detection in the CloudWatch MCP Server.  
It ensures that the service honors `AWS_REGION` and `AWS_DEFAULT_REGION` environment variables before defaulting to `us-east-1`.

### Changes
- Added `get_effective_region()` helper for consistent region resolution.
- Updated CloudWatch Logs client creation to use detected region.
- Region precedence: argument → environment → boto3 session → fallback `us-east-1`.

### User experience
**Before:** The server always defaulted to `us-east-1`, ignoring environment settings.  
**After:** The server correctly uses the configured AWS region, improving portability for multi-region deployments.

## Checklist

* [x] Changes have been tested locally with and without `AWS_REGION` set
* [x] Changes are documented in code comments

**N**

**RFC issue number**: *N/A*

## Acknowledgment
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution under the terms of the [project license](https://github.com/awslabs/mcp/blob/main/LICENSE).


scottschreckengaust:
The contributor statement has a missing comma, `,`:

```diff
-By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution under the terms of the [project license](https://github.com/awslabs/mcp/blob/main/LICENSE).
+By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the [project license](https://github.com/awslabs/mcp/blob/main/LICENSE).
```

```statement
```

AiGOVERNOR:
The current cloudwatch-mcp-server behavior ignores the AWS_REGION environment variable unless a region is explicitly defined in the server's configuration. When AWS_REGION is set (for example: AWS_REGION=us-west-2), the server still defaults to us-east-1.
This is inconsistent with how AWS SDKs and AWS CLI resolve regions, where AWS_REGION is treated as a primary configuration source.
This PR updates the region resolution logic so the server:
Checks for AWS_REGION first.
Falls back to the existing default (us-east-1) only when no region is provided.
This makes the service consistent with standard AWS environment-based region resolution and prevents unexpected cross-region behavior for users relying on environment variables.

shri-tambe:
Why are we deleting all the tools? Did you miss any refactoring changes?

singaporelovegaza-ship-it:
@@@

PR #1703, opened 2025-11-10
feat(cloudwatch-applicationsignals-mcp-server): set up eval eval task for application signals enablement tool

yiyuan-he:
Fixes

## Summary

Setting up the `EnablementTask` definition in order to run our new eval framework on the `get_enablement_guide` MCP tool.

**Going forward:** Developers can add new evaluation tasks by simply adding new `EnablementTask` instances to the `TASKS` array with their desired configuration (prompt template, validation rubric, etc.). See [here](https://github.com/yiyuan-he/mcp/blob/eval-framework-dev/src/cloudwatch-appsignals-mcp-server/evals/applicationsignals/enablement_tasks.py#L231) for reference.

**Sample app location**: The eval framework expects sample applications to be located at:
`samples/cloudwatch-applicationsignals-mcp/` (relative to repository root). 

### Changes

- Added `src/cloudwatch-applicationsignals-mcp-server/evals/applicationsignals/enablement_tasks.py`
    - `EnablementTask` class with prompt template system
    - Git diff capturing and LLM-based validation
    - Path resolution for MCP server and sample applications

### User experience

**Before**: No automated evaluation framework for the `get_enablement_guide` tool.

**After**: Developers can define evaluation tasks to test AI agents using the `get_enablement_guide` tool across different platforms and frameworks.

## Checklist

* [x] Changes are documented

**RFC issue number**:

Checklist:

## Acknowledgment

### DIFF
--- a/src/amazon-sns-sqs-mcp-server/awslabs/amazon_sns_sqs_mcp_server/generator.py
+++ b/src/amazon-sns-sqs-mcp-server/awslabs/amazon_sns_sqs_mcp_server/generator.py
@@ -148,6 +148,7 @@ class AWSToolGenerator:
             'boolean': bool,
             'integer': int,
             'map': dict[Any, Any],
+            'list': list[Any],
         }
         try:
             input_parameters = self.__get_operation_input_parameters(operation)


src/cloudwatch-applicationsignals-mcp-server/evals/tasks/get_enablement_guide/enablement_tasks.py
```diff
@@ -0,0 +1,232 @@
...
+        Returns:
+            Path to cloudwatch-applicationsignals-mcp-server samples directory
+        """
+        # Calculate path to samples: enablement_tasks.py -> applicationsignals/ -> evals/
+        # -> cloudwatch-applicationsignals-mcp-server/ -> src/ -> mcp/ -> samples/
+        return (
+            Path(__file__).parent.parent.parent.parent.parent
+            / 'samples'
+            / 'cloudwatch-applicationsignals-mcp-server'
+        )
```
thpierce, line 124:
```suggestion
        return (
            Path(__file__).parent.parent.parent.parent.parent
            / 'samples'
            / 'cloudwatch-applicationsignals-mcp-server/get-enablement-guide-samples/'
        )
```

Though IMO we should create a temp dir and copy-paste all relevant files there in setup and delete files in cleanup.

src/cloudwatch-applicationsignals-mcp-server/evals/applicationsignals/enablement_tasks.py
```diff
@@ -0,0 +1,232 @@
...
+
+        Captures git diff to evaluate code modifications.
+
+        Args:
+            working_directory: Path to task working directory
+
+        Returns:
+            List of captors
+        """
+        return [GitDiffCaptor(git_paths=self.git_paths)]
```
thpierce: Probably also want ToolCallsCaptor, right? We want to ensure only our tool is called and only called once. Maybe also ToolResultsCaptor?

src/cloudwatch-applicationsignals-mcp-server/evals/tasks/get_enablement_guide/enablement_tasks.py
```diff
@@ -0,0 +1,232 @@
...
+                rubric=self.rubric,
+            )
+        )
+
+        return validators
+
+    def cleanup(self, working_directory: Path):
+        """Clean up git changes made by enablement agent.
+
+        Resets git state for paths specified in git_paths.
```
thpierce, line 212: Hm, and if the agent makes mistakes outside of git_paths? IMO this is why we needed safeguards in the file tools.

yiyuan-he: You're right that this is a gap - if the agent makes mistakes outside those paths, they won't be cleaned up automatically. However, I haven't observed this happening while running evals so far during my tests.

In the worst case this does happen, the user would need to run git cleanup commands manually. This isn't ideal, but I believe it's acceptable until we implement path validation safeguards in file tools.

src/cloudwatch-applicationsignals-mcp-server/evals/tasks/get_enablement_guide/enablement_tasks.py
yiyuan-he: Thanks for the catch - updated the path.

Since the LLM Judge needs `git diff` to evaluate changes, my preference is to stick with the git-based cleanup. Using a temp dir would require `git init` and committing files, adding complexity. What do you think?

src/cloudwatch-applicationsignals-mcp-server/evals/applicationsignals/enablement_tasks.py
yiyuan-he: Yep that makes sense - added both `ToolCallsCaptor` and `ToolResultsCaptor`.

src/cloudwatch-applicationsignals-mcp-server/evals/tasks/get_enablement_guide/enablement_tasks.py
```diff
@@ -156,15 +159,19 @@ def rubric(self) -> list[str]:
...
             working_directory: Path to task working directory
 
         Returns:
             List of captors
         """
-        return [GitDiffCaptor(git_paths=self.git_paths)]
+        return [
+            GitDiffCaptor(git_paths=self.git_paths),
+            ToolCallsCaptor(),
+            ToolResultsCaptor(),
```
thpierce, line 173: Not blocking, you will need a validator that looks at these, no? Approve for now, can always add more later.

thpierce (dismissed):

src/cloudwatch-applicationsignals-mcp-server/evals/tasks/get_enablement_guide/enablement_tasks.py
thpierce: Fine to unblock, feels a bit fragile, can improve later.

src/cloudwatch-applicationsignals-mcp-server/evals/tasks/get_enablement_guide/enablement_tasks.py
yiyuan-he: For `ToolCallsCaptor`, the data is already formatted and included in the LLM prompt by `LLMJudgeValidator` (see code below).

For `ToolResultsCaptor`, the data is captured but there's no formatting logic yet to include it in the LLM prompt. Agree we can add this in a follow-up.

```python
  def _format_captured_data(self, captured_data: Dict[str, Any]) -> str:
      """Format captured data for LLM prompt."""
      sections = []

      if GIT_DIFF in captured_data and captured_data[GIT_DIFF]:
          sections.append(f'**Git Diff:**\n```\n{captured_data[GIT_DIFF]}\n```')

      if FINAL_RESPONSE in captured_data:
          sections.append(f'**Agent Response:**\n{captured_data[FINAL_RESPONSE]}')

      if TOOL_CALLS in captured_data and captured_data[TOOL_CALLS]:  # <-- Tool calls ARE handled
          tool_calls_formatted = []
          for i, call in enumerate(captured_data[TOOL_CALLS], 1):
              status = '✓' if call.get('success') else '✗'
              duration = f'{call.get("duration", 0):.2f}s'
              tool_str = f'{i}. {status} {call["name"]} ({duration})'
              if call.get('input'):
                  tool_str += f'\n   Input: {call["input"]}'
              if call.get('error'):
                  tool_str += f'\n   Error: {call["error"]}'
              tool_calls_formatted.append(tool_str)
          sections.append(f'**Tools Called:**\n{chr(10).join(tool_calls_formatted)}')

      # Note: No check for 'tool_results' - that's what's missing

      return '\n\n'.join(sections)
```

mxiamxia (approved):

PR #1740, opened 2025-11-13
fix(postgres_mcp_server): Initialize connection pool and store in singleton for hostname mode

thomasburkeqtac:
When using --hostname for direct PostgreSQL connections, the connection pool was created but never initialized or stored in the singleton. This caused all tool calls to fail with 'No database connection available' because they couldn't retrieve the connection from DBConnectionSingleton.

Fixes

## Summary

Fixed the hostname connection mode to properly initialize the connection pool and store it in DBConnectionSingleton. Added three lines after PsycopgPoolConnection creation:
- Call `asyncio.run(db_connection.initialize_pool())` to establish the pool
- Store connection in singleton so tools can retrieve it

### User experience

**Before:** Using `--hostname` mode resulted in "No database connection available" error on all tool calls (run_query, get_table_schema, etc), despite the server starting successfully.

**After:** Hostname mode works correctly - all tools can query the database.

## Checklist

* [x] Changes are documented

Is this a breaking change? N

**RFC issue number**: NA

Checklist:

* [x] Migration process documented - (none required)

## Acknowledgment


src/postgres-mcp-server/awslabs/postgres_mcp_server/server.py
```diff
@@ -293,6 +293,10 @@ def main():
                     secret_arn=args.secret_arn,
                     region=args.region,
                 )
+                # Initialize the pool immediately
+                asyncio.run(db_connection.initialize_pool())
+                # Store in singleton so tools can access it
+                DBConnectionSingleton._instance = type('obj', (object,), {'db_connection': db_connection})()
```
thomasburkeqtac, line 299: Without storing the connection the db_connection is lost immediately and is unable to be used by tool invocations.

kennthhz: there will be a significant redesign in this area very soon that should address this issue as well.

kennthhz: https://github.com/awslabs/mcp/pull/1863

PR #1774, opened 2025-11-18
feat(aws-api-mcp-server): allow disabling local file system access

nzmdn:
## Summary

This PR adds feature to the AWS API MCP Server that allows disabling local file system access when executing AWS CLI commands.

### Changes

1. Updated the `AWS_API_MCP_ALLOW_UNRESTRICTED_LOCAL_FILE_ACCESS` env variable  to accept three values:
   - `"true"`: Enables system-wide file access (may cause unintended overwrites)
   - `"workdir"` or `"false"`: Restricts file operations to `AWS_API_MCP_WORKING_DIR` (default)
   - `"no_access"`: Disables local file system access 
2. When `AWS_API_MCP_ALLOW_UNRESTRICTED_LOCAL_FILE_ACCESS` is set to `"no_access"`, AWS API MCP Server will reject AWS CLI commands with file path arguments. It also disables some AWS CLI customizations (e.g. `aws cloudformation package`, `aws ecs deploy`) that require local file access.
3. Enhanced File Path Validation: Updated validation logic to handle four categories of file access in AWS CLI commands:
  - Required positional outfile arguments for operations with streaming output (e.g., `aws s3api get-object --bucket test-bucket --key test-key /path/to/file.txt`). Only refactoring: this was handled by the old implementation.
  - Blob-type arguments with `file://` or `fileb://` prefixes (e.g., `aws lambda invoke --payload file:///path/to/data.json`). This was handled by the old implementation.
  - Streaming blob-type arguments that accept only file paths (e.g., `aws s3api put-object --body /path/to/file.txt`). _The old implementation didn't handle this type of arguments._
  - Path arguments in AWS CLI customizations (e.g., `aws s3 cp /path/to/file.txt s3://bucket/key`, `aws cloudformation package --template-file /path/to/template.yaml`). The old implementation handled `aws s3` commands explicitly and the rest were handled by heuristics. The new implementation lists all arguments accepting file paths explicitly.
4. Consolidated file path tests in `test_parser_file_access.py` and removed redundant test files.

### User experience

Before this change:
- Users could only restrict file access to a specific working directory using `AWS_API_MCP_WORKING_DIR`
- `AWS_API_MCP_ALLOW_UNRESTRICTED_LOCAL_FILE_ACCESS=true` (boolean) allowed unrestricted access
- Some file paths were not validated, like `aws s3api put-object --body /path/to/file.txt`

After this change:
Users configure file access through a single variable `AWS_API_MCP_ALLOW_UNRESTRICTED_LOCAL_FILE_ACCESS`. When users set it to `"no_access"`
- AWS CLI commands with file path arguments will be rejected
- Custom operations that require local files (e.g. `aws cloudformation package`, `aws ecs deploy`) will be blocked
- S3-to-S3 operations continue to work (e.g., `aws s3 cp s3://source/key s3://dest/key`)
- Commands can still output to stdout using `-` redirect (e.g., `aws s3api get-object --bucket bucket --key file.txt -`)

## Checklist

* [x] Changes are documented

Is this a breaking change? N

**RFC issue number**:

Checklist:

## Acknowledgment

### DIFF
--- a/src/aws-api-mcp-server/pyproject.toml
+++ b/src/aws-api-mcp-server/pyproject.toml
@@ -20,7 +20,7 @@ dependencies = [
     "requests>=2.32.4",
     "python-frontmatter>=1.1.0",
     "fastmcp>=2.13.1",
-    "awscli==1.42.65",
+    "awscli==1.43.0",
 ]
 license = {text = "Apache-2.0"}
 license-files = ["LICENSE", "NOTICE" ]
--- a/src/aws-api-mcp-server/uv.lock
+++ b/src/aws-api-mcp-server/uv.lock
@@ -66,7 +66,7 @@ wheels = [
 
 [[package]]
 name = "awscli"
-version = "1.42.65"
+version = "1.43.0"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "botocore" },
@@ -76,9 +76,9 @@ dependencies = [
     { name = "rsa" },
     { name = "s3transfer" },
 ]
-sdist = { url = "https://files.pythonhosted.org/packages/00/ae/1d68a0c04adb73d5d5b97ac6ccf0ca30902bceadf1cf410e2f21c7903a14/awscli-1.42.65.tar.gz", hash = "sha256:0d7f7d5cfb8e40456db311dbb7a72dd7b1eafd809937dee869a89e45b26358ce", size = 1877896, upload-time = "2025-11-03T21:56:52.747Z" }
+sdist = { url = "https://files.pythonhosted.org/packages/37/c7/6c0f664d8da301768bdf5ba330a85ca6dd075d1bf334df7b6766e060f6b1/awscli-1.43.0.tar.gz", hash = "sha256:dbc1d75eb7fbd6cb042788162228670e5f71027d4fdb37ad3edb3b238de416a4", size = 1877861, upload-time = "2025-11-19T20:29:06.258Z" }
 wheels = [
-    { url = "https://files.pythonhosted.org/packages/26/ac/a51a64d3af874b5f08498e92c9b51734d4929a6df1d7dadb0cf57e478639/awscli-1.42.65-py3-none-any.whl", hash = "sha256:cf1ea40c7f57e22b61949a21458d25d63f8f6fc72e5c5d9bc19091a09c6c0bf7", size = 4631481, upload-time = "2025-11-03T21:56:49.345Z" },
+    { url = "https://files.pythonhosted.org/packages/6a/61/5090a65b76e4d0897b02f54a4c3620139ef77619f489dc863dab4d5cf919/awscli-1.43.0-py3-none-any.whl", hash = "sha256:4247b48f4ae058a483e42cdd1febcb204057e93ca7acc5abcc04ccdb165bd235", size = 4631587, upload-time = "2025-11-19T20:29:04.572Z" },
 ]
 
 [[package]]
@@ -115,7 +115,7 @@ dev = [
 
 [package.metadata]
 requires-dist = [
-    { name = "awscli", specifier = "==1.42.65" },
+    { name = "awscli", specifier = "==1.43.0" },
     { name = "boto3", specifier = ">=1.38.18" },
     { name = "botocore", specifier = ">=1.38.18" },
     { name = "fastmcp", specifier = ">=2.13.1" },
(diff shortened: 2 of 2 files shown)


src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/parser/parser.py
```diff
@@ -90,7 +92,7 @@
     's3': ['ls', 'website', 'sync', 'cp', 'mv', 'rm', 'mb', 'rb', 'presign'],
     'cloudformation': ['package', 'deploy'],
     'cloudfront': ['sign'],
-    'cloudtrail': ['create-subscription', 'update-subscription', 'validate-logs'],
```
nzmdn, line 93: I removed `create-subscription` and `update-subscription` because they're deprecated https://docs.aws.amazon.com/awscloudtrail/latest/userguide/cloudtrail-create-and-update-a-trail-by-using-the-aws-cli.html

src/aws-api-mcp-server/README.md
```diff
@@ -207,6 +207,7 @@ See [DEPLOYMENT.md](https://github.com/awslabs/mcp/blob/main/src/aws-api-mcp-ser
...
+| `AWS_API_MCP_DISABLE_LOCAL_FILE_ACCESS`                           | ❌ No                       | `"false"`                                                | When set to `true`, rejects AWS CLI commands with local file path arguments. This takes precedence over `AWS_API_MCP_ALLOW_UNRESTRICTED_LOCAL_FILE_ACCESS`. S3 URIs (`s3://...`) and stdout redirect (`-`) remain allowed. Custom operations that require local file access (e.g., `aws cloudformation package`, `aws ecs deploy`) will be blocked when this is enabled.                                                                      |
```
rshevchuk-git: This is confusing, why do we need both. The number of configurations is already convoluted, can we merge these?

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/common/file_system_controls.py
```diff
@@ -53,9 +149,81 @@ def validate_file_path(file_path: str) -> str:
...
+                file_paths.extend(
+                    [
+                        item
+                        for item in param_value
+                        if isinstance(item, str) and not _is_remote_path(item)
+                    ]
+                )
+
+        # Check if this is a blob argument (may have file:// or fileb:// prefix)
+        elif param_name in blob_args:
```
rshevchuk-git, line 209: Should we not check for _is_remote_path on blob case?

src/aws-api-mcp-server/README.md
nzmdn: To merge them, we can introduce an env variable with three values, something like:
1. `AWS_API_MCP_LOCAL_FILE_ACCESS=restricted` (default)
2. `AWS_API_MCP_LOCAL_FILE_ACCESS=unrestricted`
3. `AWS_API_MCP_LOCAL_FILE_ACCESS=no`

But removing `AWS_API_MCP_ALLOW_UNRESTRICTED_LOCAL_FILE_ACCESS` will be a backward incompatible change, so I wouldn't go for it just now

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/common/file_system_controls.py
nzmdn: If I'm not mistaken, if an arg has `file://` prefix it cannot be a remote path. That's why I didn't add a `_is_remote_path` check

src/aws-api-mcp-server/README.md
arnewouters: Maybe we can deprecate it instead?

nzmdn: Made `AWS_API_MCP_ALLOW_UNRESTRICTED_LOCAL_FILE_ACCESS` accept three values `unrestricted` (or `true`), `workdir` (or `false`), and `no_access` and removed `AWS_API_MCP_DISABLE_LOCAL_FILE_ACCESS`

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/aws/services.py
```diff
@@ -52,11 +57,17 @@ class ConfigResult(NamedTuple):
 filter_query = re.compile(r'^\s+([-a-z0-9_.]+|tag:<key>)\s+')
 
 
+def _validate_streaming_blob_path(cli_argument: CLIArgument, value: Any, **kwargs):  # noqa: ARG001
```
arnewouters:
```suggestion
def _validate_streaming_blob_path(cli_argument: CLIArgument, value: Any, **_kwargs):
```

src/aws-api-mcp-server/tests/parser/test_parser.py
```diff
@@ -611,31 +609,6 @@ def test_client_side_filter_error():
...
-    expected_message = (
-        "Invalid file parameter '~user_that_does_not_exist/temp/test.txt' for service 's3' and operation 'cp': "
-        'should be an absolute path. Please provide a valid file path.'
-    )
-    with pytest.raises(FileParameterError) as exc_info:
-        parse(cli_command='aws s3 cp s3://my_file ~user_that_does_not_exist/temp/test.txt')
-
-    assert str(exc_info.value) == expected_message
-
-
```
arnewouters, line 638: Why did we remove these tests?

nzmdn: I moved them to test_parser_file_access.py

arnewouters: Woops I did ctrl+f but the diff for that file is not auto-loaded so didn't find them, thanks!

src/aws-api-mcp-server/README.md
```diff
@@ -206,7 +206,7 @@ See [DEPLOYMENT.md](https://github.com/awslabs/mcp/blob/main/src/aws-api-mcp-ser
...
+| `AWS_API_MCP_ALLOW_UNRESTRICTED_LOCAL_FILE_ACCESS`                | ❌ No                       | `"workdir"`                                              | Controls file system access level with three modes:<br/><br/>• `"unrestricted"` or `"true"`: Enables system-wide file access (may cause unintended overwrites)<br/>• `"workdir"` or `"false"`: Restricts file operations to `AWS_API_MCP_WORKING_DIR` (default)<br/>• `"no_access"`: Rejects all local file path arguments in AWS CLI commands and blocks commands requiring local file access (e.g., `aws cloudformation package`, `aws ecs deploy`). S3 URIs (`s3://...`) and stdout redirect (`-`) remain allowed. |
```
rshevchuk-git: This might raise some eyebrows on why we have several alternatives to configure the same thing (workdir OR false, unrestricted OR true). I understand we're deprecating previous configuration, however as the latest, up to date docs, they should reference currently expected values and not encourage using deprecated values. We should either remove references to 'false' and 'true' or add a "DEPRECATED: ..." note at the end of this section to mention new mappings. false -> workdir, true -> unrestricted. I'd probably be biased towards not mentioning it at all, but leaving it up to you.

src/aws-api-mcp-server/README.md
```diff
@@ -281,7 +281,7 @@ Our MCP server aims to support all AWS APIs. However, some of them will spawn su
 
 #### Security and Access Considerations
 
-- **No Sandboxing**: The `AWS_API_MCP_WORKING_DIR` environment variable sets a working directory. The `AWS_API_MCP_ALLOW_UNRESTRICTED_LOCAL_FILE_ACCESS` flag  by default is set to `"false"` which restricts MCP server file operations to `<AWS_API_MCP_WORKING_DIR>`. Setting to `"true"` enables system-wide file access but may cause unintended overwrites.
+- **No Sandboxing**: The `AWS_API_MCP_WORKING_DIR` environment variable sets a working directory. The `AWS_API_MCP_ALLOW_UNRESTRICTED_LOCAL_FILE_ACCESS` flag by default is set to `"workdir"` (or `"false"`) which restricts MCP server file operations to `<AWS_API_MCP_WORKING_DIR>`. Setting to `"true"` enables system-wide file access but may cause unintended overwrites. Setting to `"no_access"` disables local file access.
```
rshevchuk-git:
```suggestion
- **No Sandboxing**: The `AWS_API_MCP_WORKING_DIR` environment variable sets a working directory. The `AWS_API_MCP_ALLOW_UNRESTRICTED_LOCAL_FILE_ACCESS` flag by default is set to `"workdir"` (or `"false"`) which restricts MCP server file operations to `<AWS_API_MCP_WORKING_DIR>`. Setting to `"true"` enables system-wide file access but may cause unintended overwrites. Setting to `"no-access"` disables local file access.
```

Also kebab case here

src/aws-api-mcp-server/README.md
rshevchuk-git:
```suggestion
| `AWS_API_MCP_ALLOW_UNRESTRICTED_LOCAL_FILE_ACCESS`                | ❌ No                       | `"workdir"`                                              | Controls file system access level with three modes:<br/><br/>• `"unrestricted"` or `"true"`: Enables system-wide file access (may cause unintended overwrites)<br/>• `"workdir"` or `"false"`: Restricts file operations to `AWS_API_MCP_WORKING_DIR` (default)<br/>• `"no-access"`: Rejects all local file path arguments in AWS CLI commands and blocks commands requiring local file access (e.g., `aws cloudformation package`, `aws ecs deploy`). S3 URIs (`s3://...`) and stdout redirect (`-`) remain allowed. |
```

We use kebab case for other similar configurations (like 'no-auth')

nzmdn: Added a deprecation note

PR #1816, opened 2025-11-21
fix(s3-tables-mcp): fix time parsing

okhomin:
## Summary
Fix parsing timestamps of different formats in creating from a file and directly from MCP.

### Changes

> Added a function to parse different formats of dates of Iceberg V2 spec.
> Updated tool descriptions to provide supported field data types of Iceberg tables

### User experience

> User will be able to use dates without parsing errors.

## Checklist

* [x] Changes are documented

Is this a breaking change? No

## Acknowledgment

### DIFF
--- a/src/amazon-bedrock-agentcore-mcp-server/README.md
+++ b/src/amazon-bedrock-agentcore-mcp-server/README.md
@@ -186,3 +186,58 @@ Dictionary containing:
 - `error`: Error message (if fetch failed)
 
 Use this tool to get complete documentation pages when search snippets aren't sufficient for understanding or implementing AgentCore features.
+
+### manage_agentcore_runtime
+
+Provides comprehensive information on deploying and managing agents in AgentCore Runtime.
+
+```python
+manage_agentcore_runtime() -> Dict[str, Any]
+```
+
+**Returns:**
+Detailed deployment guide covering:
+- Code requirements and validation checklist
+- Step-by-step CLI deployment workflow (configure, launch, invoke, status, destroy)
+- Required code patterns with BedrockAgentCoreApp
+- Common issues and troubleshooting
+- Session management and cleanup procedures
+
+Use this tool when you need to deploy agents to AgentCore Runtime or troubleshoot deployment issues.
+
+### manage_agentcore_memory
+
+Provides comprehensive information on managing AgentCore Memory resources.
+
+```python
+manage_agentcore_memory() -> Dict[str, Any]
+```
+
+**Returns:**
+Complete memory management guide covering:
+- Memory resource creation and configuration
+- Short-term memory (STM) and long-term memory (LTM) concepts
+- Semantic memory strategies for facts and knowledge
+- Full CLI command reference (create, get, list, delete, status)
+- Common workflows and examples
+
+Use this tool when working with AgentCore Memory for persistent knowledge storage.
+
+### manage_agentcore_gateway
+
+Provides comprehensive information on deploying and managing MCP Gateways in AgentCore.
+
+```python
+manage_agentcore_gateway() -> Dict[str, Any]
+```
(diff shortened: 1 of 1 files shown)


src/s3-tables-mcp-server/awslabs/s3_tables_mcp_server/engines/pyiceberg.py
```diff
@@ -27,6 +25,107 @@
...
+    for row in rows:
+        converted_row = {}
+        for field_name, value in row.items():
+            # Early skip for non-string values
+            if not isinstance(value, str):
+                converted_row[field_name] = value
+                continue
+
+            # Get the field type from schema
+            field = arrow_schema.field(field_name)
```
Kurtiscwright, line 49: Is it ever possible to have a field not in the schema at this point?

src/s3-tables-mcp-server/awslabs/s3_tables_mcp_server/engines/pyiceberg.py
```diff
@@ -27,6 +25,107 @@
...
+                        dt = datetime.strptime(value_normalized, fmt)
+                        if dt.tzinfo is None:
+                            dt = dt.replace(tzinfo=timezone.utc)
+                        converted_row[field_name] = dt.astimezone(timezone.utc)
+                        break
+                    except ValueError:
+                        continue
+                else:
+                    raise ValueError(
+                        f"Could not parse timestamp with timezone: {value} for field {field_name}"
```
Kurtiscwright: Should there be examples of acceptable formats?

src/s3-tables-mcp-server/awslabs/s3_tables_mcp_server/engines/pyiceberg.py
okhomin: 1. All the fields are added dynamically. So yes, you can basically pass whatever you want.
2. This method checks only the dates and converts them to python objects.

src/s3-tables-mcp-server/awslabs/s3_tables_mcp_server/engines/pyiceberg.py
okhomin: We don't need it in the error message, but there is a description of what temporal types are supported:
```
    - date: Calendar date without timezone or time
    - time: Time of day, microsecond precision, without date or timezone
    - timestamp: Timestamp, microsecond precision, without timezone (represents date and time regardless of zone)
    - timestamptz: Timestamp, microsecond precision, with timezone (stored as UTC)
```

We could add more examples, but in my testing it was enough for LLM to understand it.

src/s3-tables-mcp-server/awslabs/s3_tables_mcp_server/engines/pyiceberg.py
Kurtiscwright: Would we need error handling then? What would happen if arrow_schema.field() encounters a value it doesn't know about?

okhomin: This method is only to convert dates, it doesn't make any validations. All other validations should be on the pyarrow side.

okhomin: image
Example of an error message

Kurtiscwright: Awesome, thank you! Rest of the PR looks good.

scottschreckengaust (approved):
LGTM

Zh111602 (approved):

PR #1836, opened 2025-11-24
feat(aws-iac-mcp-server): rename two CFN tools to follow naming consistency, add new context accessible as a tool

kdbrogan:
Refactoring existing tools to match naming consistency, adding new context accessible as tool, deleting MCP resource.
### Changes

Refactor
check_template_compliance -> check_cloudformation_template_compliance
troubleshoot_deployment -> troubleshoot_cloudformation_deployment

Add tool 
get_cloudformation_pre_deploy_validation_instructions
(context accessible as a tool) 

Delete resource 
@mcp.resource('cfn://context/template-examples-and-best-practices')
def get_template_examples()

### User experience

> Small change. MCP resource is consumed in knowledge base (also wasn't working before hand). New tool introduces additional functionality to agent's ability to run aws cli commands

## Checklist

* [X] Changes are documented

**RFC issue number**: n/a

Checklist:

## Acknowledgment

### DIFF
--- a/src/aws-iac-mcp-server/README.md
+++ b/src/aws-iac-mcp-server/README.md
@@ -107,6 +107,12 @@ Searches CDK code samples, examples, constructs, and patterns documentation.
 - `query` (required): Search query for CDK samples and constructs
 - `language` (optional): Programming language filter (default: "typescript")
 
+### cdk_best_practices
+Provides CDK best practices for application configuration, coding, constructs, security, and testing.
+
+**Parameters:**
+- None
+
 ## Prerequisites
 
 1. Install `uv` from [Astral](https://docs.astral.sh/uv/getting-started/installation/) or the [GitHub README](https://github.com/astral-sh/uv#installation)
@@ -260,6 +266,12 @@ Search CloudFormation documentation for AWS::Lambda::Function properties
 
 ```
 Find CDK code samples for serverless API with TypeScript
+```
+
+### Consult CDK best practices
+
+```
+Suggest improvements to my CDK setup based on the best practices
 ```
 
 ## Security Considerations
--- a/src/aws-iac-mcp-server/awslabs/aws_iac_mcp_server/server.py
+++ b/src/aws-iac-mcp-server/awslabs/aws_iac_mcp_server/server.py
@@ -18,6 +18,7 @@ import json
 from .sanitizer import sanitize_tool_response
 from .tools.cdk_tools import (
     SupportedLanguages,
+    cdk_best_practices_tool,
     read_cdk_documentation_page_tool,
     search_cdk_documentation_tool,
     search_cdk_samples_and_constructs_tool,
@@ -50,6 +51,7 @@ mcp = FastMCP(
                 - Use `search_cdk_samples_and_constructs` when: You need working code examples, implementation patterns, or community constructs
                 - Use `read_cdk_documentation_page` when: You have a specific documentation URL from search results and need complete content with pagination support
                 - Use `search_cloudformation_documentation` when: You need Cloudformation related official documentation, resource type information or template syntax
+                - Use `cdk_best_practices` when: You need to generate or review CDK code
 
               """,
 )
@@ -503,6 +505,50 @@ async def search_cdk_samples_and_constructs(
         List of search results with URLs, titles, and context snippets
     """
     result = await search_cdk_samples_and_constructs_tool(query, language)
+
+    # Convert CDKToolResponse to dict for JSON serialization
+    response_dict = asdict(result)
+
+    return sanitize_tool_response(json.dumps(response_dict))
+
+
+@mcp.tool()
+async def cdk_best_practices() -> str:
+    """Returns CDK best practices and security guidelines.
+
+    ## Usage
+
+    This tool provides comprehensive CDK development guidelines, security best practices, and architectural recommendations. Always run this tool when asked to generate or review CDK code and follow the guidelines returned.
+
+    ## When to Use
+
+    - Get CDK security best practices and compliance guidelines
+    - Look up architectural patterns and recommendations
+    - Get guidance on CDK application structure and organization
+    - Research performance optimization techniques
+    - Learn about proper construct usage and design patterns
+    - Understand deployment and testing best practices
+
+    ## Result Interpretation
+
+    Returns JSON with:
+    - knowledge_response: Details of the response
+      - results: Array with single result containing:
+        - rank: Always 1
+        - title: Document title or filename
+        - url: Source URL of the CDK best practices
+        - context: A summary of the CDK best practices
+    - next_step_guidance: If present, suggested next actions to take for answering user query
+
+    ## Args
+
+    No parameters required - this tool returns the complete best practices guide.
+
+    ## Returns
+
+    Complete best practices documentation as text, including security guidelines, architectural patterns, development workflow, and compliance requirements.
+    """
+    result = await cdk_best_practices_tool()
 
     # Convert CDKToolResponse to dict for JSON serialization
     response_dict = asdict(result)
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -133,7 +133,7 @@
         "filename": "src/aws-iac-mcp-server/README.md",
         "hashed_secret": "df99ad98cabfe1616640820bcfb345ef5b10077f",
         "is_verified": false,
-        "line_number": 194,
+        "line_number": 200,
         "is_secret": false
       }
     ],
@@ -906,5 +906,5 @@
       }
     ]
   },
-  "generated_at": "2025-11-25T19:13:19Z"
+  "generated_at": "2025-11-26T00:01:38Z"
 }
--- a/src/aws-iac-mcp-server/awslabs/aws_iac_mcp_server/tools/cdk_best_practices.py
+++ b/src/aws-iac-mcp-server/awslabs/aws_iac_mcp_server/tools/cdk_best_practices.py
@@ -1,0 +1,713 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from ..knowledge_models import KnowledgeResult
+
+
+# Sourced from a combination of
+# https://docs.aws.amazon.com/cdk/v2/guide/best-practices.html
+# https://docs.aws.amazon.com/cdk/v2/guide/best-practices-security.html
+# https://docs.aws.amazon.com/prescriptive-guidance/latest/aws-cdk-layers/best-practices.html
+# https://docs.aws.amazon.com/cdk/v2/guide/hello-world.html (for Getting started)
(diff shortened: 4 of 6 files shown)


src/aws-iac-mcp-server/awslabs/aws_iac_mcp_server/tools/cfn_tools.py
```diff
@@ -0,0 +1,165 @@
...
+
+    Args:
+        template_content: CloudFormation template as YAML or JSON string
+        regions: AWS regions to validate against
+        ignore_checks: Rule IDs to ignore (e.g., W2001, E3012)
+
+    Returns:
+        Validation results dictionary
+    """
+    return validate_template(
```
kumvprat: Why not port the logic here itself? 
We don't need a delegation to another function when this on doesnot do anything specific. 
Similar for the tool calls below.

If the idea is to consolidate all the tools here we can have the logic here as well.

For cdk the tool functions themselves delegate call to a client and format it's response. If the call is entirely local we can just remove this delegation and add the logic here.

If the logic is verbose, we can create one python file per CFN tool as well

kdbrogan: Updated the file structure so cfn tools are in the same directory with one file per tool, as logic is usually verbose and multiple methods.

src/aws-iac-mcp-server/awslabs/aws_iac_mcp_server/server.py
```diff
@@ -475,60 +498,6 @@ async def search_cdk_samples_and_constructs(
     return sanitize_tool_response(json.dumps(response_dict))
 
 
-@mcp.resource('cfn://context/template-examples-and-best-practices')
-def get_template_examples() -> str:
```
kumvprat, line 525: What was the use of this before ? 
Want to make sure we are not removing a functionality from the mcp server

src/aws-iac-mcp-server/awslabs/aws_iac_mcp_server/server.py
```diff
@@ -225,13 +225,36 @@ def troubleshoot_deployment(
...
+            f'#/stacks/stackinfo?stackId={stack_name})'
         )
 
     response_text = json.dumps(result, indent=2, default=str)
     return sanitize_tool_response(response_text)
 
 
+@mcp.tool()
+def get_cloudformation_pre_deploy_validation_instructions() -> str:
+    """Get instructions for CloudFormation pre-deployment validation (Nov 2025 release).
```
kumvprat: Do we need to include the `Nov 2025 Release` ?

I assume this is from the latest early validation launch, can you attach a better identifier for this feature rather than the launch date ?

src/aws-iac-mcp-server/README.md
```diff
@@ -34,7 +34,7 @@ Validates CloudFormation template syntax, schema, and resource properties using
 - `regions` (optional): List of AWS regions to validate against
 - `ignore_checks` (optional): List of cfn-lint check IDs to ignore
 
-### check_template_compliance
+### check_cloudformation_template_compliance
```
kumvprat, line 37: For th validation tool do we need to change the previous validate tool instructions and naming in this README or is pre-validation check a totally new tool ?

src/aws-iac-mcp-server/awslabs/aws_iac_mcp_server/server.py
```diff
@@ -225,13 +225,36 @@ def troubleshoot_deployment(
...
+            f'[View Stack](https://console.aws.amazon.com/cloudformation/home?region={region}'
+            f'#/stacks/stackinfo?stackId={stack_name})'
         )
 
     response_text = json.dumps(result, indent=2, default=str)
     return sanitize_tool_response(response_text)
 
 
+@mcp.tool()
+def get_cloudformation_pre_deploy_validation_instructions() -> str:
```
kumvprat, line 240: How is this tool supposed to be used ? 

I assume it's one of the below and the corresponding questions : 
- It looks like a static tool that gives LLM context around how to interpret the pre-validation checks : Tool description in this case should tell the LLM to use this in case some pre-validation errors are seen
- Tool is invoking the pre-validation checks:  How can the tool invoke these checks ?

src/aws-iac-mcp-server/awslabs/aws_iac_mcp_server/server.py
kdbrogan: The _intended_ use of this before was that in the MCP instructions, the uri would be provided with a description of what the resource was. In reality, I was never able to get the agent to access the resource, and so it was already removed from the instructions. Because the agent never had the uri, it was unaware of the existence of this resource. Removing this is not a breaking change, as it was not working in the first place. The new knowledge base should cover this documentation, so this is more of a bug fix.

src/aws-iac-mcp-server/awslabs/aws_iac_mcp_server/server.py
kdbrogan: I've updated the tool description to reflect its output. This is a static tool that returns a json output with CLI instructions and remediation steps for encountered errors. If there's a better way to implement this without flooding the context window, let me know.

src/aws-iac-mcp-server/README.md
kdbrogan: I missed the readme before. I've updated it to include the new validation tool, but it is a totally new tool.

src/aws-iac-mcp-server/awslabs/aws_iac_mcp_server/server.py
kdbrogan: Got rid of this identifier

src/aws-iac-mcp-server/awslabs/aws_iac_mcp_server/server.py
```diff
@@ -225,13 +225,48 @@ def troubleshoot_deployment(
...
+    """Get instructions for CloudFormation pre-deployment validation.
+
```
kumvprat, line 242: For this tool, the static instructions should be sources from a documentation source like https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/validate-stack-deployments.html#validate-stack-deployments-resource-limitations

Currently if users want to understand the pre-deployment validation and ask this tool for sources we wont be able to reply with any.

Either we should ingest this into our cfn knowledge base and use it or provide the documentation link to  the users

src/aws-iac-mcp-server/tests/tools/test_cloudformation_pre_deploy_validation.py
```diff
@@ -0,0 +1,43 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Tests for cloudformation_pre_deploy_validation module."""
+
+import json
```
kumvprat, line 17: Is the file name correct ? The tests here check for the new pre deploy validation tool only

kdbrogan: Renamed to test_cloudformation_pre_deploy_validation

src/aws-iac-mcp-server/awslabs/aws_iac_mcp_server/server.py
kdbrogan: ACK. planning to fast follow this change after reinvent

vishaalmehrishi (approved):

src/aws-iac-mcp-server/awslabs/aws_iac_mcp_server/client/aws_client.py
```diff
@@ -0,0 +1,62 @@
...
+        region_name: AWS region name (defaults to environment variable or 'us-east-1')
+
+    Returns:
+        Boto3 client for the specified service
+    """
+    # Default region handling
+    if not region_name:
+        region_name = environ.get('AWS_REGION', 'us-east-1')
+
+    session = Session(profile_name=environ.get('AWS_PROFILE'))
```
hfurkanbozkurt, line 46: Why not `Session()` and let boto3 resolve creds?

src/aws-iac-mcp-server/awslabs/aws_iac_mcp_server/client/aws_client.py
```diff
@@ -0,0 +1,62 @@
...
+
+    # Credential detection and client creation
+    try:
+        client = session.client(service_name, region_name=region_name, config=session_config)
```
hfurkanbozkurt, line 50: Similarly, why dont we let boto decide on the region instead of looking into environ ourselves?

src/aws-iac-mcp-server/awslabs/aws_iac_mcp_server/client/aws_client.py
```diff
@@ -0,0 +1,62 @@
...
+        return client
+
+    except Exception as e:
```
hfurkanbozkurt, line 53: What benefit does this error handling give? Cant we let errors bubble up?

src/aws-iac-mcp-server/awslabs/aws_iac_mcp_server/client/aws_client.py
```diff
@@ -0,0 +1,62 @@
...
+        print(f'Error creating {service_name} client: {str(e)}', file=sys.stderr)
```
hfurkanbozkurt, line 54: And use logging instead of print

src/aws-iac-mcp-server/awslabs/aws_iac_mcp_server/client/aws_client.py
kumvprat: The default profile is used when we do not provide a `PROFILE_NAME`

This way the user can allow the client to choose any profile they want to work with.

src/aws-iac-mcp-server/awslabs/aws_iac_mcp_server/client/aws_client.py
kumvprat: Same as above, this provides a flexibility and configuration-ability for users to choose a specific region they want to work with

src/aws-iac-mcp-server/awslabs/aws_iac_mcp_server/client/aws_client.py
kumvprat: The errors are bubbling up with simplified error messages

src/aws-iac-mcp-server/awslabs/aws_iac_mcp_server/client/aws_client.py
kumvprat: Will take this up in next set of PRs

src/aws-iac-mcp-server/awslabs/aws_iac_mcp_server/client/aws_client.py
kdbrogan: Just following up, boto3 does a fine job of resolving credentials but I was running into issues where boto3 doesn't refresh credentials if they are cached. Credentials are usually cached when you start-up the server for the first time and not again after that. So if you wanted to switch profiles mid session you had to delete the cache. I added a way that credentials could be refreshed and add additional granularity to the way boto3 resolves credentials. I can look again after re:invent, but I pretty much just scraped this from `cfn-mcp-server`

src/aws-iac-mcp-server/awslabs/aws_iac_mcp_server/client/aws_client.py
kdbrogan: https://github.com/awslabs/mcp/blob/main/src/cfn-mcp-server/awslabs/cfn_mcp_server/aws_client.py
for reference

PR #1844, opened 2025-11-25
fix(dynamodb-mcp-server): use explicit workspace_dir parameter for validation file paths

koffey-amazon:
Fixes

## Summary
Replaced implicit working directory detection with an explicit `workspace_dir` parameter in the DynamoDB MCP server to ensure validation files are written to the correct location. This change removes the `get_user_working_directory()` function and updates all dependent code to use the explicit parameter instead.

### Changes

- Replace implicit get_user_working_directory() calls with explicit workspace_dir parameter
- Update execute_access_patterns signature: workspace_dir is now first positional argument
- Update dynamodb_data_model_validation to require workspace_dir parameter
- Remove unused get_user_working_directory from model_validation_utils
- Update tests to match new function signatures
- Remove the optional ctx (Context) parameter from DynamoDB validation and command execution functions, replacing it with direct Context instantiation where needed.

### User experience

No change to user experience

## Checklist

* [x] Changes are documented

**RFC issue number**:

Checklist:

## Acknowledgment


src/dynamodb-mcp-server/awslabs/dynamodb_mcp_server/server.py
```diff
@@ -399,15 +351,14 @@ async def dynamodb_data_model_validation(
...
+        workspace_dir: Absolute path of the workspace directory
 
     Returns:
         JSON generation guide (if file missing) or validation results with transformation prompt (if file exists)
     """
     try:
         # Step 1: Get current working directory reliably
-        user_dir = get_user_working_directory()
-        data_model_path = os.path.join(user_dir, DATA_MODEL_JSON_FILE)
+        data_model_path = os.path.join(workspace_dir, DATA_MODEL_JSON_FILE)
```
akeyesamzn, line 361: I would take in the fully qualified path rather than just the directory

Taking in the directory has two downsides:
1. It couples this part of the system by requiring this code to know the file name produced by the other part of the system
2. Customers/LLMs cannot use different file names

In general, I would try to rely as little as possible on one tool knowing the name of a file that the another tool creates

LeeroyHannigan (approved):

src/dynamodb-mcp-server/awslabs/dynamodb_mcp_server/server.py
amzn-erdemkemer: That's a fair point. We'll address this in an upcoming PR to keep this fix focused. Let me know if you have any concerns with that approach.

akeyesamzn: Agreed.

PR #1858, opened 2025-12-01
feat: Add 'help' AWS CLI commands support

rshevchuk-git:
## Summary
This PR adds support for executing AWS CLI help commands in the API MCP server. Instead of simply executing the command or failing, the server now detects help requests (e.g., aws s3 ls help) and returns a structured JSON response containing the command's description and available parameters.

### Changes

* Added awslabs/aws_api_mcp_server/core/common/help_command.py which implements generate_help_document. This function leverages awscli's internal ReSTDocument and command table to extract and parse help text into a structured dictionary (description, parameters).
* Updated awslabs/aws_api_mcp_server/core/parser/parser.py and helpers.py to detect help or --help tokens in the command arguments.
* Added is_help_operation boolean flag to the IRCommand model in awslabs/aws_api_mcp_server/core/common/command.py to track if a command is a help request.
* Modified awslabs/aws_api_mcp_server/server.py and service.py to check for is_help_operation. If true, it bypasses normal execution and calls get_help_document to return the help info.

### User experience

LLMs might generate a help command, which now will be executed and not fail on validation.

## Checklist

* [ ] Changes are documented

**RFC issue number**:

Checklist:

## Acknowledgment


src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/common/help_command.py
```diff
@@ -0,0 +1,83 @@
...
+            # To avoid adding arguments like --disable-rollback and --no-disable-rollback separately
+            # we need to make sure a group name is only processed once
+            # event_handler.doc_option takes care of mentioning all arguments in a group
+            # so we can safely skip the remaining arguments in the group
+            seen_arg_groups.add(arg.group_name)
+
+    return {
+        'command': f'aws {service_name} {operation_name}',
+        'description': description,
+        'parameters': params,
```
arnewouters, line 83: Should we have some safeguards here for docs that have a very large parameter scheme that will never fit the context window?

e.g [`wafv2 update-rule-group`](https://docs.aws.amazon.com/cli/latest/reference/wafv2/update-rule-group.html)

rshevchuk-git: Updated to truncate, and tested with [wafv2 update-rule-group](https://docs.aws.amazon.com/cli/latest/reference/wafv2/update-rule-group.html) which indeed blew up the context when invoked with 'help'. Now it's not breaking anything.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/common/helpers.py
```diff
@@ -65,6 +65,11 @@ def expand_user_home_directory(args: list[str]) -> list[str]:
     return [os.path.expanduser(arg) for arg in args]
 
 
+def is_help_operation(args: list[Any]) -> bool:
```
anasstahr, line 68: Since you're passing the parameters here, wouldn't it fail on:
```
aws ec2 create-tags --resources i-xxx --tags Key=Purpose,Value=help-desk
```

rshevchuk-git: Why would it fail?

PR #1862, opened 2025-12-01
feat(api): VSS and Semantic Search MCPs for Valkey

Jonathan-Improving:
### Changes

Introduces a simple MCP wrapper around the Valkey vector search capability, and provides a full semantic search MCP for Valkey, leveraging its VSS capabilities, and encapsulating vector embeddings so that agents can interop using pure natural language and/or keywords.

#### Vector Similarity Search
Prior to the introduction of this tool users could not interact with Valkey's vector search capabilities via an MCP client, but now one can and retrieve documents stored in a Valkey instance, sought according to a vector embedding.

**Caveat**: The user must provide the vector embedding itself, and the vector embedding must have been generated by the same model that the documents stored within the Valkey instance contain (if applicable), presuming that the end-user has already seeded a Valkey instance with documents each having a vector embedding field used to VSS.

##### Results
A basic test revealed that the agent struggled to generate vector embeddings from an MCP that provides embeddings, and then pass them on _intact_ to this VSS MCP, which requires a list of floats (vector embedding) for input.  In one test the agent passed a completely fake vector embedding to the VSS MCP, and in another test the agent hung for more than 10 minutes before it was terminated.  This implies that a specifically configured agentic context is key for the VSS MCP to be useful.

#### Semantic Search
The role of a data scientist would first add documents to the Valkey database using the `add_documents` MCP, under a specifically named collection, that can be sought using semantic search.  The role of researcher would leverage the `semantic_search` MCP to seek for documents added under a specifically named collection, or query the `list_collections` MCP to find out what named collections are available for searching.

## Checklist

* [x] Changes are documented

Not a breaking change.

**RFC issue number**:

Checklist:

## Acknowledgment


src/valkey-mcp-server/awslabs/valkey_mcp_server/embeddings/__init__.py
```diff
@@ -0,0 +1,100 @@
...
+        )
+
+    elif provider_type == 'cohere':
+        api_key = os.getenv('COHERE_API_KEY')
+        if not api_key:
+            raise ValueError("COHERE_API_KEY environment variable is required for Cohere provider")
+        return CohereEmbeddings(
+            api_key=api_key,
+            model=os.getenv('COHERE_EMBEDDING_MODEL', 'embed-english-v3.0')
+        )
```
Jonathan-Improving: This section here is what remains untested.

src/valkey-mcp-server/awslabs/valkey_mcp_server/embeddings/__init__.py
```diff
@@ -0,0 +1,100 @@
...
+    CohereEmbeddings,
+    HashEmbeddings
+)
+
+
+def create_embeddings_provider() -> EmbeddingsProvider:
+    """Create an embeddings provider based on environment configuration.
+
+    Configuration via environment variables:
+    - EMBEDDINGS_PROVIDER: Provider type (ollama, openai, bedrock, cohere, hash)
```
MatthiasHowellYopp: if ollama is an option, should lmstudio not also be an option?

src/valkey-mcp-server/awslabs/valkey_mcp_server/tools/semantic_search.py
```diff
@@ -0,0 +1,388 @@
...
+        except ValkeyError:
+            index_exists = False
+
+        # Process and store each document
+        added_count = 0
+        actual_dimensions = embedding_dimensions
+
+        for doc in documents:
+            if 'id' not in doc:
+                continue  # Skip documents without ID
```
MatthiasHowellYopp, line 136: any point in logging documents that don't get indexed?

src/valkey-mcp-server/awslabs/valkey_mcp_server/embeddings/providers.py
```diff
@@ -0,0 +1,195 @@
...
+
+"""Concrete implementations of embeddings providers."""
+
+from typing import List
+import httpx
+import hashlib
+from .base import EmbeddingsProvider
+
+
+class OllamaEmbeddings(EmbeddingsProvider):
```
vasigorc, line 23: Minor. Not sure about which specific `EmbeddingsProvider` implementations must we provide. There's [Sentence-BERT](https://sbert.net/), E5 family of embeddings (for example [large](https://huggingface.co/intfloat/multilingual-e5-large)). I think that given this is for `awslabs` maybe let's keep Titan, and let users write their implementations as they find the need.

src/valkey-mcp-server/awslabs/valkey_mcp_server/embeddings/providers.py
```diff
@@ -0,0 +1,195 @@
...
+
+class BedrockEmbeddings(EmbeddingsProvider):
+    """AWS Bedrock embeddings provider.
+
+    Requires: pip install boto3
+
+    Example:
+        provider = BedrockEmbeddings(
+            region_name="us-east-1",
+            model_id="amazon.titan-embed-text-v1"
```
vasigorc: FWIW, there is a newer V2 version now: https://docs.aws.amazon.com/bedrock/latest/userguide/titan-embedding-models.html

src/valkey-mcp-server/awslabs/valkey_mcp_server/tools/semantic_search.py
```diff
@@ -0,0 +1,388 @@
...
+    try:
+        r = ValkeyConnectionManager.get_connection(decode_responses=True)
+        index_name = _get_collection_index_name(collection)
+
+        # Check if index exists
+        try:
+            r.execute_command('FT.INFO', index_name)
+            index_exists = True
+        except ValkeyError:
+            index_exists = False
```
vasigorc: Minor, we could extract this into a `def _check_index_exists`:

1. Remove the need for comment
2. Remove the need for nested indentation

src/valkey-mcp-server/awslabs/valkey_mcp_server/tools/semantic_search.py
```diff
@@ -0,0 +1,388 @@
...
+            "status": "success",
+            "added": added_count,
+            "collection": collection,
+            "total_documents": total_docs,
+            "embedding_dimensions": actual_dimensions,
+            "embeddings_provider": _embeddings_provider.get_provider_name()
+        }
+
+    except Exception as e:
+        return f"Error adding documents: {str(e)}"
```
vasigorc: We didn't tell in our docstring about this return structure. Could we rather return here a dict smilar to that from the happy path? Something like:

```python
return {
     "status": "failure",
     "added": 0,
     "failure_reason": str(e)
  }
```

src/valkey-mcp-server/awslabs/valkey_mcp_server/tools/semantic_search.py
Jonathan-Improving: Good catch, silent failures can be a nightmare

src/valkey-mcp-server/awslabs/valkey_mcp_server/embeddings/providers.py
Jonathan-Improving: Will keep _Titan_ and _Ollama_ for testing locally, the rest can go.

src/valkey-mcp-server/awslabs/valkey_mcp_server/tools/semantic_search.py
Jonathan-Improving: Yep, the other methods return a list, so they're keeping the union of list and string for now.  Might be worth thinking about that further though, I just wanted to keep the return structure clutter-free.

src/valkey-mcp-server/awslabs/valkey_mcp_server/embeddings/__init__.py
Jonathan-Improving: Might be worth removing Ollama rather than setting the precedent that we support all these embeddings providers that we clearly lack the resources to support.  I've kept Ollama specifically for testing purposes, however.

src/valkey-mcp-server/awslabs/valkey_mcp_server/embeddings/__init__.py
```diff
@@ -0,0 +1,64 @@
...
+    - OLLAMA_HOST: Ollama server URL (default: http://localhost:11434)
+    - OLLAMA_EMBEDDING_MODEL: Ollama model name (default: nomic-embed-text)
+    - AWS_REGION: AWS region for Bedrock (default: us-east-1)
+    - BEDROCK_MODEL_ID: Bedrock model ID (default: amazon.titan-embed-text-v1)
```
jbrinkman: Do we have env vars for OpenAI, Cohere or hash? If so what are they, if not, we need to explain that.

src/valkey-mcp-server/awslabs/valkey_mcp_server/embeddings/__init__.py
```diff
@@ -0,0 +1,64 @@
...
+import os
+from .base import EmbeddingsProvider
+from .providers import OllamaEmbeddings, BedrockEmbeddings
+
+
+def create_embeddings_provider() -> EmbeddingsProvider:
+    """Create an embeddings provider based on environment configuration.
+
+    Configuration via environment variables:
+    - EMBEDDINGS_PROVIDER: Provider type (ollama, openai, bedrock, cohere, hash)
```
jbrinkman: Make sure this list matches the concrete providers that are included in providers.py and remove those where we don't have concrete providers. Also, add a ", etc." so that it is clear that the provider list is not a fully enumerated set of values.

src/valkey-mcp-server/awslabs/valkey_mcp_server/embeddings/__init__.py
Jonathan-Improving: I removed those providers and corrected the remnant misleading documentation.

src/valkey-mcp-server/awslabs/valkey_mcp_server/embeddings/__init__.py
Jonathan-Improving: I improved the formatting, it wasn't quite clear.

Jonathan-Improving:
Latest enhancements include full `FT.SEARCH` support in the VSS MCP as well as structured return types for APIs

Jonathan-Improving:
Removing draft status to push this forward, one thing remains: to thoroughly test the Bedrock integration.

Jonathan-Improving:
Ready for review

src/valkey-mcp-server/awslabs/valkey_mcp_server/embeddings/__init__.py
```diff
@@ -0,0 +1,69 @@
...
+        Configured embeddings provider instance
+
+    Raises:
+        ValueError: If provider type is unknown or required credentials are missing
+    """
+    provider_type = os.getenv('EMBEDDINGS_PROVIDER', 'ollama').lower()
+
+    if provider_type == 'ollama':
+        return OllamaEmbeddings(
+            base_url=os.getenv('OLLAMA_HOST', 'http://localhost:11434'),
```
seaofawareness: * Define EMBEDDING_CFG in https://github.com/awslabs/mcp/blob/main/src/valkey-mcp-server/awslabs/valkey_mcp_server/common/config.py#L24 and use it here?
* Add new configs at https://github.com/awslabs/mcp/blob/main/src/valkey-mcp-server/README.md

src/valkey-mcp-server/awslabs/valkey_mcp_server/embeddings/__init__.py
```diff
@@ -0,0 +1,69 @@
...
+from .base import EmbeddingsProvider
+from .providers import OllamaEmbeddings, BedrockEmbeddings
+
+
+def create_embeddings_provider() -> EmbeddingsProvider:
+    """Create an embeddings provider based on environment configuration.  This list of providers can be extended
+    in the future to support additional embeddings providers such as OpenAI or Cohere etc.
+
+    Configuration via environment variables:
+    - EMBEDDINGS_PROVIDER: Provider type (ollama, bedrock)
```
seaofawareness: We should also include OpenAI embedding support, refer https://platform.openai.com/docs/guides/embeddings

src/valkey-mcp-server/awslabs/valkey_mcp_server/embeddings/__init__.py
```diff
@@ -0,0 +1,69 @@
...
+    if provider_type == 'ollama':
+        return OllamaEmbeddings(
+            base_url=os.getenv('OLLAMA_HOST', 'http://localhost:11434'),
+            model=os.getenv('OLLAMA_EMBEDDING_MODEL', 'nomic-embed-text')
+        )
+
+    elif provider_type == 'bedrock':
+        return BedrockEmbeddings(
+            region_name=os.getenv('AWS_REGION', 'us-east-1'),
+            model_id=os.getenv('BEDROCK_MODEL_ID', 'amazon.titan-embed-text-v1')
```
seaofawareness: * Update it to amazon.titan-embed-text-v2:0 or amazon.nova-2-multimodal-embeddings-v1:0
* Would we have additional parameters to let user control settings for these embeddings?

src/valkey-mcp-server/awslabs/valkey_mcp_server/embeddings/providers.py
```diff
@@ -0,0 +1,94 @@
...
+        )
+        embedding = await provider.generate_embedding("Hello world")
+    """
+
+    def __init__(self, region_name: str = "us-east-1", model_id: str = "amazon.titan-embed-text-v1"):
+        import boto3
+        self.client = boto3.client('bedrock-runtime', region_name=region_name)
```
seaofawareness: * Create boto3.Session() for isolated connection pool and configure max_pool_connections and retries
* Validate credentials are available
* Allow use of different dimensions to match model, expose it as a config/env parameter

src/valkey-mcp-server/awslabs/valkey_mcp_server/tools/semantic_search.py
```diff
@@ -0,0 +1,523 @@
...
+                if not index_exists:
+                    r.execute_command(
+                        'FT.CREATE', index_name,
+                        'ON', 'HASH',
+                        'PREFIX', '1', f'{_get_document_key(collection, "")}',
+                        'SCHEMA',
+                        'embedding', 'VECTOR', 'FLAT', '6',
+                        'TYPE', 'FLOAT32',
+                        'DIM', str(actual_dimensions),
+                        'DISTANCE_METRIC', 'L2'
```
seaofawareness: * Why not expose settings to user from env variables? Use of FLAT is limiting, HNSW with default settings should be supported for this PR.

src/valkey-mcp-server/awslabs/valkey_mcp_server/tools/semantic_search.py
```diff
@@ -0,0 +1,523 @@
...
+
+        # Valkey may need time to index, but it's never a good idea to sleep explicitly
+        # So, for now, it's okay if the total documents is inaccurate
+        # import time
+        # time.sleep(1.0)  # Wait for index to be ready
+
+        # Count total documents
+        total_docs = len(r.keys(f'{_get_document_key(collection, "")}*'))
+
+        return {
```
seaofawareness, line 183: Doc may be added though not indexed or run into error. Do you want to expose error count? Fine not to wait to get this status or provide it as a separate index_info tool.

src/valkey-mcp-server/awslabs/valkey_mcp_server/tools/vss.py
```diff
@@ -0,0 +1,135 @@
...
+            vector=[0.1, 0.2, 0.3, ...],  # 768-dimensional vector
+            count=5
+        )
+    """
+    try:
+        # Use connection with decode_responses=True for fetching document fields
+        r = ValkeyConnectionManager.get_connection(decode_responses=True)
+
+        # Access the search index
+        ft = r.ft(index)
```
seaofawareness, line 91: Handle scenario where Valkey doesn't have search module

src/valkey-mcp-server/awslabs/valkey_mcp_server/tools/semantic_search.py
```diff
@@ -0,0 +1,523 @@
...
+            ],
+            text_fields=["title", "content"]
+        )
+    """
+    if text_fields is None:
+        text_fields = ["content"]
+
+    try:
+        r = ValkeyConnectionManager.get_connection(decode_responses=True)
+        index_name = _get_collection_index_name(collection)
```
seaofawareness, line 124: Check if readonly mode is enabled and reject, refer https://github.com/awslabs/mcp/blob/main/src/valkey-mcp-server/awslabs/valkey_mcp_server/tools/hash.py#L59C7-L61

src/valkey-mcp-server/awslabs/valkey_mcp_server/tools/vss.py
```diff
@@ -0,0 +1,135 @@
...
+import struct
+import json
+import logging
+
+
+@mcp.tool()
+async def vector_search(index: str,
+                        field: str,
+                        vector: List[float],
+                        filter_expression: Optional[str] = None,
```
seaofawareness: Do you need to sanitize this string to ensure only safe/approved content is allowed to prevent unexpected executions? Any length limit checks etc?

src/valkey-mcp-server/scripts/seed_sample_data.py
```diff
@@ -0,0 +1,225 @@
...
+
+
+async def seed_data():
+    """Seed Valkey with sample product data and embeddings."""
+
+    print("🚀 Starting Valkey data seeding...")
+
+    # Initialize embeddings provider
+    try:
+        embeddings_provider = create_embeddings_provider()
```
seaofawareness: Do you want to support mock enbedding provider to enable validation without setup of actual provider?

src/valkey-mcp-server/tests/test_semantic_search_integration.py
```diff
@@ -0,0 +1,410 @@
...
+
+@pytest.mark.integration
+@pytest.mark.manual
+class TestSemanticSearchIntegration:
+    """Integration tests for semantic search functionality."""
+
+    @pytest.fixture(autouse=True)
+    def setup_valkey_config(self):
+        """Configure Valkey connection to use environment variables."""
+        original_config = VALKEY_CFG.copy()
```
seaofawareness, line 40: Error out if Valkey is not available or doesn't have search module

src/valkey-mcp-server/tests/test_semantic_search_integration.py
```diff
@@ -0,0 +1,410 @@
...
+        assert 'title' in minimal_results[0]
+        # Should not have full content fields
+        assert 'content' not in minimal_results[0] or minimal_results[0]['content'] == ""
+
+        print("\n" + "="*70)
+        print("✓ CONTENT CONTROL TESTS PASSED")
+        print("="*70)
+
+    @pytest.mark.asyncio
+    async def test_semantic_search_with_filter_expression(self):
```
seaofawareness, line 277: Add tests for invalid input

src/valkey-mcp-server/tests/test_vss.py
```diff
@@ -0,0 +1,328 @@
...
+        mock_conn.ft.return_value = mock_ft
+
+        # Execute search
+        result = await vector_search(index, field, vector, 0, count)
+
+        # Verify string message returned on error
+        assert isinstance(result, str)
+
+    @pytest.mark.asyncio
+    async def test_vector_search_connection_error(self, mock_connection):
```
seaofawareness, line 248: Add test for missing search module

seaofawareness (changes requested):
Great to see this addition, please address comments

MatthiasHowellYopp (commented):
LGTM

src/valkey-mcp-server/awslabs/valkey_mcp_server/tools/semantic_search.py
Jonathan-Improving: Do you mean in the `add_documents` MCP tool? why abort `semantic_search` which is itself read-only?

src/valkey-mcp-server/README.md
```diff
@@ -193,21 +193,68 @@ To run in readonly mode with Docker:
...
-| `VALKEY_SSL_CERTFILE` | Client's certificate file | `None` |
-| `VALKEY_CERT_REQS` | Server certificate verification | `"required"` |
-| `VALKEY_CA_CERTS` | Path to trusted CA certificates | `None` |
-| `VALKEY_CLUSTER_MODE` | Enable Valkey Cluster mode | `False` |
+### Valkey Connection
+The Valkey server can be configured using the following environment variables:
+
+| Name | Description                                                                                                                    | Default Value |
+|------|--------------------------------------------------------------------------------------------------------------------------------|---------------|
+| `VALKEY_HOST` | ElastiCache Primary Endpoint or MemoryDB Cluster Endpoint or Valkey IP or hostname                                             | `"127.0.0.1"` |
```
seaofawareness, line 202: For VSS, we have following limitations which need to be called out,
1. MemoryDB VSS is not compatible, so can't be used. Ensure right exception is returned to user as you find this mismatch. 
2. ElastiCache Serverless doesn't yet support VSS.

src/valkey-mcp-server/README.md
```diff
@@ -193,21 +193,68 @@ To run in readonly mode with Docker:
...
+Semantic search requires integration with an embeddings provider.
+The following environment variables are used to configure this:
+
+#### Common Embedding Configuration
+The first step is to configure the embeddings provider, which is used to generate embeddings for semantic search.
+This defaults to using the Bedrock embeddings provider, but can be configured to use any of the following:
+
+| Name | Description                                                                    | Default Value                  |
+|------|--------------------------------------------------------------------------------|--------------------------------|
+| `EMBEDDING_PROVIDER` | Embeddings provider to use (i.e. 'Ollama', 'Bedrock', or 'OpenAI')                         | `"bedrock"`                    |
```
seaofawareness, line 226: Did you test case insensitivity of provider names?

src/valkey-mcp-server/README.md
```diff
@@ -193,21 +193,68 @@ To run in readonly mode with Docker:
...
+
+#### Bedrock
+The following environment variables are used to configure the Bedrock embeddings provider, including a number
+of optional configuration options.  Credentials for the Bedrock service must be provided using the `AWS_ACCESS_KEY_ID` and
+`AWS_SECRET_ACCESS_KEY` environment variables or via `AWS_PROFILE`.
```
seaofawareness: Allow AWS_SESSION_TOKEN

src/valkey-mcp-server/awslabs/valkey_mcp_server/embeddings/__init__.py
```diff
@@ -0,0 +1,95 @@
...
+            base_url=EMBEDDING_CFG.get('ollama_host', 'http://localhost:11434'),
+            model=EMBEDDING_CFG.get('ollama_embedding_model', 'nomic-embed-text')
+        )
+
+    elif provider_type == 'bedrock':
+        return BedrockEmbeddings(
+            region_name=EMBEDDING_CFG.get('bedrock_region', 'us-east-1'),
+            model_id=EMBEDDING_CFG.get('bedrock_model_id', 'amazon.titan-embed-text-v1'),
+            normalize=EMBEDDING_CFG.get('bedrock_normalize'),
+            dimensions=EMBEDDING_CFG.get('bedrock_dimensions'),
```
seaofawareness: Should you provide to match amazon.titan-embed-text-v1 as default value?

src/valkey-mcp-server/awslabs/valkey_mcp_server/embeddings/__init__.py
```diff
@@ -0,0 +1,95 @@
...
+            input_type=EMBEDDING_CFG.get('bedrock_input_type'),
+            max_attempts=EMBEDDING_CFG.get('bedrock_max_attempts', 3),
+            max_pool_connections=EMBEDDING_CFG.get('bedrock_max_pool_connections', 50),
+            retry_mode=EMBEDDING_CFG.get('bedrock_retry_mode', 'adaptive')
+        )
+
+    elif provider_type == 'openai':
```
seaofawareness: Nice

src/valkey-mcp-server/awslabs/valkey_mcp_server/tools/semantic_search.py
seaofawareness: Yes

src/valkey-mcp-server/awslabs/valkey_mcp_server/tools/vss.py
```diff
@@ -0,0 +1,158 @@
...
+            # Get the document ID
+            doc_id = doc.id
+
+            # Fetch the document hash
+            doc_fields = r.hgetall(doc_id.encode() if isinstance(doc_id, str) else doc_id)
+
+            if doc_fields and b'document_json' in doc_fields:
+                try:
+                    # Parse the document_json field which contains the original document
+                    document_json_bytes = doc_fields[b'document_json']
```
seaofawareness: Valkey also supports JSON for VSS, is plan to enable it for future PR?

seaofawareness (changes requested):

src/valkey-mcp-server/README.md
Jonathan-Improving: Yes, they can be any case: upper, lower or mixed.

src/valkey-mcp-server/awslabs/valkey_mcp_server/tools/vss.py
Jonathan-Improving: Yes, for future PR.

src/valkey-mcp-server/README.md
Jonathan-Improving: So far, my findings indicate that MemoryDB is indistinct from Valkey, from a client's perspective.  But yes, we can at least document this limitation.

src/valkey-mcp-server/awslabs/valkey_mcp_server/embeddings/__init__.py
Jonathan-Improving: I don't follow, `amazon.titan-embed-text-v1` is the default both here and where `EMBEDDING_CFG` is initially configured.

src/valkey-mcp-server/awslabs/valkey_mcp_server/tools/semantic_search.py
```diff
@@ -0,0 +1,700 @@
...
+            
+            indexed_docs = int(info_dict.get('num_docs', 0))
+            
+            return {
+                "status": "success",
+                "collection": collection,
+                "stored_documents": stored_docs,
+                "indexed_documents": indexed_docs,
+                "indexing_complete": stored_docs == indexed_docs,
+                "index_name": index_name
```
seaofawareness: Should we expose following?
- percent_indexed
- hash_indexing_failures

seaofawareness (commented):
Overall LGTM, just couple comments.

src/valkey-mcp-server/awslabs/valkey_mcp_server/embeddings/__init__.py
seaofawareness: Default value already used 1536 dimension value, so should be fine.

seaofawareness (dismissed):
LGTM

src/valkey-mcp-server/awslabs/valkey_mcp_server/tools/semantic_search.py
```diff
@@ -0,0 +1,627 @@
...
+                }
+            ],
+            text_fields=["title", "content"]
+        )
+    """
+    if text_fields is None:
+        text_fields = ['content']
+
+    try:
+        r = ValkeyConnectionManager.get_connection(decode_responses=True)
```
madolson:
```suggestion
        v = ValkeyConnectionManager.get_connection(decode_responses=True)
```
i assume r is for redis. We should try to be valkey first as much as possible.

src/valkey-mcp-server/README.md
```diff
@@ -193,21 +193,69 @@ To run in readonly mode with Docker:
...
+| `VALKEY_CA_PATH` | CA certificate for verifying server                                                                                            | `None`        |
+| `VALKEY_SSL_KEYFILE` | Client's private key file                                                                                                      | `None`        |
+| `VALKEY_SSL_CERTFILE` | Client's certificate file                                                                                                      | `None`        |
+| `VALKEY_CERT_REQS` | Server certificate verification                                                                                                | `"required"`  |
+| `VALKEY_CA_CERTS` | Path to trusted CA certificates                                                                                                | `None`        |
+| `VALKEY_CLUSTER_MODE` | Enable Valkey Cluster mode                                                                                                     | `False`       |
+| `VALKEY_VECTOR_INDEX_TYPE` | Default vector index structure type to use when vector indices are created in calls to add documents (either `FLAT` or `HNSW`) | `FLAT`        |
```
zvi-code: Why is `FLAT` the default?

src/valkey-mcp-server/awslabs/valkey_mcp_server/common/connection.py
```diff
@@ -38,63 +39,77 @@ def get_connection(cls, decode_responses: bool = True) -> Union[Valkey, ValkeyCl
...
+                'ssl_certfile': VALKEY_CFG.get('ssl_certfile'),
+                'ssl_cert_reqs': ssl_cert_reqs,
+                'ssl_ca_certs': VALKEY_CFG.get('ssl_ca_certs'),
+                'decode_responses': decode_responses,
+                'lib_name': f'valkey-py(mcp-server_v{__version__})',
+            }
+
+            # Add max_connections parameter based on mode
+            if VALKEY_CFG['cluster_mode']:
+                connection_kwargs['max_connections_per_node'] = 10
```
zvi-code: What drives this number? What is the impact of higher value?
from valkey-search performance scales linearly with number of cores in the server and to get max qps high concurrency is usually required, for very large nodes we run with 300+ clients

src/valkey-mcp-server/README.md
Jonathan-Improving: No particular reason, we can change it to `HNSW` if you think that's better.

src/valkey-mcp-server/awslabs/valkey_mcp_server/common/connection.py
Jonathan-Improving: Perhaps the connections per node should be made configurable?

seaofawareness: Yes expose as a configuration with default value

src/valkey-mcp-server/README.md
seaofawareness: Let's start with HNSW as a default, ensure other HNSW configs are exposed as configuration

src/valkey-mcp-server/awslabs/valkey_mcp_server/common/connection.py
Jonathan-Improving: I've set the new default to 300 (from 10), please let me know if this is not correct.

seaofawareness (dismissed):

src/valkey-mcp-server/awslabs/valkey_mcp_server/tools/vss.py
```diff
@@ -0,0 +1,153 @@
...
+async def vector_search(
+    index: str,
+    field: str,
+    vector: List[float],
+    filter_expression: Optional[str] = None,
+    offset: int = 0,
+    count: int = 10,
+    no_content: bool = False,
+) -> Dict[str, Any]:
+    """Perform a Valkey vector search using the FT.SEARCH command.
```
zvi-code, line 45: Why not to extend the interface to expose impactful configurations like the returned fields and ef_runtime, something like:
```
async def vector_search(
    index: str,
    query_vector: List[float],
    k: int = 10,
    offset: int = 0,
    limit: int = 10,
    filter_expr: Optional[str] = None,  # e.g., "@category:{electronics}"
    return_fields: Optional[List[str]] = None,  # Selective RETURN
    nocontent: bool = False,
    ef_runtime: Optional[int] = None,
) -> SearchResults:
```

This is not a blocker anyway

src/valkey-mcp-server/awslabs/valkey_mcp_server/tools/vss.py
```diff
@@ -0,0 +1,153 @@
...
+            return {'status': 'success', 'results': []}
+
+        documents_list = []
+        if hasattr(result, 'docs'):  # type: ignore
+            for doc in result.docs:  # type: ignore
+                # Get the document ID
+                doc_id = doc.id
+
+                # Fetch the document hash
+                doc_fields = r.hgetall(doc_id.encode() if isinstance(doc_id, str) else str(doc_id))  # type: ignore
```
zvi-code: If `nocontent`, don't we only need document id?

And if `nocontent=false` , we get all the data in the query return, [see](https://github.com/valkey-io/valkey-search/blob/main/COMMANDS.md#ftsearch)

> RETURN <count> <field1> <field2> ... (options): count is the number of fields to return. Specifies the fields you want to retrieve from your documents, along with any aliases for the returned values. By default, **all fields are returned** unless the NOCONTENT option is set, in which case no fields are returned. If num is set to 0, it behaves the same as NOCONTENT.

src/valkey-mcp-server/awslabs/valkey_mcp_server/tools/index.py
```diff
@@ -0,0 +1,114 @@
...
+from awslabs.valkey_mcp_server.common.connection import ValkeyConnectionManager
+from enum import Enum
+from typing import Optional
+
+
+class IndexType(Enum):
+    """Valkey index type enumeration."""
+
+    HASH = 'HASH'
+    JSON = 'JSON'
```
zvi-code, line 26: Is JSON actually supported? I see `hgetall` does not work for json

src/valkey-mcp-server/awslabs/valkey_mcp_server/tools/vss.py
Jonathan-Improving: I do appreciate the suggestion!  Since this PR has been in review for over a month, I think it's best to consider this in a follow-up PR.

src/valkey-mcp-server/awslabs/valkey_mcp_server/tools/index.py
Jonathan-Improving: I think this is acceptable the way it is because this function is presently only used by the semantic search MCP only to create `HASH` index types.  [JSON is supported for FT.CREATE](https://valkey.io/commands/ft.create/), looks like the built syntax was slightly off, however, fixed.

src/valkey-mcp-server/awslabs/valkey_mcp_server/tools/vss.py
Jonathan-Improving: I've removed the unnecessary `HGETALL` and added an additional integration test to ensure this meets the expectation.

zvi-code (commented):
LGTM

seaofawareness (dismissed):

seaofawareness (dismissed):

src/valkey-mcp-server/scripts/seed_sample_data.py
scottschreckengaust, line 1: Can you help explain where this is called?

Jonathan-Improving: I was using that for a video demo, I don't think it needs to be a part of the repository so I will remove it.

seaofawareness (dismissed):

seaofawareness (dismissed):

src/valkey-mcp-server/awslabs/valkey_mcp_server/embeddings/__init__.py
```diff
@@ -0,0 +1,94 @@
...
+    Configuration via environment variables:
+    - EMBEDDINGS_PROVIDER: Provider type (ollama, bedrock, openai)
+
+    For Ollama provider:
+    - OLLAMA_HOST: Ollama server URL (default: http://localhost:11434)
+    - OLLAMA_EMBEDDING_MODEL: Ollama model name (default: nomic-embed-text)
+
+    For Titan / Bedrock provider:
+    - AWS_REGION: AWS region for Bedrock (default: us-east-1)
+    - BEDROCK_MODEL_ID: Bedrock model ID (default: amazon.titan-embed-text-v1)
```
scottschreckengaust: Update to current defaults

src/valkey-mcp-server/awslabs/valkey_mcp_server/embeddings/__init__.py
```diff
@@ -0,0 +1,94 @@
...
+    if provider_type == 'ollama':
+        return OllamaEmbeddings(
+            base_url=EMBEDDING_CFG.get('ollama_host', 'http://localhost:11434'),
+            model=EMBEDDING_CFG.get('ollama_embedding_model', 'nomic-embed-text'),
+        )
+
+    elif provider_type == 'bedrock':
+        return BedrockEmbeddings(
+            region_name=EMBEDDING_CFG.get('bedrock_region', 'us-east-1'),
+            model_id=EMBEDDING_CFG.get('bedrock_model_id', 'amazon.titan-embed-text-v1'),
```
scottschreckengaust: is it titan or `amazon.nova-2-multimodal-embeddings-v1:0'`?

src/valkey-mcp-server/awslabs/valkey_mcp_server/embeddings/__init__.py
Jonathan-Improving: Done

src/valkey-mcp-server/awslabs/valkey_mcp_server/embeddings/__init__.py
Jonathan-Improving: Nova, fixed.

seaofawareness (dismissed):

src/valkey-mcp-server/awslabs/valkey_mcp_server/embeddings/__init__.py
```diff
@@ -0,0 +1,96 @@
...
+    in the future to support additional embeddings providers such as OpenAI or Cohere etc.
+
+    Configuration via environment variables:
+    - EMBEDDINGS_PROVIDER: Provider type (ollama, bedrock, openai)
+
+    For Ollama provider:
+    - OLLAMA_HOST: Ollama server URL (default: http://localhost:11434)
+    - OLLAMA_EMBEDDING_MODEL: Ollama model name (default: nomic-embed-text)
+
+    For Titan / Bedrock provider:
```
seaofawareness: Fine to just refer Bedrock as titan is one of the models.

src/valkey-mcp-server/awslabs/valkey_mcp_server/embeddings/__init__.py
```diff
@@ -31,7 +31,7 @@ def create_embeddings_provider() -> EmbeddingsProvider:
     - OLLAMA_HOST: Ollama server URL (default: http://localhost:11434)
     - OLLAMA_EMBEDDING_MODEL: Ollama model name (default: nomic-embed-text)
 
-    For Titan / Bedrock provider:
+    For Titan provider:
     - AWS_REGION: AWS region for Bedrock (default: us-east-1)
     - BEDROCK_MODEL_ID: Bedrock model ID (default: amazon.nova-2-multimodal-embeddings-v1:0)
```
seaofawareness: Model refers to nova, though title talks about Titan

Jonathan-Improving: Yep, should be Bedrock as this piece is model-agnostic.

src/valkey-mcp-server/awslabs/valkey_mcp_server/common/config.py
```diff
@@ -33,6 +33,28 @@
...
+    else None,
+    'bedrock_dimensions': int(os.getenv('BEDROCK_DIMENSIONS', '0'))
+    if os.getenv('BEDROCK_DIMENSIONS')
+    else None,
+    'bedrock_input_type': os.getenv('BEDROCK_INPUT_TYPE'),
+    'bedrock_max_attempts': int(os.getenv('BEDROCK_MAX_ATTEMPTS', '3')),
+    'bedrock_max_pool_connections': int(os.getenv('BEDROCK_MAX_POOL_CONNECTIONS', '50')),
+    'bedrock_retry_mode': os.getenv('BEDROCK_RETRY_MODE', 'adaptive'),
+    'openai_api_key': os.getenv('OPENAI_API_KEY'),
+    'openai_model': os.getenv('OPENAI_MODEL', 'text-embedding-3-small'),
```
laithalsaadoon: consider a config builder function, I feel there is too much conditional logic inside a dict

src/valkey-mcp-server/awslabs/valkey_mcp_server/common/connection.py
laithalsaadoon, line 1: Is the connection manager diff related to the embeddings/VSS feature? if not, can you comment what this change is for specifically?

src/valkey-mcp-server/awslabs/valkey_mcp_server/embeddings/__init__.py
laithalsaadoon, line 1: Consider moving the function into another file instead of init (conventional/style thing we try to apply to the repo)

theagenticguy (changes requested):

src/valkey-mcp-server/awslabs/valkey_mcp_server/tools/semantic_search.py
```diff
@@ -0,0 +1,637 @@
...
+
+            indexed_docs = int(info_dict.get('num_docs', 0))
+
+            return {
+                'status': 'success',
+                'collection': collection,
+                'stored_documents': stored_docs,
+                'indexed_documents': indexed_docs,
+                'hash_indexing_failures': info_dict['hash_indexing_failures'],
+                'percent_indexed': stored_docs / indexed_docs if indexed_docs > 0 else 0,
```
laithalsaadoon: **Bug**: This calculation appears inverted. It computes `stored_docs / indexed_docs` but the variable name suggests it should be `indexed_docs / stored_docs`.

```python
# Current (likely wrong):
'percent_indexed': stored_docs / indexed_docs if indexed_docs > 0 else 0,

# Should probably be:
'percent_indexed': indexed_docs / stored_docs if stored_docs > 0 else 0,
```

Also, the denominator check should guard against the actual divisor to prevent division by zero.

src/valkey-mcp-server/awslabs/valkey_mcp_server/tools/vss.py
```diff
@@ -0,0 +1,153 @@
...
+        index: Name of the Valkey search index to use
+        field: Name of the vector field in the index to search against
+        vector: The query vector as a list of floats (must match the dimensionality of indexed vectors)
+        filter_expression: Optional filter expression to apply to the search results
+        offset: Record offset determining the window slice of results to render (default: 0)
+        count: Size of the window slice of results to render (default: 10)
+        no_content: If True, return only document IDs without content (default: False)
+
+    Returns:
+        An object indicating the results of the operation, with a "status" field set toe either "success" or "error",
```
laithalsaadoon: Minor: Typo in docstring - "toe" should be "to"

src/valkey-mcp-server/awslabs/valkey_mcp_server/tools/semantic_search.py
```diff
@@ -0,0 +1,637 @@
...
+                'type': search_results.get('type', 'valkey'),
+                'reason': search_results.get('reason', 'Unknown vector search error'),
+            }
+
+        # Extract the actual results list
+        results_list = search_results['results']
+
+        # If we don't need full content, filter to minimal fields
+        if not include_content:
+            results = []
```
laithalsaadoon, line 434: This repetitive dict-building pattern could be simplified with a comprehension:

```python
KEEP_FIELDS = {'id', 'title', 'name'}
result = {k: v for k, v in doc.items() if k in KEEP_FIELDS}
```

This is more maintainable if fields are added/removed later.

src/valkey-mcp-server/awslabs/valkey_mcp_server/tools/semantic_search.py
```diff
@@ -0,0 +1,637 @@
...
+                    if not index_exists:
+                        await create_vector_index(
+                            index_name,
+                            actual_dimensions,
+                            prefix=[_get_document_key(collection, '')],
+                            structure_type=VALKEY_CFG['vec_index_type'],
+                        )
+                        index_exists = True
+
+                embedding_bytes = struct.pack(f'{len(embedding)}f', *embedding)
```
laithalsaadoon: The `struct.pack` pattern for embedding serialization appears 3 times in this file and in `vss.py`. Consider extracting to a utility:

```python
def pack_embedding(embedding: list[float]) -> bytes:
    """Pack embedding vector to bytes for Valkey storage."""
    return struct.pack(f'{len(embedding)}f', *embedding)
```

This would live well in a `common/utils.py` or similar.

src/valkey-mcp-server/awslabs/valkey_mcp_server/embeddings/__init__.py
```diff
@@ -0,0 +1,96 @@
...
+
+"""Embeddings provider abstraction layer."""
+
+import os
+from .base import EmbeddingsProvider
+from .providers import OllamaEmbeddings, BedrockEmbeddings, OpenAIEmbeddings, HashEmbeddings
+from awslabs.valkey_mcp_server.common.config import EMBEDDING_CFG
+
+
+def create_embeddings_provider() -> EmbeddingsProvider:
```
laithalsaadoon: This 64-line factory function contains substantial business logic (provider routing, config parsing, validation). Consider moving it to a dedicated `factory.py` module - `__init__.py` is typically reserved for imports/re-exports rather than business logic. This would also improve testability.

src/valkey-mcp-server/awslabs/valkey_mcp_server/common/connection.py
Jonathan-Improving: Yes, it is.  Vector search operations can be connection-intensive.  I will add inline commentary.

seaofawareness (approved):

Jonathan-Improving:
Currently estimating effort to revise and simplify, removing semantic search features.

PR #1867, opened 2025-12-02
feat(aws-api-mcp-server): updated connect and read timeout

oleksii-k-amzn:
Fixes

## Summary
Updated read and connect timeout for AWS APIs

### Changes

### User experience
With lambda that takes 30+ seconds to run:
```
🤖 You are chatting with claude-sonnet-4

> /mcp

aws.aws-tasks-mcp-server
▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔
✓ aws.aws-tasks-mcp-server loaded in 1.12 s


> can you invoke <Lambda name> lambda function with empty payload

> I'll invoke the Lambda function with an empty payload for you.


🛠️  Using tool: call_aws from mcp server aws.aws-tasks-mcp-server
 ⋮
 ● Running call_aws with the param:
 ⋮  {
 ⋮    "name": "call_aws",
 ⋮    "arguments": {
 ⋮      "cli_command": "aws lambda invoke --function-name <Lambda name> --payload '{}' /var/folders/pb/tccrn6l561zfssgyskwj6jrr0000gq/T/aws-api-mcp/workdir/response.json"
 ⋮    }
 ⋮  }

Allow this action? Use 't' to trust (always allow) this tool for the session. [y/n/t]:

> t


 ⋮
 ● Completed in 32.170s


> The Lambda function was successfully invoked! The function returned a 200 status code, indicating successful execution. Let me check the response content:
```

## Checklist

* [x] Changes are documented

**RFC issue number**:

Checklist:

## Acknowledgment

### DIFF
--- a/src/aws-api-mcp-server/pyproject.toml
+++ b/src/aws-api-mcp-server/pyproject.toml
@@ -10,8 +10,8 @@ requires-python = ">=3.10"
 dependencies = [
     "mcp>=1.11.0",
     "pydantic>=2.10.6",
-    "boto3>=1.38.18",
-    "botocore>=1.38.18",
+    "boto3>=1.41.0",
+    "botocore[crt]>=1.41.0",
     "python-json-logger>=2.0.7",
     "setuptools>=69.0.0",
     "lxml>=5.1.0",
--- a/src/aws-api-mcp-server/uv.lock
+++ b/src/aws-api-mcp-server/uv.lock
@@ -82,13 +82,42 @@ wheels = [
 ]
 
 [[package]]
+name = "awscrt"
+version = "0.28.4"
+source = { registry = "https://pypi.org/simple" }
+sdist = { url = "https://files.pythonhosted.org/packages/4d/40/99afe81abec294594302e60ee51c5ade36c5535ad5275fa50160b8a42877/awscrt-0.28.4.tar.gz", hash = "sha256:d2835094e92d0a3d1722d03afd54983115b2172d57581a664ad6a2af3d33c12c", size = 37902030, upload-time = "2025-11-04T20:08:12.208Z" }
+wheels = [
+    { url = "https://files.pythonhosted.org/packages/62/96/cbd1822d38db89f7bb8f022c56d56d1428270d4d18f2a2d9acebb2b2af80/awscrt-0.28.4-cp310-cp310-macosx_10_15_universal2.whl", hash = "sha256:d1e205e53b08456f0f83210c20c674ebdef96e3e80f716d1bf4ad666db2c643b", size = 3391500, upload-time = "2025-11-04T20:07:14.14Z" },
+    { url = "https://files.pythonhosted.org/packages/ab/9e/65560a093d8e58d1a9e11c5e0e64e2a5f40eff8f5b66e9d7376e1c6f617b/awscrt-0.28.4-cp310-cp310-manylinux2014_aarch64.manylinux_2_17_aarch64.whl", hash = "sha256:dc11d00600888a690c1ad875759708a4d21bdf81b6c2032e0227687d27fca910", size = 3840080, upload-time = "2025-11-04T20:07:16.636Z" },
+    { url = "https://files.pythonhosted.org/packages/42/ea/0cfaaf771742a259918a0eb58377567dbd989e319ee0e8619e6c9c1774a0/awscrt-0.28.4-cp310-cp310-manylinux2014_x86_64.manylinux_2_17_x86_64.whl", hash = "sha256:13ed9b71a346146a89de85c173d007142416e6cc0358d7ca6b0d68dc1d159667", size = 4105891, upload-time = "2025-11-04T20:07:18.097Z" },
+    { url = "https://files.pythonhosted.org/packages/96/e8/bdf550ecb10dab8b30fab8fc493af241a5dd5d8a18a1eaa16f7440595b69/awscrt-0.28.4-cp310-cp310-musllinux_1_1_aarch64.whl", hash = "sha256:19adb9fa309111e20e1e850c876f093247ad084efdaa2dd654a15aef4b4bc637", size = 3762741, upload-time = "2025-11-04T20:07:19.532Z" },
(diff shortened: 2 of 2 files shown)


src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/common/config.py
```diff
@@ -150,3 +152,7 @@ def get_user_agent_extra() -> str:
 ENDPOINT_SUGGEST_AWS_COMMANDS = os.getenv(
     'ENDPOINT_SUGGEST_AWS_COMMANDS', 'https://api-mcp.global.api.aws/suggest-aws-commands'
 )
+DEFAULT_CONNECT_TIMEOUT_SECONDS = 10
+DEFAULT_READ_TIMEOUT_SECONDS = 60
```
arnewouters: Should we just stick to the [botocore defaults](https://botocore.amazonaws.com/v1/documentation/api/latest/reference/config.html) here?

> - connect_timeout (float or int) – The time in seconds till a timeout exception is thrown when attempting to make a connection. The default is 60 seconds.
> - read_timeout (float or int) – The time in seconds till a timeout exception is thrown when attempting to read from a connection. The default is 60 seconds.

oleksii-k-amzn: 60 seconds for connect timeout is too much IMO, while for read timeout 60 seconds makes sense. Additionally customer will be able to override the default config

src/aws-api-mcp-server/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+- Updated default AWS API connect and read timeout.
+- Added AWS_API_CONNECT_TIMEOUT_SECONDS and AWS_API_READ_TIMEOUT_SECONDS configuration to be able to override default timeout values.
```
arnewouters:
```suggestion

### Changed

- Updated default AWS API connect and read timeout and make it configurable (#1876)

## [1.1.8] - 2025-11-28

```

iddv (approved):
LGTM!

src/aws-api-mcp-server/README.md
```diff
@@ -202,26 +202,27 @@ See [DEPLOYMENT.md](https://github.com/awslabs/mcp/blob/main/src/aws-api-mcp-ser
...
+| `AWS_API_READ_TIMEOUT_SECONDS`                                    | ❌ No                       | 60                                                       | The timeout in seconds when attempting to read from a connection when calling AWS APIs.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       |
```
rshevchuk-git: Why should users care about these? IMO we should provide sensible defaults and not overwhelm customers with unnecessary configurations. README is hard to navigate already, given the number of configurations we describe in a single table.

oleksii-k-amzn: Because these timeout set on some average expectations. If the customer has some specific use case where this read timeout is not sufficient - think of lambda invocation that takes several minutes, as an example - we should give an ability to use override to make sure we can provide consistent experience in different conditions.
In addition there can be cases when the customers need to specify custom timeout config when they use MCP proxying for example.

rshevchuk-git: > we should give an ability to use override to make sure we can provide consistent experience in different conditions.

I don't get this, why not just remove this timeout to accommodate all use cases then? I can hardly imagine anyone using this to be honest. We haven't received any feedback or issues cut about the lack of configurability of connect/read timeouts. Having two configurations for the timeout adds mental overhead without much benefit. As a customer why do I have to care about the time API MCP waits before timing out when establishing a connection to AWS API?

>In addition there can be cases when the customers need to specify custom timeout config when they use MCP proxying for example.

Any concrete example?

oleksii-k-amzn: > Any concrete example?

as I mentioned above one of the use cases is synchronous lambda invocation for the cases when lambda run time exceeds 1 minute(think of Java example where cold start can take ~30 seconds)

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/parser/interpretation.py
nzmdn, line 29: Is this constant used elsewhere or can it be removed?

oleksii-k-amzn: it is used outside of this project(which was created as a quick workaround). it will be removed eventually. but removing it now would break one of the use cases outside of this repo

src/aws-api-mcp-server/README.md
oleksii-k-amzn: Will revert config override for now as this timeout can only be provided on socket level instead of tool level(i.e. total response timeout). Instead will create a ticket to address this

oleksii-k-amzn: https://github.com/awslabs/mcp/issues/1874

arnewouters (approved):

rshevchuk-git (approved):

PR #1870, opened 2025-12-02
feat(eks-mcp-server): add hosted MCP server documentation and registry metadata

djtung:
### Changes

- Add `server.json` metadata file used for publishing EKS hosted MCP to the [Official MCP Registry](https://github.com/modelcontextprotocol/registry/blob/main/docs/modelcontextprotocol-io/quickstart.mdx)
- Update README to announce new hosted MCP server offering from EKS
- Fix package name build script to allow for optional dependency names. We use the [mcp-proxy-for-aws](https://github.com/aws/mcp-proxy-for-aws) in the hosted setup which is called out in the README

### User experience

Users can now setup the EKS MCP server with a hosted option, eliminating local infra

## Checklist

* [X] Changes are documented

Is this a breaking change? N

**RFC issue number**:

Checklist:

## Acknowledgment

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -59,7 +59,7 @@ NOTICE                                           @awslabs/mcp-admins
 /src/cost-explorer-mcp-server                    @awslabs/mcp-admins @awslabs/mcp-maintainers  @Fedayizada
 /src/documentdb-mcp-server                       @awslabs/mcp-admins @awslabs/mcp-maintainers  @theagenticguy # @nitinahuja89
 /src/dynamodb-mcp-server                         @awslabs/mcp-admins @awslabs/mcp-maintainers  @akeyesamzn @shetsa-amzn @LeeroyHannigan
-/src/ecs-mcp-server                              @awslabs/mcp-admins @awslabs/mcp-maintainers  @guitar80ep @matthewgoodman13 @nineonine @biagic @tusharbabbar
+/src/ecs-mcp-server                              @awslabs/mcp-admins @awslabs/mcp-maintainers  @guitar80ep @matthewgoodman13 @nineonine @biagic @tusharbabbar @lewisct
 /src/eks-mcp-server                              @awslabs/mcp-admins @awslabs/mcp-maintainers  @patrick-yu-amzn @srhsrhsrhsrh
 /src/elasticache-mcp-server                      @awslabs/mcp-admins @awslabs/mcp-maintainers  @seaofawareness
 /src/finch-mcp-server                            @awslabs/mcp-admins @awslabs/mcp-maintainers  @Shubhranshu153 @pendo324


src/eks-mcp-server/server.json
```diff
@@ -0,0 +1,60 @@
...
+              "default": "us-east-1",
+              "description": "AWS region for EKS MCP service",
+              "isRequired": true
+            }
+          }
+        },
+        {
+          "name": "--service",
+          "type": "named",
+          "value": "eks-mcp"
```
srhsrhsrhsrh: Should we include that the default is inferred from the URL? I guess we don't really need it in the readme or public docs either...

src/eks-mcp-server/README.md
```diff
@@ -21,6 +25,58 @@ Integrating the EKS MCP server into AI code assistants enhances development work
...
+}
+```
+
+Replace `us-east-1` with your desired AWS region and `default` with your AWS CLI profile name. Add `--read-only` flag to restrict to read-only operations.
+
+#### IAM Permissions
+
+Your IAM role or user must have the following managed policies attached:
+
+- **For read-only operations**: `AmazonEKSMCPReadOnlyAccess`
```
srhsrhsrhsrh: Can we link https://docs.aws.amazon.com/aws-managed-policy/latest/reference/AmazonEKSMCPReadOnlyAccess.html?

src/eks-mcp-server/server.json
djtung: Sure, we can just remove this parameter. I can remove it from the README too

srhsrhsrhsrh:
Some details for reviewers: 

1. Verified that the verify package script fails without the modifications

```
❌ Package name verification failed for awslabs.eks-mcp-server
  - Package name mismatch: found 'mcp-proxy-for-aws' but expected 'awslabs.eks-mcp-server' (line 51)
```

2. `io.github.awslabs/eks-mcp-server` as the name in server.json resolves to this repo - see https://github.com/modelcontextprotocol/registry/blob/d840378a3a495bc55b70f6c7146686e854249538/docs/reference/server-json/official-registry-requirements.md#registry-api-metadata-vs-serverjson-metadata. It's not a link - it serves as an identifier based on the conventions in the MCP registry.

3. The local server references are left in README.md since it hasn't been deprecated yet.

src/eks-mcp-server/server.json
```diff
@@ -0,0 +1,55 @@
+{
+  "$schema": "https://static.modelcontextprotocol.io/schemas/2025-10-17/server.schema.json",
+  "description": "AI-powered Amazon EKS cluster management and troubleshooting",
+  "name": "io.github.awslabs/eks-mcp-server",
```
scottschreckengaust, line 4: I think this may need to be updated:

```suggestion
  "name": "aws.api.{region}.eks-mcp/mcp",
```

It is unclear if there may or may not need to be "global" endpoint (that isn't regional)

PR #1918, opened 2025-12-09
feat(aws-documentation-mcp-server): Add search intent capturing

JonLim:
Fixes

### Changes

Small change to the parameters of `search_documentation` to ask for the `search_intent` of a search query being made. This intent will be used to eventually improve the relevance of searches and overall quality of content being returned to users.

### User experience

No real change, but intents will be visible when LLMs are calling the `search_documentation` endpoint.

## Checklist

* [X] Changes are documented

Is this a breaking change? N

**RFC issue number**:

Checklist:

## Acknowledgment

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -59,7 +59,7 @@ NOTICE                                           @awslabs/mcp-admins
 /src/cost-explorer-mcp-server                    @awslabs/mcp-admins @awslabs/mcp-maintainers  @Fedayizada
 /src/documentdb-mcp-server                       @awslabs/mcp-admins @awslabs/mcp-maintainers  @theagenticguy # @nitinahuja89
 /src/dynamodb-mcp-server                         @awslabs/mcp-admins @awslabs/mcp-maintainers  @akeyesamzn @shetsa-amzn @LeeroyHannigan
-/src/ecs-mcp-server                              @awslabs/mcp-admins @awslabs/mcp-maintainers  @guitar80ep @matthewgoodman13 @nineonine @biagic @tusharbabbar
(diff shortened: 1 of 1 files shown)


src/aws-documentation-mcp-server/awslabs/aws_documentation_mcp_server/server_aws.py
```diff
@@ -165,6 +166,10 @@ async def read_documentation(
 async def search_documentation(
     ctx: Context,
     search_phrase: str = Field(description='Search phrase to use'),
+    search_intent: str = Field(
+        description='For the search_phrase parameter, describe the search intent of the user. CRITICAL: Do not include any PII or customer data, describe only the AWS-related intent for search.',
```
alexisareyn, line 170: do we have any guardrails to ensure that PII is not sent? or are we trusting agents to follow the instruction?

artb30, line 170: Is it okay for some PII to get through in some cases, is that a known limitation? When testing this I saw some PII was being sent in the intent.

JonLim: I am looking into something on our search proxy side to catch and filter out PII, if it manages to sneak through, so that we don't log them into any of our data stores. 

Nothing specific as of yet, but work in progress.

JonLim: We've been talking offline about this, and it seems like the PII might come through if your LLM wants to refer to you by name, because you've set it to remember your name. However in the examples you shared with me, it seems like it at least left off the AWS account ID that you had also provided, which is an encouraging sign, and we'll just have to scan queries as they come in and filter out any possible PII.

Thanks!

artb30 (approved):

alexisareyn (approved):
Approving and will look for changes on search side to further protect PII

PR #1920, opened 2025-12-09
feat(aws-api-mcp-server): add file access related information to call_aws tool description and update README

hfurkanbozkurt:
### Tests

Tested with MCP Inspector to validate descriptions are rendered correctly.

### Changes

Add file access related information to `call_aws` tool descrition + README updates

### User experience

Agents hopefully control their behavior better in tems of file references.

## Checklist

* [X] Changes are documented

N

Checklist:

* [ N/A] Migration process documented
* [ N/A] Implement warnings (if it can live side by side)

## Acknowledgment


src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
```diff
@@ -176,6 +184,7 @@ async def suggest_aws_commands(
     - All commands are validated before execution to prevent errors
     - Supports pagination control via max_results parameter
     - The current working directory is {WORKING_DIRECTORY}
+{_FILE_ACCESS_MSGS[FILE_ACCESS_MODE]}
```
arnewouters:
```suggestion
    - {_FILE_ACCESS_MSGS[FILE_ACCESS_MODE]}
```

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
```diff
@@ -71,6 +73,12 @@
 )
 READ_OPERATIONS_INDEX: Optional[ReadOnlyOperations] = None
 
+_FILE_ACCESS_MSGS = {
+    FileAccessMode.UNRESTRICTED: '    - You have unrestricted file access (commands can reference files anywhere)',
+    FileAccessMode.NO_ACCESS: '    - File access is disabled (commands with local file paths will be rejected)',
+    FileAccessMode.WORKDIR: '    - You have restricted file access (commands can only reference files in the current working directory)',
```
arnewouters:
```suggestion
    FileAccessMode.UNRESTRICTED: 'You have unrestricted file access (commands can reference files anywhere)',
    FileAccessMode.NO_ACCESS: 'File access is disabled (commands with local file paths will be rejected)',
    FileAccessMode.WORKDIR: 'You have restricted file access (commands can only reference files in the current working directory)',
```

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
hfurkanbozkurt: This was in case we wanted to add more information but alright with me

arnewouters (dismissed):

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
```diff
@@ -175,8 +183,8 @@ async def suggest_aws_commands(
     - For cross-region or account-wide operations, explicitly include --region parameter
     - All commands are validated before execution to prevent errors
     - Supports pagination control via max_results parameter
-    - The current working directory is {WORKING_DIRECTORY}
-    - File paths should always have forward slash (/) as a separator regardless of the system. Example: 'c:/folder/file.txt'
+    - {_FILE_ACCESS_MSGS[FILE_ACCESS_MODE]}
+    - File paths should either have forward slash (/) as a separator regardless of the system or should be relative to the current working directory. Example: 'c:/folder/file.txt' or 'file.txt' assuming current working directory is 'c:/folder/'.
```
nzmdn: >  or should be relative to the current working directory

To make this part true we need to fix a couple of issues:
- resolve relative paths using `WORKING_DIRECTORY`; the current implementation resolves it against the working directory of the MCP server process [core/common/file_system_controls.py#L145](https://github.com/hfurkanbozkurt/mcp/blob/065ede62acc734a668ce4dc43ff2ca63795f60b9/src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/common/file_system_controls.py#L145)
- remove `isabs` here [core/parser/parser.py#L856](https://github.com/hfurkanbozkurt/mcp/blob/065ede62acc734a668ce4dc43ff2ca63795f60b9/src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/parser/parser.py#L856)

> File paths should either have forward slash (/) as a separator regardless of the system or should be relative to the current working directory

Also, the whole sentence sounds like `\` can be used if the path is relative.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
```diff
@@ -175,8 +183,8 @@ async def suggest_aws_commands(
     - For cross-region or account-wide operations, explicitly include --region parameter
     - All commands are validated before execution to prevent errors
     - Supports pagination control via max_results parameter
-    - The current working directory is {WORKING_DIRECTORY}
```
nzmdn, line 178: Why is this removed?

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
nzmdn: Additionally, it makes sense not to mention paths at all when access mode is `FileAccessMode.NO_ACCESS`

hfurkanbozkurt: >resolve relative paths using WORKING_DIRECTORY; the current implementation resolves it against the working directory of the MCP server process [core/common/file_system_controls.py#L145](https://github.com/hfurkanbozkurt/mcp/blob/065ede62acc734a668ce4dc43ff2ca63795f60b9/src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/common/file_system_controls.py#L145)

The chdir is set in server so relative resolution will be against `WORKING_DIRECTORY`: https://github.com/hfurkanbozkurt/mcp/blob/065ede62acc734a668ce4dc43ff2ca63795f60b9/src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py#L387

`isabs` is definitely a miss, let me handle that!

src/aws-api-mcp-server/README.md
```diff
@@ -205,8 +205,8 @@ See [DEPLOYMENT.md](https://github.com/awslabs/mcp/blob/main/src/aws-api-mcp-ser
...
+| `AWS_API_MCP_WORKING_DIR`                                         | ❌ No                       | \<Platform-specific temp directory\>/aws-api-mcp/workdir | Working directory path for the MCP server operations. Must be an absolute path when provided. Used to resolve relative paths in commands like `aws s3 cp`. When file access mode is `"workdir"` (default), file operations are restricted to this directory. Set a broad path for your use case when using "workdir" mode. If not provided, defaults to a platform-specific directory:<br/><br/>• **Windows**: `%TEMP%\aws-api-mcp\workdir` (typically `C:\Users\<username>\AppData\Local\Temp\aws-api-mcp\workdir`)<br/>• **macOS**: `/private/var/folders/<hash>/T/aws-api-mcp/workdir`<br/>• **Linux**: `$XDG_RUNTIME_DIR/aws-api-mcp/workdir` (if set) or `$TMPDIR/aws-api-mcp/workdir` (if set) or `/tmp/aws-api-mcp/workdir` |
```
rshevchuk-git: Why are we suggesting broad path? This might have security implications

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
hfurkanbozkurt: That is a mistake while working on the line below, thanks for spotting, Fixing now.

src/aws-api-mcp-server/README.md
wesk:
```suggestion
| `AWS_API_MCP_WORKING_DIR`                                         | ❌ No                       | \<Platform-specific temp directory\>/aws-api-mcp/workdir | Working directory path for the MCP server operations. Must be an absolute path when provided. Used to resolve relative paths in commands like `aws s3 cp`. When file access mode is `"workdir"` (default), file operations are restricted to this directory. Set an appropriate path for your use case when using "workdir" mode. If not provided, defaults to a platform-specific directory:<br/><br/>• **Windows**: `%TEMP%\aws-api-mcp\workdir` (typically `C:\Users\<username>\AppData\Local\Temp\aws-api-mcp\workdir`)<br/>• **macOS**: `/private/var/folders/<hash>/T/aws-api-mcp/workdir`<br/>• **Linux**: `$XDG_RUNTIME_DIR/aws-api-mcp/workdir` (if set) or `$TMPDIR/aws-api-mcp/workdir` (if set) or `/tmp/aws-api-mcp/workdir` |
```

hfurkanbozkurt: It really depends on how users interpret this. I believe it suggests not using a single subpackage but give the root of the package for example. I dont think people go and give their home directories. Do you think we need to change/revert that part?

rshevchuk-git: If ambiguity exists, expect it to be interpreted in the least safe way. Think of Murphy’s Law.. user's might still end up shooting themselves in the foot when using this, but we should do our best to prevent this. Instead of suggesting broad paths, we should have more defensive stance, and rather say something along the lines of "pick a dir where you feel safe allowing agent full read & write access". That's why the original "Does not provide any sandboxing or security restrictions" was in there.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
```diff
@@ -71,6 +73,12 @@
 )
 READ_OPERATIONS_INDEX: Optional[ReadOnlyOperations] = None
 
+_FILE_ACCESS_MSGS = {
+    FileAccessMode.UNRESTRICTED: 'File access is unrestricted so commands can reference files anywhere.',
+    FileAccessMode.NO_ACCESS: 'File access is disabled and commands with any local file reference will be rejected.',
+    FileAccessMode.WORKDIR: 'Commands can only reference files within the current working directory; otherwise, they will be rejected.',
+}
+
```
bariskurt, line 81: Can we clearly state what the WORKING_DIRECTORY is in the FileAccessMode.WORKDIR message ? We can just append the deleted "The current working directory is {WORKING_DIRECTORY}" to this case.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
hfurkanbozkurt: Explicitly did this without relying on the chdir.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/parser/parser.py
```diff
@@ -795,13 +795,9 @@ def _validate_s3_file_paths(service: str, operation: str, parameters: dict[str,
 
 
 def _validate_s3_file_path(file_path: str, service: str, operation: str):
+    # Allow '-' for stdout/stdin streaming
```
arnewouters: Now we are allowing it for all s3 operations. We should only allow `-` if it's a target and when the operation is `cp`. It doesn't really make sense for other operations or as source.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/common/file_system_controls.py
```diff
@@ -141,9 +141,16 @@ def validate_file_path(file_path: str) -> str:
...
+    # Reject unexpanded tilde paths (e.g., ~invalid_user/path)
+    if file_path.startswith('~') and not os.path.isabs(os.path.expanduser(file_path)):
+        raise FilePathValidationError(file_path, 'contains unexpanded tilde (~) which is not allowed')
+
+    # Convert to absolute path, resolving relative paths against WORKING_DIRECTORY
     working_directory = os.path.abspath(WORKING_DIRECTORY)
+    if os.path.isabs(file_path):
+        absolute_path = os.path.abspath(file_path)
+    else:
+        absolute_path = os.path.abspath(os.path.join(working_directory, file_path))
```
arnewouters: Not sure why we need these changes?

- We expand the user directory already [here](https://github.com/awslabs/mcp/blob/main/src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/parser/parser.py#L390) before parsing
- Is joining the working_directory with the relative path needed since we set `chdir` [here](https://github.com/awslabs/mcp/blob/main/src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py#L380)?

arnewouters (changes requested):

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/common/file_system_controls.py
hfurkanbozkurt: `chdir` does this but in order to be safe, I wanted to do it here and not rely on that.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/parser/parser.py
hfurkanbozkurt: Is there any risks of allowing for all?

hfurkanbozkurt: Looking at commands, I think it is only accepted for `cp` anyway but I am happy to change if needed.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/parser/parser.py
```diff
@@ -791,16 +789,21 @@ def _validate_s3_file_paths(service: str, operation: str, parameters: dict[str,
...
+    # Allow '-' (stdout) only for s3 cp destination
     if file_path == '-':
+        if operation == 'cp' and is_destination:
+            return
         raise FileParameterError(
             service=service,
             operation=operation,
             file_path=file_path,
-            reason="streaming file ('-') is not allowed",
+            reason="streaming to stdout ('-') is only supported as destination in 'aws s3 cp'",
```
arnewouters: Does mentioning `aws s3 cp` here make sense? This error is shown to the LLM, just want to make sure this doesn't create confusion

hfurkanbozkurt: Any suggestions? I think it is good that it knows the correct behavior? Are you concerned that when we expand the behavior (if we do), this part will be obsolete?

hfurkanbozkurt:
There is another PR (https://github.com/awslabs/mcp/pull/1954) that handles S3 related changes so I am going to be removing those from this PR.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
```diff
@@ -175,8 +183,7 @@ async def suggest_aws_commands(
     - For cross-region or account-wide operations, explicitly include --region parameter
     - All commands are validated before execution to prevent errors
     - Supports pagination control via max_results parameter
-    - The current working directory is {WORKING_DIRECTORY}
-    - File paths should always have forward slash (/) as a separator regardless of the system. Example: 'c:/folder/file.txt'
+    - {_FILE_ACCESS_MSGS[FILE_ACCESS_MODE]}
```
rshevchuk-git, line 186: This is vulnerable. Let's discuss

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
```diff
@@ -71,6 +73,12 @@
 )
 READ_OPERATIONS_INDEX: Optional[ReadOnlyOperations] = None
 
+_FILE_ACCESS_MSGS = {
+    FileAccessMode.UNRESTRICTED: f'File access is unrestricted so commands can reference files anywhere; use forward slashes (/) or relative paths from the working directory ({WORKING_DIRECTORY}).',
```
nzmdn, line 77: We should still keep `regardless of the system` and an example. So it can be something like:

```
use forward slashes (/) regardless of the system (e.g. 'c:/folder/filet.txt'); relative paths are resolved relative to the working directory ({WORKING_DIRECTORY}).
```

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
hfurkanbozkurt: Discussed offline. This will be tracked separately.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
hfurkanbozkurt: Will do a mini PR after this to address this.

hfurkanbozkurt:
Small addition to this PR: https://github.com/awslabs/mcp/pull/1967

PR #1923, opened 2025-12-10
fix: add pagination support for list_state_machines in Step Functions MCP server

JulienJBO:
### Changes

Added pagination support for `list_state_machines()` API call in the Step Functions MCP server. The previous implementation only retrieved the first page of state machines (max 100), which caused issues when accounts had more than 100 state machines.

The fix implements a pagination loop using the `nextToken` parameter to retrieve all state machines across multiple pages.

### User experience

**Before:**
- Only the first 100 state machines were registered as MCP tools
- Users with more than 100 state machines would not see all their state machines available

**After:**
- All state machines are now registered as MCP tools, regardless of the total count
- Complete visibility of all Step Functions state machines in the account

## Checklist

* [x] Changes are documented

Is this a breaking change? (Y/N): **N**

**RFC issue number**: https://github.com/awslabs/mcp/issues/2050

## Acknowledgment

### DIFF
--- a/src/stepfunctions-tool-mcp-server/awslabs/stepfunctions_tool_mcp_server/server.py
+++ b/src/stepfunctions-tool-mcp-server/awslabs/stepfunctions_tool_mcp_server/server.py
@@ -336,10 +336,23 @@ def register_state_machines():
     """Register Step Functions state machines as individual tools."""
     try:
         logger.info('Registering Step Functions state machines as individual tools...')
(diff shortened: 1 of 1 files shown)


JulienJBO:
Hi !
Is there anything I can do for helping reviewers ?

JulienJBO:
@mmouniro @sten-vw @arnewouters please can you review this PR ?

PR #1925, opened 2025-12-10
fix(cloudwatch_mcp_server): Fix timezone datetime handling in GetMetricData

goranmod:
## Summary
This request fixes an issue where the `get_metric_data` tool throws an error in cases where the start and end times have a mismatch in the timezone information. This issue manifests in the following error:

```
2025-11-05 15:44:47.066 | ERROR    | awslabs.cloudwatch_mcp_server.cloudwatch_metrics.tools:get_metric_data:364 - Error in get_metric_data: can't subtract offset-naive and offset-aware datetime
```

### Changes

The change consists in applying the UTC timezone by default when a timezone is not passed explicitly to the start or end time parameters. Additionally, both values are set to have a timezone to allow for correct timezone arithmetic.

### User experience

User experience will be enhanced considering they will not receive errors on valid inputs when using the `get_metri_data` tool.

## Checklist

* [] Changes are documented

- No

**RFC issue number**: N/A

Checklist:

## Acknowledgment

### DIFF
--- a/src/cloudwatch-mcp-server/awslabs/cloudwatch_mcp_server/cloudwatch_metrics/tools.py
+++ b/src/cloudwatch-mcp-server/awslabs/cloudwatch_mcp_server/cloudwatch_metrics/tools.py
@@ -162,7 +162,10 @@ class CloudWatchMetricsTools:
         ctx: Context,
         namespace: str,
         metric_name: str,
-        start_time: Union[str, datetime],
+        start_time: Annotated[
+            Union[str, datetime],
+            Field(description='The start time for the metric data query (ISO format or datetime)'),
+        ],
         dimensions: List[Dimension] = [],
         end_time: Annotated[
             Union[str, datetime] | None,


src/cloudwatch-mcp-server/awslabs/cloudwatch_mcp_server/cloudwatch_metrics/tools.py
```diff
@@ -390,10 +390,17 @@ def _prepare_time_parameters(self, start_time, end_time, target_datapoints):
             start_time = datetime.fromisoformat(start_time.replace('Z', '+00:00'))
 
         if end_time is None:
-            end_time = datetime.utcnow()
```
agiuliano, line 393: @goranmod should we also update the description in the tool so that we can teach the LLM what is the expected format for start-end time?

goranmod: There was already steering in the `end_time` around this, and the examples clearly showed what we were expecting. I added a description for the `start_time` as well, which should make it clearer.

The issue here is that we can still get datetime as well as that the end_time is optional (as expected), so the tool was mixing different types. And we shouldn't change the expected types now, considering we would be breaking backwards compatibility.

shri-tambe (approved):

PR #1939, opened 2025-12-11
chore(aws-iac-mcp-server): read tool now proxies knowledge MCP tool

vishaalmehrishi:
### Changes

The `read_iac_documentation_page` tool now proxies the remote AWS Knowledge MCP read tool. This allows for changes to the tool signature to make agent-facing improvements.

### User experience

No change. I tested this manually with a local setup and it works as expected.

## Checklist

* [x] Changes are documented

Is this a breaking change? N

**RFC issue number**: N/A

Checklist:

## Acknowledgment

### DIFF
--- a/src/aws-iac-mcp-server/awslabs/aws_iac_mcp_server/server.py
+++ b/src/aws-iac-mcp-server/awslabs/aws_iac_mcp_server/server.py
@@ -32,6 +32,7 @@ from .tools.iac_tools import (
 from dataclasses import asdict
 from fastmcp import FastMCP
 from fastmcp.server.proxy import ProxyClient
+from loguru import logger
 from typing import Optional
 
 
@@ -529,7 +530,13 @@ def main():
     import asyncio
 
     # Create the read tool proxy before starting the server
-    asyncio.run(_create_read_tool_proxy())
+    # Don't fail the entire server startup in case there is an issue with the proxy
+    try:
+        asyncio.run(_create_read_tool_proxy())
+    except Exception as e:
+        logger.warning(
+            f'Failed to initialize read tool proxy: {e}. Continuing with server initialization.'
+        )
 
     # Run the server
     mcp.run()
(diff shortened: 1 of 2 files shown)


kumvprat (commented):
One question : If MCP user is not able to connect to remote mcp then the iac mcp initialization will fail ?

Current behavior is that the remote mcp call happens only when the specific tool is invoked.

I want to make sure that this change in behaviour is expected(or even possible, might not be if my understanding is wrong)

aemada-aws (approved):

vishaalmehrishi:
> One question : If MCP user is not able to connect to remote mcp then the iac mcp initialization will fail ?
> 
> Current behavior is that the remote mcp call happens only when the specific tool is invoked.
> 
> I want to make sure that this change in behaviour is expected(or even possible, might not be if my understanding is wrong)

Yes the initialization will fail. For this MCP, the dependency on the remote is quite significant, since the search tools will also fail to work if the remote MCP is not available. At that point only the local CFN tools will be available.

I can add error handling if we feel that we should provide the local CFN tools even when the search/read functionality isn't working.

src/aws-iac-mcp-server/awslabs/aws_iac_mcp_server/server.py
```diff
@@ -333,58 +357,6 @@ async def search_cdk_documentation(query: str) -> str:
...
-
-    Returns:
-        List of search results with URLs, titles, and context snippets
-    """
-    result = await read_iac_documentation_page_tool(url, starting_index)
-
-    # Convert dataclass to dict for JSON serialization
-    response_dict = asdict(result)
-
-    return sanitize_tool_response(json.dumps(response_dict))
```
vishaalmehrishi, line 385: The sanitisation is not being applied to the proxied read tool, since it is a pure proxy of the remote AWS knowledge MCP tool i.e. the response will be exactly what the remote tool returns. Applying sanitisation would require first parsing the ToolResult into a string, needing us to make assumptions about the structure of the response from the remote tool - the very thing we're trying to avoid by making this change (decoupling from the remote tool request/response structures).

kdbrogan (approved):
LGTM

PR #1954, opened 2025-12-13
fix(aws-api-mcp-server): allow s3 cp on stdout and improve error handling for cli customizations

arnewouters:
Fixes

### Changes

- Allow `-` as destination for s3 commands
- Improve error handling by throwing an exception if a CLI customization only has output on stderr

Technically, `-` only works for `cp` commands but now we rely on the errors from AWS CLI:

```txt
Streaming currently is only compatible with non-recursive cp commands
```

We still explicitly block `-` as source since this is for input on stdin which the MCP server doesn't support.

### User experience

`aws s3 cp s3://my-bucket/my-file.txt -` works after this change.

Error handling before (normal response that contains error text):

```json
{"response":"","error":"\nStreaming currently is only compatible with non-recursive cp commands\n"}
```

After (actual MCP error):

```txt
Error:
Error calling tool 'call_aws': Error while executing 'aws s3 mv s3://my-bucket/my-file.txt -': 
Streaming currently is only compatible with non-recursive cp commands
```

## Checklist

* [x] Changes are documented

**RFC issue number**:

Checklist:

## Acknowledgment

### DIFF
--- a/src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/parser/parser.py
+++ b/src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/parser/parser.py
@@ -797,7 +797,7 @@ def _validate_s3_file_paths(service: str, operation: str, parameters: dict[str, 
 def _validate_s3_file_path(
     file_path: str, service: str, operation: str, is_destination: bool = False
 ):
-    # `-` as destination redirects to stdout
+    # `-` as destination redirects to stdout, which we capture and wrap in an MCP response
     if file_path == '-' and is_destination:
         return
 


src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/aws/service.py
```diff
@@ -175,7 +175,12 @@ def execute_awscli_customization(
         stdout_output = stdout_capture.getvalue()
         stderr_output = stderr_capture.getvalue()
 
+        if not stdout_output and stderr_output:
+            raise AwsApiMcpError(f"Error while executing '{cli_command}': {stderr_output}")
+
         return AwsCliAliasResponse(response=stdout_output, error=stderr_output)
+    except AwsApiMcpError:
+        raise
```
rshevchuk-git: Not sure I understand what scenario are we optimizing for here? Can you give a specific example where this is needed? 

Also not sure if we need this new 'except'. Can we not adjust the error message and catch it within catch-all block, which already raises the same error?

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/parser/parser.py
```diff
@@ -791,16 +791,21 @@ def _validate_s3_file_paths(service: str, operation: str, parameters: dict[str,
...
     _validate_s3_file_path(source_path, service, operation)
-    _validate_s3_file_path(dest_path, service, operation)
+    _validate_s3_file_path(dest_path, service, operation, is_destination=True)
 
 
-def _validate_s3_file_path(file_path: str, service: str, operation: str):
+def _validate_s3_file_path(
+    file_path: str, service: str, operation: str, is_destination: bool = False
+):
+    if file_path == '-' and is_destination:
```
rshevchuk-git, line 801: I'd suggest we add a comment explaining what '-' means and why we allow this specific combination

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/parser/parser.py
```diff
@@ -791,16 +791,21 @@ def _validate_s3_file_paths(service: str, operation: str, parameters: dict[str,
 
     source_path, dest_path = paths
     _validate_s3_file_path(source_path, service, operation)
-    _validate_s3_file_path(dest_path, service, operation)
+    _validate_s3_file_path(dest_path, service, operation, is_destination=True)
 
 
-def _validate_s3_file_path(file_path: str, service: str, operation: str):
+def _validate_s3_file_path(
```
hfurkanbozkurt, line 797: Do we want this to restrict to `cp`? Imo it is not needed but wanted to re-raise as it was raised before.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/aws/service.py
arnewouters: > Not sure I understand what scenario are we optimizing for here? Can you give a specific example where this is needed?

Before we would have this as response:

```json
{"response":"","error":"\nStreaming currently is only compatible with non-recursive cp commands\n"}
```

After the changes, we get a proper error response

```txt
Error:
Error calling tool 'call_aws': Error while executing 'aws s3 mv s3://my-bucket/my-file.txt -': 
Streaming currently is only compatible with non-recursive cp commands
```

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/parser/parser.py
arnewouters: I did not restrict it here so we don't add too much custom handling and instead rely on the AWS CLI code, which will give an error for `mv`, etc.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/aws/service.py
arnewouters: > Also not sure if we need this new 'except'. Can we not adjust the error message and catch it within catch-all block, which already raises the same error?

Yes that makes more sense, will change that

sten-vw (approved):

PR #1955, opened 2025-12-13
fix(aws-api-mcp-server): convert non-compatible JSON data types before applying JMESPath expression

### LINKED ISSUE #1899: aws-api-mcp-server: s3api list-buckets query not working
### Describe the bug

JMESPath query is not working in the AWS API MCP server but works with AWS CLI

```sh
aws s3api list-buckets --query "sort_by(Buckets, &CreationDate)[-1].[Name,CreationDate]"
```

### Expected Behavior

The command to work just like it does if you run it directly with the AWS CLI

```sh
aws s3api list-buckets --query "sort_by(Buckets, &CreationDate)[-1].[Name,CreationDate]"
```

```json
[
    "bedrock-agentcore-runtime-<ACCOUNT_ID>-us-east-1-abcdefghijkl",
    "2025-11-07T10:16:12+00:00"
]
```

### Current Behavior

```json
{
  "cli_command": "aws s3api list-buckets --query \"sort_by(Buckets, &CreationDate)[-1].[Name,CreationDate]\""
}
```

```txt
Error:
Error calling tool 'call_aws': Error while executing the command: In function sort_by(), invalid type for value: {'Name': 'awouters-bucket-1', 'CreationDate': datetime.datetime(2025, 6, 16, 11, 2, 35, tzinfo=tzutc()), 'BucketArn': 'arn:aws:s3:::awouters-bucket-1'}, expected one of: ['string', 'number'], received: "unknown"
```

### Reproduction Steps

Run command with AWS API MCP server

### Possible Solution

_No response_

_No response_

### OS

macOS

### Server

aws-api-mcp-server

### Server Version

1.1.8

us-east-1

### Other information

_No response_

### Service quota

arnewouters:
Fixes #1899 

### Changes

JMESPath expressions were being applied to data types that are not compatible with JSON. With this PR we first convert the data so it is JSON compatible before applying the JMESPath expression.

For example in #1899, `CreationDate` is a datetime object instead of a string or number.

### User experience

JMESPath expressions now work correctly.

## Checklist

* [x] Changes are documented

**RFC issue number**:

Checklist:

## Acknowledgment

### DIFF
--- a/src/aws-api-mcp-server/tests/common/test_helpers.py
+++ b/src/aws-api-mcp-server/tests/common/test_helpers.py
@@ -2,10 +2,8 @@ import pytest
 from awslabs.aws_api_mcp_server.core.common.helpers import (
     get_requests_session,
     is_help_operation,
-    json_serializer,
     validate_aws_region,
 )
-from datetime import datetime
 from requests.adapters import HTTPAdapter
 from unittest.mock import MagicMock, patch
 
@@ -106,20 +104,6 @@ def test_get_requests_session():
     assert retry_config.backoff_factor == 1
     assert retry_config.status_forcelist == [429, 500, 502, 503, 504]
     assert retry_config.allowed_methods == {'HEAD', 'GET', 'OPTIONS', 'POST'}
-
-
-def test_json_serializer_datetime():
-    """Test that json_serializer converts datetime to ISO format string."""
-    dt = datetime(2025, 12, 13, 14, 30, 45)
-    result = json_serializer(dt)
-    assert result == '2025-12-13T14:30:45'
-
-
-def test_json_serializer_bytes():
-    """Test that json_serializer converts bytes to string representation."""
-    data = b'binary content'
-    result = json_serializer(data)
-    assert result == "b'binary content'"
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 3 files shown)


src/aws-api-mcp-server/tests/common/test_helpers.py
```diff
@@ -106,6 +108,20 @@ def test_get_requests_session():
     assert retry_config.allowed_methods == {'HEAD', 'GET', 'OPTIONS', 'POST'}
 
 
+def test_json_serializer_datetime():
```
rshevchuk-git: This feels like a low coverage for the change. It effectively tests built in functions, isoformat() and str(obj). Instead of testing the actual change. I'd expect more e2e tests that actually verify support for new commands

arnewouters: I was under the impression that we already had a test that covered it in `test_interpret_returns_valid_response`. That test broke for the EC2 query but after taking a closer look it doesn't actually cover the correct usecase, also noticed that we already had `Boto3Encoder` helper class which does the same thing..

I will reuse `Boto3Encoder` and add the query from the issue to the test so it covers that path.

rshevchuk-git (approved):

PR #1962, opened 2025-12-16
fix(aws-api-mcp-server): only create the default working directory and not custom ones

arnewouters:
Fixes

### Changes

### User experience

It's the user's responsibility to create the custom working directory

## Checklist

* [x] Changes are documented

**RFC issue number**:

Checklist:

## Acknowledgment

### DIFF
--- a/src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/common/config.py
+++ b/src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/common/config.py
@@ -19,6 +19,7 @@ import tempfile
 from enum import Enum
 from fastmcp.server.auth import JWTVerifier
 from fastmcp.server.dependencies import get_context
+from loguru import logger
 from pathlib import Path
 from typing import Literal, cast
 


anasstahr:
Why do we need this change? What is the issue with the MCP server creating a custom directory?

anasstahr (approved):

src/aws-api-mcp-server/tests/common/test_config.py
```diff
@@ -323,3 +325,105 @@ def test_get_file_access_mode_unknown_defaults_to_workdir(monkeypatch, env_value
     monkeypatch.setenv('AWS_API_MCP_ALLOW_UNRESTRICTED_LOCAL_FILE_ACCESS', env_value)
     result = get_file_access_mode()
     assert result == FileAccessMode.WORKDIR
+
```
rshevchuk-git, line 280: Seems like we're only covering happy paths in these tests. We should also add security tests to make sure we do not regress back to something like "/path/to/dir\n MALICIOUS INJECTION"

arnewouters: This is a valid path though, especially on Windows where we have backslashes e.g. `C:\Users\<username>\AppData\Local\Temp\aws-api-mcp\workdir`

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/common/config.py
```diff
@@ -122,6 +124,28 @@ def get_user_agent_extra() -> str:
     return user_agent_extra
 
 
+def get_working_directory() -> Path:
+    """Returns the custom working directory if AWS_API_MCP_WORKING_DIR is set, otherwise returns a default directory under the server directory."""
+    if custom_workdir := os.getenv(AWS_API_MCP_WORKING_DIR_KEY):
```
oleksii-k-amzn, line 129: Can we add a default value here or check if the env var was set

PR #1993, opened 2025-12-22
fix: bedrock kb: restrict reranking models in us-east-1

universam1:
## Summary
[The 'AMAZON' reranking model is not available in the `us-east-1` region](https://docs.aws.amazon.com/bedrock/latest/userguide/rerank-supported.html#:~:text=The%20Amazon%20Rerank%201.0%20model%20is%20not%20supported%20in%20the%20US%20East%20(N.%20Virginia)%20AWS%20Region.%20You%20can%20only%20use%20the%20Cohere%20Rerank%203.5%20model%20in%20this%20Region.), causing runtime errors as it is default parameter choice.

This ensures the tool definition accurately reflects the capabilities available in the current region.
### Changes
This change updates `server.py` to dynamically configure the available `reranking_model_name` options based on the `AWS_REGION` environment variable: 
- In `us-east-1`: The 'AMAZON' option is removed, and the default is set to 'COHERE'.
- In other regions: Both 'COHERE' and 'AMAZON' remain available, with 'AMAZON' as the default.

### User experience

Solves the initial error that when the LLM tries to use the default reranker. 

Tested with a us-east-1 KB and now the LLM correctly chooses the right reranker:

```
id   : call_MHxiWVFJaTEyZFRXOXJtdUk4MkY__vscode-xxxx
tool : mcp_aws_bedrock_k_QueryKnowledgeBases
args : {
  "knowledge_base_id": "xxx",
  "query": "how to reset a password",
  "reranking": true,
  "reranking_model_name": "COHERE"
}
```

## Checklist

* [x] Changes are documented

Is this a breaking change? (Y/N) No

**RFC issue number**:

Checklist:

## Acknowledgment

### DIFF
--- a/src/bedrock-kb-retrieval-mcp-server/awslabs/bedrock_kb_retrieval_mcp_server/knowledgebases/retrieval.py
+++ b/src/bedrock-kb-retrieval-mcp-server/awslabs/bedrock_kb_retrieval_mcp_server/knowledgebases/retrieval.py
@@ -16,6 +16,28 @@ from loguru import logger
 from typing import TYPE_CHECKING, Literal
 
 
+SUPPORTED_RERANKING_REGIONS = {
+    'us-west-2',
+    'us-east-1',
+    'ap-northeast-1',
+    'ca-central-1',
+    'eu-central-1',
+}
+
+
+def supported_reranking_models_for_region(region_name: str) -> tuple[str, ...]:
+    """Return the supported reranking model names for a region.
+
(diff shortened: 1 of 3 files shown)


universam1:
@theagenticguy @pranjbh Do you think we can remove this production blocker with this simple approach for now?

universam1:
Bump @tkoba-aws

universam1:
please review @anwesham-lab

PR #2001, opened 2025-12-24
feat: dsql mcp kiro power integration

anwesham-lab:
Co-authored-by: anwesham-lab <amukherjee973@gmail.com>

Fixes

### Changes

* Migrated kiro power from third-party combined kiro registry to MCP
  registry for better upkeep management patterns (MOST OF THE FILE
  DIFF IS COPIED FROM THE EXISTING POWER)
* Updated all DSQL examples to be in direct line with DSQL samples
  repository.
* Improve the POWER's MCP handling to correctly align with tool names
  for accurate execution.
* Small improvements to onboarding experience.

## Checklist

* [Y] Changes are documented

## Acknowledgment

### DIFF
--- a/src/aurora-dsql-mcp-server/kiro_power/POWER.md
+++ b/src/aurora-dsql-mcp-server/kiro_power/POWER.md
@@ -1,0 +1,294 @@
+---
+name: "aws-aurora-dsql"
+displayName: "Build a database with Aurora DSQL"
+description: "Build and deploy a PostgreSQL-compatible serverless distributed SQL database with Aurora DSQL - manage schemas, execute queries, and handle migrations with DSQL-specific requirements."
+keywords: ["aurora", "dsql", "postgresql", "serverless", "database", "sql", "aws", "distributed"]
+author: "AWS"
+---
+
+# Amazon Aurora DSQL Power
+
+## Overview
+
+The Amazon Aurora DSQL Power provides access to Aurora DSQL, a serverless, PostgreSQL-compatible distributed SQL database with specific constraints and capabilities. Execute queries, manage schemas, handle migrations, and work with multi-tenant data while respecting DSQL's unique limitations.
+
+Aurora DSQL is a true serverless database with scale-to-zero capability, zero operations overhead, and consumption-based pricing. It uses the PostgreSQL wire protocol but has specific limitations around foreign keys, array types, JSON columns, and transaction sizes.
+
+**Key capabilities:**
+- **Direct Query Execution**: Run SQL queries directly against your DSQL cluster via MCP tools
+- **Schema Management**: Create tables, indexes, and manage DDL operations
+- **Migration Support**: Execute schema migrations safely with proper transaction handling
+- **Multi-Tenant Patterns**: Built-in tenant isolation and data scoping
+- **IAM Authentication**: Automatic token generation using AWS credentials
+
+---
+
+## Available Steering Files
+
+This power includes the following steering files in [steering](./steering)
+- **development-guide**
+  - ALWAYS load before implementing schema changes or database operations
+  - MAY load when planning database application design
+  - DSQL Guidelines and Operational Rules
+- **language**
+  - MUST load when making language-specific implementation choices
+  - Driver selection, framework patterns, connection code for various languages
+- **dsql-examples**
+  - CAN Load when looking for specific implementation examples
+  - Specific examples and implementation patterns
+- **troubleshooting**
+  - SHOULD Load when debugging errors or unexpected behavior
+  - Common pitfalls and errors and how to solve
+- **onboarding**
+  - SHOULD load when user requests to try the power, "Get started with DSQL" or similar phrase
+  - Interactive "Get Started with DSQL" guide for onboarding users step-by-step
+- **mcp-setup**
+  - SHOULD load when MCP server isn't configured and MCP operation is being invoked
+  - Guides the user through updated MCP server configuration
+
+---
+
+## Available MCP Tools
+
+The `aurora-dsql` MCP server provides these tools:
+
+**Database Operations:**
+1. **readonly_query** - Execute SELECT queries (returns rows and metadata)
+2. **transact** - Execute DDL/DML statements in transaction (takes list of SQL statements)
+3. **get_schema** - Get table structure for a specific table
+
+**Documentation & Knowledge:**
+4. **dsql_search_documentation** - Search Aurora DSQL documentation
+5. **dsql_read_documentation** - Read specific documentation pages
+6. **dsql_recommend** - Get DSQL best practice recommendations
+
+---
+
+## Configuration
+
+Currently the Aurora DSQL MCP Server REQUIRES an existing DSQL cluster which
+all operations are executed atop. If the user requires complete onboarding
+guidance for creating a cluster, refer to the [onboarding guide](steering/onboarding.md).
+
+- **Package:** `awslabs.aurora-dsql-mcp-server@latest`
+
+**Setup Steps:**
+1. Create Aurora DSQL cluster in AWS Console
+2. Note your cluster identifier from the console
+3. Ensure AWS Credentials are configured from CLI: `aws configure`
+4. Configure environment variables in MCP server settings:
+   - `CLUSTER` - Your DSQL cluster identifier (e.g., "abcdefghijklmnopqrstuvwxyz")
+   - `REGION` - AWS region (e.g., "us-east-1")
+   - `AWS_PROFILE` - AWS CLI profile (optional)
+5. Ensure profile has required IAM permissions:
+   - `dsql:DbConnect` - Connect to DSQL cluster
+   - `dsql:DbConnectAdmin` - Admin access for DDL operations
+6. Test connection with `readonly_query` on `information_schema` as
+   detailed in basic operations.
+
+**Database Name:** Always use `postgres` (only database available in DSQL)
+
+---
+
+## Basic Operations
+
+### 1. Schema Exploration
+Use `readonly_query` with `information_schema` to list tables and explore database structure. Use
+`get_schema` to understand specific table structures including columns, types, and indexes.
+
+### 2. Query Data
+Use `readonly_query` for SELECT queries. Always include `tenant_id` in WHERE clause for multi-tenant
+applications. Use parameterized queries with `$1, $2` placeholders to prevent SQL injection.
+
+### 3. Execute Schema Changes
+Use `transact` tool with a list of SQL statements. Follow one-DDL-per-transaction rule. Always use
+`CREATE INDEX ASYNC` in separate transaction. Each DDL operation should be in its own `transact`
+call.
+
+### 4. Data Modifications
+Use `transact` for INSERT, UPDATE, DELETE operations. Respect transaction limits: 3,000 rows max,
+10 MiB max data size, 5 minutes max duration. Batch large operations appropriately.
+
+---
+
+## Common Workflows
+
+### Workflow 1: Create Multi-Tenant Schema
+
+**Goal:** Create a new table with proper tenant isolation and indexing
+
+**Steps:**
+1. Create main table with `tenant_id` column using `transact`
+2. Create async index on `tenant_id` in separate `transact` call
+3. Create composite indexes for common query patterns (separate `transact` calls)
+4. Verify schema with `get_schema`
+
+**Critical rules:**
+- Include `tenant_id` in all tables
+- Use `CREATE INDEX ASYNC` (never synchronous)
+- Each DDL in its own `transact` call
+- Store arrays/JSON as TEXT
+
+**Example:**
+```sql
+-- Step 1: Create table
+transact([
+  "CREATE TABLE entities (
+     entity_id VARCHAR(255) PRIMARY KEY,
+     tenant_id VARCHAR(255) NOT NULL,
+     name VARCHAR(255) NOT NULL
+   )"
+])
+
+-- Step 2: Create tenant index
+transact(["CREATE INDEX ASYNC idx_entities_tenant ON entities(tenant_id)"])
+
+-- Step 3: Verify schema
+get_schema("entities")
+```
+
+### Workflow 2: Safe Data Migration
+
+**Goal:** Add a new column with defaults safely across all rows
+
+**Steps:**
+1. Add column using `transact`: `transact(["ALTER TABLE ... ADD COLUMN ..."])`
+2. Populate existing rows with UPDATE in separate `transact` calls (batched under 3,000 rows)
+3. Verify migration with `readonly_query` using COUNT
+4. Create async index for new column using `transact` if needed
+
+**Critical rules:**
+- Add column first, populate later (never add DEFAULT in ALTER TABLE)
+- Batch updates under 3,000 rows per transaction
+- Each ALTER TABLE in its own transaction
+- Verify data before creating indexes
+
+**Example:**
+```sql
+-- Step 1: Add column
+transact(["ALTER TABLE entities ADD COLUMN status VARCHAR(50)"])
+
+-- Step 2: Populate with defaults (batched by tenant)
+transact([
+  "UPDATE entities
+   SET status = 'active'
+   WHERE status IS NULL AND tenant_id = $1"
+], parameters=["tenant-123"])
+
+-- Step 3: Verify migration
+readonly_query(
+  "SELECT COUNT(*) as total, COUNT(status) as with_status
+   FROM entities
+   WHERE tenant_id = $1",
+  parameters=["tenant-123"]
+)
+
+-- Step 4: Create index
+transact(["CREATE INDEX ASYNC idx_entities_status ON entities(tenant_id, status)"])
+```
+
+### Workflow 3: Application-Layer Referential Integrity
+
+**Goal:** Safely insert/delete records with parent-child relationships
+
+**Steps for INSERT:**
+1. Validate parent exists with `readonly_query`
+2. Throw error if parent not found
+3. Insert child record using `transact` with parent reference
+
+**Steps for DELETE:**
+1. Check for dependent records with `readonly_query` (COUNT)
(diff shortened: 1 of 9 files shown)


src/aurora-dsql-mcp-server/kiro_power/POWER.md
```diff
@@ -0,0 +1,290 @@
...
+- **Package:** `awslabs.aurora-dsql-mcp-server@latest`
+- **Authentication:** AWS IAM credentials with automatic token generation
+- **Required Environment Variables:**
+  - `CLUSTER` - Your DSQL cluster identifier
+  - `REGION` - AWS region (e.g., us-east-1)
+  - `AWS_PROFILE` - AWS CLI profile (optional, uses default if not set)
+
+---
+
+## Basic Operations
```
pkale, line 93: Can we include the cluster creation workflow as well as part of basic operation? Curious if you could provide the context for why we don't if that is intentional. In the current structure, can we include that as a prerequisite for using the power?

src/aurora-dsql-mcp-server/kiro_power/POWER.md
```diff
@@ -0,0 +1,290 @@
...
+6. Test connection with `readonly_query` on `information_schema`
+
+**Database Name:** Always use `postgres` (only database available in DSQL)
+
+---
+
+
+## Best Practices
+
+- **SHOULD read guidelines first** - Check [development_guide.md](steering/development-guide.md) before making schema changes
```
pkale, line 269: Are we planning to point the links to starter kit once that repo is public?

src/aurora-dsql-mcp-server/kiro_power/steering/development-guide.md
```diff
@@ -0,0 +1,351 @@
...
+- SHOULD have maximum **10 MiB data size** per write transaction
+- SHOULD expect **5-minute** transaction duration
+- ALWAYS expect repeatable read isolation
+
+---
+
+### Application-Layer Patterns
+
+**MANDATORY for Application Referential Integrity:**
+If foreign key constraints (application referential integrity) are required:
```
pkale: Should we explicitly include that foreign key is not supported?

src/aurora-dsql-mcp-server/kiro_power/POWER.md
anwesham-lab: yup! planning to update to switch links over!

src/aurora-dsql-mcp-server/kiro_power/POWER.md
anwesham-lab: basic operations are the basic operations that we can do with the MCP (and therefore, things that the Kiro power will naturally learn/understand to use the MCP for) 

cluster creation isn't possible, later down in the power, cluster creation is specified as a setup step/requirement

If people try the power as part of onboarding, then they'll get guided to create a cluster with the CLI; I can move the setup steps up above the basic operations though, I think that reading order makes more sense!

src/aurora-dsql-mcp-server/kiro_power/steering/development-guide.md
anwesham-lab: We wanted to not do explicit callouts via unsupported features, but instead provide the correct practices. In implementation, when I tested, too many callouts of foreign keys as unsupported were reasons it would either:
a) recommend against DSQL unnecessarily
b) actually try to implement with foreign keys rather than correctly designing with app layer integrity, in dsql examples, we provide an example method using application level referential integrity

I'll add the phrase after required, ", instead implementation:" in order to make clear that this is in the place of foreign keys.

anwesham-lab: phrase added

src/aurora-dsql-mcp-server/kiro_power/POWER.md
anwesham-lab: setup moved up and explicit statement about MCP server requiring existing cluster also done

src/aurora-dsql-mcp-server/kiro_power/POWER.md
```diff
@@ -0,0 +1,301 @@
+---
+name: "aws-aurora-dsql"
+displayName: "Build a database with Aurora DSQL"
```
amaksimo, line 3: nit: Do we want to add AWS Aurora DSQL?

src/aurora-dsql-mcp-server/kiro_power/POWER.md
```diff
@@ -0,0 +1,301 @@
...
+4. **dsql_search_documentation** - Search Aurora DSQL documentation
+5. **dsql_read_documentation** - Read specific documentation pages
+6. **dsql_recommend** - Get DSQL best practice recommendations
+
+**Configuration:**
+- **Package:** `awslabs.aurora-dsql-mcp-server@latest`
+- **Authentication:** AWS IAM credentials with automatic token generation
+- **Required Environment Variables:**
+  - `CLUSTER` - Your DSQL cluster identifier
+  - `REGION` - AWS region (e.g., us-east-1)
```
amaksimo: Aside: Is this actually required ~ I guess right now yes, but can we avoid it in the future?

src/aurora-dsql-mcp-server/kiro_power/POWER.md
```diff
@@ -0,0 +1,301 @@
...
+
+---
+
+## Configuration
+**Setup Steps:**
+1. Create Aurora DSQL cluster in AWS Console
+2. Note your cluster identifier from the console
+3. Ensure AWS Credentials are configured from CLI: `aws configure`
+4. Configure environment variables in MCP server settings:
+   - `CLUSTER` - Your DSQL cluster identifier
```
amaksimo: nit: Should we provide example here too? This will eliminate confusion between cluster_id and cluster endpoint, the first time I read it I figured endpoint and then was like "why do we need region too :P".

src/aurora-dsql-mcp-server/kiro_power/POWER.md
```diff
@@ -0,0 +1,301 @@
...
+   WHERE e.tenant_id = $1
+   GROUP BY e.name",
+  parameters=["tenant-123"]
+)
+```
+
+---
+
+
+## Best Practices
```
amaksimo, line 267: Should we move this up before common examples? Not sure if it makes a difference, but seems logical to me.

src/aurora-dsql-mcp-server/kiro_power/POWER.md
```diff
@@ -0,0 +1,301 @@
...
+- **MUST follow correct Application Layer Patterns** - when multi-tenant isolation or application referential itegrity are required; refer to [Application Layer Patterns](steering/development-guide.md#application-layer-patterns)
+- **REQUIRED use DELETE for truncation** - DELETE is the only supported operation for truncation
+- **SHOULD test any migrations** - Verify DDL on dev clusters before production
+- **Plan for Horizontal Scale** - DSQL is designed to optimize for massive scales without latency drops; refer to [Horizontal Scaling](steering/development-guide.md#horizontal-scaling-best-practice)
+- **SHOULD use connection pooling in production applications** - Refer to [Connection Pooling](steering/development-guide.md#connection-pooling-recommended)
+- **SHOULD debug with the troubleshooting guide:** - Always refer to the resources and guidelines in [troubleshooting.md](steering/troubleshooting.md)
+
+---
+
+## Additional Resources
```
amaksimo, line 287: Should we add connectors? I see you add them below, maybe we can link this in best practices near connection pooling?

src/aurora-dsql-mcp-server/kiro_power/steering/development-guide.md
```diff
@@ -0,0 +1,352 @@
...
+- SHOULD Add column first, then UPDATE with defaults
+- REQUIRED: Each DDL executes separately
+
+---
+
+## Database Connectivity Tools
+
+DSQL has many tools for connecting including 10 database drivers, 4, ORM libraries, and 3 specialized adapters
+across various languages as listed in the [programming guide](https://docs.aws.amazon.com/aurora-dsql/latest/userguide/aws-sdks.html)
+
```
amaksimo, line 247: Nit, we should add to use these over SDK to generate auth token if possible/available.

src/aurora-dsql-mcp-server/kiro_power/steering/language.md
```diff
@@ -0,0 +1,107 @@
+# DSQL Language-Specific Implementation Examples and Guides
+## Tenets
+- ALWAYS prefer DSQL Connector when available
+- Follow patterns outlined in [aurora-dsql-samples](https://github.com/aws-samples/aurora-dsql-samples/tree/main/)
+
+## Framework and Connection Notes for Languages and Drivers
+### Python
```
amaksimo, line 7: nit: Should we explicitly link to the preferred example? Like we are linking to the available ones, and that repo itself has a markdown explaining what is preferred and why, just wondering how this would function.

src/aurora-dsql-mcp-server/kiro_power/steering/language.md
```diff
@@ -0,0 +1,107 @@
+# DSQL Language-Specific Implementation Examples and Guides
```
amaksimo, line 1: In some places we link to dsql-samples, in others we link directly to the connector repo. Should we stick to dsql-samples here?

src/aurora-dsql-mcp-server/kiro_power/steering/language.md
```diff
@@ -0,0 +1,106 @@
...
+- Use `pgxpool` for connection pooling with max lifetime < 1 hour
+- Set `sslmode=verify-full` in connection string
+- See [aurora-dsql-samples/go/pgx](https://github.com/aws-samples/aurora-dsql-samples/tree/main/go/pgx)
+
+### JavaScript/TypeScript
+PREFER using [node-postgres](https://docs.aws.amazon.com/aurora-dsql/latest/userguide/SECTION_program-with-dsql-connector-for-node-postgres.html) or [postgres-js](https://docs.aws.amazon.com/aurora-dsql/latest/userguide/SECTION_program-with-dsql-connector-for-postgresjs.html) with the DSQL Node.js Connector
+
+**node-postgres (pg)** (recommended)
+- Use `@aws/aurora-dsql-node-postgres-connector` for automatic IAM auth
+- See [aurora-dsql-samples/javascript/node-postgres](https://github.com/awslabs/aurora-dsql-nodejs-connector/tree/main/packages/node-postgres)
```
amaksimo: Here the link says dsql-samples but we link out to the connector.

src/aurora-dsql-mcp-server/kiro_power/steering/language.md
```diff
@@ -0,0 +1,106 @@
...
+
+**postgres.js** (recommended)
+- Lightweight alternative with `@aws/aurora-dsql-node-postgres-connector`
+- Good for serverless environments
+- See [aurora-dsql-samples/javascript/postgres-js](https://github.com/awslabs/aurora-dsql-nodejs-connector/tree/main/packages/postgres-js)
```
amaksimo: Same comment here.

src/aurora-dsql-mcp-server/kiro_power/steering/mcp-setup.md
```diff
@@ -0,0 +1,102 @@
...
+      aws configure list-profiles
+      ```
+      For a non-default profile, configure the environment variable:
+      `"AWS_PROFILE"`.
+5. How permissive is the user? Is the MCP server permitted to write
+   to the database? If not, remove the `--allow-writes` flag from the
+   arguments in the MCP configuration.
+
+**Project-Scope**:
+1. Recommend disabling the power's global MCP server: `"disabled": true`
```
amaksimo, line 52: Nice thanks :D

src/aurora-dsql-mcp-server/kiro_power/steering/onboarding.md
```diff
@@ -0,0 +1,323 @@
...
+
+This guide provides steps to help users get started with Aurora DSQL in their project. It sets up their DSQL cluster with IAM authentication and connects their database to their code by understanding the context within the codebase.
+
+## Use Case
+
+These guidelines apply when users say "Get started with DSQL" or similar phrases. The user's codebase may be mature (with existing database connections) or have little to no code - the guidelines should apply to both cases.
+
+## Agent Communication Style
+
+**Keep all responses succinct:**
```
amaksimo, line 17: Should we add "do not make assumptions, ask user to clarify"? This seems to help me a lot, I always prompt that :)

I guess it is sort of this but more direct what I suggest:
```
- Ask direct questions when needed:
  - User ambiguity SHOULD result in questions.
```

src/aurora-dsql-mcp-server/kiro_power/POWER.md
anwesham-lab: Actually not really needed up here, I reference setting up the environment variables in the setup section right below this where it makes a lot more sense. I'm just deleting this paragraph in favor of the section below which includes it as steps with reason instead.

src/aurora-dsql-mcp-server/kiro_power/POWER.md
anwesham-lab: The code check hooks pick up on a fake cluster id as a secret being exposed, I had to remove it; I'll try again with just straight up alphabet as the sample lol

src/aurora-dsql-mcp-server/kiro_power/POWER.md
anwesham-lab: I thought no, since the best practices are also referenced in the development guide and the top level power is mainly focused on what you can do with the MCP server. 

Did some digging referring to the pattern in other MCP server based powers, and I think it makes sense to leave them at the end.

src/aurora-dsql-mcp-server/kiro_power/steering/language.md
anwesham-lab: Legacy, I can switch this back, it was while waiting for the samples to use connectors.

src/aurora-dsql-mcp-server/kiro_power/steering/language.md
anwesham-lab: same as above, will switch

src/aurora-dsql-mcp-server/kiro_power/POWER.md
anwesham-lab: languages.md says to use connectors for the languages which provide them and the best practices reference that file + samples uses connectors now, so I'll just add the samples repo and that should be sufficient

src/aurora-dsql-mcp-server/kiro_power/steering/onboarding.md
anwesham-lab: Big rule in steering is avoiding telling it what not to do if you can tell it what to do. 

I'll add a bullet: 
`SHOULD ask user clarifications to avoid inaccurate assumption` though, this makes sense to me.

src/aurora-dsql-mcp-server/kiro_power/POWER.md
anwesham-lab: I assume you mean the display name, I'm following the name PMs had asked for earlier + Postgres doesn't use AWS in the display name either, so for now, I'm going to opt for no.

src/aurora-dsql-mcp-server/kiro_power/steering/language.md
anwesham-lab: Yeah, it was connectors because samples didn't have the connectors when this was originally written, same comment as some other responses. Fixing. 

Note asyncpg, is not yet in the samples repo, so it will remain the connector for now and we can update with a small PR once we've changed that.

src/aurora-dsql-mcp-server/kiro_power/steering/onboarding.md
anwesham-lab: Updated section:
```
- Ask direct questions when needed:
  - ALWAYS ask clarifying questions to avoid inaccurate assumptions
  - User ambiguity SHOULD result in questions.
  - MUST clarify incompatible user decisions
  - Example:
    - "What column names would you like in this table?"
    - "What is the column name of the primary key?"
    - "JSON must be serialized. Would you like to stringify the JSON to serialize it as TEXT?"
```

src/aurora-dsql-mcp-server/kiro_power/steering/development-guide.md
anwesham-lab: They're already told to use connectors whenever possible and steered to use the samples. I think that naturally ends up with this being the deterministic outcome. Also some of the drivers do rely on SDK auth, so we can't quite say that. 

I didn't call it out separately to keep overall context window size small, but I'd ran over 100 tests to ensure it uses these tools and I called another model to validate it, and it always did, so I feel pretty confident this should be sufficient.

src/aurora-dsql-mcp-server/kiro_power/steering/language.md
anwesham-lab: I want it to access the samples README to ensure it takes the necessary setup steps (ie the packages in requirements.txt being added to the environment). When I was developing, I initially pointed directly to an example file and it would miss the installation steps which is what this avoids. It still picks the preferred example as that's what we set the environment up for and it's prescriptively marked as such.

src/aurora-dsql-mcp-server/kiro_power/steering/language.md
amaksimo: Ack

src/aurora-dsql-mcp-server/kiro_power/steering/onboarding.md
amaksimo: Yes thank you, this looks great :D

amaksimo (approved):
LGTM, nothing blocking rest can be follow-ups.

PR #2077, opened 2026-01-13
chore(docs): rename mcp.json property name from mcpServers to servers in VSCode configuration

Jobisworking:
Fixes

### Changes

Modified the documentation for .vscode/mcp.json to use the correct JSON property name. Changed it from "mcpServers" to "servers" as described in [VS Code mcp-servers documentation](https://code.visualstudio.com/docs/copilot/customization/mcp-servers#_configuration-format).

### User experience

Before: Using mcpServers will not display the Start, Tools, More options in IDE since VS code uses "servers" in configuration unlike the other IDEs that use "mcpServers".

After: The documentation will now match the documentation of VS code and will not causes issues where MCP does not get activated.

## Checklist

* [x] Changes are documented

Is this a breaking change? N

**RFC issue number**:

Checklist:

## Acknowledgment


Jobisworking:
updated the branch with the latest from main, this is a simple documentation change to align the configurations in the example.

PR #2213, opened 2026-01-19
fix(aws-api-mcp-server): decoding of binary data

arnewouters:
Fixes

### Changes

Handle binary data returned by AWS commands without throwing errors when they are not UTF-8 encoded

### User experience

Some commands that return binary data were failing

```sh
aws amp describe-rule-groups-namespace --workspace-id <WORKSPACE_ID> --name <NAME>
```

```txt
Object of type bytes is not JSON serializable
```

## Checklist

* [x] Changes are documented

**RFC issue number**:

Checklist:

## Acknowledgment

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -29,8 +29,8 @@ NOTICE                                                         @awslabs/mcp-admi
 /src/aurora-dsql-mcp-server                                    @awslabs/mcp-admins @awslabs/mcp-maintainers  @gxjx-x @imforster @wcmjunior @anwesham-lab @benjscho @pkale @amaksimo @mitchell-elholm
 /src/aws-api-mcp-server                                        @awslabs/mcp-admins @awslabs/mcp-maintainers  @awslabs/aws-api-mcp @rshevchuk-git @PCManticore @iddv @arnewouters @bidesh
 /src/aws-appsync-mcp-server                                    @awslabs/mcp-admins @awslabs/mcp-maintainers  @phani-srikar @maxi114 @neelmurt
-/src/aws-bedrock-custom-model-import-mcp-server                @awslabs/mcp-admins @awslabs/mcp-maintainers  @krokoko @alexa-perlov
-/src/aws-bedrock-data-automation-mcp-server                    @awslabs/mcp-admins @awslabs/mcp-maintainers  @krokoko @theagenticguy @alexa-perlov
+/src/aws-bedrock-custom-model-import-mcp-server                @awslabs/mcp-admins @awslabs/mcp-maintainers  @krokoko
+/src/aws-bedrock-data-automation-mcp-server                    @awslabs/mcp-admins @awslabs/mcp-maintainers  @krokoko @theagenticguy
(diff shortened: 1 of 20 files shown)


src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/common/helpers.py
```diff
@@ -45,13 +46,21 @@ def operation_timer(service: str, operation: str, region: str):
 class Boto3Encoder(json.JSONEncoder):
     """Custom JSON encoder for boto3 objects."""
 
+    def _decode_bytes(self, data: bytes) -> str:
+        """Decode bytes as UTF-8 string, falling back to base64 if not valid UTF-8."""
+        try:
+            return data.decode('utf-8')
+        except UnicodeDecodeError:
+            return base64.b64encode(data).decode('utf-8')
```
oleksii-k-amzn, line 54: I know that most APIs support either base64 or raw-in-base64-out, can this be something else except these two? Do we need to add additional try block if base64 decode fails?

And regarding raw-in-base64-out - I think we should handle this case as well, as currently we convert everything to string.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/common/helpers.py
```diff
@@ -45,13 +46,21 @@ def operation_timer(service: str, operation: str, region: str):
 class Boto3Encoder(json.JSONEncoder):
```
oleksii-k-amzn, line 46: Is the name still correct? It looks like this is more a decoder rather then encoder.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/common/helpers.py
arnewouters: We base64 encode and then decode, I think this works for everything? Anyways, if it still fails the tool call then we should just surface the error so no extra try catch block needed.

arnewouters: > raw-in-base64-out

botocore handles that under the hood AFAIU

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/common/helpers.py
arnewouters: It's a little bit confusing but I think it still makes sense since we are decoding the response and encoding it as JSON

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/common/helpers.py
oleksii-k-amzn: > if it still fails the tool call then we should just surface the error so no extra try catch block needed.

would it make more sense to use more explicit error message then to better indicate the root cause?

arnewouters: you're right, I added a new try catch block so the exception is more descriptive about the issue

oleksii-k-amzn (approved):

PR #2224, opened 2026-01-21
fix(dynamodb-mcp-server): Add exception logging

valeriodelbello-amazon:
Fixes

### Changes

Add `logger.exception()` to the `handle_exceptions` decorator to log full tracebacks when MCP tools fail in the DynamoDB MCP server.

### User experience

**Before:** Exceptions silently swallowed, no server-side logs.

**After:** Full tracebacks logged with function name for easier debugging.

## Checklist

* [X] Changes are documented

**RFC issue number**: N/A

Checklist:

## Acknowledgment

### DIFF
--- a/src/aurora-dsql-mcp-server/README.md
+++ b/src/aurora-dsql-mcp-server/README.md
@@ -15,6 +15,9 @@ steering while developing.
 
 ### Database Operations
 
+[IMPORTANT]
+The MCP Server requires a valid configuration for --cluster_endpoint, --database_user, and --region to enable database operations.
+
 - **readonly_query** - Execute read-only SQL queries against your DSQL cluster
 - **transact** - Execute SQL statements in a transaction
   - In read-only mode: Supports read operations with transactional consistency
--- a/src/aurora-dsql-mcp-server/awslabs/aurora_dsql_mcp_server/server.py
+++ b/src/aurora-dsql-mcp-server/awslabs/aurora_dsql_mcp_server/server.py
@@ -15,7 +15,6 @@
 """awslabs Aurora DSQL MCP Server implementation."""
 
 import argparse
-import asyncio
 import boto3
 import httpx
 import psycopg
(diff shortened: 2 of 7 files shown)


src/dynamodb-mcp-server/awslabs/dynamodb_mcp_server/common.py
```diff
@@ -89,6 +93,7 @@ async def wrapper(*args, **kwargs):
         try:
             return await func(*args, **kwargs)
         except Exception as e:
+            logger.exception('Error in %s', func.__name__)
```
LeeroyHannigan, line 96: Can we also log the error?

valeriodelbello-amazon: It does! As far as I understood the `logger.exception` should be invoked only inside `except` blocks and it will find the `e` automatically.

Test:

```python
import asyncio
import logging

logging.basicConfig(level=logging.DEBUG, format='%(levelname)s - %(name)s - %(message)s')

from awslabs.dynamodb_mcp_server.common import handle_exceptions

@handle_exceptions
async def process_item(item_id: str):
    """Process a DynamoDB item."""
    raise ValueError(f'Item {item_id} not found in table')

async def main():
    result = await process_item('user-123')
    print(f'Result: {result}')

asyncio.run(main())
```

Output:

```text
DEBUG - asyncio - Using selector: KqueueSelector
ERROR - awslabs.dynamodb_mcp_server.common - Error in process_item
  File ".../common.py", line 94, in wrapper
    return await func(*args, **kwargs)
  File "<string>", line 13, in process_item
ValueError: Item user-123 not found in table
Result: {'error': 'Item user-123 not found in table'}
```

LeeroyHannigan: Excellent, one last thing, does this mean we should remove try/except blocks from tools that implement `@handle_exceptions` ?

valeriodelbello-amazon: I agree for the cases where the tools are catching exceptions just to send the message to the LLM.

I assume there are still scenarios when they make sense, I can think of: wrapping an exception; condition checks driven by exceptions.

PR #2357, opened 2026-02-03
feat(aws-api-mcp-server): Implements batch calling for call_aws tool.

bariskurt:
Fixes

## Summary

This changes implements the batching mode for `call_aws` tool where an LLM can generate multiple AWS CLI commands in a single tool call. It also introduces `--region *` trick to run a given command in all available regions.

### Changes

1. `call_aws` accepts either a single CLI command or a list of CLI commands. 
2. `call_aws` returns a list of `CallAWSResponse` object that includes the actual command run, corresponding response and a possible error message.
3. If the command includes `--region *` , `call_aws` searches for available regions in the AWS account and runs the command on all those regions.

### User experience

1. `call_aws` will also accept a list of CLI commands and will always return a list of responses.
2. For an LLM, this is a quite simple change, since LLMs can read unstructured responses easily.

## Checklist

* [x] Changes are documented

N

**RFC issue number**:

Checklist:

## Acknowledgment

### DIFF
--- a/src/aws-healthomics-mcp-server/CHANGELOG.md
+++ b/src/aws-healthomics-mcp-server/CHANGELOG.md
@@ -9,6 +9,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- v0.0.27
+  - **Run Group Management Tools**: Added four new MCP tools for managing HealthOmics Run Groups
+    - **CreateAHORunGroup**: Create run groups with configurable resource limits (CPUs, GPUs, duration, concurrent runs)
+    - **GetAHORunGroup**: Retrieve detailed run group configuration and metadata
+    - **ListAHORunGroups**: List and filter run groups with pagination support
+    - **UpdateAHORunGroup**: Update run group resource limits and configuration
+    - Added optional `run_group_id` parameter to **StartAHORun** for associating runs with a run group
+    - Added optional `run_group_id` parameter to **ListAHORuns** for filtering runs by run group
+
 - v0.0.25
   - **Agent Identification**: Added support for an `AGENT` environment variable that appends `agent/<value>` to the User-Agent string on all boto3 API calls, enabling traceability and attribution of requests to specific AI agents via CloudTrail and AWS service logs
     - New `AGENT_ENV` constant in `consts.py`
--- a/src/aws-healthomics-mcp-server/README.md
+++ b/src/aws-healthomics-mcp-server/README.md
@@ -69,6 +69,13 @@ This MCP server provides tools for:
 
 1. **SearchGenomicsFiles** - Intelligent search for genomics files across S3 buckets, HealthOmics sequence stores, and reference stores with pattern matching, file association detection, and relevance scoring
 
+### Run Group Management Tools
+
+1. **CreateAHORunGroup** - Create a new run group with optional resource limits (maxCpus, maxGpus, maxDuration, maxRuns) and tags
+2. **GetAHORunGroup** - Retrieve detailed information about a specific run group
+3. **ListAHORunGroups** - List available run groups with optional name filtering and pagination
+4. **UpdateAHORunGroup** - Update an existing run group's name or resource limits
+
 ### Region Management Tools
 
 1. **GetAHOSupportedRegions** - List AWS regions where HealthOmics is available
@@ -458,6 +465,10 @@ The following IAM permissions are required:
                 "omics:GetRun",
                 "omics:ListRunTasks",
                 "omics:GetRunTask",
+                "omics:CreateRunGroup",
+                "omics:GetRunGroup",
+                "omics:ListRunGroups",
+                "omics:UpdateRunGroup",
                 "omics:ListSequenceStores",
                 "omics:ListReadSets",
                 "omics:GetReadSetMetadata",
--- a/src/aws-healthomics-mcp-server/awslabs/aws_healthomics_mcp_server/consts.py
+++ b/src/aws-healthomics-mcp-server/awslabs/aws_healthomics_mcp_server/consts.py
@@ -227,3 +227,8 @@ FASTQ_PAIR_PATTERNS = [
 
 # FASTQ file extensions
 FASTQ_EXTENSIONS = ['fastq', 'fq', 'fastq.gz', 'fq.gz']
+
+# Run group constants
+RUN_GROUP_MAX_NAME_LENGTH = 128
+RUN_GROUP_MAX_RESOURCE_LIMIT = 100000
+RUN_GROUP_ID_MAX_LENGTH = 18
--- a/src/aws-healthomics-mcp-server/awslabs/aws_healthomics_mcp_server/models/core.py
+++ b/src/aws-healthomics-mcp-server/awslabs/aws_healthomics_mcp_server/models/core.py
@@ -20,7 +20,7 @@ from awslabs.aws_healthomics_mcp_server.consts import (
 from datetime import datetime
 from enum import Enum
 from pydantic import BaseModel, field_validator, model_validator
-from typing import Any, List, Optional
+from typing import Any, Dict, List, Optional
 
 
 class WorkflowType(str, Enum):
@@ -255,3 +255,29 @@ class ContainerRegistryMap(BaseModel):
     def convert_none_to_empty_list(cls, v: Any) -> List[Any]:
         """Convert None values to empty lists for consistency."""
         return [] if v is None else v
+
+
+class RunGroupSummary(BaseModel):
+    """Summary information about a run group."""
+
+    id: str
+    arn: str
+    name: Optional[str] = None
+    maxCpus: Optional[int] = None
+    maxGpus: Optional[int] = None
+    maxDuration: Optional[int] = None
+    maxRuns: Optional[int] = None
+    creationTime: datetime
+
+
+class RunGroupDetail(RunGroupSummary):
+    """Detailed run group information including tags."""
+
+    tags: Optional[Dict[str, str]] = None
+
+
+class RunGroupListResponse(BaseModel):
+    """Response model for listing run groups."""
+
+    runGroups: List[RunGroupSummary]
+    nextToken: Optional[str] = None
--- a/src/aws-healthomics-mcp-server/awslabs/aws_healthomics_mcp_server/server.py
+++ b/src/aws-healthomics-mcp-server/awslabs/aws_healthomics_mcp_server/server.py
@@ -38,6 +38,12 @@ from awslabs.aws_healthomics_mcp_server.tools.helper_tools import (
     package_workflow,
 )
 from awslabs.aws_healthomics_mcp_server.tools.run_analysis import analyze_run_performance
+from awslabs.aws_healthomics_mcp_server.tools.run_group import (
+    create_run_group,
+    get_run_group,
+    list_run_groups,
+    update_run_group,
+)
 from awslabs.aws_healthomics_mcp_server.tools.run_timeline import generate_run_timeline
 from awslabs.aws_healthomics_mcp_server.tools.troubleshooting import diagnose_run_failure
 from awslabs.aws_healthomics_mcp_server.tools.workflow_analysis import (
@@ -90,6 +96,12 @@ This MCP server provides tools for creating, managing, and analyzing genomic wor
 - **GetAHORun**: Get details about a specific run
 - **ListAHORunTasks**: List tasks for a specific run
 - **GetAHORunTask**: Get details about a specific task
+
+### Run Group Management
+- **CreateAHORunGroup**: Create a new run group to limit compute resources for workflow runs
+- **GetAHORunGroup**: Get details of a specific run group including resource limits and tags
+- **ListAHORunGroups**: List available run groups with optional name filtering
+- **UpdateAHORunGroup**: Update an existing run group's name or resource limits
 
 ### Workflow Analysis
 - **GetAHORunLogs**: Retrieve high-level run logs showing workflow execution events
@@ -154,6 +166,12 @@ mcp.tool(name='ListAHORuns')(list_runs)
 mcp.tool(name='GetAHORun')(get_run)
 mcp.tool(name='ListAHORunTasks')(list_run_tasks)
 mcp.tool(name='GetAHORunTask')(get_run_task)
+
+# Register run group tools
+mcp.tool(name='CreateAHORunGroup')(create_run_group)
+mcp.tool(name='GetAHORunGroup')(get_run_group)
+mcp.tool(name='ListAHORunGroups')(list_run_groups)
+mcp.tool(name='UpdateAHORunGroup')(update_run_group)
 
 # Register workflow analysis tools
 mcp.tool(name='GetAHORunLogs')(get_run_logs)
--- a/src/aws-healthomics-mcp-server/awslabs/aws_healthomics_mcp_server/tools/run_group.py
+++ b/src/aws-healthomics-mcp-server/awslabs/aws_healthomics_mcp_server/tools/run_group.py
@@ -1,0 +1,260 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Run group management tools for the AWS HealthOmics MCP server."""
+
+import uuid
+from awslabs.aws_healthomics_mcp_server.consts import (
+    DEFAULT_MAX_RESULTS,
+)
+from awslabs.aws_healthomics_mcp_server.utils.aws_utils import (
+    get_omics_client,
+)
+from awslabs.aws_healthomics_mcp_server.utils.error_utils import (
+    handle_tool_error,
+)
+from loguru import logger
+from mcp.server.fastmcp import Context
+from pydantic import Field
+from typing import Any, Dict, Optional
+
+
+async def create_run_group(
+    ctx: Context,
+    name: Optional[str] = Field(None, description='Name for the run group (1-128 characters)'),
+    max_cpus: Optional[int] = Field(
+        None, description='Maximum CPUs for the run group (1-100000)', ge=1, le=100000
+    ),
+    max_gpus: Optional[int] = Field(
+        None, description='Maximum GPUs for the run group (1-100000)', ge=1, le=100000
+    ),
+    max_duration: Optional[int] = Field(
+        None, description='Maximum duration in minutes (1-100000)', ge=1, le=100000
+    ),
+    max_runs: Optional[int] = Field(
+        None, description='Maximum concurrent runs (1-100000)', ge=1, le=100000
+    ),
+    tags: Optional[Dict[str, str]] = Field(None, description='Tags to apply to the run group'),
+) -> Dict[str, Any]:
+    """Create a new HealthOmics run group.
+
+    Args:
+        ctx: MCP context for error reporting
+        name: Name for the run group (1-128 characters)
+        max_cpus: Maximum CPUs for the run group (1-100000)
+        max_gpus: Maximum GPUs for the run group (1-100000)
+        max_duration: Maximum duration in minutes (1-100000)
+        max_runs: Maximum concurrent runs (1-100000)
+        tags: Tags to apply to the run group
+
+    Returns:
+        Dictionary containing the created run group's id, arn, and tags, or error dict
+    """
+    try:
+        client = get_omics_client()
+
+        params: Dict[str, Any] = {
+            'requestId': str(uuid.uuid4()),
+        }
+
+        if name is not None:
+            params['name'] = name
+
+        if max_cpus is not None:
+            params['maxCpus'] = max_cpus
+
+        if max_gpus is not None:
+            params['maxGpus'] = max_gpus
+
+        if max_duration is not None:
+            params['maxDuration'] = max_duration
+
+        if max_runs is not None:
+            params['maxRuns'] = max_runs
+
+        if tags is not None:
+            params['tags'] = tags
+
+        logger.info(f'Creating run group with params: {params}')
+        response = client.create_run_group(**params)
+
+        return {
+            'id': response.get('id'),
+            'arn': response.get('arn'),
+            'tags': response.get('tags'),
+        }
+    except Exception as e:
+        return await handle_tool_error(ctx, e, 'Error creating run group')
+
+
+async def get_run_group(
+    ctx: Context,
+    run_group_id: str = Field(..., description='ID of the run group to retrieve'),
+) -> Dict[str, Any]:
+    """Get details of a specific HealthOmics run group.
+
+    Args:
+        ctx: MCP context for error reporting
+        run_group_id: ID of the run group to retrieve
+
+    Returns:
+        Dictionary containing the run group details, or error dict
+    """
+    try:
+        client = get_omics_client()
+
+        logger.info(f'Getting run group: {run_group_id}')
+        response = client.get_run_group(id=run_group_id)
+
+        creation_time = response.get('creationTime')
+        if creation_time is not None:
+            creation_time = creation_time.isoformat()
+
+        return {
+            'arn': response.get('arn'),
+            'id': response.get('id'),
+            'name': response.get('name'),
+            'maxCpus': response.get('maxCpus'),
+            'maxGpus': response.get('maxGpus'),
+            'maxDuration': response.get('maxDuration'),
+            'maxRuns': response.get('maxRuns'),
+            'tags': response.get('tags'),
+            'creationTime': creation_time,
+        }
+    except Exception as e:
+        return await handle_tool_error(ctx, e, 'Error getting run group')
+
+
+async def list_run_groups(
+    ctx: Context,
+    name: Optional[str] = Field(None, description='Filter by run group name'),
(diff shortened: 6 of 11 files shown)


src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
```diff
@@ -218,21 +240,60 @@ async def suggest_aws_commands(
...
+        except CliParsingError as e:
+            results.append(CallAWSResponse(cli_command=cmd, error=str(e)))
+        else:
+            for expanded_cmd in expanded_commands:
+                results.append(await _execute_single_command(expanded_cmd, ctx, max_results))
+    return results
+
+
+def _expand_regions_if_needed(cli_command: str) -> list[str]:
+    if "--region" not in cli_command:
```
rshevchuk-git: Can we extract this and _execute_single_command to the service.py? They clutter main server entry point

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
```diff
@@ -218,21 +240,60 @@ async def suggest_aws_commands(
...
     """Call AWS with the given CLI command and return the result as a dictionary."""
-    return await call_aws_helper(
-        cli_command=cli_command,
-        ctx=ctx,
-        max_results=max_results,
-        credentials=None,
-    )
+
+    commands = [cli_command] if isinstance(cli_command, str) else cli_command
+    if not isinstance(commands, list):
```
rshevchuk-git: After the ternary, `commands` is always either a `list` (from wrapping `str`) or whatever `cli_command` was (already `list[str]` per the type hint). The second check is dead code given the type annotation `str | list[str]`.

src/aws-api-mcp-server/tests/test_server.py
```diff
@@ -1000,18 +1028,79 @@ async def test_call_aws_helper_without_credentials(mock_translate, mock_validate
...
 
     mock_call_aws_helper.assert_called_once_with(
-        cli_command='aws s3api list-buckets', ctx=ctx, max_results=None, credentials=None
+        'aws s3api list-buckets', ctx, None, None
     )
-    assert result == mock_response
+    assert result == [CallAWSResponse(cli_command='aws s3api list-buckets', response=mock_response)]
+
+
+# --- FIXME ---
```
rshevchuk-git: This is the most important test for the core batch feature and it's not running.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
```diff
@@ -187,12 +189,32 @@ async def suggest_aws_commands(
     - All commands are validated before execution to prevent errors
     - Supports pagination control via max_results parameter
     - {_FILE_ACCESS_MSGS[FILE_ACCESS_MODE]}
+    - You can use `--region *` to run a command on all regions enabled in the account.
+    - Do not generate explicit batch calls for iterating over all regions, use `--region *` instead.
+
+    Single Command Mode:
+    - You can run a single AWS CLI command usign this tool.
```
rshevchuk-git: Typo `usign`

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
```diff
@@ -218,21 +240,60 @@ async def suggest_aws_commands(
 )
 async def call_aws(
     cli_command: Annotated[
-        str, Field(description='The complete AWS CLI command to execute. MUST start with "aws"')
+        str | list[str], Field(description='A single command or a list of complete AWS CLI commands to execute')
```
rshevchuk-git: Seems like there's no validation on the length of the command list. A caller could pass thousands of commands, causing resource exhaustion.

src/aws-api-mcp-server/tests/test_server.py
```diff
@@ -1181,3 +1271,65 @@ def test_get_server_auth_oauth_valid():
...
+        "aws s3 ls --region us-west-2"
+    ]
+    assert result == expected
+
+
+def test_expand_regions_if_needed_invalid_region():
+    """Test _expand_regions_if_needed with invalid --region parameter."""
+    from awslabs.aws_api_mcp_server.core.common.errors import CliParsingError
+    with pytest.raises(CliParsingError, match="--region parameter requires a value"):
+        _expand_regions_if_needed("aws s3 ls --region")
```
rshevchuk-git: This seems a bit too shallow, we need more test cases and more different potential variations of CLI to test with, including complex commands with filters, inline JSONs, etc. Done via a parametrized test

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/aws/regions.py
```diff
@@ -26,3 +29,21 @@
...
     'sagemaker-geospatial': 'us-west-2',
 }
+
+
+
+def get_active_regions() -> list[str]:
+    account_client = boto3.client('account')
+
+    active_regions = []
+    paginator = account_client.get_paginator('list_regions')
```
rshevchuk-git: What if account.ListRegions fails? or the user does not have permission to call it?

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
```diff
@@ -218,21 +240,60 @@ async def suggest_aws_commands(
...
-    )
+
+    commands = [cli_command] if isinstance(cli_command, str) else cli_command
+    if not isinstance(commands, list):
+        raise AwsApiMcpError("Tool input error. call_aws command accepts a string for a single command and list for multiple commands.")
+
+    results = []
+    for cmd in commands:
+        try:        
+            expanded_commands = _expand_regions_if_needed(cmd)
```
rshevchuk-git: This is a bit worrisome. `--region *` expands to unbounded calls with no limit — `get_active_regions()` could return 20+ regions. Combined with a batch of N commands each using `--region *`, this fans out to `N * 20+` sequential API calls with no limit or cap. If ran in a remote context, I wonder if this can cause DDoS attacks

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/aws/regions.py
bariskurt: I've never seen that.. In such a case we should return a meaningful error message such that LLM can recover in the next iteration. By recovering I mean it should call the related command again this time for a list of regions explicitly, not using `--region *`

bariskurt: I will update the CR accordingly.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
bariskurt: We can add a length check. What do you think the appropriate num of commands would be ?  My guess is around 20.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
bariskurt: `str | list[str]` is a type hint but only a hint, not a runtime type check. So it's not a dead code.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
bariskurt: Let's consider this: `--region *` may expand to multiple regions, that's true.  But if LLMs intention is to make an operation over all regions, they will be expanded one way or another. Alternatively `Sonnet45` can make multiple tool calls (one for each region) and make a DDoS effect as well.

Also, we will limit the number of CLI commands submitted (to let's say 20). 20x20 = 400  would be an absolute limit, but it's an extreme case. In reality, in the non-batching version, we recorded the p90 of number of call as 18 in 5 LLM cycles, (with Sonnet45). Same model does 5 calls due to batching. This is a big difference between practical numbers and theoretical limit. (A model can also get crazy and make even more parallel calls itself without batching)

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
bariskurt: I did not understand the ask here.

src/aws-api-mcp-server/tests/test_server.py
bariskurt: My mistake here. Will fix this case.

src/aws-api-mcp-server/tests/test_server.py
bariskurt: Will add more tests.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
rshevchuk-git: 15-20 makes sense to me

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
rshevchuk-git: What I mean is that this exception is never thrown in the real world, right? Can you provide an example of the input which would throw an exception here during runtime?

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
rshevchuk-git: This is not about the model generating this "extreme case", it's about a malicious actor manually invoking an endpoint passing input that will exhaust the system. Think of invoking call_aws in the remote AWS MCP server

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
rshevchuk-git: Move the 2 new functions to the service.py file, instead of adding them to server.py

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/aws/regions.py
```diff
@@ -26,3 +29,21 @@
...
+
+    for page in paginator.paginate():
+        page_regions = page.get('Regions', [])
+        active_regions.extend([
```
PCManticore: Small nit: `extend()` accepts a generator so you can drop the temporary list that we pass into `extend()`.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
```diff
@@ -187,12 +189,32 @@ async def suggest_aws_commands(
     - All commands are validated before execution to prevent errors
     - Supports pagination control via max_results parameter
     - {_FILE_ACCESS_MSGS[FILE_ACCESS_MODE]}
+    - You can use `--region *` to run a command on all regions enabled in the account.
+    - Do not generate explicit batch calls for iterating over all regions, use `--region *` instead.
+
+    Single Command Mode:
```
PCManticore, line 196: We keep hearing that context size of our tools is not well optimized - this PR expands `call_aws` tool description even further - can we validate that we can't get away with a shorter description?

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
```diff
@@ -187,12 +189,32 @@ async def suggest_aws_commands(
...
+                "aws s3api get-bucket-website --bucket bucket2"
+            ]
+        )
 
     Best practices for command generation:
     - Always use the most specific service and operation names
     - Always use the working directory when writing files, unless user explicitly mentioned another directory
     - Include --region when operating across regions
     - Only use filters (--filters, --query, --prefix, --pattern, etc) when necessary or user explicitly asked for it
+    - Always use the tool in batch mode whenever it's possible.
```
PCManticore, line 219: Why do we have to specify this?

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
```diff
@@ -218,21 +240,60 @@ async def suggest_aws_commands(
...
+        raise AwsApiMcpError("Tool input error. call_aws command accepts a string for a single command and list for multiple commands.")
+
+    results = []
+    for cmd in commands:
+        try:        
+            expanded_commands = _expand_regions_if_needed(cmd)
+        except CliParsingError as e:
+            results.append(CallAWSResponse(cli_command=cmd, error=str(e)))
+        else:
+            for expanded_cmd in expanded_commands:
```
PCManticore, line 269: We can also run this more efficiently using `asyncio.gather`, but specifically for multi-region queries:

```
results.extend(await asyncio.gather(*[_execute_single_command(cmd, ctx, max_results) for cmd in expanded_commands]))
```

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
bariskurt: I moved `_expand_regions_if_needed` to service.py, but it doesn't make sense to move _execute_single_command unless you also move `call_aws_helper`

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
bariskurt: Removed the line.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
bariskurt: Fixed.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
bariskurt: Added with a tool definition change to inform LLM about the limit.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/aws/regions.py
bariskurt: Done.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
bariskurt: I do believe this is the minimum required additions to the prompt. We describe the batch mode and give a single example.

Regarding the size concern, we can revisit this problem immediately after by trying to shorten the prompt and testing it comprehensively.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
bariskurt: Yes, but we don't want concurrency for the time being. As discussed offline, it can cause unexpected issues. However we can think about this as a subsequent optimization.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
arnewouters: Can we update the type here instead to enforce the size limit on the list?

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
```diff
@@ -218,21 +242,41 @@ async def suggest_aws_commands(
...
-        cli_command=cli_command,
-        ctx=ctx,
-        max_results=max_results,
-        credentials=None,
-    )
+
+    commands = [cli_command] if isinstance(cli_command, str) else cli_command
+
+    if len(commands) > MAX_BATCH_COMMANDS:
+        raise AwsApiMcpError(f'Number of batch commands exceeds the maximum limit of {MAX_BATCH_COMMANDS}.')
```
bidesh: If this is visible to the caller, this seems informative.  Should we include what is the next step the customer/agent can do as well (be instructive).

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
```diff
@@ -218,21 +242,41 @@ async def suggest_aws_commands(
...
+
+    results = []
+    for cmd in commands:
+        try:        
+            expanded_commands = expand_regions_if_needed(cmd)
```
bidesh, line 265: Feels like the region expansion happening here means the MAX_BATCH_COMMANDS check should come after.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/aws/regions.py
rshevchuk-git: Was this addressed? Can't find the handling of AccessDenied

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/common/config.py
```diff
@@ -193,7 +193,11 @@ def get_server_auth():
 )
 CONNECT_TIMEOUT_SECONDS = 10
 READ_TIMEOUT_SECONDS = 60
+<<<<<<< call_aws_batch_prod
+MAX_BATCH_COMMANDS = 20
+=======
 AWS_MAX_ATTEMPTS = int(os.getenv('AWS_MAX_ATTEMPTS', 3))
+>>>>>>> main
```
rshevchuk-git: Seems like merge artifacts

src/aws-api-mcp-server/tests/test_security_policy.py
```diff
@@ -530,10 +530,16 @@ async def test_call_aws_security_policy_deny(
...
+<<<<<<< call_aws_batch_prod
+    response_list = await call_aws.fn('aws s3 rm s3://bucket/file', ctx)
+    assert len(response_list) == 1
+    assert response_list[0].error == 'Execution of this operation is denied by security policy.'
+=======
     with pytest.raises(AwsApiMcpError) as exc_info:
         await call_aws('aws s3 rm s3://bucket/file', ctx)
 
     assert 'Execution of this operation is denied by security policy.' in str(exc_info.value)
+>>>>>>> main
```
rshevchuk-git: Seems like merge artifacts

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/common/models.py
```diff
@@ -209,3 +209,23 @@ def as_dict(self) -> dict[str, Any]:
 
 def _normalize_program(str) -> list[str]:
     return [line.strip() for line in str.splitlines() if line.strip()]
+
+class CallAWSResponse(BaseModel):
+    cli_command: str
```
bidesh, line 217: We had a scenario where a customer reached out due to added fields (with null value). Are there any risks on breaking customer's workflow due to this change?

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/common/config.py
```diff
@@ -193,7 +193,11 @@ def get_server_auth():
 )
 CONNECT_TIMEOUT_SECONDS = 10
 READ_TIMEOUT_SECONDS = 60
+<<<<<<< call_aws_batch_prod
```
bidesh: Missing merge conflict fix? This should be blocking.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/common/config.py
bariskurt: Fixed.

src/aws-api-mcp-server/tests/test_security_policy.py
bariskurt: Fixed.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
bariskurt: That's deliberate. We want a maximum of 20 CLIs that can be also expanded.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/common/config.py
bariskurt: Fixed.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/common/models.py
bariskurt: The rationale here is that LLMs can adopt to tool I/O changes instantly. Things can break if the customer algorithmically process tool output in a custom way. There may be very exceptional cases here, but quite unlikely.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
bariskurt: We update the tool description and tell LLM not to generate more than MAX_BATCH_COMMANDS commands in a single call. And here we are raising an error. I expect LLM to figure it out from here.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/aws/regions.py
```diff
@@ -26,3 +29,20 @@
     'route53domains': 'us-east-1',
     'sagemaker-geospatial': 'us-west-2',
 }
+
+
+def get_active_regions() -> list[str]:
+    account_client = boto3.client('account')
```
arnewouters: This does not respect `AWS_API_MCP_PROFILE_NAME`

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/aws/service.py
```diff
@@ -279,3 +280,27 @@ def _to_context(context: dict[str, Any] | None) -> ContextAPIModel | None:
...
+        region_idx = parts.index("--region")
+        if region_idx + 1 == len(parts):
+            raise CliParsingError("--region parameter requires a value")
+        region_value = parts[region_idx + 1]
+        if region_value.startswith("--"):
+            raise CliParsingError("--region parameter requires a value")
+        if region_value == "*":
+            base_command = " ".join(parts[:region_idx] + parts[region_idx + 2:])
+            try:
+                regions = get_active_regions()
```
arnewouters: We need to respect `--profile` here as well and fetch regions from that account

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/aws/regions.py
bariskurt: I added  profile_name parameter to the `get_active_regions` function.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/aws/service.py
bariskurt: Should be fixed by now.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/aws/service.py
```diff
@@ -279,3 +280,27 @@ def _to_context(context: dict[str, Any] | None) -> ContextAPIModel | None:
...
+        region_idx = parts.index("--region")
+        if region_idx + 1 == len(parts):
+            raise CliParsingError("--region parameter requires a value")
+        region_value = parts[region_idx + 1]
+        if region_value.startswith("--"):
+            raise CliParsingError("--region parameter requires a value")
+        if region_value == "*":
+            base_command = " ".join(parts[:region_idx] + parts[region_idx + 2:])
+            try:
+                regions = get_active_regions(AWS_API_MCP_PROFILE_NAME)
```
arnewouters: We need to fetch the profile from the command as well, I'm also a bit concerned that we are manually parsing the command here again. Can we use a smarter approach that uses the `IRTranslation`?

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/aws/regions.py
bariskurt: Currently, if list_regions fail we cannot expand the regions and return an error message back to LLM.
Is there any other behaviour you want to see here ?

src/aws-api-mcp-server/tests/aws/test_service.py
```diff
@@ -804,3 +805,97 @@ def test_execute_awscli_customization_raises_error(mock_get_driver):
...
+    'command',
+    [
+        'aws s3 ls --region us-east-1',
+        'aws s3api list-buckets --region ap-south-1 --output json',
+    ],
+)
+def test_expand_regions_if_needed_specific_region(command):
+    """Test expand_regions_if_needed with a specific region."""
+    result = expand_regions_if_needed(command)
+    assert result == [command]
```
arnewouters, line 837: Can we combine this into one test since the body is exactly the same?

I would also create separate tests for weird cases like `us-east-1*`, `a*b`, `--profileabc` to validate the regex a bit more.

src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/aws/service.py
bariskurt: This is done as we discussed offline.

src/aws-api-mcp-server/tests/aws/test_service.py
```diff
@@ -804,3 +805,112 @@ def test_execute_awscli_customization_raises_error(mock_get_driver):
...
+    [
+        'aws s3 ls',
+        'aws account list-regions',
+        'aws s3 ls --region us-east-1',
+        'aws s3api list-buckets --region ap-south-1 --output json',
+    ],
+)
+def test_expand_regions_if_needed_without_expansion(command):
+    """Test expand_regions_if_needed with no --region parameter."""
+    result = expand_regions_if_needed(command)
```
arnewouters, line 821: We should assert on get_active_regions being called with `None` somewhere here I think

src/aws-api-mcp-server/tests/aws/test_service.py
bariskurt: Test cases are added.

src/aws-api-mcp-server/tests/aws/test_service.py
bariskurt: Done.

rshevchuk-git (approved):
LGTM. Thanks for addressing the comments.

PR #2366, opened 2026-02-04
fix: patch traversal path vulnerability

HaoOliv:
Fixes

## Summary
- The original validate_file_path function used path.resolve(strict=True) but didn't verify the resolved path was within an allowed base directory

### Changes
1. Base Directory Enforcement: Added validation to ensure all resolved file paths are within the allowed base directory

- Dynamic Base Directory: Leveraged existing _get_base_directory() function that:
Defaults to current working directory in production (secure by default)
Uses / during testing (allows temp files for tests)
Supports DOCUMENT_BASE_DIR environment variable for custom configurations

### User experience

No impact

## Checklist

* [x] Changes are documented

N
**RFC issue number**:

Checklist:

## Acknowledgment

### DIFF
--- a/src/document-loader-mcp-server/tests/test_server.py
+++ b/src/document-loader-mcp-server/tests/test_server.py
@@ -385,6 +385,34 @@ async def test_validate_file_path_general_exception():
 
 
 @pytest.mark.asyncio
+async def test_path_traversal_blocked():
+    """Test that path traversal attempts are blocked by base directory enforcement."""
+    import os
+    import tempfile
+    from pathlib import Path
+
+    # Create a temp file
+    with tempfile.NamedTemporaryFile(suffix='.pdf', delete=False) as temp_file:
+        temp_file.write(b'%PDF-1.4 test')  # Minimal PDF header
+        temp_file_path = temp_file.name
+
+    try:
+        # Mock base directory to a restricted location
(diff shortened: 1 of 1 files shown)


andywidjaja:
@HaoOliv Thanks for addressing the concern by adding base directory enforcement.

Could you please add a unit test to your PR?  Something like the following:

```
@pytest.mark.asyncio
async def test_path_traversal_blocked():
    """Test that path traversal attempts are blocked by base directory enforcement."""
    import os
    import tempfile
    from unittest.mock import patch
    from pathlib import Path
    
    # Create a temp file
    with tempfile.NamedTemporaryFile(suffix='.pdf', delete=False) as temp_file:
        temp_file.write(b'%PDF-1.4 test')  # Minimal PDF header
        temp_file_path = temp_file.name
    
    try:
        # Mock base directory to a restricted location
        with patch('awslabs.document_loader_mcp_server.server._get_base_directory') as mock_base:
            mock_base.return_value = Path('/restricted/directory')
            
            ctx = MockContext()
            error = validate_file_path(ctx, temp_file_path)
            
            assert error is not None
            assert 'Access denied: path outside allowed directory' in error
            print('✓ Path traversal blocked by base directory enforcement')
    finally:
        if os.path.exists(temp_file_path):
            os.unlink(temp_file_path)

```

PR #2526, opened 2026-02-27
feat(postgres): add --secret_arn fallback for clusters without MasterUserSecret

leoalord:
## Problem

Clusters using CDK-managed secrets or `addRotationSingleUser()` don't populate `MasterUserSecret`. The postgres MCP server discovers credentials exclusively from `MasterUserSecret.SecretArn`, so `secret_arn` resolves to `None` and every Data API call fails.

## Solution

Add an optional `--secret_arn` CLI argument and `connect_to_database` tool parameter that provides a fallback when `MasterUserSecret` is not present on the cluster. Zero behavioral change when the flag is omitted.

Also fixes a pre-existing bug where `main()` references `args.hostname` instead of `args.db_endpoint`.

## Testing

Tested against Aurora PostgreSQL Serverless v2 with CDK-managed secret (no `manageMasterUserPassword`). Server starts, auto-connects at startup, and queries succeed via the RDS Data API using the provided secret ARN.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/postgres-mcp-server/awslabs/postgres_mcp_server/server.py
+++ b/src/postgres-mcp-server/awslabs/postgres_mcp_server/server.py
@@ -277,6 +277,7 @@ def connect_to_database(
     db_endpoint: Annotated[str, Field(description='database endpoint')],
     port: Annotated[int, Field(description='Postgres port')],
     database: Annotated[str, Field(description='database name')],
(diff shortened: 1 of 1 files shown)


kennthhz:
I agree we should add a MCP tool parameter to caller (LLM) to specify secret manager ARN instead of always assume it is from the MasterUsrSecret property. However, this should not be done as -- parameter as args. we support supporting different cred for each cluster or instance via MCP tool API. we don't want to add any program args anymore because it assume a single DB or cluster.

PR #2538, opened 2026-03-01
fix(mysql-mcp-server): allow REPLACE() string function in readonly mode

### LINKED ISSUE #2178: (mysql-mcp): (REPLACE() string function blocked as mutating operation in SELECT queries)
### Describe the bug

# Bug Report: REPLACE() Function Incorrectly Rejected in Readonly Mode

## Summary
The MySQL MCP server incorrectly rejects SELECT queries containing the `REPLACE()` string function when readonly mode is enabled. The `REPLACE()` function is a safe, read-only MySQL string function that should be allowed in SELECT statements, but it's being blocked because `REPLACE` is included in the mutating keywords detection list.

## Problem
When executing SELECT queries that use the `REPLACE()` string function (e.g., `SELECT REPLACE(column, ' ', '_') FROM table`), the MCP server rejects them with an error indicating mutating operations are not allowed, even though these are perfectly safe read-only queries.

## Root Cause
The issue is in the mutating keyword detection logic. The file `mutable_sql_detector.py` includes `REPLACE` in the `MUTATING_KEYWORDS` set:

**Problematic Code:**
```python
# Line 23 in mutable_sql_detector.py
MUTATING_KEYWORDS = {
    'INSERT',
    'UPDATE',
    'DELETE',
    'REPLACE',  # ← This catches both REPLACE statement AND REPLACE() function
    'TRUNCATE',
    # ... other keywords
}
```

The regex pattern `MUTATING_PATTERN` matches `REPLACE` as a word boundary, which incorrectly flags:
- ✅ `REPLACE INTO table ...` (correctly rejected - this is a mutating statement) | OR just if the REPLACE has "(" right after, it supposed to be fine also. ("REPLACE(...." ✅, REPLACE without "(" - ❌) 
- ❌ `SELECT REPLACE(column, 'old', 'new') FROM table` (incorrectly rejected - this is a safe function)

## Expected Behavior
SELECT queries using the `REPLACE()` string function should be allowed in readonly mode, as they are read-only operations that do not modify data.

## Actual Behavior
SELECT queries containing `REPLACE()` are rejected with an error message indicating mutating operations are not allowed.

### ❌ Currently Rejected (but should be allowed):
```sql

-- Example: Simple string replacement in SELECT
SELECT REPLACE(product_name, 'Old', 'New') as normalized_name
FROM products
WHERE category = 'electronics';

-- Example 3: Multiple REPLACE functions
SELECT 
    REPLACE(REPLACE(email, '@old-domain.com', '@new-domain.com'), '.', '_') as normalized_email
FROM users;
```

### ✅ Correctly Rejected (mutating statement):
```sql
-- This should continue to be rejected
REPLACE INTO users (id, name) VALUES (1, 'John');
```

## Suggested Fix
The detection logic should distinguish between:
1. **`REPLACE` statement** (mutating) - `REPLACE INTO table ...`
2. **`REPLACE()` function** (read-only) - `SELECT REPLACE(column, ...) FROM table`

**Proposed Solution:**
- Update the regex pattern to only match `REPLACE` when it's followed by `INTO` (indicating a REPLACE statement)
- Or exclude `REPLACE` from `MUTATING_KEYWORDS` and add a specific pattern for `REPLACE INTO` statements

## Impact
This bug prevents users from:
- Normalizing string values in WHERE clauses
- Transforming data in SELECT statements
- Using common MySQL string manipulation functions in readonly queries

## Environment
- **MCP Server Version:** Latest (commit 4a07e72fa7fd514661d867b6b6a0b551e7348ceb)
- **File:** `src/mysql-mcp-server/awslabs/mysql_mcp_server/mutable_sql_detector.py`
- **Line:** 23 (MUTATING_KEYWORDS definition)

## References
- [MySQL REPLACE() Function Documentation](https://dev.mysql.com/doc/refman/8.0/en/string-functions.html#function_replace)
- [MySQL REPLACE Statement Documentation](https://dev.mysql.com/doc/refman/8.0/en/replace.html)
- Problematic code: https://github.com/awslabs/mcp/blob/4a07e72fa7fd514661d867b6b6a0b551e7348ceb/src/mysql-mcp-server/awslabs/mysql_mcp_server/mutable_sql_detector.py#L23

## Expected Behavior
When readonly mode is enabled, SELECT queries containing the `REPLACE()` string function should execute successfully. The `REPLACE()` function is a MySQL string manipulation function that:
- Does not modify database data
- Only transforms values in the query result set
- Is safe to use in readonly mode

Examples that should work:
- `SELECT REPLACE(column, 'old', 'new') FROM table`
- `WHERE REPLACE(bucket_name, ' ', '_') = 'normalized_value'`
- Any SELECT query using `REPLACE()` as a string function

Only `REPLACE INTO` statements (which modify data) should be rejected in readonly mode.

### Current Behavior

Image
## Actual Behavior
SELECT queries containing `REPLACE()` are rejected with an error message indicating mutating operations are not allowed.

### Reproduction Steps

Reproduction Steps - do SELECT query with REPLACE() inside it. 
SELECT REPLACE(product_name, 'Old', 'New') as normalized_name
FROM products
WHERE category = 'electronics';

## Suggested Fix
Distinguish between `REPLACE` statement (mutating) and `REPLACE()` function (read-only):

1. Remove `REPLACE` from `MUTATING_KEYWORDS` set
2. Add specific regex pattern to detect `REPLACE INTO` statements only:
   ```python
   REPLACE_STATEMENT_REGEX = re.compile(r'(?i)\bREPLACE\s+INTO\b')
   ```
3. Check for `REPLACE INTO` separately in `detect_mutating_keywords()` function

This allows `REPLACE()` function in SELECT queries while still blocking `REPLACE INTO` statements.

Even - REPLACE and the ( should be fine. 

_No response_

### OS

linux

### Server

mysql-mcp-server

### Server Version

_No response_

us-east-1

### Other information

_No response_

### Service quota

Br1an67:
Fixes #2178

## Summary

Allow the `REPLACE()` string function in SELECT queries when readonly mode is enabled.

### Changes

The `REPLACE` keyword in `MUTATING_KEYWORDS` incorrectly matched both the mutating `REPLACE INTO ...` DML statement and the safe `REPLACE()` string function. Added a negative lookahead `(?!\s*\()` to the regex pattern so that `REPLACE` followed by a parenthesis (function call) is not flagged, while `REPLACE INTO` (DML) is still correctly blocked.

Added test suite for `mutable_sql_detector` module covering:
- `REPLACE INTO` statement detection (blocked)
- `REPLACE()` function in SELECT (allowed)
- Nested `REPLACE()` calls (allowed)
- `REPLACE  (` with spaces (allowed)
- Other mutating keywords (INSERT, UPDATE, DELETE, DDL)
- SQL injection risk detection

### User experience

**Before:** `SELECT REPLACE(product_name, 'Old', 'New') FROM products` was rejected with "mutating operation not allowed" in readonly mode.

**After:** The query executes successfully. Only `REPLACE INTO ...` statements are blocked.

## Checklist

* [x] Changes are documented

Is this a breaking change? N

## Acknowledgment

### DIFF
--- a/src/mysql-mcp-server/awslabs/mysql_mcp_server/mutable_sql_detector.py
+++ b/src/mysql-mcp-server/awslabs/mysql_mcp_server/mutable_sql_detector.py
@@ -34,8 +34,14 @@ MUTATING_KEYWORDS = {
     'UNINSTALL PLUGIN',
 }
 
+# Build the mutating keyword pattern.
+# REPLACE needs special handling: `REPLACE INTO ...` is mutating, but
+# `REPLACE(col, 'a', 'b')` is a safe string function.  A negative
+# lookahead for `(` distinguishes the two.
+_REPLACE_PATTERN = r'REPLACE(?!\s*\()'
+_OTHER_KEYWORDS = sorted(k for k in MUTATING_KEYWORDS if k != 'REPLACE')
 MUTATING_PATTERN = re.compile(
-    r'(?i)\b(' + '|'.join(re.escape(k) for k in MUTATING_KEYWORDS) + r')\b'
+    r'(?i)\b(' + '|'.join([_REPLACE_PATTERN] + [re.escape(k) for k in _OTHER_KEYWORDS]) + r')\b'
 )
 
 # -- Regex for DDL statements --
--- a/src/mysql-mcp-server/tests/test_mutable_sql_detector.py
+++ b/src/mysql-mcp-server/tests/test_mutable_sql_detector.py
@@ -1,0 +1,99 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Tests for mutable_sql_detector module."""
+
+from awslabs.mysql_mcp_server.mutable_sql_detector import (
+    check_sql_injection_risk,
+    detect_mutating_keywords,
+)
+
+
+class TestDetectMutatingKeywords:
+    """Tests for detect_mutating_keywords function."""
+
+    def test_replace_statement_blocked(self):
+        """REPLACE INTO is a mutating DML statement and should be detected."""
+        result = detect_mutating_keywords('REPLACE INTO users (id, name) VALUES (1, "John")')
+        assert 'REPLACE' in result
+
+    def test_replace_function_allowed(self):
+        """REPLACE() string function in SELECT should not be detected as mutating."""
+        result = detect_mutating_keywords(
+            "SELECT REPLACE(product_name, 'Old', 'New') FROM products"
+        )
+        assert 'REPLACE' not in result
+        assert result == []
+
+    def test_replace_function_nested(self):
+        """Nested REPLACE() function calls should not be detected as mutating."""
+        result = detect_mutating_keywords(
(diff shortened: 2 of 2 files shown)


Br1an67:
Hi — just a gentle bump on this. Happy to make any changes if needed!

shahar-shabtay-ni:
Looks good thanks! when will you merge it? 
(I opened the bug)

PR #2550, opened 2026-03-03
feat(eks): add kubeconfig/OIDC authentication support

karancode:
## Summary
- Add an alternative `kubeconfig` authentication mode for the EKS MCP server, enabling users without AWS IAM. credentials(e.g., OIDC-authenticated, namespace-scoped Kubernetes users) to use Kubernetes tools.
- Introduce `EKS_AUTH_MODE` env var (`iam` | `kubeconfig`) and `--auth-mode` CLI arg to select the authentication path
- Default behavior (`iam` mode) is fully backward compatible - no changes to existing users.
 
#### Motivation
Customers using OIDC authentication to access Kubernetes/EKS cannot use the EKS MCP server because they don't have an IAM user account and are not assuming a role within AWS. Their access is limited to a Kubernetes namespace, with permissions granted through Kubernetes RBAC roles/groups. These users need an alternative authentication path that works with their existing kubeconfig credentials.

#### Core auth changes

- **`k8s_apis.py`**: Added `K8sApis.from_api_client()` classmethod that creates an instance from a pre-configured `kubernetes.client.ApiClient`, bypassing the manual endpoint/token/CA construction.
- **`k8s_client_cache.py`**: Added `EKS_AUTH_MODE` env var support with `_get_kubeconfig_client()` method that uses `kubernetes.config.new_client_from_config(context=cluster_name)` which handles all kubeconfig auth methods (OIDC exec plugins, certificates, bearer tokens) transparently.

#### Configuration
- **`server.py`**: Added `--auth-mode` CLI arg, updated `SERVER_INSTRUCTIONS` with authentication modes section, added startup logging for auth mode.
- **`k8s_handler.py`**: Updated `cluster_name` Field descriptions across 6 tools to document dual interpretation (EKS cluster name vs kubeconfig context name).

#### Tests (17 new, 270 total passing)
- `TestK8sApisFromApiClient`: 4 tests for the new classmethod (instance creation, user agent, cleanup, operations)
- `TestK8sClientCacheAuthMode`: 7 tests for mode selection (default, explicit, invalid fallback, TTL, case insensitivity)
- `TestK8sClientCacheKubeconfigMode`: 5 tests for kubeconfig path (client creation, KUBECONFIG env, caching, error handling, IAM isolation)
- Fixed existing `test_server.py` mocks to include `auth_mode` attribute

#### Documentation
- **`README.md`**: Added `--auth-mode` CLI arg docs, `EKS_AUTH_MODE`/`KUBECONFIG` env var docs, "Kubeconfig/OIDC Authentication" section with example MCP config, updated prerequisites and access requirements

### User experience

**Before:** Users without AWS IAM credentials (e.g., OIDC-authenticated, namespace-scoped Kubernetes users) cannot use the EKS MCP server at all. Every Kubernetes API call requires AWS credentials for STS presigned URL generation.

**After:** Users set `EKS_AUTH_MODE=kubeconfig` (env var or `--auth-mode kubeconfig` CLI arg) and the server authenticates using their existing kubeconfig file, supporting OIDC exec plugins, certificates, and bearer tokens. The `cluster_name` parameter is interpreted as a kubeconfig context name. No AWS credentials needed for Kubernetes operations.

## Checklist

* [x] Changes are documented

**RFC issue number**: https://github.com/awslabs/mcp/issues/2551 

Checklist:

## Acknowledgment

### DIFF
--- a/src/eks-mcp-server/awslabs/eks_mcp_server/k8s_client_cache.py
+++ b/src/eks-mcp-server/awslabs/eks_mcp_server/k8s_client_cache.py
@@ -170,19 +170,21 @@ class K8sClientCache:
         from kubernetes.config import list_kube_config_contexts
 
         kubeconfig_path = os.environ.get('KUBECONFIG', None)
-        contexts, active_context = list_kube_config_contexts(config_file=kubeconfig_path)
+        raw_contexts, _ = list_kube_config_contexts(config_file=kubeconfig_path)
+        contexts: list[dict] = [dict(ctx) for ctx in raw_contexts]
 
         # First, try exact context name match
         for ctx in contexts:
-            if ctx['name'] == cluster_name:
+            if ctx.get('name') == cluster_name:
                 return cluster_name
 
         # Search for a context whose cluster field contains the cluster name
-        matches = []
+        matches: list[str] = []
         for ctx in contexts:
-            ctx_cluster = ctx.get('context', {}).get('cluster', '')
+            ctx_detail = ctx.get('context', {})
+            ctx_cluster = ctx_detail.get('cluster', '') if isinstance(ctx_detail, dict) else ''
             if ctx_cluster.endswith(f'/{cluster_name}') or ctx_cluster == cluster_name:
-                matches.append(ctx['name'])
+                matches.append(str(ctx.get('name', '')))
 
         if len(matches) == 1:
             logger.info(f'Resolved cluster name "{cluster_name}" to context "{matches[0]}"')
@@ -193,7 +195,7 @@ class K8sClientCache:
                 f'Please specify the full context name.'
             )
         else:
-            available = [ctx['name'] for ctx in contexts]
+            available = [str(ctx.get('name', '')) for ctx in contexts]
             raise ValueError(
                 f'No kubeconfig context found for cluster "{cluster_name}". '
                 f'Available contexts: {available}'
--- a/src/eks-mcp-server/tests/test_server.py
+++ b/src/eks-mcp-server/tests/test_server.py
@@ -39,6 +39,7 @@ async def test_server_initialization():
 
     # Test kubeconfig mode appends the addendum
     server_kc = create_server(auth_mode='kubeconfig')
+    assert server_kc.instructions is not None
     assert 'Kubeconfig Authentication Mode (Active)' in server_kc.instructions
     assert 'Only Kubernetes tools are available' in server_kc.instructions
     # Test that the server has the correct dependencies


karancode:
thanks @srhsrhsrhsrh, fixed.

src/eks-mcp-server/awslabs/eks_mcp_server/k8s_handler.py
```diff
@@ -97,7 +97,7 @@ async def apply_yaml(
         ),
         cluster_name: str = Field(
             ...,
-            description='Name of the EKS cluster where the resources will be created or updated.',
+            description='Name of the EKS cluster (or kubeconfig context name when EKS_AUTH_MODE=kubeconfig) where the resources will be created or updated.',
```
patrick-yu-amzn: Does the client (AI agent) get to see this EKS_AUTH_MODE flag and make the right decision?

patrick-yu-amzn: Given this is a breaking change of behavior that impacts existing customers, we have to keep the tool parameter identical (i.e. use cluster names even with kubeconfig).

karancode: thanks @patrick-yu-amzn. regarding the first one, I am thinking to make the server instructions dyanmic based on auth_mode. Basically clarifying that only k8s tools are available and removing the "common workflows" as those references to unavailable tools. This way the agent won't attempt calling tools which don't (as already skipped the AWS handlers registration).

karancode: for the cluster_name, will add context resolution in the `_get_kubeconfig_client` method. will load kubeconfig and check for the context with cluster field matching cluster_name. 
for cases with multiple or no matches, will return an error.

src/eks-mcp-server/awslabs/eks_mcp_server/k8s_client_cache.py
```diff
@@ -151,14 +151,65 @@ def _get_cluster_credentials(self, cluster_name: str):
...
+        for ctx in contexts:
+            if ctx.get('name') == cluster_name:
+                return cluster_name
+
+        # Search for a context whose cluster field contains the cluster name
+        matches: list[str] = []
+        for ctx in contexts:
+            ctx_detail = ctx.get('context', {})
+            ctx_cluster = ctx_detail.get('cluster', '') if isinstance(ctx_detail, dict) else ''
+            if ctx_cluster.endswith(f'/{cluster_name}') or ctx_cluster == cluster_name:
```
srhsrhsrhsrh, line 186: Should this always be an exact match to avoid the case where a context including the target cluster_name but with a different prefix gets matched first?

karancode: The `endswith('/{cluster_name}')` uses the `/` delimiter which prevents partial suffix matches (e.g., dev-my-cluster won't match my-cluster). Additionally, if multiple contexts match, we raise an error asking the user to specify the full context name. The exact context name match is tried first as a fast path.

Issue #2354, opened 2026-02-03
Bug: `Optional[int]` type hint incorrectly generates `string` type in JSON Schema


SeanWei-lab:
## Description

When using `Optional[int]` type hints in MCP tool functions, the `awslabs.mcp_lambda_handler` package incorrectly generates JSON Schema with `type: "string"` instead of `type: "integer"`.

## Environment

- Package: `awslabs.mcp_lambda_handler`
- Python version: 3.x
- Type hint: `Optional[int]` (which is `Union[int, None]`)

### Expected Behavior

When a function parameter is annotated with `Optional[int]`, the generated JSON Schema should be:
```json
{
  "type": "integer",
  "description": "..."
}
```

Or with nullable support:

```json
{
  "type": ["integer", "null"],
  "description": "..."
}
```

### Current Behavior

The generated JSON Schema incorrectly uses `string` type:

```json
{
  "type": "string",
  "description": "..."
}
```

### Reproduction Steps

```python
from typing import Optional
from awslabs.mcp_lambda_handler import MCPLambdaHandler

mcp = MCPLambdaHandler(
    name="test-mcp",
    version="1.0.0",
)

@mcp.tool()
def test_int_param(days: Optional[int] = 30, status: Optional[str] = "new") -> str:
    """Test integer parameter
    
    Args:
        days: Number of days, default 30
        status: Status, default "new"
    
    Returns:
        Test result
    """
    return f"days={days} (type: {type(days).__name__}), status={status}"

# Check generated schema
import json
print(json.dumps(mcp.tools, indent=2))
```

**Output:**

```json
{
  "test_int_param": {
    "name": "test_int_param",
    "description": "Test integer parameter",
    "inputSchema": {
      "type": "object",
      "properties": {
        "days": {
          "type": "string",  // ❌ Should be "integer"
          "description": "Number of days, default 30"
        },
        "status": {
          "type": "string",  // ✅ Correct
          "description": "Status, default \"new\""
        }
      },
      "required": ["days", "status"]
    }
  }
}
```
**Function annotations (correct):**

```python
{
  'days': typing.Optional[int],
  'status': typing.Optional[str],
  'return': <class 'str'>
}
```

### Possible Solution

Add handling for `Union` types before the default case:

```python
from typing import Union, get_origin, get_args

def get_type_schema(type_hint: Any) -> Dict[str, Any]:
    # Handle basic types
    if type_hint is int:
        return {'type': 'integer'}
    elif type_hint is float:
[...]

PR #2561, opened 2026-03-04
fix(mcp-lambda-handler): resolve Optional[T] to correct JSON Schema type instead of defaulting to string

ayushtr-aws:
Fixes

### Changes

The `get_type_schema` function had no handler for `Union` types, so `Optional[int]` (which is Union[int, None]) fell through to the default case and was incorrectly mapped to `"string"`. The fix adds a Union handler that strips NoneType and recursively resolves the actual inner type, so `Optional[int]` now correctly produces `"integer"`.

### User experience

When a user defines a tool with `Optional[int]` parameters like:

`@mcp.tool()
def test_int_param(days: Optional[int] = 30, status: Optional[str] = "new") -> str:
`
Before this change, the generated JSON Schema showed `days` as a string:
`
"days": {
  "type": "string",
  "description": "Number of days, default 30"
}
`
After this change, it correctly reflects the integer type:
`
"days": {
  "type": "integer",
  "description": "Number of days, default 30"
}
`
This matters because MCP clients validate parameter types against the schema. With the old behavior, a client sending `days: 30` as an integer would see a type mismatch against the `"string"` schema, causing confusion or requiring workarounds.

## Checklist

* [x] Changes are documented

Is this a breaking change? N

**RFC issue number**: #2354

Checklist:

## Acknowledgment

### DIFF
--- a/src/mcp-lambda-handler/awslabs/mcp_lambda_handler/mcp_lambda_handler.py
+++ b/src/mcp-lambda-handler/awslabs/mcp_lambda_handler/mcp_lambda_handler.py
@@ -216,6 +216,18 @@ class MCPLambdaHandler:
 
                 # Get origin type (e.g., Dict from Dict[str, int])
                 origin = get_origin(type_hint)
+
+                # Handle Union types (including Optional[T] which is Union[T, None])
+                if origin is Union:
+                    args = get_args(type_hint)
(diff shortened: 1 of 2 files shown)


Lukas-Xue (approved):
LGTM

isadeks (commented):
LGTM

ClintEastman02 (approved):
LGTM

PR #2586, opened 2026-03-05
feat(document-loader-mcp-server): add configurable MAX_FILE_SIZE_MB and slide image extraction via soffice

awsmadi:
### Changes

**1. Configurable `MAX_FILE_SIZE_MB` env var**
- The 50MB file size limit was hardcoded, making it impossible to adjust for environments with larger documents or tighter constraints.
- Added `_get_max_file_size()` which reads `MAX_FILE_SIZE_MB` from the environment on each request, validates it (must be a positive integer in megabytes), and falls back to the 50MB default with a warning on invalid values.
- Removed the stale module-level `MAX_FILE_SIZE` constant — the limit is now evaluated dynamically so env var changes take effect without restarting the server.
- Documented the previously undocumented `DOCUMENT_BASE_DIR` env var in the README.

**2. `extract_slides_as_images` tool**
- New MCP tool that extracts individual slides/pages as PNG images from PPTX, PPT, or PDF files.
- Conversion pipeline: PPTX/PPT → soffice (headless) → PDF → pdf2image → PNG per page. PDF input skips the soffice step.
- Added `_find_soffice()` which checks `$PATH` first, then known macOS app bundle locations (`/Applications/LibreOffice.app`, `/Applications/OpenOffice.app`, and their `~/Applications` equivalents).
- Configurable DPI (72–600), output directory, and timeout (default 120s).
- Returns `SlidesExtractionResponse` with slide image paths, count, and output directory.
- Added `pdf2image>=1.16.0` dependency and regenerated `uv.lock`.
- Documented the tool and its system prerequisites (LibreOffice, poppler-utils) with per-platform installation instructions in the README.

### User experience

**Before:**
- File size limit was locked at 50MB with no way to change it. Users with large documents (financial reports, research papers) received `File too large: X bytes (max: 52428800 bytes)` with no recourse.
- No way to extract presentation slides as images — only text extraction was available via `read_document`, which loses layout, charts, and visual content.

**After:**
- Users can set `MAX_FILE_SIZE_MB=100` (or any positive integer) in their MCP client config to raise or lower the limit. Default behavior is unchanged (50MB).
- Invalid values (non-numeric, negative, zero) log a warning and fall back to the default rather than crashing.
- New `extract_slides_as_images` tool lets LLMs visually analyze presentation slides and PDF pages. Example output for a 4-slide PPTX:
  ```
  Status: success
  Slides: 4
  slide_1.png (61KB), slide_2.png (121KB), slide_3.png (113KB), slide_4.png (124KB)
  ```
- When soffice is not installed, the tool returns a clear error with installation instructions per platform.
- Works with both LibreOffice and OpenOffice on Linux, macOS (including app bundles not on `$PATH`), and Windows.

## Checklist

* [x] Changes are documented

Is this a breaking change? N

**RFC issue number**:

Checklist:

## Acknowledgment


src/document-loader-mcp-server/awslabs/document_loader_mcp_server/server.py
```diff
@@ -144,10 +184,11 @@ def validate_file_path(ctx: Context, file_path: str) -> Optional[str]:
...
             return f'Path is not a file: {file_path}'
 
-        # Check file size
+        # Check file size — read dynamically so env var changes take effect
+        max_file_size = _get_max_file_size()
         file_size = path.stat().st_size
-        if file_size > MAX_FILE_SIZE:
-            return f'File too large: {file_size} bytes (max: {MAX_FILE_SIZE} bytes)'
+        if file_size > max_file_size:
+            return f'File too large: {file_size} bytes (max: {max_file_size} bytes)'
```
andywidjaja: This should be updated to:
- Convert bytes to human-readable MB
- Include the hint about MAX_FILE_SIZE_MB

Something like:

"File too large: 189.4MB (max: 50MB). To increase the limit, set the MAX_FILE_SIZE_MB environment variable (e.g., MAX_FILE_SIZE_MB=200)."

andywidjaja:
@awsmadi 

Once my earlier comment about providing a more intuitive error message is addressed, please do the following:

In test_document_parsing.py around line 490, the test asserts 'File too large' in response.error_message.  This would still pass, but could also verify the hint text is present.

src/document-loader-mcp-server/awslabs/document_loader_mcp_server/server.py
```diff
@@ -16,15 +16,18 @@
 import asyncio
 import os
 import pdfplumber
+import shutil
+import subprocess
```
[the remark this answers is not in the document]
andywidjaja: @awsmadi Recommend to add inline comment to supress the Bandit warning with justification since shell=False (the default)

import subprocess  # nosec B404 - subprocess used with fixed command, no shell=True

src/document-loader-mcp-server/awslabs/document_loader_mcp_server/server.py
awsmadi: Addressed

awsmadi:
> @awsmadi
> 
> Once my earlier comment about providing a more intuitive error message is addressed, please do the following:
> 
> In test_document_parsing.py around line 490, the test asserts 'File too large' in response.error_message. This would still pass, but could also verify the hint text is present.

Addressed

awsmadi:
All review feedback has been addressed:
- Human-readable file size error messages with `MAX_FILE_SIZE_MB` hint ✅
- `# nosec B404` inline comment for subprocess import ✅

All CI checks are passing. Could I get a re-review when you have a chance? Thanks!

PR #2623, opened 2026-03-09
fix: add Windows compatibility for signal handlers

awsmasudur:
# Pull Request Details

### Title:
```
Fix Windows compatibility - add platform check for signal handlers
```

### Description:
```markdown
## Problem
The aws-msk-mcp-server crashes on Windows with `NotImplementedError` when attempting to use signal handlers. The error occurs because `anyio.open_signal_receiver()` with Unix signals (`SIGINT`, `SIGTERM`) is not supported on Windows.

### Error Stack Trace:
```
exceptiongroup.ExceptionGroup: unhandled errors in a TaskGroup (1 sub-exception)
  | Traceback (most recent call last):
  |   File "...\awslabs\aws_msk_mcp_server\server.py", line 55, in signal_handler
  |     with open_signal_receiver(signal.SIGINT, signal.SIGTERM) as signals:
  |   File "...\anyio\_backends\_asyncio.py", line 2093, in __enter__
  |     self._loop.add_signal_handler(sig, self._deliver, sig)
  |   File "...\asyncio\events.py", line 553, in add_signal_handler
  |     raise NotImplementedError
  | NotImplementedError
```

## Root Cause
In `server.py`, the `signal_handler()` function unconditionally uses `open_signal_receiver()` which calls `asyncio.add_signal_handler()`. This function is Unix-only and raises `NotImplementedError` on Windows.

## Solution
Added a platform check to skip signal handler setup on Windows:
- Import `platform` module
- Check `platform.system() == 'Windows'` at the start of `signal_handler()`
- Return early with a log message on Windows
- Maintain full functionality on Unix/Linux/macOS

## Why This Approach?
1. **Minimal change** - Only adds 4 lines of code
2. **Safe** - No behavior change on Unix systems
3. **Consistent** - Similar to how other cross-platform Python applications handle this
4. **Graceful** - MCP stdio-based servers don't strictly need signal handlers as the parent process manages lifecycle

## Testing
- ✅ Tested on Windows 11 with Python 3.10
- ✅ Server starts successfully without crashes
- ✅ Successfully connects to MCP clients (Kiro)
- ✅ AWS API calls work correctly (verified with `list_topics`, `get_global_info`)
- ✅ Graceful shutdown still works via stdio closure

## Related
- Similar issue: https://github.com/crossbario/autobahn-python/issues/471
- AnyIO documentation notes Windows doesn't support signals: https://anyio.readthedocs.io/en/2.x/signals.html
```

- Add platform check to skip signal handler setup on Windows
- Signal handlers using anyio.open_signal_receiver() are not supported on Windows
- Server now works on Windows, Linux, and macOS
- Tested on Windows 11 with successful AWS MSK API calls

---

## Contributor Statement


src/aws-msk-mcp-server/awslabs/aws_msk_mcp_server/server.py
```diff
@@ -46,12 +47,20 @@
...
 
     The anyio.open_signal_receiver returns an async generator that yields
     signal numbers whenever a specified signal is received. The async for
     loop waits for signals and processes them as they arrive.
+    
+    Note: Signal handlers are not supported on Windows, so this function
+    returns immediately on Windows platforms.
     """
+    # Signal handlers are not supported on Windows
+    if platform.system() == 'Windows':
```
mehbey: Thank you for fixing this, 
I tried to see how other MCP servers are handling this case and I noticed that the problem is not handed there as well. E.g `src/aws-iot-sitewise-mcp-server/awslabs/aws_iot_sitewise_mcp_server/server.py`

Just an idea to consider:

looking at the documentation here: https://docs.python.org/3/library/platform.html seems like platform.system() can return user-facing OS name and they recommend using https://docs.python.org/3/library/os.html#os.uname but I don't think it matters for this case. Just adding it here to make sure your aware of it.

Another alternative would have been just to catch the `NotImplementedError` that way you will not depend on the specific name returned for the above call.

src/aws-msk-mcp-server/awslabs/aws_msk_mcp_server/server.py
```diff
@@ -46,12 +47,20 @@
...
+        logger.info('Signal handlers not supported on Windows - skipping signal handler setup')
+        return
```
mehbey, line 71: Does this have a test coverage in [test_server.py](https://github.com/awsmasudur/mcp/blob/fix/windows-signal-handler-compatibility/src/aws-msk-mcp-server/tests/test_server.py)?

mehbey (changes requested):

src/aws-msk-mcp-server/awslabs/aws_msk_mcp_server/server.py
awsmasudur: Thank you for the feedback.

I agree that catching `NotImplementedError` is a cleaner and more robust approach. It's more Pythonic and doesn't depend on platform-specific string comparisons. It will also work on any platform that doesn't support signal handlers, not just Windows.

I've updated the code to use try/except instead of the platform check.

src/aws-msk-mcp-server/awslabs/aws_msk_mcp_server/server.py
awsmasudur: I've added test coverage in `test_server.py`. 
The new test `test_signal_handler_not_implemented` verifies that the signal handler gracefully handles `NotImplementedError` (which occurs on platforms like Windows that don't support signal handlers). It confirms that:
- The exception is caught properly
- An appropriate log message is generated
- The function returns without crashing

The test is now included in the PR.

mehbey (dismissed):
LGTM, Thank you!

src/aws-msk-mcp-server/awslabs/aws_msk_mcp_server/server.py
```diff
@@ -51,14 +51,21 @@ async def signal_handler(scope: CancelScope):
...
-            # 'Enter' to unblock the stdio.
-            os._exit(0)
+    try:
+        with open_signal_receiver(signal.SIGINT, signal.SIGTERM) as signals:
+            async for _ in signals:  # Shutting down regardless of the signal type
+                print('Shutting down MCP server...')
+                # Force immediate exit since MCP blocks on stdio.
+                # You can also use scope.cancel(), but it means after Ctrl+C, you need to press another
+                # 'Enter' to unblock the stdio.
+                os._exit(0)
```
isaurab007, line 68: Is there a reason why we have a `print` statement here over a logger? Also do you need a flush to actually print before  os._exit(0)?

awsmasudur: @isaurab007 , thanks for checking this.

I've updated the code to:
1. Use `logger.info()` instead of `print()` for consistency with the rest of the codebase
2. Add `sys.stdout.flush()` and `sys.stderr.flush()` before `os._exit(0)` to ensure the shutdown message is actually written before the immediate exit

I also updated the corresponding test case to verify the new behavior. All 327 tests are passing locally.

Changes pushed in commits:
- 21ebe29: Use logger instead of print and flush output before os._exit(0)
- 08b5df2: Add sys import and update test to match logger.info() usage

Thanks for the thorough review!

isaurab007 (approved):
I am approving based on previous comments and responses. Ideally, I'd love to see some evidence of graceful shutdown of the application on Windows.

mehbey (approved):

awsmasudur:
Hi, any update on this?

PR #2627, opened 2026-03-10
feat(neptune): Add endpoint_url support to NeptuneAnalytics for local/custom endpoints

thompsonbry:
# feat(neptune): Add endpoint_url support to NeptuneAnalytics for local/custom endpoints

## Summary

Add optional `endpoint_url` parameter to `NeptuneAnalytics`, matching the existing pattern in
`NeptuneDatabase`.

## Motivation

The `NeptuneDatabase` class already supports custom endpoints via `host`/`port`/`use_https`
parameters, which construct an `endpoint_url` for the boto3 client. The `NeptuneAnalytics`
class has no equivalent — it only accepts a `graph_identifier` and relies on boto3's default
service endpoint resolution.

### `graph_store/analytics.py`

```python
class NeptuneAnalytics(NeptuneGraph):

    def __init__(
        self,
        graph_identifier: str,
        credentials_profile_name: Optional[str] = None,
        endpoint_url: Optional[str] = None,
    ) -> None:
        """Create a new Neptune Analytics graph wrapper instance."""
        self.graph_identifier = graph_identifier
        try:
            if not credentials_profile_name:
                session = boto3.Session()
            else:
                session = boto3.Session(profile_name=credentials_profile_name)

            client_params = {}
            if endpoint_url:
                client_params['endpoint_url'] = endpoint_url

            self.client = session.client(
                'neptune-graph', config=USER_AGENT_CONFIG, **client_params
            )
        except Exception as e:
            # ... existing error handling ...
```

### `neptune.py`

```python
elif endpoint.startswith('neptune-graph://'):
    graphId = endpoint.replace('neptune-graph://', '')
    self.graph = NeptuneAnalytics(
        graphId, endpoint_url=kwargs.get('endpoint_url')
    )
```

### `server.py`

Add env var support:

```python
NEPTUNE_ANALYTICS_ENDPOINT_URL = os.environ.get('NEPTUNE_ANALYTICS_ENDPOINT_URL')
```

Pass through to NeptuneServer when set.

### Environment variables

| Variable | Description | Example |
|---|---|---|
| `NEPTUNE_ANALYTICS_ENDPOINT_URL` | Custom endpoint URL for Neptune Analytics | `http://localhost:9100` |

This is consistent with the existing `NEPTUNE_USE_HTTPS` env var pattern.

## Security considerations

- `endpoint_url` is a standard boto3 parameter — it changes where requests are sent but does
  not bypass SigV4 authentication. Requests to the custom endpoint are still signed.
- For containers that don't validate SigV4, the user must separately configure boto3
  credentials (e.g., dummy credentials). This is the same pattern used with other local AWS
  service emulators (LocalStack, DynamoDB Local).
- The `NeptuneDatabase` class already supports this pattern via its `host`/`port`/`use_https`
  parameters, so this change is consistent with existing security posture.
- No credentials are exposed or logged.

## Testing

- Existing tests continue to pass (no behavioral change when `endpoint_url` is not set)
- Manual testing against Neptune Analytics with a non-default endpoint on `http://localhost:9100`

---

### DIFF
--- a/src/amazon-neptune-mcp-server/awslabs/amazon_neptune_mcp_server/graph_store/analytics.py
+++ b/src/amazon-neptune-mcp-server/awslabs/amazon_neptune_mcp_server/graph_store/analytics.py
@@ -75,9 +75,7 @@ class NeptuneAnalytics(NeptuneGraph):
                 client_params['endpoint_url'] = endpoint_url
                 config = Config(inject_host_prefix=False).merge(USER_AGENT_CONFIG)
 
-            self.client = session.client(
-                'neptune-graph', config=config, **client_params
-            )
+            self.client = session.client('neptune-graph', config=config, **client_params)
 
         except Exception as e:
             logger.exception(
--- a/src/amazon-neptune-mcp-server/awslabs/amazon_neptune_mcp_server/neptune.py
+++ b/src/amazon-neptune-mcp-server/awslabs/amazon_neptune_mcp_server/neptune.py
@@ -79,9 +79,7 @@ class NeptuneServer:
                     graphId = 'g-1234567890'
                     endpoint_url = f'http://{value}'
 
-                self.graph = NeptuneAnalytics(
-                    graphId, endpoint_url=endpoint_url
-                )
+                self.graph = NeptuneAnalytics(graphId, endpoint_url=endpoint_url)
                 logger.debug('Creating Neptune Graph session for %s', endpoint)
             else:
                 raise ValueError(


src/amazon-neptune-mcp-server/awslabs/amazon_neptune_mcp_server/server.py
```diff
@@ -65,7 +65,9 @@ def get_graph():
             't',
         )
 
-        _graph = NeptuneServer(endpoint, use_https=use_https)
+        endpoint_url = os.environ.get('NEPTUNE_ANALYTICS_ENDPOINT_URL', None)
+
+        _graph = NeptuneServer(endpoint, use_https=use_https, endpoint_url=endpoint_url)
```
cornerwings: I think endpoint/endpoint_url are bit confusing, perhaps endpoint_override? Even better would be to extend the approach of how endpoint is being specified for neptune. So for example, instead of `neptune-graph://<graphId>`, we can also support `neptune-graph://<endpoint>`. GraphId is guaranteed to conform to this regex (`^g-[a-z0-9]{10}$`), we can try to match with this and if not treat it as endpoint. This avoids the need for additional environment variable.

phanindhra876: Using two endpoint fields is confusing. We can extract the graphId from endpoint field used already. For an endpoint of form `neptune-graph://g-123457890.us-east-1.neptune-graph.amazonaws.com`, we can split it into graphId `g-1234567890` and endpoint_url(boto3 parameter name) `https://us-east-1.neptune-graph.amazonaws.com`.

src/amazon-neptune-mcp-server/awslabs/amazon_neptune_mcp_server/graph_store/analytics.py
```diff
@@ -56,7 +62,13 @@ def __init__(
...
                 session = boto3.Session(profile_name=credentials_profile_name)
 
-            self.client = session.client('neptune-graph', config=USER_AGENT_CONFIG)
+            client_params = {}
+            if endpoint_url:
+                client_params['endpoint_url'] = endpoint_url
+
+            self.client = session.client(
+                'neptune-graph', config=USER_AGENT_CONFIG, **client_params
+            )
```
phanindhra876: When using locally, boto3 client constructs the endpoint to be `https://{graphId}.<endpoint>` which may not work out of the box for localhost. Conditionally set `inject_host_prefix` to false when initializing the client to avoid this when connecting to local NeptuneAnalytics server.

src/amazon-neptune-mcp-server/awslabs/amazon_neptune_mcp_server/server.py
thompsonbry: refactored per suggestions

src/amazon-neptune-mcp-server/awslabs/amazon_neptune_mcp_server/graph_store/analytics.py
thompsonbry: refactored per suggestions

test-plan-pr2627.md
cornerwings, line 1: Lets remove this file

src/amazon-neptune-mcp-server/awslabs/amazon_neptune_mcp_server/neptune.py
```diff
@@ -66,8 +68,34 @@ def __init__(self, endpoint: str, use_https: bool = True, port: int = 8182, *arg
                 logger.debug('Creating Neptune Database session for %s', endpoint)
             elif endpoint.startswith('neptune-graph://'):
                 # This is a Neptune Analytics Graph
-                graphId = endpoint.replace('neptune-graph://', '')
-                self.graph = NeptuneAnalytics(graphId)
+                # Support two forms:
+                #   neptune-graph://g-1234567890  (graph ID only, uses default service endpoint)
+                #   neptune-graph://g-1234567890.us-east-1.neptune-graph.amazonaws.com
+                #       (graph ID + custom endpoint, extracted by splitting on first dot)
+                #   neptune-graph://localhost:9100/g-1234567890
```
cornerwings: graphId is ignored in local setups, I think this part can be just simplified to have a hardcoded placeholder

src/amazon-neptune-mcp-server/awslabs/amazon_neptune_mcp_server/neptune.py
```diff
@@ -66,8 +68,34 @@ def __init__(self, endpoint: str, use_https: bool = True, port: int = 8182, *arg
...
+
+                # Check if value is a bare graph ID
+                if GRAPH_ID_PATTERN.match(value):
+                    graphId = value
+                elif '/' in value:
+                    # Form: host:port/graphId (local container)
+                    parts = value.rsplit('/', 1)
+                    endpoint_url = endpoint_url or f'http://{parts[0]}'
+                    graphId = parts[1]
+                elif '.' in value:
```
cornerwings: This check can be tightened to say if endpoint starts with graphId (honestly maybe this can be omitted as well unless we have some use case to override graph endpoints)

src/amazon-neptune-mcp-server/awslabs/amazon_neptune_mcp_server/neptune.py
```diff
@@ -66,8 +68,34 @@ def __init__(self, endpoint: str, use_https: bool = True, port: int = 8182, *arg
...
+                #       (graph ID + custom endpoint, extracted by splitting on first dot)
+                #   neptune-graph://localhost:9100/g-1234567890
+                #       (local endpoint with graph ID as path)
+                value = endpoint.replace('neptune-graph://', '')
+                endpoint_url = kwargs.get('endpoint_url')
+
+                # Check if value is a bare graph ID
+                if GRAPH_ID_PATTERN.match(value):
+                    graphId = value
+                elif '/' in value:
```
cornerwings: This can be moved to else block, with whole thing becoming endpoint_url and have a hardcoded graphId as placeholder (g-1234567890)

test-plan-pr2627.md
thompsonbry: done

src/amazon-neptune-mcp-server/awslabs/amazon_neptune_mcp_server/neptune.py
thompsonbry: Simplified the URI parsing as suggested. Now just two cases:
- neptune-graph://g-1234567890 — bare graph ID, default service endpoint
- neptune-graph://<anything-else> — treated as custom endpoint URL, uses placeholder graph ID

This covers both the managed NA case (bare graphId) and local containers (localhost:9100) without needing to parse intermediate forms. The full-endpoint form (g-xxx.region.neptune-graph.amazonaws.com) is dropped since there's no current use case for overriding the service endpoint
while also specifying a real graph ID.

Hope I am tracking your suggestion.

src/amazon-neptune-mcp-server/awslabs/amazon_neptune_mcp_server/graph_store/analytics.py
```diff
@@ -56,7 +69,20 @@ def __init__(
...
+
+            # When using a custom endpoint (e.g., local container), disable
+            # host prefix injection so boto3 doesn't prepend the graph
+            # identifier to the hostname.
+            if endpoint_url:
+                self.client.meta.events.unregister('before-sign.neptune-graph.*',
```
cornerwings: Sorry for piecemeal comments (missed this somehow in last rev), this can be set directly in config (at L73). Something like below.

```
config = USER_AGENT_CONFIG
if endpoint_url:
   ...
   host_prefix_config = Config(inject_host_prefix=False)
   config = host_prefix_config.merge(config)
```

thompsonbry: wilco

PR #2628, opened 2026-03-10
fix: healthomics mcp create workflow registry mapping model to be in s…

nbulsara:
fix: healthomics mcp create workflow registry mapping model to be in sync with API spec

### Changes

The `container_registry_map` parameter for `CreateAHOWorkflow` and `CreateAHOWorkflowVersion` required `upstreamRepositoryPrefix` and `ecrAccountId` in each `registryMapping` entry, even though the HealthOmics API treats these as optional fields. This change aligns the MCP server's validation with the API spec:

- `validate_container_registry_params` now returns a cleaned/normalized `Dict` instead of `None`, omitting `upstreamRepositoryPrefix` and `ecrAccountId` from registry mappings when they are not provided, and including them only when explicitly set.
- Both `create_workflow` and `create_workflow_version` use the validated map returned from the validator rather than passing the raw input directly.
- The `imageMappings` key is always normalized to an empty list when not provided by the caller.
- Tests updated to assert against the normalized map structure.

### User experience

Before: Providing a `container_registry_map` without `upstreamRepositoryPrefix` or `ecrAccountId` in a registry mapping would fail validation, even though the HealthOmics API accepts those fields as optional.

After: Users can provide registry mappings with only the required `upstreamRegistryUrl` and `ecrRepositoryPrefix`. The optional `upstreamRepositoryPrefix` and `ecrAccountId` are included in the API call only when explicitly provided.

## Checklist

* [x] Changes are documented

Is this a breaking change? N

**RFC issue number**: 2625

Checklist:

## Acknowledgment

### DIFF
--- a/src/aws-healthomics-mcp-server/awslabs/aws_healthomics_mcp_server/models/core.py
+++ b/src/aws-healthomics-mcp-server/awslabs/aws_healthomics_mcp_server/models/core.py
@@ -233,8 +233,8 @@ class RegistryMapping(BaseModel):
 
     upstreamRegistryUrl: str
     ecrRepositoryPrefix: str
-    upstreamRepositoryPrefix: Optional[str]
-    ecrAccountId: Optional[str]
+    upstreamRepositoryPrefix: Optional[str] = None
+    ecrAccountId: Optional[str] = None
 
 
 class ImageMapping(BaseModel):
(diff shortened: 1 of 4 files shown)


markjschreiber (approved):
LGTM!

Issue #1376, opened 2025-09-27
Introduce optional profile and region arguments to every tool call


codemug:
### Is this related to an existing feature request or issue?

No

### Summary

The current implementation of AWS MCP servers only supports interaction with a single AWS account per server instance. This restriction arises because the servers depend on the default AWS credential chain, with the selected profile determined by the AWS_PROFILE environment variable set at startup. Once the server starts, the profile cannot change.

This limitation makes multi-account workflows cumbersome. Users must either run multiple MCP server instances (one per account) or repeatedly restart the server with a different profile. Neither approach scales well especially when using streamable_http transport when MCP tools are consumed by agents, as it results in redundant tool registrations and hinders cross-account operations.

This RFC proposes extending AWS MCP servers to accept optional profile and region arguments in each tool call, enabling dynamic switching between accounts without restarting or duplicating servers.

### Use case

Organizations typically operate across multiple AWS accounts for reasons such as security boundaries, cost allocation, or team isolation. A single MCP server locked to one account is insufficient in these scenarios. Examples include:

- Cross-VPC debugging: Troubleshooting connectivity across VPCs in different accounts interconnected through VPC endpoints, peering, or transit gateways.
- IAM validation: Verifying cross-account role assumptions and permissions.
- Cost analysis: Exploring and aggregating costs holistically across accounts.
- Incident analysis: An AI agent investigating failures involving EC2 instances or services spread across multiple accounts.

### Proposal

Introduce the ability to specify an optional profile and an optional region parameter for every AWS MCP tool call.

* If provided, the server should temporarily assume the specified profile and region for the scope of that call.
* If omitted, the server should fall back to the default profile and region in the credential chain

The mechanism should integrate seamlessly with existing AWS SDK behaviors (e.g., environment variables, shared config, role assumption).

This approach eliminates the need for multiple MCP servers and supports dynamic, multi-account workflows for agents and tools.

### Out of scope

* Authentication model changes: This RFC does not propose modifying the way AWS credentials are sourced (e.g., SSO, federated login).
* Advanced federation support: While useful, support for AWS SSO, OIDC federation, or external identity providers is not covered.
* Global orchestration: Aggregating results across accounts within a single tool call is out of scope; calls remain atomic per account/region.

### Potential challenges

* Security implications: Allowing arbitrary profile selection could introduce risks if profiles are not properly permission-scoped or validated.
* Performance overhead: Switching credentials per call may add latency compared to a single static profile.
[...]

PR #2672, opened 2026-03-12
feat(aws-healthomics-mcp-server): add optional aws_profile and aws_region parameters to all tools

ahvigil:
Ref #1376

### Changes

Adds optional `aws_profile` and `aws_region` parameters to every HealthOmics MCP tool that makes AWS API calls. When provided, these override the server's default credential chain and region for that specific tool invocation. When omitted, behavior is unchanged (falls back to `AWS_PROFILE` and `AWS_REGION` environment variables).

This is a partial implementation of the RFC in #1376, scoped to `aws-healthomics-mcp-server` only. It follows the precedent set by `ccapi-mcp-server`, which already supports per-call `region` override.

**Modified layers:**
- `utils/aws_utils.py`: `get_aws_session()` and all client factories (`get_omics_client`, `get_ecr_client`, `get_logs_client`, etc.) accept optional `region_name`/`profile_name`
- All 58 tool functions that make AWS calls: accept `aws_profile`/`aws_region` and thread them to client factories
- Search engines and orchestrator: thread params through to client creation
- Tests: fix existing assertions for new signatures, add 9 new tests for override behavior

### User experience

**Before:** Users must run a separate MCP server instance per AWS account, each configured with a different `AWS_PROFILE` environment variable. This results in duplicate tool registrations, prevents cross-account operations in a single conversation, and doesn't scale well for organizations with many accounts.

**After:** A single MCP server instance can operate across all configured AWS accounts. Users (or agents) specify `aws_profile` and/or `aws_region` per tool call to target different accounts:

```json
{
  "tool": "ListAHORuns",
  "arguments": {
    "aws_profile": "prod-account",
    "aws_region": "us-west-2",
    "max_results": 10
  }
}
```

## Checklist

* [x] Changes are documented

Is this a breaking change? N

**RFC issue number**: #1376

Checklist:

## Acknowledgment

---
> 🤖 *Generated by Claude (Opus 4.6)*

### DIFF
--- a/src/aws-dataprocessing-mcp-server/awslabs/aws_dataprocessing_mcp_server/utils/aws_helper.py
+++ b/src/aws-dataprocessing-mcp-server/awslabs/aws_dataprocessing_mcp_server/utils/aws_helper.py
@@ -42,13 +42,39 @@ class AwsHelper:
 
     @staticmethod
     def get_aws_region() -> str:
-        """Get the AWS region from the environment if set."""
-        aws_region = os.environ.get(
-            'AWS_REGION',
-        )
-        if not aws_region:
-            return 'us-east-1'
-        return aws_region
+        """Get the AWS region from environment, AWS config, or default.
+
+        This method follows the standard AWS SDK credential/config resolution chain:
+        1. AWS_REGION environment variable (explicit override)
+        2. AWS_DEFAULT_REGION environment variable (legacy override)
+        3. boto3.Session().region_name (reads ~/.aws/config, instance metadata, etc.)
+        4. 'us-east-1' as last resort fallback
+
+        Returns:
+            The AWS region as a string
+        """
+        # Check AWS_REGION environment variable first
+        aws_region = os.environ.get('AWS_REGION')
+        if aws_region:
+            return aws_region
+
+        # Check AWS_DEFAULT_REGION environment variable
+        aws_region = os.environ.get('AWS_DEFAULT_REGION')
+        if aws_region:
+            return aws_region
+
+        # Use boto3 Session to read from AWS config files and other sources
+        try:
+            session = boto3.Session()
+            aws_region = session.region_name
+            if aws_region:
+                return aws_region
+        except Exception:
+            # If boto3 Session fails, continue to fallback
+            pass
+
+        # Last resort fallback
+        return 'us-east-1'
 
     @staticmethod
     def get_aws_profile() -> Optional[str]:
--- a/src/aws-dataprocessing-mcp-server/tests/utils/test_aws_helper.py
+++ b/src/aws-dataprocessing-mcp-server/tests/utils/test_aws_helper.py
@@ -41,14 +41,57 @@ class TestAwsHelper:
         AwsHelper._aws_partition = None
 
     def test_get_aws_region_with_env_var(self):
-        """Test that get_aws_region returns the region from the environment variable."""
+        """Test that get_aws_region returns the region from AWS_REGION environment variable."""
         with patch.dict(os.environ, {'AWS_REGION': 'us-west-2'}):
             assert AwsHelper.get_aws_region() == 'us-west-2'
 
-    def test_get_aws_region_without_env_var(self):
-        """Test that get_aws_region returns None when the environment variable is not set."""
+    def test_get_aws_region_with_default_env_var(self):
+        """Test that get_aws_region returns the region from AWS_DEFAULT_REGION environment variable."""
+        with patch.dict(os.environ, {'AWS_DEFAULT_REGION': 'eu-west-1'}, clear=True):
+            assert AwsHelper.get_aws_region() == 'eu-west-1'
+
+    def test_get_aws_region_prioritizes_aws_region_over_default(self):
+        """Test that AWS_REGION takes precedence over AWS_DEFAULT_REGION."""
+        with patch.dict(
+            os.environ, {'AWS_REGION': 'us-west-2', 'AWS_DEFAULT_REGION': 'eu-west-1'}
+        ):
+            assert AwsHelper.get_aws_region() == 'us-west-2'
+
+    def test_get_aws_region_from_boto3_session(self):
+        """Test that get_aws_region falls back to boto3.Session().region_name when env vars not set."""
+        # Mock boto3.Session to return a region
+        mock_session = MagicMock()
+        mock_session.region_name = 'ap-southeast-2'
+
         with patch.dict(os.environ, {}, clear=True):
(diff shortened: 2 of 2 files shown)


ahvigil:
i think this covers most things but `GENOMICS_SEARCH_S3_BUCKETS` is still set as a static value on server start which leaves a bit of a gap in the file search tools, not sure if its worth increasing scope here to make that adjustable as well

markjschreiber:
> i think this covers most things but `GENOMICS_SEARCH_S3_BUCKETS` is still set as a static value on server start which leaves a bit of a gap in the file search tools, not sure if its worth increasing scope here to make that adjustable as well

We recently added the ability to search ad-hoc S3 buckets with the `adhoc_s3_buckets: Optional[List[str]]` parameter so you should be able to say something like "search for small sets of reads in my 'sequence-reads' bucket".

The profile and region is a great addition. I will review.

ahvigil:
thanks @markjschreiber, I've installed precommit hooks and addressed linting complaints. Added a `_get_partition` mock to address failing tests, it looks like this previously relied on sharing populated `@lru_cache` entry across all calls, now we get different cache entries for `get_partition()` vs `get_partition(profile_name=None, region_name=None)` which was producing a new cache miss that wasn't mocked.

[8deded9](https://github.com/awslabs/mcp/pull/2672/commits/8deded99c8caf7a57e8b1fca66343bf3272ab899)

edit: missed a bunch of docstring updates and a few pyright issues, updated in 5f7cab15

src/aws-healthomics-mcp-server/awslabs/aws_healthomics_mcp_server/tools/helper_tools.py
```diff
@@ -146,6 +146,14 @@ async def package_workflow(
 
 async def get_supported_regions(
     ctx: Context,
+    aws_profile: Optional[str] = Field(
```
markjschreiber: Strictly speaking it's not necessary to use a profile or region to do this. I support there is an extreme edge case where one profile doesn't have permissions to perform the SSM lookup but another does?

src/aws-healthomics-mcp-server/awslabs/aws_healthomics_mcp_server/utils/aws_utils.py
```diff
@@ -306,9 +388,16 @@ def get_account_id() -> str:
         raise
 
 
-@lru_cache(maxsize=1)
-def get_partition() -> str:
-    """Get the current AWS partition (memoized).
+@cache
```
markjschreiber: Nit, is there a reason to use a `@cache` rather than making the `@lru_cache` a larger but reasonable `maxsize`?

ahvigil: i lazily chose `@cache` here as I wasn't sure what a reasonable value size threshold here would be, assuming that in practice there will be a small-ish number of combinations in a session so size shouldn't blow up. AFAIK the implementation for unbounded cache is also slightly more efficient because it doesn't need to maintain a linked list of cache accesses (probably irrelevant here though).

If you'd prefer a bounded cache I could update this to a plain `@lru_cache` (using default max_size=128). I think that should be enough to handle vast majority of use cases.

ahvigil:
thanks for the review!

> The genomics search functions hold cached results for efficiency of repeated searches. If a search is performed by different profiles I don't think there is a mechanism to segregate the caches or purge the cache. This could lead to confusing search results and leak details of object names/ sizes across profiles.

`S3SearchEngine` and `GenomicsSearchOrchestrator` seem to have instance level caches ([here](https://github.com/awslabs/mcp/blob/main/src/aws-healthomics-mcp-server/awslabs/aws_healthomics_mcp_server/search/s3_search_engine.py#L69-L71) and [here](https://github.com/awslabs/mcp/blob/main/src/aws-healthomics-mcp-server/awslabs/aws_healthomics_mcp_server/search/genomics_search_orchestrator.py#L1158)), and these get reinstantiated for every tool call. As we never switch profiles _within the same tool call_, I don't think the search caches ever have to worry about seeing requests across profiles.

Are these meant to be persisted across calls? Maybe caches should be defined as class variables? Or are you thinking of another more persistent cache somewhere I haven't looked?

src/aws-healthomics-mcp-server/awslabs/aws_healthomics_mcp_server/tools/helper_tools.py
ahvigil: makes sense. even in the case where a profile is missing permissions and causes the lookup to fail, I see theres a fallback implemented here.

ahvigil: resolved in 75ab16e5

markjschreiber:
> thanks for the review!
> 
> > The genomics search functions hold cached results for efficiency of repeated searches. If a search is performed by different profiles I don't think there is a mechanism to segregate the caches or purge the cache. This could lead to confusing search results and leak details of object names/ sizes across profiles.
> 
> `S3SearchEngine` and `GenomicsSearchOrchestrator` seem to have instance level caches ([here](https://github.com/awslabs/mcp/blob/main/src/aws-healthomics-mcp-server/awslabs/aws_healthomics_mcp_server/search/s3_search_engine.py#L69-L71) and [here](https://github.com/awslabs/mcp/blob/main/src/aws-healthomics-mcp-server/awslabs/aws_healthomics_mcp_server/search/genomics_search_orchestrator.py#L1158)), and these get reinstantiated for every tool call. As we never switch profiles _within the same tool call_, I don't think the search caches ever have to worry about seeing requests across profiles.

Good point. Currently the design is safe and stateless. It would be optimal if we could somehow enforce that this can never change, not sure how though. Can you add some code comments to the relevant code that invokes these to warn that these classes should always be created with each call to the search tool?

src/aws-healthomics-mcp-server/awslabs/aws_healthomics_mcp_server/utils/aws_utils.py
ahvigil: switched back to @lru_cache (using default max_size) 5f7cab15

ahvigil:
added some comments to indicate we need to create new instances to keep search caches separate.
8c680a37f9cdd62b5ced702b01c43d5a963a2d8b

looks like CI checks are also passing now 🎉

markjschreiber (approved):
LGTM! Thanks @ahvigil

PR #2699, opened 2026-03-16
feat(agentcore): integrate 9 code interpreter tools to AgentCore MCP Server

Bhavik-T:
Fixes

### Changes

Adds **Code Interpreter** support to the existing `amazon-bedrock-agentcore-mcp-server`, providing 9 new MCP tools that enable AI agents to execute code, run commands, and manage files in sandboxed AgentCore
environments.

**Session Lifecycle (4):**
- `start_code_interpreter_session` — Create a new sandboxed session
- `stop_code_interpreter_session` — Terminate a session
- `get_code_interpreter_session` — Get session details/status
- `list_code_interpreter_sessions` — List sessions (with optional status filter)

**Code Execution (3):**
- `execute_code` — Run Python code with stdout/stderr/result capture
- `execute_command` — Run shell commands
- `install_packages` — Install pip packages

**File Operations (2):**
- `upload_file` — Upload files (text or base64) to a session
- `download_file` — Download files from a session

Integrates into the unified server alongside existing agent runtime tools, using the `bedrock-agentcore` SDK `CodeInterpreter` client. All tools return structured error responses.
Optional `AUTO_STOP_SESSIONS=true` env var for automatic cleanup on server shutdown.

6 new source files under `tools/code_interpreter/`, 55 unit tests across 4 test files, updated `server.py` and `CODEOWNERS`. **136/136 tests passing** (55 new + 81 existing), ruff lint + format clean, pyright
clean.

### User experience

**Before:** The AgentCore MCP server only exposes agent browser tools. Users who want to execute code in a sandboxed environment have no built-in option.

**After:** Users can add `code_interpreter` to their `TOOL_GROUPS` environment variable to get 9 tools for creating sandboxed sessions, executing Python code and shell commands, installing packages, and
uploading/downloading files — all through the same unified MCP server.

## Checklist

* [x] Changes are documented

Is this a breaking change? No

**RFC issue number**: N/A

## Acknowledgment

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -18,7 +18,7 @@ NOTICE                                                         @awslabs/mcp-admi
 
 ## Alphabetical listing of MCP servers
 
-/src/amazon-bedrock-agentcore-mcp-server                       @awslabs/mcp-admins @awslabs/mcp-maintainers  @theagenticguy @scottschreckengaust @theumbrella1 @Vivekbhadauria1 @jesseturner21 @AlexanderRichey @kevin-orellana @Bhavik-T
+/src/amazon-bedrock-agentcore-mcp-server                       @awslabs/mcp-admins @awslabs/mcp-maintainers  @theagenticguy @scottschreckengaust @theumbrella1 @Vivekbhadauria1 @jesseturner21 @AlexanderRichey @kevin-orellana @Bhavik-T @sundargthb
 /src/amazon-kendra-index-mcp-server                            @awslabs/mcp-admins @awslabs/mcp-maintainers  @krokoko @scottschreckengaust
 /src/amazon-keyspaces-mcp-server                               @awslabs/mcp-admins @awslabs/mcp-maintainers  @jcshepherd
 /src/amazon-mq-mcp-server                                      @awslabs/mcp-admins @awslabs/mcp-maintainers  @kenliao94 @hashimsharkh


kevin-orellana (commented):
## Deep Code Review

Thorough review grounded in the actual SDK source (bedrock-agentcore v1.4.6 from PyPI) and the existing browser tools architecture on main. Overall this is a solid integration — well-structured, good test coverage, correct use of the SDK convenience methods. 9 findings below, ordered by impact.

---

### 1. Error Handling Diverges from Browser Tools Pattern (Important)

**Files**: `execution.py`, `session.py`, `files.py` — all tool functions

Browser tools raise exceptions and let the MCP protocol signal errors:

```python
# browser/session.py pattern
except Exception as e:
    error_msg = f'Error starting browser session: {e}'
    logger.error(error_msg)
    await ctx.error(error_msg)
    raise  # FastMCP translates to MCP error response
```

Code interpreter tools catch everything and return dicts with `is_error: true`:

```python
# code_interpreter/execution.py pattern
except Exception as e:
    return ExecutionResult(is_error=True, stderr=..., message=...).model_dump()
```

The consequence: MCP clients see code interpreter errors as **successful tool responses** that happen to contain error info in the body. An LLM may not react to `{"is_error": true, "message": "Session expired"}` the same way it would to an MCP protocol-level error.

**Suggestion**: For infrastructure errors (session not found, credentials expired, SDK exceptions), raise so the MCP framework handles them. Reserve the `is_error` return pattern for *execution* errors (code threw an exception, command returned non-zero exit code) — those are expected outcomes, not tool failures:

```python
async def execute_code(ctx: Context, session_id: str, code: str, ...) -> ExecutionResult:
    try:
        client = get_client(region)
        set_session_context(client, session_id)
        raw = client.execute_code(**kwargs)
        parsed = _parse_invoke_response(raw)
        return ExecutionResult(...)  # execution errors reported via is_error
    except Exception as e:
        logger.error(f'Code execution failed: {e}', exc_info=True)
        await ctx.error(str(e))
        raise  # infrastructure errors raised as MCP errors
```

---

### 2. Missing MCP `Context` Parameter (Important)

**Files**: `session.py`, `execution.py`, `files.py` — all tool functions

All browser tools accept `ctx: Context` from FastMCP:

```python
# browser/session.py
async def start_browser_session(self, ctx: Context, browser_identifier: str = ..., ...):
```

Code interpreter tools don't:

```python
# code_interpreter/session.py
async def start_code_interpreter_session(
    code_interpreter_identifier: str | None = None, ...
):
```

Without `Context`, there's no way to signal errors through the MCP protocol (`await ctx.error()`), send progress updates for long operations (`await ctx.info()`), or access request metadata.

**Fix**: Add `ctx: Context` as the first parameter to each tool function. FastMCP injects it automatically — no changes needed at the registration site.

```python
from mcp.server.fastmcp import Context

async def execute_code(
    ctx: Context,
    session_id: str,
    code: str,
    ...
) -> dict[str, Any]:
```

---

### 3. Shared Mutable Client State (Design Concern)

**File**: `client.py` — `set_session_context` and `get_client`

The per-region client cache returns the same `CodeInterpreter` instance for all tool calls in a region, and `set_session_context()` mutates `client.session_id` / `client.identifier` directly on that shared instance before each SDK call:

```python
client = get_client(region)          # shared cached instance
set_session_context(client, session_id)  # mutates shared state
raw = client.execute_code(**kwargs)       # uses client.session_id internally
```

This works today because asyncio is single-threaded and there are no `await` points between `set_session_context` and the synchronous SDK call. However, it's fragile:

- If the SDK goes async in a future version, coroutine interleaving would cause session cross-talk
- The SDK's `invoke()` silently auto-starts a **new session** if `session_id` is `None` — so if anything clears the client state (e.g., `stop()` from another call), the next `execute_code` runs on a surprise new session instead of failing
- The browser tools avoid this entirely by passing `sessionId` and `browserIdentifier` as **arguments** to each API call via `client.data_plane_client`, with session state tracked in a separate `BrowserConnectionManager`

Not blocking for merge, but worth tracking. A more robust approach would be either (a) one client per session instead of per region, or (b) calling the lower-level `client.data_plane_client.invoke_code_interpreter(sessionId=..., codeInterpreterIdentifier=...)` directly, bypassing the stateful convenience layer.

---

### 4. `stop_all_sessions` Only Stops the Last Session Per Region

**File**: `client.py` — `stop_all_sessions`

The cleanup function reads `client.session_id` to decide what to stop:

```python
async def stop_all_sessions():
    for region, client in _clients.items():
        if client.session_id:      # only the LAST session set on this client
            client.stop()
```

`client.session_id` is mutable state that only holds whichever session was last set via `set_session_context()`. If a user starts sessions A and B in the same region, `client.session_id` points to whichever was used most recently — the other session leaks.

For the `AUTO_STOP_SESSIONS` feature to work correctly, you'd need to track active session IDs explicitly (e.g., a `set` per region). If multi-session-per-region isn't a supported scenario for this initial launch, a code comment noting the limitation would be helpful.

---

### 5. Truthiness Checks Should Use `is not None`

**File**: `session.py` — `start_code_interpreter_session` and `list_code_interpreter_sessions`

```python
# start_code_interpreter_session
if session_timeout_seconds:          # falsy for 0
    kwargs['session_timeout_seconds'] = session_timeout_seconds

# list_code_interpreter_sessions
if max_results:                      # falsy for 0
    kwargs['max_results'] = max_results
```

While `0` is likely invalid for timeouts, it's a subtle correctness issue. **Fix**: `if session_timeout_seconds is not None:` and `if max_results is not None:`.

---

### 6. `stop_code_interpreter_session` — `get_session` After `stop()` Is Unnecessary

**File**: `session.py` — `stop_code_interpreter_session`

After calling `client.stop()` (which clears `client.session_id = None` and `client.identifier = None`), the code makes a second API call to verify status via `get_session()`. The explicit arguments protect against the cleared state, but it's an extra API call that could fail independently. The browser tools don't do this — `stop_browser_session` just returns `TERMINATED` after the stop call succeeds.

**Suggestion**: Remove the verification and report directly:

```python
set_session_context(client, session_id, identifier)
stopped = client.stop()

response = CodeInterpreterSessionResponse(
    session_id=session_id,
    status='TERMINATED' if stopped else 'UNKNOWN',
    code_interpreter_identifier=identifier,
    message=f'Session {session_id} stopped.',
)
```

---

### 7. Unnecessary `.model_dump()` Calls

**Files**: `execution.py`, `session.py`, `files.py` — all return sites

Browser tools return Pydantic models directly (FastMCP serializes them natively). Code interpreter tools manually convert with `.model_dump()`, which adds unnecessary conversion and loses type information.

**Fix**: Return the Pydantic model directly and update return type annotations:

```python
async def start_code_interpreter_session(...) -> CodeInterpreterSessionResponse:
    ...
    return CodeInterpreterSessionResponse(
        session_id=returned_session_id or client.session_id or '',
        status='READY',
        ...
    )
```

---

### 8. Redundant Binary Decode in `download_file`

**File**: `files.py` — `download_file`

The SDK's `download_file()` (v1.4.6) already tries UTF-8 decoding and only returns `bytes` when that fails. The PR then tries UTF-8 decoding again:

```python
if isinstance(result, bytes):
    try:
        file_content = result.decode('utf-8')     # will always fail — SDK already tried this
    except UnicodeDecodeError:
        file_content = base64.b64encode(result).decode('ascii')
        is_binary = True
```

The `result.decode('utf-8')` will always raise `UnicodeDecodeError` because the SDK only returns `bytes` when UTF-8 decoding already failed. The code works correctly (falls through to base64), but the try branch is dead code.

**Suggestion**: Simplify to:

```python
if isinstance(result, bytes):
    # SDK only returns bytes when UTF-8 decode already failed
    file_content = base64.b64encode(result).decode('ascii')
    is_binary = True
else:
    file_content = result
```

---

### 9. No Binary Upload Path Documented

**File**: `files.py` — `upload_file`

The SDK's `upload_file` accepts `Union[str, bytes]` and handles base64 encoding for binary content. The MCP tool only accepts `content: str`, which is reasonable for JSON transport, but there's no guidance for users who need to upload binary files. Since `download_file` returns base64 for binaries, it would be consistent to accept base64 input too.

**Suggestion**: Either add an `encoding` parameter (`encoding='base64'` for binary files) or a docstring note: "Text content only; for binary files, use `execute_command` with base64 decoding."

Bhavik-T:
Addressed all 9 review findings + improved test coverage. Rebased on latest main.

src/amazon-bedrock-agentcore-mcp-server/awslabs/amazon_bedrock_agentcore_mcp_server/tools/code_interpreter/client.py
```diff
@@ -0,0 +1,160 @@
...
+use a separate per-region client cache.
+"""
+
+import os
+from bedrock_agentcore.tools.code_interpreter_client import CodeInterpreter
+from loguru import logger
+
+
+MCP_INTEGRATION_SOURCE = 'awslabs-mcp-code-interpreter-server'
+DEFAULT_IDENTIFIER = 'aws.codeinterpreter.v1'
```
sundargthb, line 28: Q: Every `start_code_interpreter_session` call with no explicit identifier silently lands in Sandbox mode. I came across [this article](https://www.beyondtrust.com/blog/entry/pwning-aws-agentcore-code-interpreter). Our response in the article classified this as intended functionality and explicitly recommends VPC mode for sensitive workloads. Do we want to add a warning log on session start when the identifier resolves to `aws.codeinterpreter.v1`, and surface it in the tool's docstring so agents and human operators are informed at the point of use.

src/amazon-bedrock-agentcore-mcp-server/awslabs/amazon_bedrock_agentcore_mcp_server/tools/code_interpreter/session.py
```diff
@@ -0,0 +1,252 @@
...
+            CODE_INTERPRETER_IDENTIFIER env var or 'aws.codeinterpreter.v1'.
+        name: Optional human-readable name for the session.
+        session_timeout_seconds: Session timeout in seconds.
+            Defaults to DEFAULT_SESSION_TIMEOUT_SECONDS (900).
+        region: AWS region. Defaults to AWS_REGION env var or 'us-east-1'.
+
+    Returns:
+        CodeInterpreterSessionResponse with session_id, status, and message.
+    """
+    identifier = code_interpreter_identifier or get_default_identifier()
```
sundargthb, line 64: same comment as client.py's default identifier

src/amazon-bedrock-agentcore-mcp-server/awslabs/amazon_bedrock_agentcore_mcp_server/tools/code_interpreter/files.py
sundargthb, line 1: The MCP server exposes `upload_file` and `download_file` but has no `list_files` tool. For developers using this in Claude Code, can we add the ability to inspect the sandbox filesystem  — agents won't know what files exist before trying to read or reference them.

A `list_files(session_id, directory_path=None)` tool wrapping `client.invoke("listFiles", {"directoryPath": ...})`.

src/amazon-bedrock-agentcore-mcp-server/awslabs/amazon_bedrock_agentcore_mcp_server/tools/code_interpreter/session.py
```diff
@@ -0,0 +1,252 @@
...
+        error_msg = f'Failed to start session: {type(e).__name__}: {e}'
+        logger.error(error_msg, exc_info=True)
+        await ctx.error(error_msg)
+        raise
+
+
+async def stop_code_interpreter_session(
+    ctx: Context,
+    session_id: str,
+    code_interpreter_identifier: str | None = None,
```
sundargthb, line 103: Do we need this `code_interpreter_identifier` parameter (given per-session client isolation and it doesn't look to be passed to stop call) ?

src/amazon-bedrock-agentcore-mcp-server/awslabs/amazon_bedrock_agentcore_mcp_server/tools/code_interpreter/client.py
kevin-orellana: Two different concerns here:
1. Omission of identifier leads to sandbox mode as default - here, we might opt for public. @Bhavik-T , can we make this public mode by default (the managed tool). Can be another PR.
2. Adding a warning log - we already have public guidance here regarding sandbox mode. Adding a warning log is something we can consider, but our documentation is up to date (sandbox has limited network connectivity)

src/amazon-bedrock-agentcore-mcp-server/awslabs/amazon_bedrock_agentcore_mcp_server/tools/code_interpreter/files.py
kevin-orellana: This is a strong feature request add - we will back-fill this. CC @Bhavik-T

src/amazon-bedrock-agentcore-mcp-server/awslabs/amazon_bedrock_agentcore_mcp_server/tools/code_interpreter/session.py
kevin-orellana: Requiring it dilutes the value prop here which is code-interpreter ready to go and experiment with.

PR #2747, opened 2026-03-23
fix(browser): handle non-2xx HTTP responses in browser_navigate

EA-b13:
Fixes T314: Navigate errors on HTTP error status codes (20 test failures)

### Changes

`browser_navigate` now gracefully handles non-2xx HTTP responses (404, 500, etc.) instead of throwing an error.

__Root cause:__ This is an AgentCore Browser-specific issue. The cloud browser's Chromium configuration converts non-2xx HTTP responses into network-level `net::ERR_HTTP_RESPONSE_CODE_FAILURE` errors. Standard Playwright does NOT throw on non-2xx — it returns a `Response` object with the status code ([docs](https://playwright.dev/docs/api/class-page#page-goto)). This was verified by running the same navigation locally: `page.goto("https://httpbin.org/status/404")` returns `Response(status=404)` with no exception.

When the AgentCore browser throws this error, the page content IS loaded — the tool now recovers and returns the page title, URL, and accessibility tree snapshot to the agent.

__Files changed:__

- __`navigation.py`__ — Catch `PlaywrightError` for `ERR_HTTP_RESPONSE_CODE_FAILURE` specifically. Use `(title unavailable)` fallback when `page.title()` fails due to execution context destruction (the browser may replace the page with its own error screen). Snapshot capture works regardless via CDP. Non-HTTP Playwright errors (DNS, connection refused) still propagate as errors. Updated docstring.
- __`test_unit_interaction.py`__ — Added 4 new unit tests: navigate to 404 returns snapshot, navigate to 500 returns snapshot, non-HTTP errors still error, and title-unavailable fallback when execution context is destroyed.

### User experience

__Before:__ Navigating to a URL that returns a non-2xx HTTP status code (e.g., `https://httpbin.org/status/404`) threw an error like `Error navigating to URL: net::ERR_HTTP_RESPONSE_CODE_FAILURE`. The agent could not see the page content, even though the browser had loaded it. This caused 20 test failures in the 1,000-test integration suite.

__After:__ The tool returns the page snapshot with status `non-2xx (error page loaded)`:

```javascript
Navigated to chrome://new-tab-page/
Title: (title unavailable)
Status: non-2xx (error page loaded)

--- PAGE CONTENT START ---
- heading "This httpbin.org page can't be found" [level=1]
- strong
- button "Reload" [ref=e1]
--- PAGE CONTENT END ---
```

The agent can now see and interact with error pages, just like any other page.

## Checklist

- I have reviewed the [contributing guidelines](https://github.com/awslabs/mcp/blob/main/CONTRIBUTING.md)
- I have performed a self-review of this change
- Changes have been tested
- Changes are documented

Is this a breaking change? __N__

__RFC issue number__: N/A

Checklist:

- Migration process documented
- Implement warnings (if it can live side by side)

## Testing

__Test results:__ 248 unit tests pass, 50 live integration tests pass, ruff lint/format clean, 100% coverage on navigation.py.

__Test results on MCP Inspector:__

Request:

```json
{
  "method": "tools/call",
  "params": {
    "name": "browser_navigate",
    "arguments": {
      "session_id": "01KMCX9TZS1ZXSR12T8WGG85A1",
      "url": "https://httpbin.org/status/404"
    },
    "_meta": {
      "progressToken": 1
    }
  }
}
```

Response:

```json
{
  "content": [
    {
      "type": "text",
      "text": "Navigated to chrome://new-tab-page/\nTitle: (title unavailable)\nStatus: non-2xx (error page loaded)\n\n--- PAGE CONTENT START ---\n- heading \"This httpbin.org page can't be found\" [level=1]\n- strong\n- button \"Reload\" [ref=e1]\n--- PAGE CONTENT END ---"
    }
  ],
  "isError": false
}
```

__Local Playwright verification__ (confirms this is AgentCore-specific, not a Playwright issue):

```javascript
Testing: https://httpbin.org/status/404 → ✅ RETURNED: status=404 (no throw)
Testing: https://httpbin.org/status/500 → ✅ RETURNED: status=500 (no throw)
Testing: https://httpbin.org/status/403 → ✅ RETURNED: status=403 (no throw)
```

## Acknowledgment


kevin-orellana (dismissed):
Looks good. Clean fix for the AgentCore-specific ERR_HTTP_RESPONSE_CODE_FAILURE issue — non-HTTP errors still propagate correctly, title fallback is defensive, and snapshot recovery via CDP is the right approach. Tests cover all branches.

sundargthb (dismissed):
LGTM

kevin-orellana (dismissed):
Approving the fix. Verified locally on this branch merged with current `main` (head `95b1b4c8`), running the same gates CI would:

- `uv run --frozen pytest -m 'not live and not local_browser'` - 821 passed, 50 deselected
- `uv run --frozen pyright` - 0 errors, 0 warnings
- `uv run --frozen ruff check` - all checks passed
- `uv run --frozen ruff format --check` - 153 files already formatted

I also drove `browser_navigate` through a behavioral harness rather than relying on the test count alone. Confirmed:

1. 200 responses are unchanged.
2. A 404 arriving as `net::ERR_HTTP_RESPONSE_CODE_FAILURE` now returns `Navigated to ...` with `Status: non-2xx (error page loaded)` and usable element refs, so the agent can act on the error page instead of getting a failure string.
3. `goto` returning `None` still yields `Status: unknown`.
4. `ERR_NAME_NOT_RESOLVED`, `ERR_CONNECTION_REFUSED`, `ERR_CERT_AUTHORITY_INVALID` and `ERR_ABORTED` all still error out, and so do timeouts. Worth calling out that `PlaywrightTimeoutError` subclasses `playwright.async_api.Error`, so the `else: raise` is the only thing keeping this catch from swallowing timeouts. Good that it is there.
5. `page.title()` raising falls back to `(title unavailable)`, and an `asyncio.CancelledError` from `title()` still propagates rather than being swallowed.
6. Scheme validation still short-circuits before any navigation is attempted.

I mutation-tested the new tests as well: reverting only `navigation.py` to `main` while keeping the 4 new tests fails 3 of them, so they are genuinely pinned to the behavior they claim to cover.

Two non-blocking observations, neither of which should hold this PR:

1. `_navigate_history`, behind `browser_navigate_back` and `browser_navigate_forward`, has the same underlying problem. `go_back()` / `go_forward()` onto a non-2xx page raise the identical error and get reported as `Error navigating back in session ...`. It is less severe than the `browser_navigate` case because `error_with_snapshot` still attaches a snapshot, so the agent is not blind, just told the navigation failed when it did not. Perhaps a follow-up PR.
2. The `'net::ERR_HTTP_RESPONSE_CODE_FAILURE' in error_str` check is a substring match, so a compound message mentioning both a DNS failure and that marker would be reported as success. Contrived, and not a shape Playwright emits today, so I would leave it as is.

One thing is needed before this can merge: the branch conflicts with `main` on `.github/CODEOWNERS`. That is the only conflicting file. I checked with `git merge-tree` and both `navigation.py` and `test_unit_interaction.py` auto-merge cleanly. `main` has since added `@jariy17` and `@Sudip2403` to the agentcore line, which this branch predates, so the resolution is the union of both:

```
/src/amazon-bedrock-agentcore-mcp-server   @awslabs/mcp-admins @awslabs/mcp-maintainers  @theagenticguy @scottschreckengaust @theumbrella1 @Vivekbhadauria1 @jesseturner21 @AlexanderRichey @kevin-orellana @Bhavik-T @sundargthb @jariy17 @Sudip2403 @EA-b13
```

Adding you as a code owner is fine by me.

Two sequencing notes so this does not take an extra round trip:

1. `dismiss_stale_reviews_on_push` is enabled here, so this approval and @sundargthb's will both drop the moment the conflict resolution is pushed. Please push it and then ping me, and I will re-approve.
2. Better that you push the resolution rather than me. `require_last_push_approval` is also enabled, so if I push it my own approval no longer counts toward the requirement and we would need two other reviewers.

Also flagging that CI has never actually executed on this PR. Fork PRs sit at `action_required` until a maintainer approves each workflow run individually, so the current check state is not a real signal either way. I will approve the runs once the conflict is resolved.

kevin-orellana (approved):
I went ahead and resolved the `.github/CODEOWNERS` conflict for you, since the PR allows maintainer edits. Pushed as `7dd7ff6e` (fast-forward from `95b1b4c8`, no history rewritten). The branch is now `MERGEABLE`.

The resolution keeps everyone: `main` had added `@jariy17` and `@Sudip2403` to the agentcore owner line after this branch forked, and this branch added `@EA-b13`, so the merged line is the union of both. Nothing else in the file changed.

Confirmed the net diff against `main` is still exactly the three intended files at +119/-7, unchanged from before the merge:

- `.github/CODEOWNERS` (+1/-1)
- `tools/browser/navigation.py` (+32/-6)
- `tests/browser/test_unit_interaction.py` (+86/-0)

Re-ran all four gates on the resolved merge:

- `uv run --frozen pytest -m 'not live and not local_browser'` - 821 passed, 50 deselected
- `uv run --frozen pyright` - 0 errors, 0 warnings
- `uv run --frozen ruff check` - all checks passed
- `uv run --frozen ruff format --check` - 153 files already formatted

Re-approving, since the push dismissed the earlier approvals as expected.

@sundargthb - sorry, your approval got dismissed by the resolution push. Could you re-approve? Two things need it: the repo requires 2 approvals, and `require_last_push_approval` is enabled, which means my own approval cannot satisfy that rule now that I am the one who pushed. So your re-approval is what actually unblocks the merge.

One unrelated thing I noticed while verifying, not caused by this PR and not something to fix here: `tools/memory/guide.py` on `main` has unresolved merge-conflict markers committed into it around lines 45-53, referencing `66f98ac2` and `45897c56`. It sits inside a docstring so it does not break imports, which is presumably how it slipped through. Worth a separate issue.

Bhavik-T (approved):

scottschreckengaust (approved):
LGTM

PR #2797, opened 2026-03-26
fix(aws-api-mcp-server): add error sanitization

arnewouters:
Fixes

### Changes

### User experience

## Checklist

* [ ] Changes are documented

**RFC issue number**:

Checklist:

## Acknowledgment


src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/common/errors.py
```diff
@@ -679,3 +676,57 @@ def __init__(self, reason: str, profile_name: str | None = None):
...
+    @classmethod
+    def _resolve_message(cls, exc: BaseException) -> str:
+        """Walk the exception's MRO to find the most specific safe message."""
+        for exception_class in type(exc).__mro__:
+            if exception_class in cls._SAFE_MESSAGES:
+                return cls._SAFE_MESSAGES[exception_class]
+        return cls._DEFAULT_MESSAGE.format(exception_name=type(exc).__name__)
+
+
+def sanitized_exceptions(func):
```
iddv, line 720: Note: `SanitizedException` extends `Exception`, not the original exception type. This means any upstream code that catches specific types (e.g., `except OSError`, `except FileNotFoundError`) will no longer match — the exception type is now `SanitizedException`. If awscli or any caller of these patched functions has typed exception handlers, they'll miss and the `SanitizedException` will bubble up further than the original would have. Worth verifying that the callers of `get_file`, `unpack_scalar_cli_arg`, `resolve_given_outfile_path`, and `_get_file_contents` don't have specific `except` clauses that depend on catching the original exception type.

Also: consider `raise SanitizedException(exc) from None` instead of `raise SanitizedException(exc)` — currently the original exception (with expanded credential values) is implicitly chained via `__context__` and could appear in full tracebacks if logged.

arnewouters: > SanitizedException will bubble up further than the original would have

I checked the paths before and this is fine for the current state of the code, but also FastMCP catches any exceptions and returns them as a tool error

> full tracebacks if logged

We don't log full tracebacks but if we would that's still okay I think since these only be viewable on the platform where the MCP server is running

PR #2802, opened 2026-03-27
feat(codeInterpreter): add list-files tool to MCP

EA-b13:
Fixes: N/A — New feature addition

### Changes

Adds a new `list_files` MCP tool to the Code Interpreter tools in the Amazon Bedrock AgentCore MCP server. This tool allows users to list files and directories within a sandboxed code interpreter session, complementing the existing `upload_file` and `download_file` tools.

**Files changed (6):**

- **`models.py`** — Added `FileListResult` Pydantic response model with `files` (list of paths), `content` (raw text output), and `message` fields
- **`files.py`** — Added `list_files()` async tool function that calls `client.invoke('listFiles', ...)` via the `InvokeCodeInterpreter` API with an optional `directoryPath` argument, plus a `_parse_list_files_response()` helper that handles both `text` and `resource_link` content blocks from the EventStream response
- **`__init__.py`** — Registered `files.list_files` as an MCP tool; updated docstring and registration function doc from "9 tools" to "10 tools"
- **`server.py`** — Updated `AGENTCORE_MCP_INSTRUCTIONS` to mention `list_files`; updated tool count log message from 9 to 10
- **`test_files.py`** — Added 13 new tests: 7 for `_parse_list_files_response` (text blocks, resource_link blocks, name fallback, empty stream, missing stream key, total-line filtering, mixed content priority) and 6 for `list_files` (happy path, directory_path param, empty directory, resource_link response, SDK exception, unregistered session)
- **`test_init.py`** — Updated tool registration count assertion from 9 to 10

### User experience

**Before:** Users could upload and download files in Code Interpreter sandbox sessions, but had no way to discover what files existed in the sandbox without executing a shell command like `ls`.

**After:** Users can call the `list_files` tool with a `session_id` and optional `directory_path` to get a structured list of files/directories in the sandbox. Example response:
```json
{
  "files": ["/log", "/node_modules", "/package.json", "/.ipython"],
  "content": "",
  "message": "Found 4 file(s) in default working directory."
}
```

## Checklist

* [x] Changes are documented

Is this a breaking change? N

**RFC issue number**: N/A

Checklist:

## Acknowledgment


kevin-orellana (dismissed):
Looks good. Parser handles both response formats well, test coverage is thorough.

One nit: the `region` parameter on `list_files()` is accepted but never used — it's not passed to `get_session_client` or anywhere else. Consider removing it to avoid implying multi-region support that isn't wired up, or wire it through if it's intended for a follow-up.

kevin-orellana:
Thanks for this — `list_files` is a real gap, and the shape of the tool is right. Rather than leave you a review round-trip, I pushed the fixes directly to this branch (`12ef0dd1..c0247261`, fast-forward, your commits untouched) since `maintainerCanModify` was enabled. Happy to revert any of it if you'd rather own the change.

## What I changed

**1. `isError` was ignored (the blocking one).**

`_parse_list_files_response` never looked at `event_result['isError']`, so an error response was parsed as if it were a listing. Concretely, for a bad path the tool returned:

```python
FileListResult(
    files=["Error: directory '/nope' does not exist"],
    message='Found 1 file(s) in /nope.',
)
```

That is a failure reported as success, and the error text is now a "file path" an agent can hand straight to `download_file`. The parser returns an error flag, `list_files` short-circuits with `files=[]`, surfaces the sandbox text via `ctx.error`, and `FileListResult` carries `is_error` (matching `ExecutionResult`, which already has it).

**2. The text fallback treated every non-blank line as a file name.**

The only filter was a literal `total ` prefix, so `ls -l` output produced entries like `-rw-r--r-- 1 root root 12 Jan  1 00:00 a.csv` as "file names". New `_parse_listing_line` helper detects a 10-character mode field (`-rw-r--r--`, `drwxr-xr-x`) and takes the trailing name. It joins the remaining fields rather than taking `fields[-1]`, so names containing spaces survive, and bare-name listings pass through unchanged.

**3. No flat-dict fallback, and a test pinned the gap.**

`test_parse_no_stream_key` asserted `files == []` for a response without a `stream` key — encoding "we don't handle this" as a passing test. The parser now mirrors `execution.py::_parse_invoke_response`, which already has `elif isinstance(result, dict)` and `elif isinstance(result, str)` branches for exactly this case, and the test asserts the parsed names.

Also bumped the README tool count 9 → 10 in the three places it appears (summary line, `### Code Interpreter` heading + key-tools list, architecture tree) and added the CHANGELOG entry. Worth noting: this branch is forked from a `main` that predates the README rewrite in #3529, so I merged current `main` in to make those edits land on the right text — that's the merge commit in the range.

## Verification

Ran the four gates via `uv run --frozen` in `src/amazon-bedrock-agentcore-mcp-server`:

| Gate | Result |
|---|---|
| `pytest -m 'not live and not local_browser'` | 838 passed, 50 deselected |
| `pyright` | 0 errors, 0 warnings |
| `ruff check` | All checks passed |
| `ruff format --check` | 153 files already formatted |

Also mutation-tested the new tests: reverted `files.py` to your version while keeping the tests, and confirmed `test_list_files_reports_sandbox_error` fails on the `is_error` assertion and `test_list_files_long_format_output` fails with the raw `-rw-r--r-- ... sub` rows. The regression guards actually guard.

## Known limitation I did not fix

If the sandbox emits prose around the listing — e.g.

```
Contents of /work:
a.csv
b.csv
2 items
```

— the header and footer still come through as file names. I left this alone deliberately: any heuristic that strips them (drop first/last line, filter lines with colons) will eventually eat a real file name, and a wrong filter is worse than a stray entry the caller can see. The `resource_link` path, which is what the service actually returns, is unaffected. Flagging it so it's a known edge rather than a surprise.

## Retraction

My April review asked you to drop the unused `region` parameter. That was wrong. Checking the directory: 6 of the 9 pre-existing tools accept `region` without reading it — `stop_code_interpreter_session`, all three in `execution.py`, and `upload_file`/`download_file`. Every one of them resolves its client through `get_session_client(session_id)`, so region is already fixed at session creation and there is nothing left for the parameter to do. The three that *do* use it (`start_code_interpreter_session`, `get_code_interpreter_session`, `list_code_interpreter_sessions`) are region-level rather than session-bound. `list_files` is session-bound, so keeping the parameter is both the convention here and the right call for signature consistency. Sorry for the noise.

## What this needs now

Because I pushed, `dismiss_stale_reviews_on_push` clears existing review state, and `require_last_push_approval` means my own approval can no longer satisfy the ruleset on this head. This needs two approvals from other maintainers — @Bhavik-T, @sundargthb, if either of you has a moment, `files.py` is the only file with real logic in it.

kevin-orellana:
Follow-up, and a correction to my last comment.

I said patch coverage was clean. That was read off an incomplete Codecov report and was wrong. What happened: `codecov/project` initially failed at 91.17% (-1.88%) with `Files 1031 → 262` and `Lines 86826 → 71333` fewer — a partial upload, because one build job died on a Docker Hub pull timeout and Codecov finalized at 18:36:27Z while the coverage-uploading workflow ran until 18:37:55Z. I diagnosed that as an infra artifact, which it was, and took the passing `codecov/patch` at face value in the same breath. Once the reruns completed and all 1031 files uploaded, the two checks swapped: `project` passed at 93.03% (-0.01%), and `patch` failed at **88.88% of diff hit** against a 93.04% target — 3 missing lines and 6 partial branches, every one of them in code I added.

Reproduced locally with `--cov-branch`: `153-154, 182, 191->185, 196->185, 204->203, 206->209`. Pushed `82eccd64` with seven tests, one per uncovered path:

| Path | What was untested |
|---|---|
| `153-154` | long-format row with no date columns (busybox-style `ls -l`) |
| `153->156` | mode-like first field with too few columns, kept verbatim |
| `182` | stream event with no `result` key |
| `191->185` | content block that is neither `text` nor `resource_link` |
| `196->185` | `resource_link` carrying neither `uri` nor `name` |
| `204->203` | flat-dict response containing a non-text block |
| `206->209` | response that is neither dict nor str |

`files.py` is now at 100% statement and branch coverage (109 stmts, 48 branches, 0 missing). Two of these are worth more than a coverage number: the empty-`resource_link` case confirms such a block adds no phantom entry, and `test_parse_mode_like_line_too_short_is_kept_verbatim` pins the deliberate choice not to truncate a legitimate 10-character file name that happens to begin with an `ls` type character.

Gates on `82eccd64`:

| Gate | Result |
|---|---|
| `pytest -m 'not live and not local_browser'` | 845 passed, 50 deselected |
| `pyright` | 0 errors, 0 warnings |
| `ruff check` | All checks passed |
| `ruff format --check` | 153 files already formatted |

The `s3-tables-mcp-server` build failure was unrelated to this PR — `docker: Error response from daemon: Get "https://registry-1.docker.io/v2/": context deadline exceeded` pulling `cyclonedx/cyclonedx-cli:latest`. It passed on rerun.

This push moves the head again, so the review requirement is unchanged: two approvals from maintainers other than me. @Bhavik-T, @sundargthb — `files.py` is still the only file with logic in it; `82eccd64` is tests only.

kevin-orellana:
CI is green on `82eccd64` — 137 check runs, no failures. Closing out the coverage thread above: `codecov/patch` is now **100.00% of diff hit** (target 93.04%), up from the 88.88% that the earlier partial upload had hidden, and `codecov/project` is 93.04% (+0.00%).

Nothing is outstanding on this PR except review. @Bhavik-T, @sundargthb — `files.py` is the only file with logic in it, and `82eccd64` on top of it is tests only.

Bhavik-T (approved):

Bhavik-T (approved):

kevin-orellana (approved):

PR #2808, opened 2026-03-27
feat: replace direct client imports with get_client() factory and add get_region() override in cloudwatch-applicationsignals

AsakerMohd:
Fixes N/A

### Changes

Replace direct singleton client imports with a pluggable `get_client()` factory pattern and add `get_region()` / `set_region_override()` for runtime region overrides across all tool modules.

- **aws_clients.py**: Added `get_region()`, `set_region_override()`, `clear_region_override()`. Updated docstrings to reflect generic consumer-facing factory pattern.
- **9 tool modules** (server.py, change_tools.py, group_tools.py, trace_tools.py, slo_tools.py, service_tools.py, canary_utils.py, sli_report_client.py, audit_utils.py): Replaced direct singleton imports (e.g. `from .aws_clients import logs_client`) with `get_client('logs')` and `AWS_REGION` with `get_region()`.
- **Developer guardrails**: Added `scripts/check_direct_boto3.sh` lint script, `no-direct-boto3` pre-commit hook, and "AWS Client Usage" section in README.md.

### User experience

**Before**: Consumers of this package could not override AWS credentials or region at runtime. All tool modules imported singleton clients directly from `aws_clients`, which were created once at module load time using the default credential chain.

**After**: Consumers can register a custom client factory via `set_client_factory()` and override the region via `set_region_override()`. All tools transparently use the overridden credentials/region. Local/CLI usage is unchanged — without a factory registered, `get_client()` falls back to the default singleton clients.

## Checklist

* [x] Changes are documented

Is this a breaking change? N

**RFC issue number**: N/A

Checklist:

## Acknowledgment


src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/aws_clients.py
```diff
@@ -12,109 +12,190 @@
...
 
-def _initialize_aws_clients():
-    """Initialize AWS clients with proper configuration."""
-    # Add caller suffix if MCP_RUN_FROM is set
+# ---------------------------------------------------------------------------
+# Region override — allows consumers of this package to override the region
+# at runtime. Falls back to _DEFAULT_REGION.
+# ---------------------------------------------------------------------------
+
+_region_override: Optional[str] = None
```
mxiamxia: **Thread safety concern:** `_region_override` and `_client_factory` (line 92) are plain module-level globals mutated via `set_*()` with no synchronization. Since this is an async MCP server handling concurrent requests, there's a race condition:

1. Request A calls `set_region_override('us-west-2')`
2. Request B arrives and calls `get_region()` — gets `us-west-2` (Request A's value)
3. Request A finishes and calls `clear_region_override()`
4. Request B's subsequent `get_region()` calls now return the default mid-request

The same issue applies to `_client_factory`. These are request-scoped values stored in process-global state.

**Suggestion:** Use `contextvars.ContextVar` instead — it's the standard Python mechanism for request-scoped state in async code and gives each coroutine its own isolated copy:

```python
import contextvars

_region_override_var: contextvars.ContextVar[Optional[str]] = contextvars.ContextVar(
    '_region_override', default=None
)
_client_factory_var: contextvars.ContextVar[Optional[Callable]] = contextvars.ContextVar(
    '_client_factory', default=None
)

def set_region_override(region: str) -> None:
    _region_override_var.set(region)

def get_region() -> str:
    override = _region_override_var.get()
    return override if override is not None else _DEFAULT_REGION
```

A context manager API would also be safer to prevent forgotten `clear_*()` calls:
```python
@contextmanager
def region_override(region: str):
    token = _region_override_var.set(region)
    try:
        yield
    finally:
        _region_override_var.reset(token)
```

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/aws_clients.py
```diff
@@ -12,109 +12,190 @@
...
+
+
+def set_region_override(region: str) -> None:
+    """Override the AWS region for all subsequent get_region() calls."""
+    global _region_override
+    _region_override = region
+    logger.info(f'Region override set: {region}')
```
mxiamxia, line 102: **Semantic inconsistency:** In default mode (no custom factory), `set_region_override()` changes what `get_region()` returns, but `get_client()` still returns pre-built singleton clients that were created with `_DEFAULT_REGION` at module load time. This creates a mismatch:

```python
set_region_override('eu-west-1')
region = get_region()        # → 'eu-west-1'
client = get_client('logs')  # → still connected to us-east-1!
```

A developer constructing ARNs or filter parameters with `get_region()` could easily send the API call to a different region via the client. This is a subtle bug source.

**Possible fixes:**
1. Require a custom factory before allowing region override (cleanest):
```python
def set_region_override(region: str) -> None:
    if _client_factory is None:
        raise RuntimeError(
            "set_region_override() requires a custom client factory. "
            "Call set_client_factory() first."
        )
    ...
```
2. Rebuild singleton clients when region changes (heavier)
3. At minimum, document this limitation prominently in the docstring here

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/aws_clients.py
```diff
@@ -12,109 +12,190 @@
...
+    Returns:
+        A boto3 client for the specified service.
+    """
+    if _client_factory is not None:
+        logger.debug(f'get_client({service_name}): using custom factory')
+        return _client_factory(service_name)
+
+    # Fall back to module-level singletons
+    logger.debug(f'get_client({service_name}): using default singleton (local creds)')
+    return _singleton_clients[service_name]
```
mxiamxia, line 211: **Missing test coverage:** This PR introduces a new public API surface (`get_client`, `get_region`, `set_client_factory`, `set_region_override`, `clear_*`) and refactors credential handling across 9 tool modules, but includes no tests. For a change of this scope touching auth/credential plumbing, tests should be a hard requirement before merge.

Suggested minimum test cases:
- `test_get_client_returns_singleton_in_default_mode` — same object on repeated calls
- `test_get_client_uses_factory_when_set` — factory is invoked, singleton is bypassed
- `test_get_client_unknown_service_raises` — clear error on typo (currently raises raw `KeyError`)
- `test_get_region_returns_default` / `test_get_region_returns_override` / `test_clear_region_override_reverts`
- `test_clear_client_factory_reverts_to_singleton`
- `test_backward_compat_exports_still_work` — `from .aws_clients import logs_client` etc.
- Integration: at least one tool-level test verifying the factory-injected client is actually used end-to-end

PR #2833, opened 2026-04-01
fix(mcp-lambda-handler): handle PEP 604 union types in schema generation

### LINKED ISSUE #2354: Bug: `Optional[int]` type hint incorrectly generates `string` type in JSON Schema

mango766:
## Summary

Fixes #2354

On Python 3.10+, `int | None` creates a `types.UnionType` instead of `typing.Union`. The `get_type_schema()` function in `MCPLambdaHandler.tool()` only checked for `typing.Union`, causing PEP 604 union types (like `int | None`) to fall through to the default `{"type": "string"}` case.

This means `Optional[int]` parameters (when resolved to `int | None` by `get_type_hints()`) incorrectly generate `{"type": "string"}` instead of `{"type": "integer"}`.

## Fix

Add an `isinstance` check for `types.UnionType` (Python 3.10+) alongside the existing `typing.Union` check, so both syntaxes are handled correctly.

## Test plan

- [ ] `Optional[int]` generates `{"type": "integer"}` in the JSON schema
- [ ] `int | None` generates `{"type": "integer"}` in the JSON schema
- [ ] `Optional[str]` still generates `{"type": "string"}`
- [ ] `Union[int, str]` still works (takes first non-None type)

### DIFF
--- a/src/mcp-lambda-handler/awslabs/mcp_lambda_handler/mcp_lambda_handler.py
+++ b/src/mcp-lambda-handler/awslabs/mcp_lambda_handler/mcp_lambda_handler.py
@@ -16,6 +16,8 @@ import functools
 import inspect
 import json
 import logging
+import sys
+import types
 from awslabs.mcp_lambda_handler.session import DynamoDBSessionStore, NoOpSessionStore, SessionStore
 from awslabs.mcp_lambda_handler.types import (
     Capabilities,
@@ -218,7 +220,10 @@ class MCPLambdaHandler:
                 origin = get_origin(type_hint)
 
                 # Handle Union types (including Optional[T] which is Union[T, None])
-                if origin is Union:
+                # Also handle PEP 604 union syntax (e.g., int | None) on Python 3.10+
+                if origin is Union or (
+                    sys.version_info >= (3, 10) and isinstance(type_hint, types.UnionType)
+                ):
                     args = get_args(type_hint)
                     non_none_args = [arg for arg in args if arg is not type(None)]
                     if len(non_none_args) == 1:


mikegc-aws (approved):
Verified locally — the fix is correct and all 81 existing tests pass with the change applied (Python 3.10.17).

**Analysis:**

The bug: on Python 3.10+, `int | None` (PEP 604 syntax) creates a `types.UnionType` whose `get_origin()` returns `types.UnionType`, *not* `typing.Union`. The current code only checks `origin is Union`, so PEP 604 union types fall through to the default `{"type": "string"}`.

Nuance: `get_type_hints()` normalizes PEP 604 annotations back to `typing.Union` on Python 3.10, so `Optional[int]` and `int | None` used as *annotations* both work today via the existing path. But on Python 3.12+ the normalization behavior differs, and if `get_type_schema()` is ever called with a runtime-constructed PEP 604 type, the fix is necessary. The fix is correct and defensive.

**Tested:**
- `Optional[int]` → `{"type": "integer"}` ✅
- `int | None` → `{"type": "integer"}` ✅  
- `Optional[str]` → `{"type": "string"}` ✅
- All existing tests pass ✅

**Note:** The PR doesn't add test coverage for PEP 604 types. Consider adding a test like:

```python
def test_tool_decorator_pep604_type_hints():
    handler = MCPLambdaHandler('test-server')
    
    @handler.tool()
    def foo(a: int | None, b: str | None, c: float | None) -> str:
        """Test tool.
        
        Args:
            a: An int
            b: A str
            c: A float
        """
        return 'ok'

    props = handler.tools['foo']['inputSchema']['properties']
    assert props['a']['type'] == 'integer'
    assert props['b']['type'] == 'string'
    assert props['c']['type'] == 'number'
```

Adding the `backlog` label to prevent the stale bot from closing this.

mikegc-aws:
Thanks for this fix — the PEP 604 `int | None` handling looks correct and the `sys.version_info >= (3, 10)` guard is the right approach. The only thing blocking merge is the **Require Contributor Statement** check, which is currently failing. Could you add the contributor statement (per CONTRIBUTING.md) so CI can go green? Once that's signed and the check passes, this should be ready to merge.

PR #2877, opened 2026-04-07
fix(amazon-bedrock-agentcore-mcp-server): skip signal handlers on Windows

### LINKED ISSUE #2752: amazon-bedrock-agentcore-mcp-server: crashes on Windows due to asyncio.add_signal_handler NotImplementedError
### Describe the bug

The MCP server fails to start on native Windows.

When running:
uvx awslabs.amazon-bedrock-agentcore-mcp-server@latest

The server initializes (browser tools register successfully) but then crashes during startup with:

NotImplementedError: loop.add_signal_handler()

This happens in server_lifespan() where asyncio signal handlers are registered:

loop = asyncio.get_running_loop()
for sig in (signal.SIGTERM, signal.SIGINT):
    loop.add_signal_handler(
        sig, lambda cm=_browser_cm: asyncio.ensure_future(cm.cleanup())
    )

On Windows, asyncio does not support add_signal_handler(), so the server crashes before becoming usable.

I verified this also occurs when launching via Claude Code MCP (stdio transport).

### Expected Behavior

The server should start successfully on Windows, or at minimum:

- Skip signal handler registration on Windows, or
- Catch NotImplementedError, or
- Clearly document that native Windows is unsupported

Currently the server crashes during startup, making it unusable on Windows.

### Current Behavior

The server starts initialization but fails during startup on Windows.

Observed logs:
- Browser tools register successfully
- Then the process crashes before becoming available

Exception:
NotImplementedError raised from asyncio when calling:
loop.add_signal_handler()

As a result:
- MCP server never becomes ready
- Claude Code detects the server but cannot use any tools

### Reproduction Steps

1. Use a Windows machine (e.g. Windows 10/11)

2. Run the MCP server:
   uvx awslabs.amazon-bedrock-agentcore-mcp-server@latest

3. Observe startup logs:
   - Browser tools register successfully

4. Server crashes with:
   NotImplementedError: loop.add_signal_handler()

(Optional)
5. Add the server to Claude Code via MCP:
   claude mcp add --transport stdio bedrock-agentcore-mcp-server -- uvx awslabs.amazon-bedrock-agentcore-mcp-server@latest

6. Open Claude Code and run /mcp:
   - Server is detected but fails to start and is unusable

### Possible Solution

_No response_

_No response_

### OS

Windows

### Server

other

### Server Version

_No response_

eu-west-1

### Other information

Guard signal handler registration on Windows, or catch NotImplementedError when calling asyncio loop.add_signal_handler().

For example:

loop = asyncio.get_running_loop()
for sig in (signal.SIGTERM, signal.SIGINT):
    try:
        loop.add_signal_handler(
            sig, lambda cm=_browser_cm: asyncio.ensure_future(cm.cleanup())
        )
    except NotImplementedError:
        pass

Alternatively, only register signal handlers on non-Windows platforms, e.g. with a platform check such as sys.platform != "win32".

This would allow the server to continue starting on Windows instead of crashing during startup.

### Service quota

syukawa-gh:
Fixes #2752

### Changes

The MCP server crashes on Windows with `NotImplementedError` because `asyncio.add_signal_handler()` is not supported on Windows event loops.

- Guard `add_signal_handler` calls with `sys.platform != 'win32'` check
- On Windows, signal handlers are skipped — cleanup still happens via the `finally` block in the lifespan context manager

### User experience

Before: Server crashes immediately on Windows with `NotImplementedError: loop.add_signal_handler()`.
After: Server starts and runs on Windows. Browser session cleanup is handled by the `finally` block.

## Checklist

* [ ] Changes are documented

Is this a breaking change? N

## Acknowledgment

### DIFF
--- a/src/amazon-bedrock-agentcore-mcp-server/awslabs/amazon_bedrock_agentcore_mcp_server/server.py
+++ b/src/amazon-bedrock-agentcore-mcp-server/awslabs/amazon_bedrock_agentcore_mcp_server/server.py
@@ -17,6 +17,7 @@
 import asyncio
 import os
 import signal
+import sys
 from .tools import docs, gateway, memory, runtime
 from .utils import cache
 from collections.abc import AsyncIterator
@@ -100,11 +101,12 @@ _code_interpreter_cleanup = None
 async def server_lifespan(server: FastMCP) -> AsyncIterator[None]:
     """Manage server lifecycle — browser cleanup task, code interpreter cleanup, and graceful shutdown."""
     if _browser_cm is not None and _browser_sm is not None:
-        loop = asyncio.get_running_loop()
-        for sig in (signal.SIGTERM, signal.SIGINT):
-            loop.add_signal_handler(
-                sig, lambda cm=_browser_cm: asyncio.ensure_future(cm.cleanup())
-            )
+        if sys.platform != 'win32':
+            loop = asyncio.get_running_loop()
+            for sig in (signal.SIGTERM, signal.SIGINT):
+                loop.add_signal_handler(
+                    sig, lambda cm=_browser_cm: asyncio.ensure_future(cm.cleanup())
+                )
 
         from .tools.browser import cleanup_stale_sessions
 


kevin-orellana:
Thanks for this, and apologies for the very long silence — this deserved a response in April, not now.

Closing because the code this patches no longer exists. The Unix-only `asyncio.add_signal_handler` call was removed outright in #3280 (`41fc0f4f`, merged 2026-04-28), which also closed #2752. Rather than guarding the call for Windows, we deleted it, so there is nothing left to guard. Verified on current `main`: no `add_signal_handler`, `SIGINT`, `SIGTERM`, or `win32` references remain anywhere in `server.py`.

Your report and fix were correct for the code as it stood. The removal approach was chosen because the handler was not doing anything the MCP stdio transport does not already handle on shutdown, so the simplest correct fix was to drop it rather than add a platform branch. That is a difference in approach, not a problem with your diagnosis — you identified a real crash on a real platform.

The branch has also since gone into conflict with `main`, which is a secondary reason to close rather than rebase for a change that is now a no-op.

PR #2924, opened 2026-04-08
feat(cloudwatch-applicationsignals): improve Synthetics functionalities with list_canaries and kb

yig-yi:
* Improve CloudWatch Synthetics canary failure analysis with a knowledge base-powered recommendation engine for known runtime and environment issues.
* Add mcp tool `list_canaries` to discover canaries in the account

### Changes

**New**:
* `list_canaries`: discover canaries with status, schedule, and runtime version
* Knowledge base system: curated KB entries covering runtime bugs, env issues, and configuration issues. 

**Improvement**:
* `analyze_canary_failures`: enhance with knowledge base matching and description parameter for targeted issue lookup
*  `audit_services`: automatically detects Synthetics canaries linked to audited services and reports their health

### User experience

**before**:
When a Synthetics canary failed, users got raw error messages and basic pattern matching for a limit error types. To diagnose runtime/env issues, they still had to piece together root cause from scattered log messages, HAR files, and screenshots and reach out to support team for help. Besides, they don't have visibility into canary health when auditing services or service groups. 

**after**:
Users now get automatic knowledge base-powered diagnosis with fix recommendations, and canaries health will be automatically surfaced alongside service metrics in the audit

## Checklist

* [X] Changes are documented

unit tests have been added 
```
--------------------------------------------------------------------------------------------------------------------------
TOTAL                                                                           3377     96   1356    163    94%
================================================================ 612 passed in 11.76s ================================================================
```

**RFC issue number**: N/A

Checklist:

## Acknowledgment


src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/group_tools.py
```diff
@@ -729,6 +729,66 @@ async def audit_group_health(
...
+                    for dep in dep_resp.get('ServiceDependents', []):
+                        dep_attrs = dep.get('DependentKeyAttributes', {})
+                        if dep_attrs.get('ResourceType') == 'AWS::Synthetics::Canary':
+                            canary_names.add(dep_attrs.get('Identifier', ''))
+                except Exception:
+                    pass
+
+            canary_names.discard('')
+            if canary_names:
+                from .aws_clients import synthetics_client as synth_client
```
zhoucy-syn: make import to the top of the file?

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/canary_utils.py
```diff
@@ -890,6 +891,85 @@ async def get_canary_code(canary: dict, region: str = 'us-east-1') -> dict:
         return {'error': f'Canary code analysis failed: {str(e)}'}
 
 
+async def check_canaries_for_service(normalized_targets, unix_start, unix_end, region: str = 'us-east-1'):
```
zhoucy-syn: I think this method is mostly the same logic as the new change in group_tools? can we extract them into a shared method

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/group_tools.py
yig-yi: the logic has been centralized to `check_canaries_for_service` so we don't need this import anymore.

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/canary_utils.py
yig-yi: removed the duplicate logic from group_tools and replaced with this util function

zhoucy-syn (approved):
gltm, but looks like the checks failed. https://github.com/awslabs/mcp/actions/runs/24262539191/job/70849691580?pr=2924 can you take a look

yig-yi:
> gltm, but looks like the checks failed. https://github.com/awslabs/mcp/actions/runs/24262539191/job/70849691580?pr=2924 can you take a look

The PR name exceeded the limit. Updated the pr name and the check passed.

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/server.py
```diff
@@ -1478,12 +1546,131 @@ async def analyze_canary_failures(canary_name: str, region: str = AWS_REGION) ->
...
+
+        canaries = []
+        paginator_token = None
+
+        while True:
+            kwargs = {'MaxResults': 20}
+            if paginator_token:
+                kwargs['NextToken'] = paginator_token
+
+            response = synthetics_client.describe_canaries(**kwargs)
```
syed-ahsan-ishtiaque, line 1632: **Bug: `region` parameter is ignored.**

`synthetics_client` is a module-level singleton bound to `AWS_REGION` at import time. The `region` parameter in the function signature has no effect — calling `list_canaries(region="eu-west-1")` will silently return canaries from the default region.

This should either create a regional client using the passed `region`, or the parameter should be removed to avoid misleading callers.

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/canary_utils.py
```diff
@@ -890,6 +891,85 @@ async def get_canary_code(canary: dict, region: str = 'us-east-1') -> dict:
...
+    try:
+        canary_names = set()
+        for target in normalized_targets:
+            svc = (target.get('Data') or {}).get('Service') or {}
+            key_attrs = {k: v for k, v in svc.items() if v}
+            if not key_attrs.get('Name') or not key_attrs.get('Environment'):
+                continue
+            try:
+                resp = applicationsignals_client.list_service_dependents(
+                    StartTime=datetime.fromtimestamp(unix_start, tz=timezone.utc),
```
syed-ahsan-ishtiaque: **Bug: `region` parameter ignored for canary discovery.**

Both `applicationsignals_client` (here) and `synthetics_client` (used for `get_canary_runs` below) are module-level singletons bound to `AWS_REGION`. The `region` parameter is only passed through to the `analyze_canary_failures` deep-dive call at the bottom of this function.

This means canary **discovery** and **status checks** always happen in the default region, even when `region` is explicitly passed as something else. Only the deep analysis for the worst canary respects it — creating a confusing split where half the function uses the right region and half doesn't.

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/server.py
```diff
@@ -1478,12 +1546,131 @@ async def analyze_canary_failures(canary_name: str, region: str = AWS_REGION) ->
         except Exception as e:
             result += f'Note: Could not retrieve canary code: {str(e)}\n'
 
+        # --- Knowledge Base Recommendations ---
+        try:
+            # Build resource_metrics from telemetry_data if available
+            kb_resource_metrics: dict[str, float] = {}
+            if 'telemetry_data' in dir() and isinstance(telemetry_data, dict) and 'error' not in telemetry_data:
```
syed-ahsan-ishtiaque: **Bug: `'telemetry_data' in dir()` is fragile and non-idiomatic.**

Using `dir()` to check local variable existence is unreliable across Python implementations and hard to follow. The standard pattern is to initialize `telemetry_data = {}` (or `None`) at the top of the try block, then check against the sentinel here.

```python
# At the top of the try block:
telemetry_data = {}
# ...
# Here:
if isinstance(telemetry_data, dict) and 'error' not in telemetry_data:
```

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/canary_recommendation_engine.py
```diff
@@ -0,0 +1,247 @@
...
+                prefix = version[:-1]
+                parts = prefix.rsplit('-', 1)
+                if len(parts) == 2:
+                    family, min_ver = parts[0], parts[1]
+                    canary_parts = context.runtime_version.rsplit('-', 1)
+                    if len(canary_parts) == 2:
+                        canary_family, canary_ver = canary_parts[0], canary_parts[1]
+                        if canary_family == family:
+                            try:
+                                if float(canary_ver) >= float(min_ver):
```
syed-ahsan-ishtiaque, line 237: **Bug: `float()` comparison breaks on multi-dot version strings.**

`float(canary_ver)` works for `"11.0"` vs `"8.0"`, but will raise `ValueError` for versions like `"11.0.1"`. The exception is silently caught by the `except ValueError: pass`, causing a legitimate version match to score 0.0 instead of 1.0.

Consider using `packaging.version.Version` or `tuple(int(x) for x in ver.split('.'))` for robust version comparison.

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/canary_utils.py
```diff
@@ -890,6 +891,85 @@ async def get_canary_code(canary: dict, region: str = 'us-east-1') -> dict:
...
+                    EndTime=datetime.fromtimestamp(unix_end, tz=timezone.utc),
+                    KeyAttributes=key_attrs,
+                    MaxResults=100,
+                )
```
syed-ahsan-ishtiaque: **Bug: `list_service_dependents` response is not paginated.**

This call uses `MaxResults=100` but never checks for or follows `NextToken`. Services with more than 100 dependents will silently miss canaries.

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/server.py
```diff
@@ -1478,12 +1546,131 @@ async def analyze_canary_failures(canary_name: str, region: str = AWS_REGION) ->
...
+            for msg in all_collected_error_messages:
+                if msg and msg not in kb_error_messages:
+                    kb_error_messages.append(msg)
+
+            failure_context = FailureContext(
+                error_messages=kb_error_messages,
+                state_reasons=[selected_reason],
+                runtime_version=canary.get('RuntimeVersion', ''),
+                log_patterns=all_collected_log_patterns,
+                resource_metrics=kb_resource_metrics,
```
syed-ahsan-ishtiaque: **Bug: `resource_metrics` is collected but never consumed.**

`kb_resource_metrics` is carefully extracted from `telemetry_data` and passed into `FailureContext.resource_metrics`, but no scoring method in `CanaryRecommendationEngine` ever reads this field. The `_score_error_patterns`, `_score_symptoms`, `_score_runtime_version`, and `_score_environment` methods don't reference it.

This is dead code that creates a false impression of functionality. Either implement a scorer that uses it, or remove the collection to reduce complexity.

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/server.py
```diff
@@ -978,7 +987,7 @@ async def audit_service_operations(
 
 
 @mcp.tool()
-async def analyze_canary_failures(canary_name: str, region: str = AWS_REGION) -> str:
+async def analyze_canary_failures(canary_name: str, region: str = AWS_REGION, description: str = '') -> str:
```
syed-ahsan-ishtiaque: **Security: No input length validation on `description`.**

The `description` string is passed directly into regex matching (`re.search(pattern.regex, msg, re.IGNORECASE)`) with no length cap. Some KB patterns use `.*` constructs (e.g., `page\\.goto:.*Timeout.*exceeded`, `limit.*increase.*\\d+`) which could exhibit slow matching with very long inputs.

Consider capping `description` to a reasonable length (e.g., 1000 chars) before it enters the matching pipeline.

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/canary_utils.py
```diff
@@ -890,6 +891,85 @@ async def get_canary_code(canary: dict, region: str = 'us-east-1') -> dict:
...
+            result += f'✅ {len(healthy)} healthy canaries: {", ".join(n for n, _, _, _ in healthy)}\n'
+
+        # Deep analysis for worst canary (0% or lowest success)
+        if failing:
+            worst = min(failing, key=lambda x: x[2])
+            worst_name = worst[0]
+            result += f'\n🔍 Deep analysis for worst canary ({worst_name}):\n'
+            try:
+                deep = await analyze_canary_failures(worst_name, region)
+                # Extract just the key findings (first 500 chars)
```
syed-ahsan-ishtiaque: **Code Quality + Token Efficiency: Full `analyze_canary_failures` called for ~150 tokens of output.**

This triggers a *complete* deep analysis — S3 artifact analysis, IAM role checks, CloudWatch Logs queries, canary code retrieval, and now KB matching — then truncates the entire result to 600 characters (~150 tokens). That's multiple AWS API calls, S3 reads, and log analysis for a truncated snippet.

Additionally, the hard `[:600]` truncation can cut mid-word or mid-sentence, producing garbled output.

Consider either:
1. A lightweight summary function that only fetches the essentials (last run status + state reason), or
2. At minimum truncating at a sentence/line boundary instead of a hard character cutoff.

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/canary_knowledge_base_loader.py
```diff
@@ -0,0 +1,84 @@
...
+                if entry is not None:
+                    self._entries[entry.id] = entry
+
+        logger.info(f'Knowledge base loaded: {len(self._entries)} entries')
+        self._loaded = True
+
+    def _parse_and_validate(self, file_path: Path) -> Optional[KBEntry]:
+        """Parse a single JSON file, validate against schema, return KBEntry or None."""
+        try:
+            with open(file_path, 'r') as f:
```
syed-ahsan-ishtiaque, line 71: **Code Quality: Blocking synchronous I/O in an async server.**

`open()` + `json.load()` blocks the asyncio event loop while reading and parsing JSON files. This is called for all 16 KB files on first use.

In practice the impact is small (one-time lazy load of small files), but in an async codebase this should ideally use `aiofiles` or `asyncio.to_thread()` to avoid blocking concurrent request handling.

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/canary_utils.py
```diff
@@ -890,6 +891,85 @@ async def get_canary_code(canary: dict, region: str = 'us-east-1') -> dict:
...
+                if not runs:
+                    canary_statuses.append((name, 'no_data', 0, 0))
+                    continue
+                passed = sum(1 for r in runs if r.get('Status', {}).get('State') == 'PASSED')
+                failed = sum(1 for r in runs if r.get('Status', {}).get('State') == 'FAILED')
+                total = len(runs)
+                success_pct = (passed / total * 100) if total > 0 else 0
+                canary_statuses.append((name, 'ok' if success_pct >= 80 else 'failing', success_pct, failed))
+            except Exception:
+                canary_statuses.append((name, 'error', 0, 0))
```
syed-ahsan-ishtiaque, line 950: **Code Quality: Silent exception swallowing.**

This bare `except Exception:` with no logging makes it impossible to diagnose failures when a canary status check fails. Other error handlers in this same function use `logger.warning(...)`. This should be consistent:

```python
except Exception as e:
    logger.warning(f'Failed to get runs for canary {name}: {e}')
    canary_statuses.append((name, 'error', 0, 0))
```

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/server.py
```diff
@@ -1478,12 +1546,131 @@ async def analyze_canary_failures(canary_name: str, region: str = AWS_REGION) ->
...
+            paginator_token = response.get('NextToken')
+            if not paginator_token:
+                break
+
+        if not canaries:
+            return 'No canaries found in this account/region.'
+
+        result = f'Found {len(canaries)} canaries:\n\n'
+
+        for c in canaries:
```
syed-ahsan-ishtiaque: **Token/Context: `list_canaries` has no output pagination or truncation.**

This loop accumulates output for every canary in the account into a single MCP tool response string. Each canary produces ~20-26 tokens of formatted output. At the default account limit of 500 canaries, this is ~13,000 tokens in a single tool response that lands in the LLM's context window.

Accounts with limit increases (up to 1000+ canaries) would be even worse.

Since this is an MCP tool whose output goes directly into an LLM's context window, consider:
1. Adding a `max_results` parameter (defaulting to e.g. 50) with a note about using pagination for more
2. Or returning a summary count + top-N with a prompt to narrow the query

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/canary_recommendation_engine.py
```diff
@@ -0,0 +1,247 @@
...
+                if rec.title:
+                    lines.append(f'   Recommendation: {rec.title}')
+                if rec.problem:
+                    lines.append(f'     Problem: {rec.problem[:200]}')
+                for step in rec.solution:
+                    lines.append(f'   - {step.step}')
+                    if step.description:
+                        lines.append(f'     {step.description}')
+                    if step.command:
+                        lines.append(f'     Command: {step.command}')
```
syed-ahsan-ishtiaque, line 334: **Token/Context: `command` fields are emitted in full with no truncation.**

`root_cause` is truncated to 300 chars and `problem` to 200 chars, but `step.command` is unlimited. Some KB entries have very long commands — e.g., RUNTIME-002 includes inline CloudFormation YAML (~500 chars in a single command field).

With up to 5 matches x multiple recommendations x multiple steps, untruncated commands can add 3,000+ characters to the tool response. Since this all lands in the LLM's context window, consider either:
1. Truncating commands with a `... (see docs)` suffix, or
2. Only including the first command/step and linking to docs for the full remediation

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/server.py
```diff
@@ -1088,6 +1102,22 @@ async def analyze_canary_failures(canary_name: str, region: str = AWS_REGION) ->
         if not unique_reasons:
             result += '✅ No consecutive failures to analyze\n'
             result += '💡 Canary appears to be recovering or healthy\n'
+
+            # Still run KB lookup when user provided an issue description
+            if description:
```
syed-ahsan-ishtiaque, line 1111: **Token/Context: Healthy canary path can produce unexpectedly heavy output.**

When a canary is healthy but the user provides a `description`, this block runs the full KB matching pipeline and appends `format_recommendations()` output. A single matching entry can add 600-700 tokens, and 5 matches can add ~3,000 tokens.

Callers expecting a lightweight "canary is fine" response may be surprised by the token impact. Consider either:
1. Limiting the healthy path to returning just KB entry titles + confidence as a summary (not full remediation steps), or
2. Documenting this behavior in the tool docstring so callers know that `description` on a healthy canary can trigger a heavy response

wangzlei (dismissed):

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/canary_recommendation_engine.py
yig-yi: The commands are unlike the root_cause/problem so they are intentionally un-truncated. The AI agent is supposed to present the complete fix to the customer. It's not that helpful for the agent to return an incomplete remediation that the customer can't act on. 

For the RUNTIME-002 mentioned in the comment, it takes a little bit more tokens but not that bad. It's still under 1k token.

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/canary_utils.py
yig-yi: The region is already being passed to clients https://github.com/awslabs/mcp/blob/main/src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/aws_clients.py#L82 

We don't really need it at each API level. Code reference: https://github.com/awslabs/mcp/blob/8f8460d377c692b27764daf20dfc3510706f98ca/src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/trace_tools.py#L649

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/server.py
yig-yi: `synthetics_client` is a regional client using the passed `region` https://github.com/awslabs/mcp/blob/main/src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/aws_clients.py#L93

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/canary_recommendation_engine.py
yig-yi: Synthetics runtime versions format is like `syn-language -framework-majorversion. minorversion`. The multi-dot scenario doesn't apply here.

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/canary_utils.py
yig-yi: Thought more about this. Gonna remove the deep analysis for the canary here and let customers decide if they want to call `analyze_canary_failures` for detailed analysis of any canary.

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/server.py
yig-yi: Yeah, this is the expected improvement. Healthy canaries can also have issues that customer wants to investigate. For example, the issues in RUNTIME-006 that sensitive headers visible in HTTPSteps logs and reports because the configuration is being ignored.

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/server.py
yig-yi: good point. Added the maximum number for the total canaries to list to avoid overwhelming context windows

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/server.py
yig-yi: good point. Added the description cap considering customer don't usually put that long description.

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/canary_knowledge_base_loader.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+class CanaryKnowledgeBaseLoader:
+    """Singleton loader that reads JSON KB entries once and caches them."""
+
+    _instance: Optional[CanaryKnowledgeBaseLoader] = None
+
+    @classmethod
+    async def get_instance(cls) -> CanaryKnowledgeBaseLoader:
+        """Return the singleton instance, loading KB entries off the event loop on first call."""
```
syed-ahsan-ishtiaque, line 35: **Bug: Race condition in async `get_instance()` — concurrent callers can get an unloaded instance.**

The `await` on the next line yields control to the event loop. Consider this interleaving:

1. Coroutine A checks `_instance is None` → True
2. Coroutine A sets `_instance = cls()` (unloaded, empty `_entries`)
3. Coroutine A hits `await asyncio.to_thread(...)` → yields
4. Coroutine B checks `_instance is None` → **False** (set in step 2)
5. Coroutine B returns `_instance` — **which hasn't finished loading**
6. Coroutine B's recommendation engine gets zero matches (empty `_entries`)

This silently produces no KB recommendations for the unlucky concurrent caller. No crash, just missing results.

**Fix** — assign `_instance` only after load completes:

```python
if cls._instance is None:
    inst = cls()
    await asyncio.to_thread(inst.load)
    cls._instance = inst
```

Two coroutines could still both load (theoretical double-load), but that's harmless — idempotent file reads, second write wins with the same data. The critical thing is no caller ever sees a half-loaded instance.

src/cloudwatch-applicationsignals-mcp-server/tests/test_canary_recommendation_engine.py
```diff
@@ -0,0 +1,429 @@
...
+class TestScoreEnvironment:
+    """Tests for _score_environment."""
+
+    def test_overlap(self):
+        entry = _make_entry(tags=['timeout', 'puppeteer'], category='test')
+        ctx = _make_context(environment_indicators=['timeout', 'lambda'])
+        engine = CanaryRecommendationEngine(MagicMock())
+        score = engine._score_environment(entry, ctx)
+        assert score > 0.0
+
```
syed-ahsan-ishtiaque, line 296: **Nit: Test name and assertion are stale after the top-5 → top-2 change.**

`get_recommendations` now returns `results[:2]`, but this test is still named `test_max_five_results` and asserts `<= 5`. The assertion trivially passes for a top-2 cap but doesn't verify the actual limit.

```python
def test_max_two_results(self):
    ...
    assert len(results) <= 2
```

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/canary_knowledge_base_loader.py
yig-yi: I don't think the concurrent caller race condition applies. The first [get_instance() call](https://github.com/yig-yi/mcp/blob/309dbdc70917d7a04dcffafc535d6d6c110d639c/src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/server.py#L1116) is inside the `if not unique_reasons:` branch which hits return result right after. The [second call](https://github.com/yig-yi/mcp/blob/309dbdc70917d7a04dcffafc535d6d6c110d639c/src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/server.py#L1579) handles the case where there are failure reasons. These two paths are exclusive in the single invocation

src/cloudwatch-applicationsignals-mcp-server/tests/test_canary_recommendation_engine.py
yig-yi: good catch. Updated the test in latest commit. 
```
TOTAL                                                                               3825    121   1548    181    94%
=========================================== 705 passed in 8.41s ============================================
```

wangzlei (dismissed):

syed-ahsan-ishtiaque (dismissed):

src/cloudwatch-applicationsignals-mcp-server/tests/test_canary_recommendation_engine.py
```diff
@@ -0,0 +1,408 @@
...
+            confidence_score=0.8,
+            error_pattern_score=1.0,
+            symptom_score=0.5,
+            runtime_version_score=1.0,
+            environment_score=0.0,
+        )
+        engine = CanaryRecommendationEngine(MagicMock())
+        output = engine.format_recommendations([match])
+        assert 'AWS Docs' in output
+        assert 'https://docs.aws.amazon.com' in output
```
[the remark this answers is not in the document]
yig-yi: updated the test to fix the issue

syed-ahsan-ishtiaque (approved):

wangzlei (approved):

PR #2545, opened 2026-03-02
fix(openapi-mcp-server): add upper bound to fastmcp dependency

abhu85:
## Summary

- Add `<3.0.0` upper bound to the fastmcp dependency to prevent installation of fastmcp 3.x which introduced breaking changes

## Problem

All published versions of `awslabs.openapi-mcp-server` fail to start when installed via `uvx` because the package has no upper bound on its `fastmcp` dependency. `uvx` resolves `fastmcp` to its latest version (currently 3.0.2), which introduced breaking changes that removed `RouteType`, `FastMCPOpenAPI`, and `get_prompts()` - all of which the package depends on.

## Root Cause

The package declared `fastmcp>=2.14.0` with no upper version bound. fastmcp v3.0.0 was a major breaking release that removed APIs the openapi-mcp-server relies on.

## Solution

Add an upper bound `<3.0.0` to the fastmcp dependency constraint, changing from:
```
fastmcp>=2.14.0
```
to:
```
fastmcp>=2.14.0,<3.0.0
```

This implements the "SHORT TERM" solution suggested in #2533.

## Test Plan

- [x] Verified fastmcp 2.14.5 is installed with the new constraint (not 3.x)
- [x] Ran all 410 tests in openapi-mcp-server - all pass
- [x] Confirmed the fix matches the workaround that users have validated works

```bash
cd src/openapi-mcp-server
python3 -m venv .venv
source .venv/bin/activate
pip install -e ".[test]"
pip show fastmcp | grep Version
# Output: Version: 2.14.5

pytest tests/ -v
# Output: 410 passed
```

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the [project license](https://github.com/awslabs/mcp/blob/main/LICENSE).

Fixes #2533

PR #2941, opened 2026-04-09
fix(openapi): pin fastmcp <3.0.0 to prevent startup crash

### LINKED ISSUE #2533: OpenAPI MCP Server: v0.2.13 crashes with AttributeError: 'FastMCPOpenAPI' object has no attribute 'get_prompts'
### Describe the bug

Running `awslabs.openapi-mcp-server` v0.2.13 (latest) via `uvx` causes an immediate
crash due to a deprecated `FastMCPOpenAPI` class that is incompatible with the current
version of `fastmcp`. The server shuts down before accepting any connections, making it
completely non-functional. In MCP clients such as Kilo Code it fails silently with
`MCP error -32000: Connection closed`.

### Expected Behavior

The server should start successfully and expose API endpoints as MCP tools, allowing
LLM clients to interact with the specified OpenAPI specification.

### Current Behavior

The server crashes immediately on startup with:

    AttributeError: 'FastMCPOpenAPI' object has no attribute 'get_prompts'

It also emits the following deprecation warnings before crashing:

    DeprecationWarning: fastmcp.server.openapi is deprecated. Import from fastmcp.server.providers.openapi instead.

    DeprecationWarning: FastMCPOpenAPI is deprecated. Use FastMCP with OpenAPIProvider instead:
        provider = OpenAPIProvider(openapi_spec=spec, client=client)
        mcp = FastMCP('name', providers=[provider])

    ERROR | awslabs.openapi_mcp_server.server:create_mcp_server_async:313 - Failed to list tools: 'OpenAPITool' object has no attribute 'get'

    ERROR | awslabs.openapi_mcp_server.server:main:524 - Server shutting down due to error in tool/resource registration.

### Reproduction Steps

1. Install uv/uvx (https://astral.sh/uv)
2. Run the following command:

    uvx awslabs.openapi-mcp-server@latest \
      --api-name petstore \
      --api-url https://petstore3.swagger.io/api/v3 \
      --spec-url https://petstore3.swagger.io/api/v3/openapi.json \
      --log-level DEBUG

3. Observe the server crash with the AttributeError above.

### Possible Solution

Migrate from the deprecated `FastMCPOpenAPI` class to the new pattern as suggested
in the deprecation warning:

    provider = OpenAPIProvider(openapi_spec=spec, client=client)
    mcp = FastMCP('name', providers=[provider])

- `awslabs.openapi-mcp-server` v0.2.13 is the latest published version.
- The issue is caused by the internal use of the deprecated `FastMCPOpenAPI` class
  which no longer has a `get_prompts()` method in the current version of `fastmcp`.
- Tested using the public Petstore demo API which requires no authentication,
  so this is not an auth or network issue.
- Python version: 3.11.11 (Clang 19.1.6)
- httpx version: 0.28.1
- MCP client tested: Kilo Code (VS Code extension)

### OS

MacOS

### Server

other

### Server Version

_No response_

/A - this issue is not AWS region specific. The server crashes locally before making any AWS or external API calls.

### Other information

_No response_

### Service quota

Gahrcoder:
## Problem

The openapi-mcp-server crashes on startup for all new installations because `pyproject.toml` specifies `fastmcp>=2.14.0` with no upper bound. When `uvx` resolves dependencies, it installs fastmcp 3.x which removed several APIs the server depends on:

- `RouteType` (removed in 3.0)
- `FastMCPOpenAPI` (removed in 3.0)
- `get_prompts()` (removed in 3.0)

This means **every new user** who installs the openapi-mcp-server gets a crash on startup.

## Fix

Pin `fastmcp>=2.14.0,<3.0.0` to prevent the breaking 3.x release from being installed.

## Related

Fixes #2533. PR #2687 bumps the fastmcp version but does not add an upper bound, so it does not actually fix the root cause.

### DIFF
--- a/src/openapi-mcp-server/pyproject.toml
+++ b/src/openapi-mcp-server/pyproject.toml
@@ -28,7 +28,7 @@ classifiers = [
     "Programming Language :: Python :: 3.13",
 ]
 dependencies = [
-    "fastmcp>=2.14.0",
+    "fastmcp>=2.14.0,<3.0.0",
     "httpx>=0.28.1",
     "pydantic>=2.11.7",
     "typing-extensions>=4.14.1",


prajwalendra:
@Gahrcoder , please follow this existing PR to address the same issue. https://github.com/awslabs/mcp/pull/2545

@scottschreckengaust , @krokoko

Gahrcoder:
Thanks for the pointer\! Looks like PR #2545 addresses the same root cause with the same fix. Happy to close this one in favor of that approach.

PR #2946, opened 2026-04-09
feat(agentcore): Add operational Memory tools sub-package

sundargthb:
## Summary

Replaces the static `tools/memory.py` guide with a full operational sub-package (`tools/memory/`) providing 21 MCP tools that make real boto3 API calls to Amazon Bedrock AgentCore Memory control plane and data plane services.

### Changes

The Memory primitive was previously a single static guide tool (`manage_agentcore_memory`) that returned a text blob. Developers using the MCP server from their IDE had no way to actually interact with Memory resources — they had to switch to the AWS console, CLI, or write SDK code manually.

This PR brings Memory to parity with Browser (25 tools) and Code Interpreter (9 tools) by providing operational tools that AI coding agents can call directly.

### Source files (`tools/memory/`)

|File              |Description                                                                                           |
|------------------|------------------------------------------------------------------------------------------------------|
|`__init__.py`     |`register_memory_tools(mcp)` entry point with cost tier docs                                          |
|`memory_client.py`|Dual cached boto3 clients: `bedrock-agentcore-control` + `bedrock-agentcore`, with user-agent tracking|
|`models.py`       |15 Pydantic response models for all tool returns                                                      |
|`error_handler.py`|`handle_memory_error()` — ClientError → ErrorResponse                                                 |
|`controlplane.py` |5 tools: create, get, update, delete, list memory resources                                           |
|`events.py`       |6 tools: create/get/delete/list events, list actors, list sessions                                    |
|`records.py`      |7 tools: get/delete/list records, semantic search, batch create/update/delete                         |
|`extraction.py`   |2 tools: list and start extraction jobs                                                               |
|`guide.py`        |1 tool: comprehensive static reference guide                                                          |

### Test files (`tests/memory/`)

|File                        |Coverage                                                          |
|----------------------------|------------------------------------------------------------------|
|`conftest.py`               |Shared fixtures: `mock_ctx`, `mock_boto3_client`, `client_factory`|
|`test_unit_memory_client.py`|Client caching, region resolution, user-agent verification        |
|`test_unit_controlplane.py` |All 5 control plane tools                                         |
|`test_unit_events.py`       |All 6 event/actor/session tools                                   |
|`test_unit_records.py`      |All 7 record tools including semantic search and batch ops        |
|`test_unit_extraction.py`   |Both extraction tools                                             |
|`test_unit_guide.py`        |Guide content assertions (CLI, schema, IAM, cost tiers)           |

### Modified existing files

|File                                      |Change                                                                                                                                                          |
|------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------|
|`server.py`                               |Replaced `mcp.tool()(memory.manage_agentcore_memory)` with `register_memory_tools(mcp)` in try/except block; removed `memory` from flat import                  |
|`tests/test_tools.py`                     |`TestMemoryTool` now tests `GuideTools` class and `MEMORY_GUIDE` constant instead of deleted flat module                                                        |
|`tests/browser/test_integ_mcp_protocol.py`|Updated `BASE_TOOLS` to include 21 memory tool names; `_build_server()` calls `register_memory_tools()`; added `test_memory_tools_require_memory_id` schema test|

### Deleted files

|File             |Reason                                                                             |
|-----------------|-----------------------------------------------------------------------------------|
|`tools/memory.py`|Replaced by `tools/memory/` sub-package. Cannot coexist with the package directory.|

### Why this matters

We recently had an incident where a developer installed the MCP server for documentation access, and an AI agent autonomously discovered Code Interpreter tools and started spinning up billable sessions without the developer realizing — resulting in surprise AWS billing alerts.

Memory tools have similar cost implications. An AI agent could autonomously create Memory resources, store events (triggering background extraction), or run semantic searches — all of which incur AWS charges.

### Safeguards implemented

**1. Tool descriptions include cost warnings.** Every tool that creates billable resources or invokes compute has an explicit `COST WARNING` in its docstring. AI agents read these docstrings before calling tools, so they understand the implications:

```
COST WARNING: Creating events triggers background long-term memory
extraction if strategies are configured. This consumes compute
resources and incurs AWS charges.
```

**2. Tools are grouped by cost tier in the guide and `__init__.py` docstring:**

|Tier                          |Tools                                                                                                                                                                                         |Risk                                         |
|------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|---------------------------------------------|
|**Read-only (no cost)**       |memory_get, memory_list, memory_get_event, memory_list_events, memory_list_actors, memory_list_sessions, memory_get_record, memory_list_records, memory_list_extraction_jobs, get_memory_guide|Safe to call freely                          |
|**Billable (incur charges)**  |memory_create, memory_update, memory_create_event, memory_retrieve_records, memory_batch_create_records, memory_batch_update_records, memory_start_extraction_job                             |Agent should confirm with user before calling|
|**Destructive (irreversible)**|memory_delete, memory_delete_event, memory_delete_record, memory_batch_delete_records                                                                                                         |Permanent data loss, no undo                 |

**3. Opt-out via `_is_service_enabled()`.** Memory tools are registered behind the existing `AGENTCORE_DISABLE_TOOLS=memory` / `AGENTCORE_ENABLE_TOOLS` mechanism, same as Browser and Code Interpreter. Developers who only want documentation tools can disable memory entirely.

**4. `__init__.py` docstring documents cost tiers.** This is visible during code review and makes it clear which tools are billable:

```python
"""Memory tools sub-package for the unified AgentCore MCP server.

Tools that create billable resources or incur compute costs:
- memory_create: Provisions Memory infrastructure (AWS charges)
- memory_create_event: Triggers background extraction (compute charges)
- memory_retrieve_records: Semantic search (embedding charges)
...

Read-only tools (no cost):
- memory_get, memory_list, ...
"""
```

## User-Agent Tracking

Both boto3 clients include a `user_agent_extra` string: `agentcore-mcp-server/0.1.0 memory`. This lets AWS service-side logs distinguish API calls from the MCP server vs CLI vs SDK vs console, broken down by primitive (`memory` vs `browser` vs `code_interpreter`). No phone-home telemetry — just standard HTTP User-Agent headers that AWS already captures.

## Architecture

Follows the established sub-package pattern (Pattern A) from Browser and Code Interpreter:

- Tool group classes with `__init__(client_factory)` and `register(mcp)` methods
- Per-region cached boto3 clients (separate caches for control plane and data plane)
- `Annotated[type, Field(description='...')]` parameters on all tools
- Pydantic `BaseModel` responses (never raw dicts)
- Structured error handling via `handle_memory_error()` (never raw exceptions to MCP client)
- All tool methods are `async def` with `ctx: Context` as first parameter

## Testing

```bash
# Unit tests (mocked boto3, no AWS credentials needed)
uv run --frozen pytest tests/memory/ -v --cov --cov-branch --cov-report=term-missing

# Full test suite
uv run --frozen pytest --cov --cov-branch --cov-report=term-missing

# Lint
uv run ruff check awslabs/amazon_bedrock_agentcore_mcp_server/tools/memory/ tests/memory/

# Type check
uv run pyright

# MCP Inspector (interactive verification)
npx @modelcontextprotocol/inspector uv --directory <path-to-server> run server.py
```

All tests use mocked boto3 — no real API calls, no AWS charges during CI.

## Checklist

- [x] Apache 2.0 license headers on every file
- [x] ruff clean (99 char lines, single quotes, Google docstrings, D102 compliance)
- [x] pyright clean (isinstance narrowing for union return types)
- [x] All tools return Pydantic BaseModel responses
- [x] All tools handle errors via error_handler (never raw exceptions)
- [x] Billable tools have COST WARNING in docstrings
- [x] Guide covers CLI commands, agentcore.json schema, IAM permissions, cost tiers, troubleshooting, migration notes
- [x] User-agent tracking on both boto3 clients
- [x] `_is_service_enabled('memory')` gating with ImportError fallback
- [x] Old `tools/memory.py` deleted
- [x] `server.py`, `test_tools.py`, `test_integ_mcp_protocol.py` updated
- [x] MCP Inspector verified — tool discovery and guide invocation work end-to-end

## Checklist

* [x ] I have reviewed the [contributing guidelines](https://github.com/awslabs/mcp/blob/main/CONTRIBUTING.md)
* [x ] I have performed a self-review of this change
* [x ] Changes have been tested
* [x ] Changes are documented

Is this a breaking change? N

**RFC issue number**:

Checklist:

## Acknowledgment


src/amazon-bedrock-agentcore-mcp-server/awslabs/amazon_bedrock_agentcore_mcp_server/tools/memory/guide.py
```diff
@@ -0,0 +1,367 @@
...
+- AWS credentials configured (AWS_PROFILE, AWS_ACCESS_KEY_ID, or IAM role)
+- AWS_REGION environment variable (defaults to us-east-1)
+- No additional installation — tools use boto3 bundled with the MCP server
+
+### For CLI commands referenced in this guide
+The `agentcore` CLI is a separate tool for project scaffolding, deployment,
+and management. Install it before using any `agentcore` commands:
+
+```bash
+npm install -g @anthropic-ai/agentcore-cli
```
jesseturner21, line 49: nit: wrong install should be @aws/agentcore-cli

src/amazon-bedrock-agentcore-mcp-server/awslabs/amazon_bedrock_agentcore_mcp_server/tools/memory/guide.py
```diff
@@ -0,0 +1,367 @@
...
+AgentCore Memory gives AI agents the ability to remember past interactions.
+It supports **short-term memory** (turn-by-turn events within a session) and
+**long-term memory** (extracted insights persisted across sessions via
+strategies like semantic, summarization, user preference, and episodic).
+
+---
+
+## Prerequisites
+
+### For MCP tools (this sub-package)
```
jesseturner21, line 35: Similar to [runtime pr](https://github.com/awslabs/mcp/pull/2947#pullrequestreview-4101901971) Just adding for other reviewers

This is just a thought but can we could put this guide in the [agentcore-cli](https://github.com/aws/agentcore-cli) repo and then have this tool pull it in.

I think this guide would be a helpful reference for devs too.
We only have one place manage.
This could be something we explore!

Curious what your thoughts are?

jesseturner21 (changes requested):
Only change needed is install command fix

src/amazon-bedrock-agentcore-mcp-server/awslabs/amazon_bedrock_agentcore_mcp_server/tools/memory/guide.py
sundargthb: Thanks for the catch.. fixed it

src/amazon-bedrock-agentcore-mcp-server/awslabs/amazon_bedrock_agentcore_mcp_server/tools/memory/guide.py
sundargthb: i like the idea.. single source of truth for the guide content and devs can reference it directly without needing the MCP server. For this PR I’d prefer to keep it as the inline static string since that’s what Browser does today, and it keeps things working offline with no network dependency. But I think we should explore the fetch from CLI repo approach as a follow-up — ideally for all primitives at once so they all work the same way (after cli is updated with all). Can we file an issue to track it?​​​​​​​​​​​​​​​​

jesseturner21 (dismissed):

kevin-orellana (commented):
## Review — Memory Tools Sub-package

Overall this is well-structured and follows the Browser sub-package pattern closely. Memory correctly uses `is not None` for all optional parameter checks (Runtime PR doesn't — see that review). A few items to address:

---

### 1. PII in logs — `records.py:213`

```python
logger.info(f'Retrieving memory records from {memory_id}: "{search_query[:50]}..."')
```

This logs the first 50 chars of the user's semantic search query at INFO level. Memory search queries are likely to contain PII or sensitive context (e.g., "What did John say about the contract?"). Every other tool in this PR logs only resource IDs, which is the safer pattern.

**Suggestion:** Log only `memory_id` and `namespace`, or drop to `logger.debug`.

---

### 2. `event_expiry_duration` range conflict

- `controlplane.py` Field description says **3–365**
- `guide.py` schema section says **7–365**

These should match. Please verify against the actual API and align both.

---

### 3. `clear_clients()` is dead code — `memory_client.py`

`clear_clients()` exists but is never called — not in `server_lifespan`, not in `register_memory_tools`, nowhere. Either wire it into server shutdown or remove it. Dead code that looks like it should be called is worse than no code.

---

### 4. Docstring says "22 tools", actual count is 21

`__init__.py` line 17: `Provides 22 memory tools` — but `test_unit_init.py` asserts 21, `BASE_TOOLS` in the integration test lists 21, and counting the registered tools in the code gives 21. Fix the docstring.

---

### 5. Cross-PR merge conflict with Runtime PR #2947

Both PRs modify `server.py`, `test_integ_mcp_protocol.py`, `test_tools.py`, and both add `utils/user_agent.py` (identical content). The Runtime PR reverts your memory registration back to the old single-tool `manage_agentcore_memory` pattern. Whoever merges second will have conflicts.

**Suggestion:** Coordinate merge order. This PR is cleaner — recommend merging this first, then rebasing #2947 on top.

sundargthb:
@kevin-orellana 

1. PII in logs — records.py:213
Fixed — removed the search query from the log. Now logs only memory_id and namespace, consistent with every other tool in the sub-package.

2. event_expiry_duration range conflict
Good catch. API accepts 3–365, CLI enforces 7–365. Updated the guide’s schema section to 3–365 since our tools call the API directly. CLI --expiry description still says 7–365 (correct for CLI context).

3. clear_clients() dead code
Removed. boto3 clients don’t hold open connections so there’s no shutdown cleanup to do — unlike Code Interpreter which has active sessions to stop.

4. Docstring says 22, actual is 21
Fixed the docstring in __init__.py.

5. Cross-PR merge conflict with Runtime
Agreed — let’s merge this one first, then rebase the Runtime PR on top. utils/user_agent.py is identical in both so that file won’t conflict. I’ll rebase Runtime after this lands.​​​​​​​​​​​​​​​​

jesseturner21 (dismissed):
lgtm

kevin-orellana (commented):
Two minor findings:

**1. `filter` parameter shadows Python builtin** (`events.py`, `extraction.py`)

`memory_list_events` and `memory_list_extraction_jobs` both use `filter` as a parameter name, which shadows the Python builtin `filter()`. Consider renaming to `event_filter` / `extraction_filter` or `filter_criteria`.

**2. `get_memory_guide` return type inconsistent with other tools** (`guide.py`)

Every other tool in the sub-package returns a Pydantic model (`MemoryResponse`, `EventResponse`, `ErrorResponse`, etc.), but `get_memory_guide` returns `dict[str, str]`. Consider wrapping in a Pydantic model for consistency with the rest of the package.

Otherwise looks good — clean pattern adherence, solid test coverage, and the cost tier documentation is well done.

kevin-orellana (dismissed):
Approved

jesseturner21 (approved):

kevin-orellana (approved):

src/amazon-bedrock-agentcore-mcp-server/awslabs/amazon_bedrock_agentcore_mcp_server/tools/memory/guide.py
```diff
@@ -0,0 +1,383 @@
...
+### For CLI commands referenced in this guide
+The `agentcore` CLI is a separate tool for project scaffolding, deployment,
+and management. Install it before using any `agentcore` commands:
+
+```bash
+<<<<<<< HEAD
+<<<<<<< HEAD
+npm install -g @aws/agentcore-cli
+=======
+npm install -g @anthropic-ai/agentcore-cli
```
spencercorwin, line 49: FYI @sundargthb looks like this needs to be fixed on main :)

PR #2947, opened 2026-04-09
feat(agentcore): Add operational Runtime tools sub-package

sundargthb:
## Summary

Replaces the single-file static guide (`tools/runtime.py`) with a full operational sub-package (`tools/runtime/`) providing 14 MCP tools that make real boto3 API calls against AgentCore Runtime. Follows the established Browser and Code Interpreter sub-package patterns.

The old `runtime.py` was a static `manage_agentcore_runtime()` function returning a deployment guide string referencing the deprecated `bedrock-agentcore-starter-toolkit` pip CLI. The new sub-package provides real CRUD operations, agent invocation, session management, and an updated guide referencing the current `@aws/agentcore` npm CLI.

### Changes

|Action|Path                                         |Notes                                                      |
|------|---------------------------------------------|-----------------------------------------------------------|
|DELETE|`awslabs/.../tools/runtime.py`               |Old static guide (deprecated CLI refs)                     |
|ADD   |`awslabs/.../tools/runtime/__init__.py`      |Registration entry point with cost classification          |
|ADD   |`awslabs/.../tools/runtime/runtime_client.py`|Cached boto3 factories (control + data plane)              |
|ADD   |`awslabs/.../tools/runtime/models.py`        |14 Pydantic response models                                |
|ADD   |`awslabs/.../tools/runtime/error_handler.py` |ClientError → ErrorResponse mapping                        |
|ADD   |`awslabs/.../tools/runtime/lifecycle.py`     |Create/Get/Update/Delete/List runtimes + versions          |
|ADD   |`awslabs/.../tools/runtime/endpoints.py`     |Endpoint CRUD + list                                       |
|ADD   |`awslabs/.../tools/runtime/invocation.py`    |Invoke agent + stop session                                |
|ADD   |`awslabs/.../tools/runtime/guide.py`         |Static reference guide (new CLI, cost tiers, IAM)          |
|ADD   |`tests/runtime/__init__.py`                  |Package marker                                             |
|ADD   |`tests/runtime/conftest.py`                  |Shared fixtures                                            |
|ADD   |`tests/runtime/test_unit_runtime_client.py`  |Client caching + user-agent tests                          |
|ADD   |`tests/runtime/test_unit_lifecycle.py`       |Lifecycle tool tests                                       |
|ADD   |`tests/runtime/test_unit_endpoints.py`       |Endpoint tool tests                                        |
|ADD   |`tests/runtime/test_unit_invocation.py`      |Invoke + stop session tests                                |
|ADD   |`tests/runtime/test_unit_guide.py`           |Guide content assertion tests                              |
|MODIFY|`awslabs/.../server.py`                      |Replace runtime registration block                         |
|MODIFY|`tests/test_tools.py`                        |Remove old `TestRuntimeTool` class                         |
|MODIFY|`tests/browser/test_integ_mcp_protocol.py`   |Update `BASE_TOOLS` set + `_build_server()` for sub-package|

### Background

We recently had an incident where a developer installed the MCP server for documentation search, and an AI coding agent autonomously discovered the Code Interpreter tools and started spinning up billable sandbox sessions without the developer realizing — resulting in surprise AWS billing alerts.

### Safeguards applied to Runtime tools

**1. Tool descriptions include cost implications.** Every tool that creates billable resources has an explicit cost warning in its docstring — the text the LLM reads before deciding to call the tool:

- `invoke_agent_runtime`: *“⚠️ BILLABLE OPERATION: This creates or reuses a microVM session that incurs AWS compute charges for the duration of the session.”*
- `stop_runtime_session`: *“This is a cost-saving operation — it prevents runaway charges from idle sessions.”*
- `create_agent_runtime`: *“You are not billed until sessions are invoked, but the runtime definition and its resources persist until deleted.”*

**2. Tools are classified by cost tier in `__init__.py`.**  The module docstring explicitly groups all 14 tools into four tiers so cost implications are visible during code review:

|Tier                |Tools                                                                          |Risk                                                  |
|--------------------|-------------------------------------------------------------------------------|------------------------------------------------------|
|**Per-use billable**|`invoke_agent_runtime`                                                         |Creates/uses microVM sessions — compute charges accrue|
|**One-time setup**  |`create_agent_runtime`, `create_agent_runtime_endpoint`                        |Creates infrastructure, no per-use charge             |
|**Cost-saving**     |`stop_runtime_session`, `delete_agent_runtime`, `delete_agent_runtime_endpoint`|Terminates resources, prevents runaway costs          |
|**Read-only**       |All get/list tools + `get_runtime_guide`                                       |Zero cost, safe for autonomous agent use              |

**3. The guide tool documents cost tiers.** `get_runtime_guide()` includes a “Cost Tiers for MCP Tools” section so agents can self-assess before calling billable tools.

**4. The `_is_service_enabled()` gating mechanism applies.** Runtime tools are registered behind the existing `_is_service_enabled('runtime')` check. Developers can disable them with `AGENTCORE_DISABLE_TOOLS=runtime` or restrict to specific services with `AGENTCORE_ENABLE_TOOLS=docs,browser`.

### What this does NOT do (and why)

- **No confirmation prompts before billable calls.** MCP tools are invoked programmatically by AI agents — there is no interactive confirmation mechanism in the protocol. Cost awareness is handled through tool descriptions that the LLM reads.
- **No per-tool opt-in.** The sub-package registers all 14 tools as a unit. Granular control would require a different mechanism (e.g., tool-level env var flags), which is out of scope for this PR but could be added later.

## Architecture

- **Two boto3 clients**: `bedrock-agentcore-control` (control plane CRUD) and `bedrock-agentcore` (data plane invoke/stop), cached per-region with user-agent tracking (`agentcore-mcp-server/0.1.0 runtime-control` and `agentcore-mcp-server/0.1.0 runtime`)
- **Union return types**: Every tool returns `Union[SpecificResponse, ErrorResponse]` for pyright compatibility
- **Centralised error handling**: `ClientError` → `ErrorResponse` with AWS error code, message, and HTTP status
- **No `from .conftest import`**: Test helper functions are inlined to avoid pytest import issues

## server.py Modification

Replace the old runtime registration block:

```python
# OLD (line 138)
if _is_service_enabled('runtime'):
    mcp.tool()(runtime.manage_agentcore_runtime)
```

With:

```python
# NEW
if _is_service_enabled('runtime'):
    try:
        from .tools.runtime import register_runtime_tools
        register_runtime_tools(mcp)
        logger.info('Runtime tools registered')
    except ImportError as e:
        logger.error(
            f'Runtime tools disabled — failed to import dependencies: {e}.'
        )
    except Exception as e:
        logger.error(
            f'Runtime tools disabled — initialization failed: {e}. '
            f'Set AGENTCORE_DISABLE_TOOLS=runtime to suppress.'
        )
```

Also remove `runtime` from the top-level import: `from .tools import docs, gateway, memory` (was `docs, gateway, memory, runtime`).

## Tool Inventory (14 tools)

|Tool                           |Category    |Description                                  |
|-------------------------------|------------|---------------------------------------------|
|`create_agent_runtime`         |Setup       |Create a new runtime definition              |
|`get_agent_runtime`            |Read-only   |Get runtime details and configuration        |
|`update_agent_runtime`         |Config      |Update runtime, creates new immutable version|
|`delete_agent_runtime`         |Teardown    |Delete runtime and all versions              |
|`list_agent_runtimes`          |Read-only   |List all runtimes in the account             |
|`list_agent_runtime_versions`  |Read-only   |List versions of a specific runtime          |
|`create_agent_runtime_endpoint`|Setup       |Create a custom endpoint (dev/staging/prod)  |
|`get_agent_runtime_endpoint`   |Read-only   |Get endpoint details                         |
|`update_agent_runtime_endpoint`|Config      |Point endpoint to a different version        |
|`delete_agent_runtime_endpoint`|Teardown    |Delete an endpoint                           |
|`list_agent_runtime_endpoints` |Read-only   |List endpoints for a runtime                 |
|`invoke_agent_runtime`         |**Billable**|Invoke a hosted agent (creates/uses microVM) |
|`stop_runtime_session`         |Cost-saving |Terminate a microVM session early            |
|`get_runtime_guide`            |Read-only   |Static reference guide                       |

## pyproject.toml

No changes needed — all dependencies already present:

- `bedrock-agentcore>=1.1.0` (brings boto3/botocore transitively)
- `pydantic>=2.10.6`, `loguru>=0.7.0`, `mcp[cli]>=1.23.0`

## Testing

```bash
# Runtime unit tests only
uv run pytest tests/runtime/ -v

# Full suite including integration tests
uv run pytest --cov --cov-branch
```

All 44 runtime tests pass. Tests mock boto3 — no real AWS API calls. Coverage includes success cases, ClientError handling, pagination, VPC config, payload encoding, user-agent verification, env var defaults, and guide content assertions.

## Migration Notes

The old `tools/runtime.py` referenced the deprecated `bedrock-agentcore-starter-toolkit` pip package with commands like `agentcore launch`, `agentcore destroy`, `agentcore configure`, and `.bedrock_agentcore.yaml`. The new guide references the current `@aws/agentcore` npm CLI with `agentcore create`, `agentcore deploy`, `agentcore remove`, and `agentcore.json`.

## Checklist

* [X ] I have reviewed the [contributing guidelines](https://github.com/awslabs/mcp/blob/main/CONTRIBUTING.md)
* [X ] I have performed a self-review of this change
* [X ] Changes have been tested
* [X ] Changes are documented

Is this a breaking change? N

**RFC issue number**:

Checklist:

* [x ] Migration process documented
* [x ] Implement warnings (if it can live side by side)

## Acknowledgment


src/amazon-bedrock-agentcore-mcp-server/awslabs/amazon_bedrock_agentcore_mcp_server/tools/runtime/guide.py
```diff
@@ -0,0 +1,264 @@
...
+RUNTIME_GUIDE = r"""
+# Amazon Bedrock AgentCore Runtime — Quick Reference
+
+## What is AgentCore Runtime?
+
+A serverless hosting environment for AI agents and tools. Each user
+session runs in an isolated microVM with dedicated CPU, memory, and
+filesystem. Supports HTTP, MCP, A2A, and AGUI protocols.
+
+---
```
jesseturner21, line 30: This is just a thought but can we could put this guide in the [agentcore-cli](https://github.com/aws/agentcore-cli) repo and then have this tool pull it in. 

1. I think this guide would be a helpful reference for devs too.
2. We only have one place manage.

This could be something we explore!

Curious what your thoughts are?

jesseturner21 (approved):
LGTM

src/amazon-bedrock-agentcore-mcp-server/awslabs/amazon_bedrock_agentcore_mcp_server/tools/runtime/guide.py
sundargthb: I like that idea — having one place to manage it makes sense long term. For this PR I’d keep it as the static string since it works today with no network dependency, but let’s explore pulling it from the CLI repo in a follow-up. Main thing to figure out is how to handle offline/fallback scenarios so the tool always returns something useful even if the fetch fails.

jesseturner21 (dismissed):
lgtm

kevin-orellana (commented):
## Review — Runtime Tools Sub-package

Good structure following the Browser pattern. However, there are functional bugs in `invoke_agent_runtime` (the most sensitive tool) and a systematic pattern bug across multiple files. All issues below were reproduced locally.

---

### 1. Truthiness checks silently drop valid values (multiple files)

Several tools use `if value:` instead of `if value is not None:` for optional parameters. This silently drops empty strings (and theoretically `"0"`). Reproduced locally — passing `description=""` to `create_agent_runtime` does not send the field.

**Affected locations:**
- `lifecycle.py` — `create_agent_runtime`: `description`, `server_protocol`
- `lifecycle.py` — `update_agent_runtime`: `server_protocol`
- `endpoints.py` — `create_agent_runtime_endpoint`: `agent_runtime_version`, `description`
- `endpoints.py` — `update_agent_runtime_endpoint`: `agent_runtime_version`
- `invocation.py` — `invoke_agent_runtime`: `runtime_session_id`

Note: the Memory PR (#2946) uses `is not None` everywhere — that's the correct pattern. `update_agent_runtime` in this PR already uses `is not None` for `description` (line 348), making the inconsistency within the same file.

**Fix:** Change all `if param:` to `if param is not None:` for optional parameters.

---

### 2. `invoke_agent_runtime` blocks the asyncio event loop — `invocation.py:94-112`

The entire boto3 call and response reading (`.read()`, `for chunk in body:`) are synchronous operations inside an `async def`. This blocks the MCP server from processing any other requests for the full duration of the agent invocation + response streaming.

For a billable operation that can take seconds to minutes, this is a real problem.

**Fix:** Wrap in `asyncio.to_thread()`:
```python
resp = await asyncio.to_thread(client.invoke_agent_runtime, **kwargs)
```
And similarly for the response body reading.

---

### 3. Non-UTF8 binary responses crash — `invocation.py:104`

```python
response_body = body.read().decode('utf-8')
```

If the agent returns binary content (e.g., `application/octet-stream`), this throws `UnicodeDecodeError`. The exception is caught by the generic handler and returns `ErrorResponse`, even though the API call itself succeeded. Reproduced locally.

**Fix:** Use `decode('utf-8', errors='replace')` or check `content_type` before decoding.

---

### 4. EventStream dict handling produces garbage — `invocation.py:105-112`

The actual boto3 streaming response for `invoke_agent_runtime` yields dicts like `{'chunk': {'bytes': b'hello'}}`. The current code hits the `hasattr(body, '__iter__')` branch and calls `str(chunk)` on each dict, producing:

```
"{'chunk': {'bytes': b'hello '}}{'chunk': {'bytes': b'world'}}"
```

This is raw Python repr, not usable content. Reproduced locally.

**Fix:** Handle the EventStream format explicitly:
```python
for event in body:
    if 'chunk' in event and 'bytes' in event['chunk']:
        chunks.append(event['chunk']['bytes'].decode('utf-8'))
```

---

### 5. `update_agent_runtime` and `update_agent_runtime_endpoint` misclassified as "Read-only" — `__init__.py:42-43`

The cost classification docstring places both update operations under "Read-only (no cost)":

```python
**Read-only (no cost):**
- get_agent_runtime, list_agent_runtimes, list_agent_runtime_versions
- get_agent_runtime_endpoint, list_agent_runtime_endpoints
- update_agent_runtime, update_agent_runtime_endpoint  # <-- wrong
- get_runtime_guide
```

`update_agent_runtime` creates a new immutable version and triggers container deployment. `update_agent_runtime_endpoint` can trigger endpoint redeployment. These should be in a "Configuration" tier. This directly undermines the cost-safety goals of the PR.

The guide's cost tier section (lines 136-141) actually handles this better by omitting update tools, but it should list them under the correct tier instead.

---

### 6. Cross-PR merge conflict with Memory PR #2946

Both PRs modify `server.py`, `test_integ_mcp_protocol.py`, `test_tools.py`, and both add `utils/user_agent.py`. This PR reverts #2946's memory registration back to the old single-tool pattern. Direct conflicts on merge.

**Suggestion:** Merge #2946 first (it's cleaner, no functional bugs), then rebase this PR on top.

kevin-orellana (commented):
## Follow-up: additional finding from functional testing

### 7. `code_entry_point.split(',')` produces empty strings — `lifecycle.py:156,325`

Trailing or double commas in `code_entry_point` produce empty strings in the entry point list:

```python
"main.py,".split(',')        # → ['main.py', '']
"a,,b".split(',')            # → ['a', '', 'b']
",main.py".split(',')        # → ['', 'main.py']
```

These empty strings get sent to the API as part of `entryPoint`. Affects both `create_agent_runtime` (line 156) and `update_agent_runtime` (line 325).

**Fix:** Filter and strip:
```python
code_cfg['entryPoint'] = [s.strip() for s in code_entry_point.split(',') if s.strip()]
```

This is consistent with how `subnets` and `security_groups` already use `.strip()` (lines 169-171), though those don't filter empties either.

sundargthb:
@kevin-orellana 
Good catches across the board, all 7 addressed:

1. Switched all optional param checks to `is not None` — also caught `agent_runtime_version` in `get_agent_runtime` which had the same issue. Left `if next_token:` and `if lc:` as-is since empty pagination tokens and empty dicts are never valid.
2. Wrapped both the `boto3` call and response body reading in `asyncio.to_thread()` so we don't block the event loop during potentially long invocations.
3. All `.decode()` calls now use `errors='replace'` so binary responses get replacement characters instead of crashing.
4. Extracted `_read_response_body()` that handles the `EventStream` dict format (`{'chunk': {'bytes': b'...'}}`) explicitly, with bytes and string fallbacks.
5. Moved both update tools to a new **"Configuration"** tier in the docstring. Good catch — that directly contradicted the cost-safety intent.
6. Agreed — will rebase on #2946 once it merges. The shared `utils/user_agent.py` is already integrated on this branch.
7. All `split(',')` calls now filter and strip: `[s.strip() for s in x.split(',') if s.strip()]`. Applied to `entry_point`, `subnets`, and `security_groups` in both create and update.

jesseturner21 (approved):
lgtm

kevin-orellana (approved):
Looks good. Follows the established Browser/CI sub-package pattern, solid test coverage, cost tier documentation is clear. No blockers.

PR #3011, opened 2026-04-11
fix(aws-iac-mcp-server): add retry with exponential backoff for Knowledge MCP 429 rate limiting

### LINKED ISSUE #2949: aws-knowledge MCP aggressive rate limit causing MCP startup failure
### Describe the bug

The remote AWS Knowledge MCP server at `https://knowledge-mcp.global.api.aws` has a very aggressive per-IP rate limit (~1 request per 15 seconds). This causes MCP clients to fail during startup because the standard MCP handshake requires multiple requests in quick succession (`initialize` → `tools/list`).

The second request consistently gets a 429 response:
```json
{"jsonrpc":"2.0","id":1,"result":null,"error":{"code":-32004,"message":"Too many requests. Please try again later."}}
```

The server was working fine prior to 2026-04-10. The rate limit appears to have been tightened recently.

Tested rate limit window from a single IP:

| Gap between requests | HTTP status |
|---|---|
| 2s | 429 |
| 5s | 429 |
| 10s | 429 |
| 15s | 200 ✓ |
| 20s | 200 ✓ |
| 30s | 200 ✓ |

### Expected Behavior

The MCP handshake sequence (`initialize` + `tools/list` + `notifications/initialized`) should complete without hitting rate limits. These are small, lightweight requests that happen once at startup. The rate limit should allow at least a burst of 3-5 requests to accommodate the standard MCP initialization flow.

### Current Behavior

1. Client sends `initialize` — 200 OK
2. Client immediately sends `tools/list` — 429 Too Many Requests
3. Client treats this as a connection failure and marks the server as unavailable
4. Manual retry sometimes works if enough time has passed, but is unreliable

### Reproduction Steps

```bash
# First request succeeds after a cooldown period
curl -s -w "\nHTTP %{http_code}\n" --max-time 10 -X POST "https://knowledge-mcp.global.api.aws" \
  -H "Content-Type: application/json" \
  -d '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-03-26","capabilities":{},"clientInfo":{"name":"test","version":"1.0"}}}'

# Second request within 15s gets 429
sleep 5
curl -s -w "\nHTTP %{http_code}\n" --max-time 10 -X POST "https://knowledge-mcp.global.api.aws" \
  -H "Content-Type: application/json" \
  -d '{"jsonrpc":"2.0","id":2,"method":"tools/list","params":{}}'
```

### Possible Solution

- Allow a small burst window (e.g. 5 requests within 5 seconds) to accommodate the MCP handshake, then enforce the per-minute/per-hour rate limit for actual tool calls.
- Alternatively, return a `Retry-After` header with the 429 response so clients can implement proper backoff.

- CloudFront POP: AKL53-P1 (Auckland, New Zealand)
- The server was working without issues prior to 2026-04-10
- The `fastmcp` stdio proxy workaround from the README may help some clients, but the remote server should handle the standard MCP handshake natively

### OS

Ubuntu 24.04.4 LTS (WSL2)

### Server

aws-knowledge-mcp-server

### Server Version

1.0.0 (AWSDocumentationMCPProdGateway)

Global (tested from New Zealand IP, CloudFront POP AKL53-P1)

### Other information

MCP Client: OpenCode 1.4.3 (also tested with raw curl)

SARAMALI15792:
Fixes #2949

### Changes

- Added `_is_rate_limit_error()` helper to `aws-iac-mcp-server` that detects HTTP 429 responses via `httpx.HTTPStatusError` status code check plus a string-fallback for wrapped exceptions
- Replaced the `try/except` block in `search_documentation()` with a retry loop (max 3 attempts, 16 s base backoff, +/-10% jitter, capped at 60 s) that retries on 429 and raises immediately on any other error
- Added `httpx>=0.27.0` as an explicit dependency in `pyproject.toml` (previously only a transitive dep via fastmcp)
- Updated `aws-knowledge-mcp-server` README with a new "Rate Limiting" section, a troubleshooting note in the Testing section, and FAQ #8

### User experience

**Before:** `search_documentation()` in `aws-iac-mcp-server` would raise an exception immediately on any error, including HTTP 429. Since each call creates a new `fastmcp` client, the full MCP handshake (`initialize` -> `tools/list`) runs on every invocation - consuming multiple requests from the per-IP quota before the actual tool call. A single rate-limited response meant the caller received an unhandled exception.

**After:** On a 429 response, `search_documentation()` silently waits and retries up to 3 times with exponential backoff (16 s -> 32 s -> 60 s max). Only after all retries are exhausted does the exception propagate. Non-rate-limit errors are still raised immediately. Users of `aws-knowledge-mcp-server` directly (via MCP clients) also now have documentation explaining the ~1 req/15 s per-IP window and the recommended `fastmcp` proxy workaround.

## Checklist

* [x] Changes are documented

Is this a breaking change? N

**RFC issue number**: N/A

Checklist:

## Acknowledgment


amirasyraf:
I don't think this fix should be on the client-side. The retry and backoff will just make loading the MCP very slow

The proper solution is server side - for AWS to fix their unnecessarily aggressive rate limit

SARAMALI15792:
Hey @amirasyraf, thanks for taking the time to review this — really appreciate the feedback!

You're absolutely right that the **ideal long-term fix is server-side**, and ideally AWS would relax their rate limits or provide better tooling around this. However, since that's outside our control and users are actively hitting this problem today, a client-side mitigation is the most practical path forward in the interim.

A few things worth clarifying about how this implementation works:

- **No impact on normal loading** — the retry logic only activates when a `429 Too Many Requests` response is received. If you're not being rate limited, you'll never notice this code exists
- **Exponential backoff is AWS best practice** — AWS themselves recommend this pattern in their official SDK documentation for handling throttling gracefully
- **It's a safety net, not a workaround** — rather than crashing or silently failing when rate limited, the client recovers automatically without user intervention

That said, I completely understand the concern about potential slowdowns in edge cases. To address that, I'm happy to make the retry behavior **fully configurable** — for example, an opt-in flag in the config so users who prefer to handle it differently (or who want fast-fail behavior) can disable it entirely.

I'd love to hear your thoughts — if client-side handling isn't the right approach here, what alternative would you suggest? Open to any direction the maintainers prefer.

amirasyraf:
@SARAMALI15792 The retry logic will absolutely activate every time the MCP is started. At least with OpenCode, this means every time you launch a new session. Which adds 20+ seconds delay/block before it can use all other MCPs. I'm not sure how other agent harnesses handle it, but it's consistent with OpenCode

The only acceptable client-side alternative "fix" is to have the MCP continuously run in the background. This is what I'm doing now via a simple MCP proxy that runs as a service when my OS boots up. "fix" in quotes because it's not really a fix, but a mitigation.

But that still doesn't address the burst requests that agents will make while using the MCP. Those burst requests will inevitably get rate limited as well. Edit: well actually this PR might at least address this issue, but also means the AI agent will need to wait 20+ seconds. Will this impact harness timeouts?

I don't see any other real fix for this other than server-side.

SARAMALI15792:
@amirasyraf The retry only fires during explicit tool invocations, not during MCP initialization (`initialize` / `tools/list`). So startup delay only applies if OpenCode calls `search_documentation` as part of session setup. Do you have logs confirming that? Or is the 20s delay from the per-call handshake overhead itself — each invocation creates a new `fastmcp` Client doing a full `initialize → tools/list` round-trip before the actual tool call, consuming ~3 quota units per search regardless of this change.

The harness timeout concern is valid. With `BASE_BACKOFF=16s`, two retries = up to ~48s blocking — that will breach most harness timeouts. Two fixes available:
- Drop `BASE_BACKOFF` to `5s` (max ~17s total)
- Expose `AWS_KNOWLEDGE_MAX_RETRIES` and `AWS_KNOWLEDGE_BASE_BACKOFF_SECONDS` as env vars so operators can tune per harness

Alternatively, the stronger fix is reusing a persistent `fastmcp` Client instead of creating one per call — cuts quota usage from ~3 units to 1 per invocation, reducing rate limit frequency before retry is even needed.

If fast-fail is preferred over retry, I can drop the backoff entirely and raise a clear `RateLimitError` pointing to the proxy workaround instead — no blocking, no harness timeout risk.

Which direction do the maintainers prefer?

PR #3013, opened 2026-04-13
feat(cloudwatch-logs): add execute_cwl_insights_batch tool

shri-tambe:
Add multi-region, multi-log-group, supporting longer duration CloudWatch Logs Insights batch query tool. 

## Summary

Adds a new execute_cwl_insights_batch tool to the CloudWatch MCP server that enables running a single Logs Insights query across multiple regions and log groups in one call. The tool handles the complexity of CloudWatch's per-region, 50-log-group-per-query limits by automatically chunking log groups, executing queries in parallel across regions, and merging results. This will allow customers to use longer duration queries that could be more than 60 minutes by diving those into multiple Queries.

## User experience

Before: Users had to manually issue separate execute_logs_insight_query calls for each region and manage log group batching themselves when querying across multiple regions or more than 50 log groups.

After: Users call execute_cwl_insights_batch with a list of regions and log group patterns. The tool automatically chunks log groups, runs queries in parallel across regions with rate limiting, handles the 10k record limit via adaptive time-range splitting, and returns merged results annotated with _region, _account, and _logGroups metadata.
## Checklist

* [X] Changes are documented

**RFC issue number**: NA

Checklist:

## Acknowledgment


src/cloudwatch-mcp-server/awslabs/cloudwatch_mcp_server/cloudwatch_logs/cwl_insights_batch.py
```diff
@@ -0,0 +1,562 @@
...
+    """Result of a multi-region Logs Insights query."""
+
+    summary: MultiRegionQuerySummary = Field(..., description='Execution summary')
+    results: List[Dict] = Field(default_factory=list, description='Merged result records')
+
+
+# ---------------------------------------------------------------------------
+# Per-region rate-limiting context
+# ---------------------------------------------------------------------------
+class _RegionContext:
```
goranmod, line 89: AI assisted:

  - cwl_insights_batch.py:~480 — _RegionContext created per-call, not shared across concurrent calls
    - Rationale: The carefully documented 25% concurrency budget (7 concurrent queries, 1.4 TPS per region) is only enforced within a single invocation. If the MCP server handles concurrent tool calls, multiple invocations targeting the same region will each create independent
  semaphores, potentially exceeding account limits and causing throttling for the customer's other workloads.
    - Suggestion: Use a module-level dict[str, _RegionContext] keyed by region name, lazily initialized:

      _region_contexts: Dict[str, _RegionContext] = {}
      _region_contexts_lock = asyncio.Lock()

      async def _get_region_context(region: str) -> _RegionContext:
          async with _region_contexts_lock:
              if region not in _region_contexts:
                  _region_contexts[region] = _RegionContext()
              return _region_contexts[region]

src/cloudwatch-mcp-server/awslabs/cloudwatch_mcp_server/cloudwatch_logs/cwl_insights_batch.py
```diff
@@ -0,0 +1,562 @@
...
+        kwargs = remove_null_values(
+            {
+                'logGroupNames': log_groups,
+                'startTime': start_ts,
+                'endTime': end_ts,
+                'queryString': query_string,
+                'limit': limit,
+            }
+        )
+        resp = logs_client.start_query(**kwargs)
```
goranmod: AI assisted: 
- 
- cwl_insights_batch.py:148–170 — Synchronous boto3 calls block the async event loop
    - Rationale: logs_client.start_query() and logs_client.get_query_results() are synchronous network calls. In an async function dispatched via asyncio.gather, these block the event loop, preventing true parallel execution. With many chunks, this serializes what should be concurrent
  work and degrades MCP server responsiveness. The existing tools have the same issue, but this PR amplifies it by design (parallel multi-region execution).
    - Suggestion: Wrap sync calls with asyncio.get_event_loop().run_in_executor(None, ...):

      loop = asyncio.get_event_loop()
      resp = await loop.run_in_executor(None, lambda: logs_client.start_query(**kwargs))
      # ...
      resp = await loop.run_in_executor(None, lambda: logs_client.get_query_results(queryId=query_id))

      Or consider using aioboto3 for native async support. Note: this is a pre-existing pattern in the codebase, so it could be addressed as a follow-up rather than blocking this PR.

src/cloudwatch-mcp-server/awslabs/cloudwatch_mcp_server/cloudwatch_logs/cwl_insights_batch.py
```diff
@@ -0,0 +1,562 @@
...
+MAX_RETRIES = 2
+MAX_RECHUNK_DEPTH = 4  # cap recursive time-range splits
+RETRY_BACKOFF_BASE = 2  # exponential backoff: 2s, 4s for attempts 1, 2
+LARGE_RESULT_WARNING_THRESHOLD = 100_000  # warn when result set exceeds this
+
+# ---------------------------------------------------------------------------
+# Throttle defaults – consume ≤ 25 % of the account/region hard limits so
+# the customer's other workloads are not starved.
+#
+# Account/region hard limits (most regions):
```
goranmod: Does it make sense to add the link to the limits rather than mentioning "most regions"? Would you envision setting this as a parameter at a future time?

shri-tambe: Yep, let me update doc link. I'd not keep separate parameters for this since this will require customer to know limits which might create one extra step. Maybe something to consider in future.

goranmod (dismissed):

pr-body.md
```diff
@@ -0,0 +1,44 @@
+Add multi-region, multi-log-group CloudWatch Logs Insights batch query tool.
```
gcacace: Please remove this file from the PR

src/cloudwatch-mcp-server/awslabs/cloudwatch_mcp_server/cloudwatch_logs/cwl_insights_batch.py
```diff
@@ -0,0 +1,583 @@
...
+                    query_string,
+                    limit,
+                    max_timeout,
+                    summary,
+                    region,
+                    account_label,
+                )
+            )
+
+    all_rows_nested = await asyncio.gather(*tasks)
```
gcacace, line 573: What happens to the in-flight queries if the tool execution get cancelled by the user (e.g.: if it's taking too long and user wants to abort)? Would we be leaving all the tasks running and not cancelling?

shri-tambe: Good catch! We do not honor cancellation. Let me fix.

gcacace (approved):

goranmod (approved):

Issue #1824, opened 2025-11-22
Support for AWS `aws login` command


moritalous:
### Describe the feature

I would like AWS MCP servers to support authentication using the new `aws login` command announced by AWS on November 19, 2025.

Currently, AWS MCP servers recommend authentication methods using the `AWS_PROFILE` environment variable, but credentials generated by `aws login` do not appear to be properly recognized.

### Use Case

**Background:**
`aws login` is a new feature that automatically generates and rotates (every 15 minutes) short-term credentials using AWS Management Console credentials. This provides the following benefits:

- No need to manage long-term access keys
- Secure authentication (using OAuth 2.0 + PKCE)
- Simple setup (browser authentication only)

**Problem:**
When using AWS MCP servers, even after authenticating with `aws login` and configuring as follows, the credentials are not properly recognized:

```bash
# Authenticate with AWS CLI 2.32.0
aws login
```

```json
# MCP configuration example (mcp.json)
{
  "mcpServers": {
    "aws-api": {
      "command": "uvx",
      "args": ["awslabs.aws-api-mcp-server@latest"],
      "env": {
        "AWS_REGION": "us-east-1"
      }
    }
  }
}
```

**Expected behavior:**
The AWS SDK credential chain should automatically detect credentials generated by `aws login`, and the MCP server should operate normally.

**Actual behavior:**
Credentials are not detected, and the MCP server fails to start or AWS API access fails.

### Proposed Solution

Please enable AWS MCP servers to properly recognize credentials generated by `aws login`.

**Implementation approach (speculation):**
Supporting `aws login` likely requires adding the `botocore[crt]` dependency:
```bash
uv add botocore[crt]
```

Adding this dependency should enable the AWS SDK credential chain to properly recognize credentials generated by `aws login`.

### Other Information

**Reference links:**
- [AWS Official Announcement](https://aws.amazon.com/about-aws/whats-new/2025/11/console-credentials-aws-cli-sdk-authentication/)
- [AWS Security Blog: Simplified developer access to AWS with 'aws login'](https://aws.amazon.com/blogs/security/simplified-developer-access-to-aws-with-aws-login/)
- [AWS CLI User Guide: Sign in for authentication](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-sign-in.html)

### Acknowledgements

- [x] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

sebsto:
+1

[...]

Issue #481, opened 2025-05-30
Support non-profile credentials


clareliguori:
### Describe the feature

Many of the MCP servers explicitly set the profile name for Boto sessions and set it to `default` profile name by default, preventing users from using other credential sources like session credentials in the environment. If there is no credentials file in the environment, then you get an error like `botocore.exceptions.ProfileNotFound: The config profile (default) could not be found`.

Additionally, the design guidelines (snippet below) show explicitly setting the profile_name by retrieving the value of "AWS_PROFILE" env var. This is unnecessary because Boto will do this for you.

```python
import boto3
import os

# Bedrock Runtime Client
aws_region: str = os.environ.get('AWS_REGION', 'us-east-1')

try:
    if aws_profile := os.environ.get('AWS_PROFILE'):
        bedrock_runtime_client = boto3.Session(
            profile_name=aws_profile, region_name=aws_region
        ).client('bedrock-runtime')
    else:
        bedrock_runtime_client = boto3.Session(region_name=aws_region).client('bedrock-runtime')
except Exception as e:
    logger.error(f'Error creating bedrock runtime client: {str(e)}')
    raise
```

### Use Case

While I am testing MCP servers, I want to use temporary credentials stored in environment variables.  These should be automatically picked up by Boto.

### Proposed Solution

I'm happy to go search and replace across all the MCP servers in the repo.

### Acknowledgements

- [x] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

clareliguori:
Would it be better to submit a PR per MCP server, or is one mega one OK?

laithalsaadoon:
The reason we chose to let the users choose a custom profile name (and for
own use cases) was to use separate principals and even AWS accounts from
our host terminal and the MCP processes. For example, when using a common
knowledge base in a shared AWS account, separate from the AWS session and
account that's active in the host terminal.

Furthermore, I've found some MCP clients don't always start an MCP server
with all the current terminal environment variables (hence why other open
source AWS-related MCP servers simply accept copy-pasted creds in the MCP
config json files).

Should we instead use a different environment variable name in such
instances? What do you think?


Thank you,
Laith Al-Saadoon

clareliguori:
Ok, if I'm understanding correctly, you want a user-specified profile name to take precedence over environment variable creds and other creds that usually have precedence over the shared credentials file.

https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html#configuring-credentials

[...]

PR #2685, opened 2026-03-16
fix(redshift-mcp-server): handle missing DBName field in discover_clusters

abhu85:
## Summary

- Fixes `KeyError: 'DBName'` when `discover_clusters()` processes clusters without the `DBName` field present
- Uses safe dictionary access via `.get('DBName', 'dev')` with default to 'dev' (Redshift's default database name)
- Adds regression test to verify clusters without DBName are handled gracefully

## Problem

The `describe_clusters` API may return cluster metadata without a `DBName` field in certain scenarios (e.g., clusters created via certain methods, or clusters in specific states). The current implementation used direct dictionary access `cluster['DBName']` which raises `KeyError` when the field is missing.

Per [AWS API Reference](https://docs.aws.amazon.com/redshift/latest/APIReference/API_Cluster.html), `DBName` is marked as "Required: No".

## Root Cause

Line 417 in `redshift.py`:
```python
'database_name': cluster['DBName'],  # Direct access - raises KeyError if missing
```

Other fields in the same block already used safe access patterns (`.get()`), but `DBName` was missed.

## Fix

Changed to safe dictionary access with default value:
```python
'database_name': cluster.get('DBName', 'dev'),  # Returns 'dev' if missing
```

This matches the existing Serverless workgroup behavior which already defaults to 'dev'.

## Tests Run

All 13 discover tests pass, including new test for missing DBName scenario.

Fixes #2331

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the [project license](https://github.com/awslabs/mcp/blob/main/LICENSE).

PR #3194, opened 2026-04-21
fix(redshift-mcp-server): refresh expired AWS credentials automatically

scaredcat:
## Problem

`RedshiftClientManager` caches boto3 clients at first use and never refreshes them. When the underlying credentials are rotated while the MCP server is running, a cached client keeps signing requests with the stale credentials and every subsequent call fails with `ExpiredToken` until the process is restarted — a real pain point for long-running MCP server sessions.

This specifically affects credential sources that botocore resolves **once** and never refreshes on its own — most notably keys written to the **shared credentials file** (e.g. `ada` rewriting `~/.aws/credentials`). Sources that yield *refreshable* credentials (SSO, assume-role, web identity, container) already refresh inside a cached client via botocore, and environment-variable credentials cannot be refreshed in-process at all.

Related: #481, #1824

## Fix

Added **reactive** credential-refresh handling, centralized in a small per-service client so every call site is covered without per-call boilerplate.

`_CredentialRefreshingClient` wraps a single boto3 Redshift-family service client (`redshift`, `redshift-serverless`, `redshift-data`). It lazily creates and caches its client, and routes every operation through a shared retry helper: on an `ExpiredToken`/`ExpiredTokenException` error it discards the cached client and retries the call once against a freshly built one, which re-reads credentials from the default chain. Non-expiry errors propagate unchanged.

### Design decisions

- **Reactive catch-and-retry, not proactive inspection.** We catch the two documented error codes rather than inspecting private botocore internals (`_expiry_time`) — simpler and not tied to botocore implementation details.
- **Self-contained client, thin manager.** Each `_CredentialRefreshingClient` owns its own lifecycle (build, cache, invalidate, retry); `RedshiftClientManager` is a thin registry that hands out one client per service. Dependencies point inward — the manager depends on the clients, not the reverse.
- **Transparent proxy.** `__getattr__` forwards operations to the underlying client wrapped in the retry; non-callable attributes (e.g. `meta`) pass through unchanged; private names are not proxied.
- **Pagination collected eagerly under retry.** `get_paginator(...).paginate()` returns all pages as a list within a single retry. Collecting eagerly means an expiry mid-listing restarts the (idempotent, read-only) listing once with no pages already handed out — so no duplicates. Both paginated listings here are small and fully materialized by callers, so nothing is lost by not streaming.
- **Additive at the call sites.** `discover_clusters` / `_execute_statement` are unchanged from `main` — the behavior comes entirely from the client the manager hands out.

### Unit tests

Full `redshift-mcp-server` suite passes (**202 tests**), including coverage for client creation/caching, credential invalidation, the retry helper (expired → invalidate → retry → success), non-expiry pass-through, non-operation attribute pass-through, eager pagination + retry, and — critically — that a credential refresh **never re-submits a statement** (expiry while polling and while fetching results both retry only the idempotent describe/result calls).

### E2E tests — real infrastructure (`us-east-1`, provisioned `ra3.large` + serverless 8 RPU)

Deployed via CloudFormation (provisioned cluster + serverless workgroup), tested, and torn down. Last run: 2026-08-07.

| # | Scenario | Provisioned | Serverless |
|---|----------|:-----------:|:----------:|
| 1 | discover_clusters | ✅ | ✅ |
| 2 | discover_databases | ✅ | ✅ |
| 3 | discover_schemas | ✅ | ✅ |
| 4 | discover_tables | ✅ | ✅ |
| 5 | discover_columns | ✅ | ✅ |
| 6 | execute_query (SELECT) | ✅ | ✅ |

### E2E test — credential refresh (against real infra)

| # | Scenario | Result |
|---|----------|:------:|
| 1 | Initial call establishes cached clients | ✅ |
| 2 | `invalidate()` then `execute_query` succeeds with a freshly built client | ✅ |
| 3 | Injected `ExpiredToken` at the boto3 layer → client invalidates, rebuilds, and recovers against real infra | ✅ |

CloudFormation template and E2E scripts: https://gist.github.com/scaredcat/c0465de49d7e796ac64d54aacd6285c9

---

### DIFF
--- a/src/redshift-mcp-server/awslabs/redshift_mcp_server/redshift.py
+++ b/src/redshift-mcp-server/awslabs/redshift_mcp_server/redshift.py
@@ -44,8 +44,10 @@ from awslabs.redshift_mcp_server.models import (
 from awslabs.redshift_mcp_server.sql_guard import assert_executable
 from botocore.config import Config
 from botocore.exceptions import ClientError
+from functools import partial
 from loguru import logger
 from sqlglot import exp
+from types import SimpleNamespace
 
 
 def _sql_identifier(value: str) -> str:
@@ -57,69 +59,139 @@ def _sql_identifier(value: str) -> str:
 _ACCESS_DENIED = {'AccessDeniedException', 'UnauthorizedAccess', 'AccessDenied'}
 
 
+# ClientError codes that indicate expired credentials.
+_EXPIRED_TOKEN = {'ExpiredToken', 'ExpiredTokenException'}
+
+
+def _error_code(error: ClientError) -> str | None:
+    """Return the AWS error code carried by a ClientError, if any."""
+    return error.response.get('Error', {}).get('Code')
+
+
+class _CredentialRefreshingClient:
+    """A boto3 client for one service that refreshes credentials once on ExpiredToken.
+
+    Lazily creates and caches its boto3 client. On an ExpiredToken/ExpiredTokenException
+    error it discards the cached client and retries once against a freshly built one,
+    which re-reads credentials from the default chain.
+
+    This covers credential sources botocore resolves once and never refreshes, notably
+    keys written to the shared credentials file, where a cached client keeps signing with
+    them past expiry. Sources that yield refreshable credentials (SSO, assume-role, web
+    identity, container) already refresh inside a cached client, and credentials taken
+    from environment variables cannot be refreshed in-process at all.
+    """
+
+    def __init__(
+        self,
+        service_name: str,
+        config: Config,
+        aws_region: str | None = None,
+        aws_profile: str | None = None,
+    ):
+        """Initialize for a single Redshift-family boto3 service."""
+        self._service_name = service_name
+        self._config = config
+        self._aws_region = aws_region
+        self._aws_profile = aws_profile
+        self._client = None
+
+    @property
+    def service_name(self) -> str:
+        """The boto3 service this client is bound to."""
+        return self._service_name
+
+    def _build(self):
+        """Return the live boto3 client, creating and caching it on first use."""
+        if self._client is None:
+            try:
+                # Session works with None values - uses default credentials/region chain
+                session = boto3.Session(
+                    profile_name=self._aws_profile, region_name=self._aws_region
+                )
+                self._client = session.client(self._service_name, config=self._config)
+                logger.info(
+                    f'Created {self._service_name} client with profile: {self._aws_profile or "default"}, region: {self._aws_region or "default"}'
+                )
+            except Exception as e:
+                logger.error(f'Error creating {self._service_name} client: {str(e)}')
+                raise
+
+        return self._client
+
+    def invalidate(self):
+        """Discard the cached client so the next call recreates it with fresh credentials."""
+        self._client = None
+        logger.info(f'Invalidated cached {self._service_name} client for credential refresh')
+
+    def _retry(self, description: str, call, client=None):
+        """Run call against the live client, refreshing credentials once on expiry."""
+        try:
+            return call(client if client is not None else self._build())
+        except ClientError as e:
+            if _error_code(e) not in _EXPIRED_TOKEN:
+                raise
+            logger.warning(f'Expired credentials on {description}, refreshing and retrying')
+            self.invalidate()
+            return call(self._build())
+
+    def __getattr__(self, name: str):
+        """Proxy an attribute, wrapping operations with a retry on expired credentials."""
+        if name.startswith('_'):
+            raise AttributeError(name)
+
+        client = self._build()
+        attribute = getattr(client, name)
+        if not callable(attribute):
+            return attribute
+
+        def call(*args, **kwargs):
+            return self._retry(name, lambda c: getattr(c, name)(*args, **kwargs), client)
+
+        return call
+
+    def get_paginator(self, operation: str) -> SimpleNamespace:
+        """Return an object exposing paginate(**kwargs), like a boto3 paginator."""
+        return SimpleNamespace(paginate=partial(self._paginate, operation))
+
+    def _paginate(self, operation: str, **kwargs) -> list[dict]:
+        """Return every page of the operation, retried once on expired credentials.
+
+        Pages are collected eagerly so a refresh cannot re-yield pages already handed
+        out. Both paginated listings here are small and fully materialized by callers.
+        """
+        return self._retry(
+            f'paginating {operation}',
+            lambda client: list(client.get_paginator(operation).paginate(**kwargs)),
+        )
+
+
 class RedshiftClientManager:
-    """Manages AWS clients for Redshift operations."""
+    """Builds and holds the self-refreshing clients used for Redshift operations."""
 
     def __init__(
         self, config: Config, aws_region: str | None = None, aws_profile: str | None = None
     ):
-        """Initialize the client manager."""
-        self.aws_region = aws_region
-        self.aws_profile = aws_profile
-        self._redshift_client = None
-        self._redshift_serverless_client = None
-        self._redshift_data_client = None
-        self._config = config
+        """Create one self-refreshing client per Redshift service."""
+        self._redshift = _CredentialRefreshingClient('redshift', config, aws_region, aws_profile)
+        self._redshift_serverless = _CredentialRefreshingClient(
+            'redshift-serverless', config, aws_region, aws_profile
+        )
+        self._redshift_data = _CredentialRefreshingClient(
+            'redshift-data', config, aws_region, aws_profile
+        )
 
-    def redshift_client(self):
-        """Get or create the Redshift client for provisioned clusters."""
-        if self._redshift_client is None:
-            try:
-                # Session works with None values - uses default credentials/region chain
-                session = boto3.Session(profile_name=self.aws_profile, region_name=self.aws_region)
-                self._redshift_client = session.client('redshift', config=self._config)
-                logger.info(
-                    f'Created Redshift client with profile: {self.aws_profile or "default"}, region: {self.aws_region or "default"}'
-                )
-            except Exception as e:
-                logger.error(f'Error creating Redshift client: {str(e)}')
-                raise
+    def redshift_client(self) -> _CredentialRefreshingClient:
+        """Get the Redshift client for provisioned clusters (auto-refreshes on expiry)."""
+        return self._redshift
 
-        return self._redshift_client
+    def redshift_serverless_client(self) -> _CredentialRefreshingClient:
+        """Get the Redshift Serverless client (auto-refreshes on expiry)."""
+        return self._redshift_serverless
 
-    def redshift_serverless_client(self):
-        """Get or create the Redshift Serverless client."""
-        if self._redshift_serverless_client is None:
-            try:
-                # Session works with None values - uses default credentials/region chain
-                session = boto3.Session(profile_name=self.aws_profile, region_name=self.aws_region)
-                self._redshift_serverless_client = session.client(
-                    'redshift-serverless', config=self._config
-                )
-                logger.info(
-                    f'Created Redshift Serverless client with profile: {self.aws_profile or "default"}, region: {self.aws_region or "default"}'
-                )
-            except Exception as e:
-                logger.error(f'Error creating Redshift Serverless client: {str(e)}')
-                raise
-
-        return self._redshift_serverless_client
-
-    def redshift_data_client(self):
-        """Get or create the Redshift Data API client."""
-        if self._redshift_data_client is None:
-            try:
-                # Session works with None values - uses default credentials/region chain
-                session = boto3.Session(profile_name=self.aws_profile, region_name=self.aws_region)
-                self._redshift_data_client = session.client('redshift-data', config=self._config)
-                logger.info(
-                    f'Created Redshift Data API client with profile: {self.aws_profile or "default"}, region: {self.aws_region or "default"}'
-                )
-            except Exception as e:
-                logger.error(f'Error creating Redshift Data API client: {str(e)}')
-                raise
-
-        return self._redshift_data_client
+    def redshift_data_client(self) -> _CredentialRefreshingClient:
+        """Get the Redshift Data API client (auto-refreshes on expiry)."""
+        return self._redshift_data
 
 
 class RedshiftSessionManager:
@@ -523,7 +595,7 @@ async def discover_clusters() -> list[RedshiftCluster]:
         logger.info(f'Found {len(clusters)} provisioned clusters')
 
     except ClientError as e:
-        if e.response.get('Error', {}).get('Code') not in _ACCESS_DENIED:
+        if _error_code(e) not in _ACCESS_DENIED:
             raise
         provisioned_error = e
         logger.warning(f'Skipping provisioned; IAM lacks permission: {e}')
@@ -569,7 +641,7 @@ async def discover_clusters() -> list[RedshiftCluster]:
         logger.info(f'Found {serverless_count} serverless workgroups')
 
     except ClientError as e:
-        if e.response.get('Error', {}).get('Code') not in _ACCESS_DENIED:
+        if _error_code(e) not in _ACCESS_DENIED:
             raise
         serverless_error = e
         logger.warning(f'Skipping serverless; IAM lacks permission: {e}')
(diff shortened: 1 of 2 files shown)


grayhemp:
I agree, the problem is real. However, this PR doen't seem to be covering it all. The cluster discovery will fail as well when the token is expired. That needs to be a part of the fix too. Also, I think it's okay to make it all a bit more simple in the first iteration of the fix by not implementing a retry logic in the MCP itself but rather letting the agent/chat decide if it needs to retry or not as it gets an expiration exception. What needs to be done is the cached session/client expiration on these exceptions.

grayhemp:
1. Alternatively, I wonder  if  we  should consider something like this to check the expired  creds?

```
creds = session.get_credentials().get_frozen_credentials()
creds.expiry  # <-- date/time
```

2. Also , I wonder  why you picked such a sophisticated error check. Wouldn't the following do the trick? 

```
except ClientError as e:
    if e.response['Error']['Code'] == 'ExpiredToken':
        ...
```

scaredcat:
> 1. Alternatively, I wonder  if  we  should consider something like this to check the expired  creds?
> 
> 
> ```
> creds = session.get_credentials().get_frozen_credentials()
> creds.expiry  # <-- date/time
> ```


Thanks for the feedback. I agree that checking expiry time is a much simpler approach and have updated the PR as such.

src/redshift-mcp-server/awslabs/redshift_mcp_server/redshift.py
```diff
@@ -50,14 +51,45 @@ def __init__(
...
+            if credentials is None:
+                return False
+            resolved = credentials.get_frozen_credentials()
+            if resolved.token is None:
+                return False
+            expiry = getattr(credentials, '_expiry_time', None)
+            if expiry is None:
+                return False
+            return expiry <= datetime.now(timezone.utc)
+        except Exception:
```
grayhemp: I don't like that. We should not be swallowing exceptions like this. Let' the exception happen and report the real error. No cheating.

src/redshift-mcp-server/awslabs/redshift_mcp_server/redshift.py
```diff
@@ -50,14 +51,45 @@ def __init__(
         self._redshift_client = None
         self._redshift_serverless_client = None
         self._redshift_data_client = None
+        self._session = None
         self._config = config
 
+    def _create_session(self):
```
grayhemp: This function is unnecessary IMHO.

grayhemp (changes requested):
Left a couple of comments in the code.

Additionally, you still need to act on 
- https://github.com/awslabs/mcp/pull/3194#issuecomment-4402002080 
- and partially on https://github.com/awslabs/mcp/pull/3194#issuecomment-4411994834 (item 1).

scaredcat:
> Left a couple of comments in the code.
> 
> Additionally, you still need to act on
> 
>     * [#3194 (comment)](https://github.com/awslabs/mcp/pull/3194#issuecomment-4402002080)
> 
>     * and partially on [#3194 (comment)](https://github.com/awslabs/mcp/pull/3194#issuecomment-4411994834) (item 1).

Thanks for the review. I've addressed those by checking for valid credentials before returning the client (for each of the client type getters) as every operation call uses the client getter first.

For item 1, [_expiry_time](https://github.com/boto/botocore/blob/855395bf1b34616f99ff68ecec09f6f08757987d/botocore/credentials.py#L428) is the property leveraged as it's not exposed by the SDK in any other method.

src/redshift-mcp-server/awslabs/redshift_mcp_server/redshift.py
```diff
@@ -50,31 +51,56 @@ def __init__(
         self._redshift_client = None
         self._redshift_serverless_client = None
         self._redshift_data_client = None
+        self._session = None
         self._config = config
 
+    def _are_credentials_expired(self) -> bool:
```
grayhemp: Could you please point me to the documentation describing the following algorithm? How can I understand that it is complete and nothing is missing?

src/redshift-mcp-server/awslabs/redshift_mcp_server/redshift.py
```diff
@@ -50,31 +51,56 @@ def __init__(
         self._redshift_client = None
         self._redshift_serverless_client = None
         self._redshift_data_client = None
+        self._session = None
```
grayhemp: If we introduced self._session on the client manager class level, why do we need to keep and maintain self._redshift_client, self._redshift_data_client, self._redshift_serverless_client?

grayhemp (changes requested):
After the review notes addressed, could you please run and update the description with:

- The feature specific E2E testing, eg. https://github.com/awslabs/mcp/pull/2685#issuecomment-4158568897.
- The full E2E testing, eg. https://github.com/awslabs/mcp/pull/2685#issuecomment-4158602947.

Your E2E setup should include both provisioned and serverless clusters.

ps. Yes, I will need to add a proper eval framework in the near future to not run it manually in every PR.

grayhemp:
@scaredcat ping.

src/redshift-mcp-server/awslabs/redshift_mcp_server/redshift.py
scaredcat: *This response was AI-assisted.*


The individual client variables are intentional performance caches. `session.client()` is not a cheap call — each invocation creates a new client object, which involves parsing the service model, resolving the endpoint, and building a new HTTP connection pool. Since MCP tools are called frequently, we want to avoid that overhead on every invocation.


The `self._session` serves a different purpose: it's the credential freshness signal. When credentials expire, we invalidate the session *and* all cached clients together (via `invalidate_clients()`), so they get lazily recreated with fresh credentials on the next access.


In short:

- `self._session` → credential lifecycle management

- `self._redshift_client` / `_data_client` / `_serverless_client` → performance caches tied to that session's lifetime


Both are needed for correctness and performance.

src/redshift-mcp-server/awslabs/redshift_mcp_server/redshift.py
scaredcat: The algorithm mirrors how botocore itself detects credential expiry in the [`RefreshableCredentials` class](https://github.com/boto/botocore/blob/dbfa9d0d2bdad5476556e632d96126a0b2a0e4ce/botocore/credentials.py#L388):


1. **`token is None` → not temporary credentials** — Static IAM credentials (long-lived access keys) don't have a session token. Only temporary/session credentials (from STS assume-role, SSO, instance profiles, etc.) have a token and can expire. If there's no token, there's nothing to expire. This matches botocore's `Credentials` vs `RefreshableCredentials` distinction.


2. **`_expiry_time` attribute** — This is the internal field botocore's `RefreshableCredentials` uses to track when credentials expire ([L428](https://github.com/boto/botocore/blob/dbfa9d0d2bdad5476556e632d96126a0b2a0e4ce/botocore/credentials.py#L428)). It's set from the `expiry_time` in the credential metadata.


3. **Comparing against `datetime.now(timezone.utc)`** — This mirrors botocore's own [`_is_expired()`](https://github.com/boto/botocore/blob/dbfa9d0d2bdad5476556e632d96126a0b2a0e4ce/botocore/credentials.py#L562) which calls `refresh_needed(refresh_in=0)`, effectively checking whether `_seconds_remaining() < 0` i.e. `_expiry_time <= now`.


The check is conservative: if any step returns `None` (no session, no credentials, no token, no expiry), we return `False` (not expired) and let boto3's own refresh machinery handle it on the next API call. We only proactively invalidate when we can *positively confirm* expiry.

scaredcat:
I've replied to comments and updated PR overview with details of E2E tests executed.

src/redshift-mcp-server/awslabs/redshift_mcp_server/redshift.py
grayhemp: Makes sense. Thanks.

src/redshift-mcp-server/awslabs/redshift_mcp_server/redshift.py
grayhemp: 1.

> botocore itself detects credential expiry in the RefreshableCredentials class:

I see.

Just for clarity: Why didn't we go this native way instead of mirroring? What are the tradeoffs?

2. 

> expiry = getattr(credentials, '_expiry_time', None)

This is accessing an internal/private property of credentials bypassing the public API. Is there a better way of doing the same? There is no guarantee it will be kept the same in the future versions.

grayhemp:
Left a couple of clarification questions.

src/redshift-mcp-server/awslabs/redshift_mcp_server/redshift.py
scaredcat: Good points — both concerns are now addressed. I've reworked the approach to be **fully reactive** instead of proactive:

1. **No more accessing private botocore internals** — we no longer inspect `_expiry_time` or any internal credential attributes. The proactive expiry check is gone entirely.

2. **Why not the native `RefreshableCredentials` path?** — Because botocore's lazy refresh only fires during an API call, and the MCP server caches the *clients* (not just the session). Even if botocore internally refreshes credentials, our cached `session.client(...)` objects hold references to the old credential resolver. The cleanest fix is to catch the expiry error reactively and recreate the clients.

The new approach is simple:
- `with_credential_retry(fn)` wraps AWS API calls
- On `ExpiredToken` or `ExpiredTokenException` → invalidate all cached clients → retry `fn()` once (fresh clients pick up new credentials from the default chain)
- Non-expired errors propagate immediately, no swallowing

This addresses your earlier feedback as well — `discover_clusters` is now also covered (both the Redshift and Serverless paginator calls are wrapped), not just the Data API statements.

Re-tested today (2026-07-10) against real provisioned + serverless infrastructure — all E2E tests pass including the credential refresh scenario.

scaredcat:
## Update (2026-07-10): Reworked to reactive approach

Based on @grayhemp's feedback about avoiding private botocore internals (`_expiry_time`) and keeping the implementation simple, I've reworked the credential refresh to use a **reactive** catch-and-retry pattern:

**What changed:**
- Removed the proactive `_are_credentials_expired()` check and `self._session` tracking
- Added `with_credential_retry(fn)` — catches `ExpiredToken`/`ExpiredTokenException`, invalidates cached clients, retries once
- Wrapped all AWS API call sites: `discover_clusters` (both provisioned and serverless paginators) and `_execute_statement` (Data API)
- Non-expired errors propagate immediately — no exception swallowing

**Why this is better:**
- No private API access — only uses public `ClientError` error codes
- Simpler — ~10 lines of retry logic vs the previous expiry-checking machinery
- Covers all code paths (addresses the original feedback that cluster discovery wasn't covered)

**Re-tested today** against real infrastructure (provisioned ra3.large + serverless 8 RPU in us-east-1). All 11 E2E tests pass plus 8 credential-refresh-specific tests. Updated the PR description with full results.

grayhemp (changes requested):
```py
    response = client_manager.with_credential_retry(
        lambda: client_manager.redshift_data_client().execute_statement(**request_params)
    )
```

Shall we make it all a part of the client manager instead of wrapping every client construction?

scaredcat:
Thanks for the patience here, @grayhemp — I've reworked this substantially since the last review and rebased onto latest `main`. Where it landed, mapped to your notes:

**No more private botocore internals.** The proactive `_expiry_time` inspection is gone entirely. Refresh is now purely reactive: we catch `ClientError` with code `ExpiredToken`/`ExpiredTokenException` (public, documented) and retry once. That resolves both the "no guarantee `_expiry_time` stays stable across versions" concern and the "why mirror botocore instead of going native" question — we no longer inspect or mirror anything internal.

**No exception swallowing.** Only those two expiry codes trigger invalidate-and-retry; every other `ClientError` is re-raised unchanged, so the real error still surfaces.

**Centralized, so every call path is covered.** Retry now lives in a small per-service adapter (`_ServiceClientAdapter`) returned by the client accessors. `discover_clusters` (both provisioned and serverless paths) is covered too — not just Data API statements — and no call site has to remember to wrap anything.

**Rebased** onto current `main` and reconciled with the new typed `RedshiftCluster` models.

**Re-tested** end-to-end against real provisioned + serverless clusters (deploy → test → teardown), including a test that injects an `ExpiredToken` and confirms reactive recovery. The description and the linked gist are updated with the current scripts.

Ready for another look when you get a chance.

---
*This comment was AI-assisted.*

grayhemp:
@scaredcat 

The latest solution is "a bit too much" IMHO. Let's simplify it like this: [redshift-credential-refresh.diff.md](https://github.com/user-attachments/files/30568011/redshift-credential-refresh.diff.md). 

Please review and test it very carefully, it's a significant change.

@saeedma8 would be good to have your thorough review of it as well.

scaredcat:
Thanks @grayhemp — this is a nice simplification and I've taken most of it: the `__getattr__` dispatch, the lazy paginator with the first-page-only retry (the `yielded` guard), and your "never re-submit a statement" tests. The diff against `main` is considerably smaller now, and `discover_clusters`/`_execute_statement` are untouched — the feature is additive at the call sites.

One deliberate difference: rather than the manager owning build + cache + `invalidate_clients` with the client as a stateless facade over those callbacks, I kept a self-contained `_RetryingRedshiftClient` that owns its own lifecycle (lazy build, cache, invalidate, retry), and `RedshiftClientManager` is a thin registry that hands out one client per service. Reasoning:

- Dependencies point inward — the client doesn't depend on the manager; the manager depends on the clients. The facade-over-callbacks shape couples them the other way.
- `invalidate` stays the client's own concern rather than a public `invalidate_clients()` on the manager that exists only to serve the retry path.
- Client behavior lives in one place instead of split across the manager and the facade.

Functionally it's the same reactive retry + lazy pagination you proposed — it's just where the lifecycle lives. Happy to switch to the facade split if you feel strongly, but I think this keeps the responsibilities cleaner.

Rebased onto latest `main`. Re-tested: full unit suite + E2E against real provisioned + serverless clusters (deploy → test → teardown), including the reactive-refresh path. Gist updated with the current scripts.

---
*This comment was AI-assisted.*

grayhemp:
@scaredcat I like your pivot from the manager to client facade.

Attaching a refinement patch [redshift-cred-refresh-refining.patch](https://github.com/user-attachments/files/30806772/redshift-cred-refresh-refining.patch). Use git apply.

Changes in the patch:

- __getattr__ wraps only callables, not properties.
- Added _retry helper, shared by the operation and paginate paths.
- _paginate collects pages eagerly. get_paginator keeps its shape, so existing discovery tests are untouched.
- _RetryingRedshiftClient -> _CredentialRefreshingClient.
- Bare raise.
- _error_code() helper.
- service_name property, so tests read it instead of _service_name.
- Client tests driven through public operations with boto3.Session patched - 1 import instead of 9.
- Class docstring reworded: keys in the shared credentials file are the case this fixes. SSO, assume-role, etc credentials already auto-refresh via boto3 internal means.

Your items to do:

- Make sure all the unit tests are in a fine shape.
- Run E2E tests to make sure everything is working properly.
- Fix the PR description, it is outdated.
- Rebase.

scaredcat:
Applied your refinement patch as-is, @grayhemp — thanks, it reads cleaner: the shared `_retry` helper, `_error_code`, the callable check in `__getattr__`, `service_name`, and the docstring scope note (the shared-credentials-file case is exactly the one I keep hitting with `ada`). Rebased onto latest `main`, re-ran the full unit suite + E2E (provisioned + serverless, plus the injected-`ExpiredToken` refresh path against real infra), and updated the PR description, which was outdated.

On **point A** (collecting pages eagerly): agreed. Eager collection is marginally less memory-efficient than a lazy iterator in the abstract, but both listings here (`describe_clusters`, `list_workgroups`) are small and fully materialized by their callers, so there's no practical cost — and even at much larger scale it just uses a bit more memory while still working. The bonus is that eager collection makes a mid-listing expiry cleanly recoverable: since no page has been handed to the caller yet, a refresh restarts the whole (idempotent, read-only) listing with no risk of duplicates. Happy with it as-is.

Also kept the bare `raise` — confirmed it still surfaces the original error to the caller (it re-raises the same exception; the whole server already relies on this to propagate tool errors up to the MCP layer), so no concern there.

---
*This comment was AI-assisted.*

scaredcat:
bumping PR

grayhemp:
Will be taking a look and running E2E tests. The solution looks good after a skimming. Standby...

PR #3197, opened 2026-04-21
fix(aws-api-mcp-server): cache boto3 clients and reuse heavyweight objects

antenore:
Fixes memory growth in long-running aws-api-mcp-server sessions.

## Summary

Third attempt. #2449, #2452, and #2828 were all auto-closed by the stale bot without maintainer review. Same leak, same fix, rebased on current `main` and tightened against Copilot's review comments on #2828.

### Changes

- `interpretation.py`: LRU cache (30 entries) of `boto3.client()` keyed on `(service, region, credentials, endpoint)`. Cached clients are evicted on auth errors so stale credentials do not survive token rotation.
- `config.py`: per-profile cache of the resolved region string. `get_region()` was constructing a fresh `boto3.Session` on every `parse()` just to read one attribute.
- `helpers.py` + `server.py`: `@functools.cache` on `get_requests_session()`; `suggest_aws_commands` now reuses it instead of building a new session per request.
- `parser.py`: module-level cache of `ServiceCommand._get_command_table()` results. Each call re-parses service model JSON and pymalloc never returns those arenas to the OS.
- `tests/conftest.py`: root fixture that clears every module cache between tests and closes the cached requests session before dropping the reference.

Design rationale and the previous Copilot review thread on the cache key, user agent, and test isolation are in #2452 and #2828.

### User experience

Before: aws-api-mcp-server RSS grows several MB per `call_aws` invocation in long-running sessions. The allocator does not return pages, so the only remedy is restarting the MCP host.

After: steady-state RSS. Clients, sessions, regions, and service models are built once per unique key and reused.

714 tests pass on current upstream `main`, ruff and pyright clean.

@arnewouters, you were the only reviewer who engaged on #2452. If you have time, the points you raised there are all applied here.

## Checklist

Is this a breaking change? N

## Acknowledgment

### DIFF
--- a/src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/common/config.py
+++ b/src/aws-api-mcp-server/awslabs/aws_api_mcp_server/core/common/config.py
@@ -40,17 +40,26 @@ class FileAccessMode(str, Enum):
     NO_ACCESS = 'no-access'
 
 
+_region_cache: dict[str, str] = {}
+
+
 def get_region(profile_name: str | None = None) -> str:
     """Get the region depending on configuration."""
     if AWS_REGION:
         return AWS_REGION
 
+    cache_key = profile_name or '__default__'
+    if cache_key in _region_cache:
+        return _region_cache[cache_key]
+
     fallback_region = 'us-east-1'
-
     if profile_name:
-        return boto3.Session(profile_name=profile_name).region_name or fallback_region
+        region = boto3.Session(profile_name=profile_name).region_name or fallback_region
+    else:
+        region = boto3.Session().region_name or fallback_region
 
-    return boto3.Session().region_name or fallback_region
+    _region_cache[cache_key] = region
+    return region
 
 
 def get_server_directory():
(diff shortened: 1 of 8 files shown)


antenore:
@arnewouters , you were the only reviewer who engaged on https://github.com/awslabs/mcp/pull/2452. If you have time, the points you raised there are all applied here.

antenore:
Addressed Copilot's five inline comments in c28a2986:

1. **Secret in cache key** — replaced the raw `access_key_id`/`secret_access_key`/`session_token` tuple with a SHA-256 fingerprint. Declined on #2828; applying it this round since the concern is cheap to remove.
2. **`user_agent_extra` not in key** — added to the cache key. Stable in stdio so the hot path is unaffected; correct telemetry per MCP client under streamable-http.
3. **Missing client cache tests** — added `tests/parser/test_interpretation_cache.py` with 10 cases covering fingerprint invariants, cache hits/misses by service/region/credentials/user agent, LRU eviction, and client eviction on auth vs non-auth errors.
4. **Setup did not clear `_service_command_table_cache`** — real bug, fixed in `tests/conftest.py`.
5. **Duplicated command-table lookup** — extracted `_get_cached_service_command_table` in `parser.py`, all three call sites use it.

724 tests pass, ruff and pyright clean.

PR #3266, opened 2026-04-22
fix(aurora-dsql): case-insensitive table name matching in get_schema

anwesham-lab:
## Summary

- Apply `LOWER()` on both sides of the `WHERE` clause in `get_schema` so identifier matching follows PostgreSQL's case-folding convention. Previously, `get_schema('Associate')` returned empty results because `information_schema.columns` stores folded (lowercase) identifiers, but the query matched literally.
- Support schema-qualified table names (`schema.table`) by splitting on `.` and filtering on both `table_schema` and `table_name` via a new `GET_QUALIFIED_SCHEMA_SQL` constant. This allows `get_schema('data.associate')` to return only columns from the intended schema rather than all schemas containing a table with that name.
- Add two new tests covering both the schema-qualified and plain table name paths.

## Motivation

When an agent calls `get_schema('data.Associate')`, the tool returned empty results (3 times for 3 tables). The agent then jumped to `information_schema.tables`, saw the lowercase `associate`, concluded "DSQL is case-sensitive", and ran the lowercase version — arriving at the correct answer via an incorrect diagnosis. The root cause was the literal `WHERE table_name = %s` match in the MCP server, not DSQL case-sensitivity.

## Test plan

- [x] New unit test `test_get_schema_with_schema_qualified_name` verifies `get_schema('data.Associate')` uses `GET_QUALIFIED_SCHEMA_SQL` with split args `['data', 'Associate']`
- [x] New unit test `test_get_schema_without_schema_uses_table_name_only` verifies `get_schema('Associate')` uses `GET_SCHEMA_SQL` with `['Associate']`
- [x] Manual verification: call `get_schema` with mixed-case table names against a live DSQL cluster

## Checklist

* [y] I have reviewed the [contributing guidelines](https://github.com/awslabs/mcp/blob/main/CONTRIBUTING.md)
* [y] I have performed a self-review of this change
* [y] Changes have been tested
* [y] Changes are documented

**RFC issue number**:

Checklist:

## Acknowledgment

### DIFF
--- a/src/aurora-dsql-mcp-server/awslabs/aurora_dsql_mcp_server/consts.py
+++ b/src/aurora-dsql-mcp-server/awslabs/aurora_dsql_mcp_server/consts.py
@@ -31,9 +31,8 @@ BEGIN_READ_ONLY_TRANSACTION_SQL = 'BEGIN TRANSACTION READ ONLY'
 COMMIT_TRANSACTION_SQL = 'COMMIT'
 ROLLBACK_TRANSACTION_SQL = 'ROLLBACK'
 BEGIN_TRANSACTION_SQL = 'BEGIN'
-GET_SCHEMA_SQL = (
-    'SELECT column_name, data_type FROM information_schema.columns WHERE table_name = %s'
-)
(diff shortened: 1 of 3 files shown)


amaksimo (approved):
Thanks for the test too

Benjscho (approved):

PR #3268, opened 2026-04-22
feat(applicationsignalsmcp): audit banners, canary auto-discovery, and SLO console links

goravsoni:
Fixes

N/A — no open issue. These are small standalone improvements to the CloudWatch Application Signals MCP server.

### Changes

Three independent quality-of-life improvements to the CloudWatch Application Signals MCP server, each scoped to a single tool. All three are additive, low-risk, and reviewable / revertable in isolation.

**1. Flag uninstrumented and canary targets in `audit_services` output**

`audit_services` supports both wildcard targets (`*`, `payment-*`) and explicit named targets. The wildcard expansion path already filters non-application services before auditing. The explicit-name path does not. When a caller names a target that turns out to be an auto-discovered, uninstrumented entry — or a CloudWatch Synthetics canary (`cwsyn-*`) whose name resembles a real service — the tool returns metrics for that entity with no indication the semantics are different from an instrumented application service.

For example, a canary's `Duration` metric measures canary script execution time (often seconds), not application latency (milliseconds). Comparing the two side-by-side is misleading.

Fix: two new helpers in `service_audit_utils.py`:

- `detect_uninstrumented_targets(...)` — inspects explicitly-named (non-wildcard) service targets after normalization. Flags names starting with `cwsyn-` via a fast path (no API call), and for the remaining explicit names makes one `list_services` call to check `InstrumentationType` in `AttributeMaps` for `UNINSTRUMENTED` or `AWS_NATIVE`. Non-fatal on API failure.
- `format_uninstrumented_warning(...)` — renders the flag list into a banner block explaining what the flagged target actually is and why its metrics should not be compared to instrumented application services. Returns an empty string when nothing is flagged, so output is byte-identical for clean targets.

`server.py` calls both helpers inside `audit_services` after target normalization and splices the warning at the top of the response banner. The audit itself always proceeds — the warning is informational, never blocking. Wildcard path is untouched.

**2. Auto-discover canaries when `analyze_canary_failures` is called without a name**

`analyze_canary_failures` previously required `canary_name` as a mandatory argument. A caller that does not already know a specific canary name has three bad options: ask upstream for a name, guess, or skip the tool entirely.

Fix: `canary_name` now defaults to `''`. When empty, the tool calls `synthetics_client.describe_canaries()`, classifies each by last run state (`ERROR`/`STOPPED` or last `CanaryRun.Status.State == FAILED` → failing; otherwise healthy), and either returns a healthy-summary page or recursively analyzes each failing canary with its name concatenated under a separator banner. Healthy canaries are listed by name in the header so the caller still sees the full inventory.

Existing callers that pass an explicit `canary_name` hit the unchanged code path. The signature change is additive-compatible (positional argument now has a default).

**3. Add AWS console deep-link per SLO in `list_slos` output**

`list_slos` renders each SLO with name, ARN, operation, creation timestamp, and key attributes. Drilling into a specific SLO in the console required manual navigation.

Fix: new `console_url_slo(slo_name, region)` helper in `utils.py` returns a deterministic deep-link URL:

```
https://{region}.console.aws.amazon.com/cloudwatch/home?region={region}#application-signals:slo?slo={url_encoded_name}
```

`list_slos` imports `AWS_REGION` and the new helper, and renders `Console: <url>` on a new line per SLO, right after `Created: ...`.

Purely additive text. No schema, signature, or behavior change for any caller. URLs are deterministic — no network calls, no API dependency.

### User experience

**Before**

- `audit_services` with a canary or uninstrumented target name returned metrics with no flag indicating the target was a canary or uninstrumented. A client interpreting the response could present a canary's `Duration` alongside an application service's `Latency` as if they were comparable.
- `analyze_canary_failures` required a canary name up front. A client without one either asked the user, invented a name, or skipped the tool.
- `list_slos` showed SLO names as plain text. Drilling into a specific SLO required manually constructing a console URL or navigating there.

**After**

- `audit_services` prepends a clearly-labelled warning to the response banner when a named target is a canary (`cwsyn-*`) or has `InstrumentationType: UNINSTRUMENTED` / `AWS_NATIVE`, so the response can frame the target correctly.
- `analyze_canary_failures` can be called with no arguments. It lists the full inventory, classifies each canary as healthy or failing, and analyzes only the failing ones.
- `list_slos` shows a `Console:` line per SLO with a direct deep-link to the SLO detail page.

## Checklist

* [] Changes are documented

**RFC issue number**: N/A

Checklist:

## Acknowledgment

### DIFF
--- a/src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/server.py
+++ b/src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/server.py
@@ -54,7 +54,12 @@ from .group_tools import (
     list_group_services,
     list_grouping_attribute_definitions,
 )
-from .service_audit_utils import normalize_service_targets, validate_and_enrich_service_targets
+from .service_audit_utils import (
+    detect_uninstrumented_targets,
+    format_uninstrumented_warning,
+    normalize_service_targets,
+    validate_and_enrich_service_targets,
+)
 from .service_tools import (
     get_service_detail,
     list_monitored_services,
@@ -414,6 +419,14 @@ async def audit_services(
             normalized_targets, applicationsignals_client, unix_start, unix_end
         )
 
+        # Detect explicitly-named uninstrumented / canary targets and build a
+        # warning block. Wildcard targets are already filtered upstream, so
+        # this only catches the case where the LLM (or user) named a canary
+        # or uninstrumented service directly.
+        uninstrumented_flagged = detect_uninstrumented_targets(
+            normalized_targets, applicationsignals_client, unix_start, unix_end
+        )
+
         # Parse auditors with service-specific defaults
         auditors_list = parse_auditors(auditors, ['slo', 'operation_metric'])
 
@@ -423,6 +436,10 @@ async def audit_services(
             f'🎯 Scope: {len(normalized_targets)} service target(s) | Region: {region}\n'
             f'⏰ Time: {unix_start}–{unix_end}\n'
         )
+
+        # Add uninstrumented-target warning at the top of the banner so it is
+        # prominent in the response.
+        banner += format_uninstrumented_warning(uninstrumented_flagged)
 
         # Add filtering statistics if services were filtered
         if filtering_stats['total_services'] > 0:
--- a/src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/server.py
+++ b/src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/server.py
@@ -995,8 +995,11 @@ async def audit_service_operations(
 
 
 @mcp.tool()
-async def analyze_canary_failures(canary_name: str, region: str = AWS_REGION) -> str:
+async def analyze_canary_failures(canary_name: str = '', region: str = AWS_REGION) -> str:
     """Comprehensive canary failure analysis with deep dive into issues.
+
+    **Auto-discovery**: If no canary_name is provided, automatically discovers all canaries in the account,
+    checks their last run status, and analyzes any that are currently failing.
 
     Use this tool to:
     - Deep dive into canary failures with root cause identification
@@ -1007,6 +1010,7 @@ async def analyze_canary_failures(canary_name: str, region: str = AWS_REGION) ->
     - Identify performance degradation and availability issues across service dependencies
 
     Key Features:
+    - **Auto-Discovery**: When called without a canary name, discovers and analyzes all failing canaries
     - **Failure Pattern Analysis**: Identifies recurring failure modes and temporal patterns
     - **Artifact Deep Dive**: Analyzes canary logs, screenshots, and network traces for root causes
     - **Service Correlation**: Links canary failures to upstream/downstream service issues using Application Signals
@@ -1029,7 +1033,8 @@ async def analyze_canary_failures(canary_name: str, region: str = AWS_REGION) ->
     - Historical failure patterns and recovery recommendations
 
     Args:
-        canary_name (str): Name of the CloudWatch Synthetics canary to analyze
+        canary_name (str, optional): Name of the CloudWatch Synthetics canary to analyze.
+            If empty, auto-discovers all canaries and analyzes failing ones.
         region (str, optional): AWS region where the canary is deployed.
 
     Returns:
@@ -1041,6 +1046,63 @@ async def analyze_canary_failures(canary_name: str, region: str = AWS_REGION) ->
             - Historical pattern analysis and trend insights
     """
     try:
+        # Auto-discovery: if no canary_name provided, find and analyze failing canaries
+        if not canary_name:
+            logger.info('No canary_name provided — auto-discovering canaries')
+            try:
+                describe_resp = synthetics_client.describe_canaries()
+                all_canaries = describe_resp.get('Canaries', [])
+            except Exception as e:
+                logger.error(f'Failed to describe canaries: {e}', exc_info=True)
+                return f'Error discovering canaries: {str(e)}'
+
+            if not all_canaries:
+                return 'No canaries found in this account/region. Create a canary first using CloudWatch Synthetics.'
+
+            # Check each canary's last run status
+            failing_canaries = []
+            healthy_canaries = []
+            for c in all_canaries:
+                c_name = c.get('Name', 'Unknown')
+                last_run = c.get('Status', {})
+                state = last_run.get('State', 'UNKNOWN')
+                state_reason = last_run.get('StateReason', '')
+                if state in ('ERROR', 'STOPPED') or 'error' in state_reason.lower():
+                    failing_canaries.append(c_name)
+                else:
+                    # Also check last run via get_canary_runs
+                    try:
+                        runs_resp = synthetics_client.get_canary_runs(Name=c_name, MaxResults=1)
+                        last_runs = runs_resp.get('CanaryRuns', [])
+                        if last_runs and last_runs[0].get('Status', {}).get('State') == 'FAILED':
+                            failing_canaries.append(c_name)
+                        else:
+                            healthy_canaries.append(c_name)
+                    except Exception:
+                        healthy_canaries.append(c_name)
+
+            if not failing_canaries:
+                result = f'✅ All {len(all_canaries)} canaries are healthy.\n\n'
+                result += 'Canaries checked:\n'
+                for name in healthy_canaries:
+                    result += f'  • {name} — PASSING\n'
+                return result
+
+            # Analyze each failing canary
+            result = f'🔍 Auto-Discovery: Found {len(failing_canaries)} failing canary(ies) out of {len(all_canaries)} total.\n\n'
+            if healthy_canaries:
+                result += f'✅ Healthy canaries ({len(healthy_canaries)}): {", ".join(healthy_canaries)}\n\n'
+
+            for fc_name in failing_canaries:
+                result += f'{"=" * 60}\n'
+                result += f'Analyzing failing canary: {fc_name}\n'
+                result += f'{"=" * 60}\n'
+                # Recursively call with the specific canary name
+                single_result = await analyze_canary_failures(canary_name=fc_name, region=region)
+                result += single_result + '\n\n'
+
+            return result
+
         # Get recent canary runs
         response = synthetics_client.get_canary_runs(Name=canary_name, MaxResults=5)
         runs = response.get('CanaryRuns', [])
--- a/src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/service_audit_utils.py
+++ b/src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/service_audit_utils.py
@@ -229,3 +229,151 @@ def validate_and_enrich_service_targets(
         enriched_targets.append(t)
 
     return enriched_targets
+
+
+# Prefixes / instrumentation types that indicate the target is NOT an
+# instrumented application service. These deserve a warning in the audit
+# banner so the user (and the LLM) does not treat the target as equivalent
+# to an instrumented application service.
+_CANARY_PREFIX = 'cwsyn-'
+_UNINSTRUMENTED_TYPES = {'UNINSTRUMENTED', 'AWS_NATIVE'}
+
+
+def detect_uninstrumented_targets(
+    normalized_targets: List[dict],
+    applicationsignals_client,
+    unix_start: int,
+    unix_end: int,
+) -> List[dict]:
+    """Detect non-wildcard service targets that are uninstrumented or are canaries.
+
+    The wildcard expansion path already filters these out in audit_utils.
+    This helper catches the other path: the user (or LLM) explicitly names a
+    service that happens to be a CloudWatch Synthetics canary (`cwsyn-*`) or
+    carries InstrumentationType UNINSTRUMENTED / AWS_NATIVE in its
+    AttributeMaps. We do not block the call; we surface a warning so the
+    response frames the target correctly.
+
+    Args:
+        normalized_targets: Service targets after normalization + enrichment.
+        applicationsignals_client: Boto3 Application Signals client.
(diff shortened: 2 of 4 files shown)


src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/server.py
```diff
@@ -1024,6 +1046,63 @@ async def analyze_canary_failures(canary_name: str, region: str = AWS_REGION) ->
...
+
+            # Check each canary's last run status
+            failing_canaries = []
+            healthy_canaries = []
+            for c in all_canaries:
+                c_name = c.get('Name', 'Unknown')
+                last_run = c.get('Status', {})
+                state = last_run.get('State', 'UNKNOWN')
+                state_reason = last_run.get('StateReason', '')
+                if state in ('ERROR', 'STOPPED') or 'error' in state_reason.lower():
```
syed-ahsan-ishtiaque, line 1070: **Bug (high): `STOPPED` canaries and any `StateReason` containing the substring `error` are misclassified as failing**

Two issues on this line:

1. `STOPPED` is a normal, operator-initiated state in the Synthetics API (`CanaryState` enum values: CREATING/READY/STARTING/RUNNING/UPDATING/STOPPING/STOPPED/ERROR/DELETING). A user pausing a canary is not a failure — pulling it into a "failing" bucket and then recursively running the full failure analysis pipeline on it produces a misleading report and burns API calls. The PR description explicitly says the tool "analyzes any that are currently failing."
2. `'error' in state_reason.lower()` substring-matches benign text. `StateReason` is a free-form string and can contain the word "error" in non-error contexts (e.g. "last run completed without error"). False positives will silently show up here.

Suggested fix:

```python
if state == 'ERROR':
    failing_canaries.append(c_name)
elif state == 'STOPPED':
    # User-paused — not a failure; surface separately if you want.
    healthy_canaries.append(c_name)
else:
    # Fall through to last-run check below.
    ...
```

If you do want to surface STOPPED canaries, consider a third bucket (`stopped_canaries`) rather than conflating with failures.

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/server.py
```diff
@@ -1024,6 +1046,63 @@ async def analyze_canary_failures(canary_name: str, region: str = AWS_REGION) ->
...
+                    # Also check last run via get_canary_runs
+                    try:
+                        runs_resp = synthetics_client.get_canary_runs(Name=c_name, MaxResults=1)
+                        last_runs = runs_resp.get('CanaryRuns', [])
+                        if last_runs and last_runs[0].get('Status', {}).get('State') == 'FAILED':
+                            failing_canaries.append(c_name)
+                        else:
+                            healthy_canaries.append(c_name)
+                    except Exception:
+                        healthy_canaries.append(c_name)
```
syed-ahsan-ishtiaque, line 1082: **Code Quality: silent `except Exception` that misclassifies canaries as healthy**

When `get_canary_runs` fails — throttling, permissions, transient API error — this bare `except` drops the exception and appends the canary to `healthy_canaries`. So a canary that is actually failing can be reported as PASSING because we never managed to look at its last run.

At minimum, log the failure. Better: put the canary in a third "unknown" bucket so the user is not given a false green.

```python
except Exception as e:
    logger.warning(f'get_canary_runs failed for {c_name}: {e}')
    unknown_canaries.append(c_name)
```

And surface `unknown_canaries` in the output so the user knows the check was incomplete for those.

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/server.py
```diff
@@ -1024,6 +1046,63 @@ async def analyze_canary_failures(canary_name: str, region: str = AWS_REGION) ->
             - Historical pattern analysis and trend insights
     """
     try:
+        # Auto-discovery: if no canary_name provided, find and analyze failing canaries
+        if not canary_name:
+            logger.info('No canary_name provided — auto-discovering canaries')
+            try:
+                describe_resp = synthetics_client.describe_canaries()
```
syed-ahsan-ishtiaque, line 1053: **Bug (medium): `describe_canaries()` is not paginated**

`synthetics.describe_canaries` returns a `NextToken` and is paged. Default page size is 20; max is 100. For any account with more canaries than the default page, this silently misses the rest — the tool will confidently claim "All N canaries are healthy" while ignoring the ones on page 2.

Suggested:

```python
all_canaries = []
next_token = None
while True:
    kwargs = {'MaxResults': 100}
    if next_token:
        kwargs['NextToken'] = next_token
    describe_resp = synthetics_client.describe_canaries(**kwargs)
    all_canaries.extend(describe_resp.get('Canaries', []))
    next_token = describe_resp.get('NextToken')
    if not next_token:
        break
```

You probably also want an upper cap (e.g. stop at ~200 canaries and tell the user to narrow the scope) so auto-discovery cannot accidentally fan out over a huge inventory.

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/server.py
```diff
@@ -1024,6 +1046,63 @@ async def analyze_canary_failures(canary_name: str, region: str = AWS_REGION) ->
...
+            result = f'🔍 Auto-Discovery: Found {len(failing_canaries)} failing canary(ies) out of {len(all_canaries)} total.\n\n'
+            if healthy_canaries:
+                result += f'✅ Healthy canaries ({len(healthy_canaries)}): {", ".join(healthy_canaries)}\n\n'
+
+            for fc_name in failing_canaries:
+                result += f'{"=" * 60}\n'
+                result += f'Analyzing failing canary: {fc_name}\n'
+                result += f'{"=" * 60}\n'
+                # Recursively call with the specific canary name
+                single_result = await analyze_canary_failures(canary_name=fc_name, region=region)
```
syed-ahsan-ishtiaque, line 1101: **Token/Context: recursive fan-out produces an unbounded response**

For each failing canary we recursively call `analyze_canary_failures`, whose single-canary output is already a detailed multi-section report (telemetry insights, run history, exception details, HAR/screenshot analysis, recommendations). With N failing canaries the LLM receives N concatenated reports in one tool response.

On a noisy account (say 10 failing canaries) this can easily blow past 20k tokens and dominate the context window. Two concrete mitigations:

1. Cap the number analyzed (e.g. top 5 by recency), and list the rest by name with a note that the caller can re-invoke with an explicit `canary_name` for more detail.
2. Run the N analyses in parallel with `asyncio.gather` so the wall-clock doesn't scale linearly — right now this is also a sequential-await perf issue.

Example:

```python
MAX_AUTO_ANALYZED = 5
to_analyze = failing_canaries[:MAX_AUTO_ANALYZED]
deferred = failing_canaries[MAX_AUTO_ANALYZED:]
results = await asyncio.gather(
    *[analyze_canary_failures(canary_name=n, region=region) for n in to_analyze]
)
if deferred:
    result += f'\nℹ️ {len(deferred)} additional failing canary(ies) not analyzed (limit {MAX_AUTO_ANALYZED}): {", ".join(deferred)}\n'
```

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/server.py
```diff
@@ -414,6 +419,14 @@ async def audit_services(
             normalized_targets, applicationsignals_client, unix_start, unix_end
         )
 
+        # Detect explicitly-named uninstrumented / canary targets and build a
+        # warning block. Wildcard targets are already filtered upstream, so
+        # this only catches the case where the LLM (or user) named a canary
+        # or uninstrumented service directly.
+        uninstrumented_flagged = detect_uninstrumented_targets(
```
syed-ahsan-ishtiaque, line 426: **Code Quality: duplicate `list_services` API call in the hot path**

`validate_and_enrich_service_targets` (called on line 418) already calls `applicationsignals_client.list_services(StartTime=..., EndTime=..., MaxResults=100)` for any target that lacks `Environment`. Then `detect_uninstrumented_targets` on this line makes a second call with the same arguments to scan `InstrumentationType`.

When the caller passes a target without `Environment` (the common case — the PR description notes the LLM may name a service by bare name), we now do two identical 100-result scans. This doubles the latency and cost for no gain.

Two options:

1. Have `validate_and_enrich_service_targets` return the `ServiceSummaries` it fetched and pass them into `detect_uninstrumented_targets`.
2. Introduce a small memoized fetcher (e.g. a `list_services_cached(start, end)` helper) and have both call sites go through it.

Option 1 is less cute but zero-risk.

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/service_audit_utils.py
```diff
@@ -229,3 +229,151 @@ def validate_and_enrich_service_targets(
...
+            needs_api_check.append((name, env))
+
+    # For the remaining targets, look them up in list_services to check
+    # InstrumentationType. We do one API call and scan its results.
+    if needs_api_check:
+        try:
+            response = applicationsignals_client.list_services(
+                StartTime=datetime.fromtimestamp(unix_start, tz=timezone.utc),
+                EndTime=datetime.fromtimestamp(unix_end, tz=timezone.utc),
+                MaxResults=100,
```
syed-ahsan-ishtiaque, line 302: **Bug (medium): `MaxResults=100` with no `NextToken` loop → false negatives**

`list_services` returns a `NextToken`; this single call caps at 100 results. If the account has more than 100 services during the window, any named target that happens to live beyond the first page silently falls through — `by_name.get(name)` returns `None`, nothing gets flagged, and the user sees no warning for a target that is actually uninstrumented.

For an auto-discovery warning that is supposed to catch user error, silently failing when the caller's account is large is worse than failing loudly.

Suggested:

```python
summaries = []
next_token = None
while True:
    kwargs = {
        'StartTime': datetime.fromtimestamp(unix_start, tz=timezone.utc),
        'EndTime': datetime.fromtimestamp(unix_end, tz=timezone.utc),
        'MaxResults': 100,
    }
    if next_token:
        kwargs['NextToken'] = next_token
    response = applicationsignals_client.list_services(**kwargs)
    summaries.extend(response.get('ServiceSummaries', []))
    next_token = response.get('NextToken')
    if not next_token:
        break
    # Optional: bail after a safety cap, e.g. 10 pages.
```

Or — better — share the summaries that `validate_and_enrich_service_targets` already fetched (see my comment on `server.py:426`) and avoid the extra call entirely.

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/service_audit_utils.py
```diff
@@ -229,3 +229,151 @@ def validate_and_enrich_service_targets(
...
+            logger.debug(f'detect_uninstrumented_targets: list_services failed: {e}')
+            return flagged
+
+        # Index summaries by name for O(1) lookup.
+        by_name = {}
+        for summary in summaries:
+            key_attrs = summary.get('KeyAttributes') or {}
+            sname = key_attrs.get('Name')
+            if sname:
+                by_name.setdefault(sname, summary)
```
syed-ahsan-ishtiaque, line 317: **Bug (medium): `by_name` keyed by Name only — collisions across environments cause wrong/missed flags**

In Application Signals, a service name is unique within an Environment, not globally. It's common to see e.g. `payment-service` in both `eks:prod/default` and `eks:staging/default`. This loop does `by_name.setdefault(sname, summary)`, so only the first occurrence wins, and then the lookup below uses `by_name.get(name)` without consulting the target's own `Environment`.

Consequences:
- A prod target named `payment-service` that happens to be instrumented can be flagged as uninstrumented because the staging copy (which happens to be `AWS_NATIVE`) was indexed first.
- Symmetric: a truly uninstrumented prod target is missed because the instrumented staging copy was indexed first.

Suggested fix — key by `(Name, Environment)`:

```python
by_key = {}
for summary in summaries:
    key_attrs = summary.get('KeyAttributes') or {}
    sname = key_attrs.get('Name')
    senv = key_attrs.get('Environment') or ''
    if sname:
        by_key.setdefault((sname, senv), summary)

for name, env in needs_api_check:
    summary = by_key.get((name, env))
    if not summary:
        # Fall back to name-only if env is unknown on the target side.
        summary = next((s for (n, _e), s in by_key.items() if n == name), None)
    ...
```

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/slo_tools.py
```diff
@@ -355,6 +356,7 @@ async def list_slos(
             result += f'  ARN: {slo_arn}\n'
             result += f'  Operation: {operation_name}\n'
             result += f'  Created: {created_time}\n'
+            result += f'  Console: {console_url_slo(slo_name, AWS_REGION)}\n'
```
syed-ahsan-ishtiaque, line 359: **Bug (low): console URL uses module-global `AWS_REGION`, not the SLO's own region**

`list_service_level_objectives` is called with `IncludeLinkedAccounts=True` by default (line 285). Linked-account SLOs are a separate account but, more importantly, SLO ARNs embed the region — and in multi-region setups the SLO may live in a region other than the one `AWS_REGION` points at. When that happens this URL sends the user to the wrong regional console (which will show "SLO not found" in that region).

The SLO's ARN is already in `slo.get('Arn')` — parse the region from it and use that:

```python
def _region_from_arn(arn: str, fallback: str) -> str:
    # arn:aws:application-signals:<region>:<account>:slo/<name>
    try:
        return arn.split(':')[3] or fallback
    except (IndexError, AttributeError):
        return fallback

slo_region = _region_from_arn(slo_arn, AWS_REGION)
result += f'  Console: {console_url_slo(slo_name, slo_region)}\n'
```

If SLOs are always same-region with the caller in your deployments, fine to leave — just flag the limitation in the docstring.

PR #3284, opened 2026-04-24
feat(mcp-lambda-handler): upgrade MCP protocol from 2024-11-05 to 2025-11-25

feelgom:
## Summary

This PR upgrades the `mcp-lambda-handler` package from MCP protocol version `2024-11-05` to the latest `2025-11-25`, bringing it in line with the current MCP specification. The `mcp-lambda-handler` is unique among the servers in this repo in that it implements the MCP protocol directly (without depending on the `mcp` SDK), so these changes were applied manually based on a spec-level diff.

#### Protocol Version & Negotiation
- The server now advertises `2025-11-25` as its protocol version.
- **Version negotiation** is implemented per the spec: if the client sends a supported version (`2024-11-05`, `2025-03-26`, `2025-06-18`, `2025-11-25`), the server echoes it back; if the client sends an unsupported version, the server responds with the latest version it supports, allowing the client to decide whether to continue or disconnect.

#### Capabilities Structure (Breaking change in spec)
- The `Capabilities` dataclass now uses the 2025-11-25 format: `{"tools": {"listChanged": false}}` instead of the old `{"tools": {"list": true, "call": true}}`.
- Capabilities are now **dynamically populated** — `tools` is only advertised when tools are registered, and `resources` is only advertised when resources are registered. This avoids claiming capabilities the server cannot fulfill.

#### HTTP Headers
- Replaced the legacy `MCP-Version: 0.6` header with the spec-compliant `MCP-Protocol-Version: 2025-11-25` header on all responses.

#### Error Codes
- Resource-not-found errors now use the MCP-specific error code `-32002` instead of the generic `-32601` (Method not found), matching the 2025-11-25 specification.

#### New Content Type: `AudioContent`
- Added `AudioContent` dataclass (`type: "audio"`) introduced in 2025-11-25.
- The `_convert_result_to_content` method now detects common audio byte signatures (MP3/ID3, WAV, OGG, FLAC) and returns `AudioContent` instead of `ImageContent`.

#### Server Instructions
- Added an optional `instructions` parameter to `MCPLambdaHandler.__init__()`, which is included in the `initialize` response when set. This allows server authors to provide guidance to LLMs on how to use the server.

#### `tools/call` Response
- The `tools/call` success response now includes `"isError": false` explicitly, as recommended by the spec.

### Files changed

| File | Changes |
|------|---------|
| `types.py` | Added `AudioContent`, made `Capabilities` fields optional, added `instructions` to `InitializeResult` |
| `mcp_lambda_handler.py` | Protocol version constants, version negotiation, dynamic capabilities, audio detection, `MCP-Protocol-Version` header, `-32002` error code, `isError` field, `instructions` support |
| `test_lambda_handler.py` | Updated existing tests for new capabilities format and error codes; added 7 new tests covering protocol negotiation, instructions, dynamic capabilities, header validation, `isError`, and audio streams |

## Test plan

- [x] All 90 existing + new tests pass (`uv run pytest tests/test_lambda_handler.py -v` → 90 passed)
- [x] Verify backward compatibility: clients sending `protocolVersion: "2024-11-05"` should receive that version back in the response
- [x] Verify that a client sending an unknown future version (e.g., `"9999-12-31"`) receives `"2025-11-25"` back
- [x] Verify `MCP-Protocol-Version` header is present on all HTTP responses
- [x] Verify capabilities only advertise `tools`/`resources` when they are actually registered


mikegc-aws (commented):
Thanks for this upgrade — the PR description is excellent and the test coverage is thorough.

One concern around spec compliance with the version negotiation:

**The issue:** When a client sends `protocolVersion: "2024-11-05"` and the server echoes it back, the spec interprets that as "I will speak 2024-11-05 for this session." But the server then responds with 2025-11-25 format throughout — capabilities use `{listChanged: false}` instead of `{list: true, call: true}`, tool responses include `isError: false`, and resource errors use `-32002`. The negotiation handshake is correct, but the server doesn't honor the negotiated version in its responses.

This creates a risk at scale: older clients see "we're compatible" but then receive a format they may not understand. The capabilities change is particularly dangerous — a client checking `capabilities.tools.call == True` to decide whether to invoke tools will get `undefined` and silently skip the feature.

**Suggested fix:** The simplest spec-compliant approach would be to drop multi-version negotiation and only advertise `2025-11-25`. Replace the negotiation logic with:

```python
negotiated_version = LATEST_PROTOCOL_VERSION  # Always respond with our version
```

This is explicitly allowed by the spec — the server responds with the latest version it supports, and the client decides whether to continue or disconnect. It's honest ("I only speak 2025-11-25") rather than ambiguous ("I'll speak your version" but then not actually adapting the response format).

The alternative — actually branching response formats per negotiated version — is possible but adds significant complexity and testing burden for a Lambda handler library, especially since Lambda deployments are pinned by version anyway.

What do you think?

andrew-ji-cupix:
Great catch — agreed. The "echo back but only respond in new format" path is ambiguous in exactly the way you describe (the `capabilities.tools.call` silent-skip case is the scary one), and an honest single-version advertise is the spec-compliant minimum for a Lambda handler library like this.

Pushed a commit (e1291f5) that:
- Drops the multi-version negotiation logic and removes `SUPPORTED_PROTOCOL_VERSIONS`.
- Replaces it with `negotiated_version = LATEST_PROTOCOL_VERSION` so the server always responds with `2025-11-25`.
- Removes the now-unused `current_protocol_version` ContextVar — the `MCP-Protocol-Version` response header is always `LATEST_PROTOCOL_VERSION`, which incidentally resolves the Copilot review notes at lines 346/420 (header/`protocolVersion` no longer drift).
- Updates the negotiation test (renamed to `test_initialize_always_advertises_latest_protocol_version`) to assert that older or unknown client versions both receive `2025-11-25` back.

All 90 tests still pass. PTAL.

mikegc-aws (approved):
Thanks for the thorough work on this upgrade, @feelgom — the PR description is one of the best I've seen (clear spec-level reasoning, files-changed table, test plan with checkboxes). Really appreciated.

The latest commit (e1291f5) addresses my earlier concern cleanly. Always advertising `2025-11-25` is the right call for a Lambda handler library — honest about what format it actually emits, and the comment in the code explains the rationale well.

Everything looks good. Approving.

One minor non-blocking note for future consideration: the `test_handle_audio_byte_streams` test re-registers the same tool name in a loop. It works because the closure is invoked immediately after overwrite, but it's a confusing pattern for readers. Could be refactored to call `_convert_result_to_content` directly if you ever revisit the tests.

Thanks again for your time on this — solid contribution. 🎉

feelgom:
@mikegc-aws @Lukas-Xue  This PR was automatically closed by the stale bot, but it has already been approved by @mikegc-aws and all review feedback has been addressed (see commit e1291f5). Could you please reopen this and proceed with the merge? Happy to make any additional changes if needed.

feelgom:
Gentle reminder — this PR has been approved by @mikegc-aws (see review on May 26), all feedback addressed in commit https://github.com/awslabs/mcp/commit/e1291f59ccf7c94d9bd62d80d23a65f971346551, and 90 tests pass. Would appreciate a reopen + merge when you get a chance. Thanks!

PR #3315, opened 2026-04-28
fix(dsql): Harden injection filters and add bind-parameter support

anwesham-lab:
## Summary

Harden the Aurora DSQL MCP server's injection defenses and add bind-parameter support to eliminate the SQL injection surface at the protocol level.

### Changes

1. **Bind-parameter support for `readonly_query` and `transact`.** Added optional `params` and `params_list` arguments that use `%s` placeholders bound by the psycopg driver. This is the primary defense — values never reach the SQL string. The internal `execute_query` already accepted params; this exposes the capability to MCP callers. Fully backwards-compatible (defaults to `None`).

2. **Injection-shape checks now run in write mode.** `check_sql_injection_risk` now accepts a `read_only` flag. In write mode, only injection-shape patterns fire (tautologies, stacked queries, time-based injection, comment injection). Keyword-level patterns (DROP, TRUNCATE, GRANT, COPY, UNION SELECT) are skipped in write mode because they overlap with legitimate DDL/DML. The mutating-keyword check remains read-only-gated as before.

3. **Structured labels on `check_sql_injection_risk` returns.** Callers can branch on `issues[0]['label']` (e.g. `'numeric_tautology'`, `'stacked_query'`) instead of regex-parsing human-readable messages.

### Test coverage

165 tests pass (107 pre-existing + 58 new). New tests cover:
- Write-mode enforcement: injection rejected, legitimate DML passes, legitimate DDL passes (DROP TABLE, TRUNCATE, GRANT, UNION SELECT)
- Structured label contract
- Params pass-through, backwards compatibility, multi-placeholder, length mismatch
- Pre-existing pattern corpus (tautologies, UNION, DROP, stacked, time-based, file ops, bypass)
- False-positive guards on legitimate SQL

## Checklist

* [x] Changes are documented

**RFC issue number**: n/a

## Acknowledgment

### DIFF
--- a/src/aurora-dsql-mcp-server/awslabs/aurora_dsql_mcp_server/mutable_sql_detector.py
+++ b/src/aurora-dsql-mcp-server/awslabs/aurora_dsql_mcp_server/mutable_sql_detector.py
@@ -111,23 +111,39 @@ TRANSACTION_CONTROL_REGEX = re.compile(
 )
 
 # -- Suspicious pattern detection (SQL injection, stacked queries, etc.) --
-SUSPICIOUS_PATTERNS = [
-    r"(?i)'.*?--",  # comment injection
-    r'(?i)\bor\b\s+\d+\s*=\s*\d+',  # numeric tautology
-    r"(?i)\bor\b\s*'[^']+'\s*=\s*'[^']+'",  # string tautology
-    r'(?i)\bunion\b.*\bselect\b',  # UNION SELECT
-    r'(?i)\bdrop\b',  # DROP
-    r'(?i)\btruncate\b',  # TRUNCATE
-    r'(?i)\bgrant\b|\brevoke\b',  # GRANT or REVOKE
-    r';\s*(?!($|\s*--|\s*/\*))(?=\S)',  # stacked queries, excluding semicolons followed by comments or whitespace
-    r'(?i)\bsleep\s*\(',  # time-based injection
-    r'(?i)\bpg_sleep\s*\(',  # PostgreSQL time-based injection
-    r'(?i)\bload_file\s*\(',  # file read
-    r'(?i)\binto\s+outfile\b',  # file write
-    r'(?i)\bcopy\s+.*\s+from\b',  # PostgreSQL COPY FROM
-    r'(?i)\bcopy\s+.*\s+to\b',  # PostgreSQL COPY TO
-    r'(?i)\b(begin|commit|rollback)\b.*;\s*\w+',  # Transaction control followed by other statements
+# Each entry is (pattern, label). The label is surfaced back to callers via
+# check_sql_injection_risk so operators can diagnose false positives without
+# having to read server logs.
+#
+# Patterns are split into two tiers:
+#   INJECTION_PATTERNS — shapes that indicate injection attempts and never
+#       appear in legitimate single-statement SQL. Safe to check in both
+#       read-only and write modes.
+#   KEYWORD_PATTERNS — keyword-level checks that overlap with legitimate
+#       DDL/DML (DROP, TRUNCATE, GRANT, COPY, UNION SELECT). Only checked
+#       in read-only mode, where these operations are prohibited.
+INJECTION_PATTERNS = [
+    (r"(?i)'.*?--", 'comment_injection'),
+    (r'(?i)\bor\b\s+\d+\s*=\s*\d+', 'numeric_tautology'),
+    (r"(?i)\bor\b\s*'[^']+'\s*=\s*'[^']+'", 'string_tautology'),
+    (r';\s*(?!($|\s*--|\s*/\*))(?=\S)', 'stacked_query'),
+    (r'(?i)\bsleep\s*\(', 'sleep_time_based'),
+    (r'(?i)\bpg_sleep\s*\(', 'pg_sleep_time_based'),
+    (r'(?i)\bload_file\s*\(', 'load_file'),
+    (r'(?i)\binto\s+outfile\b', 'into_outfile'),
+    (r'(?i)\b(begin|commit|rollback)\b.*;\s*\w+', 'transaction_control_with_extra_statement'),
 ]
+
+KEYWORD_PATTERNS = [
+    (r'(?i)\bunion\b.*\bselect\b', 'union_select'),
+    (r'(?i)\bdrop\b', 'drop'),
+    (r'(?i)\btruncate\b', 'truncate'),
+    (r'(?i)\bgrant\b|\brevoke\b', 'grant_revoke'),
+    (r'(?i)\bcopy\s+.*\s+from\b', 'copy_from'),
+    (r'(?i)\bcopy\s+.*\s+to\b', 'copy_to'),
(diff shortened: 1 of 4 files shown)


anwesham-lab:
### Review iterations 1-3

**Iteration 1** (`2a3757d5`): Initial hardening.
- 5 new injection detection patterns covering non-equality boolean injection and subquery exfiltration
- Injection/transaction-bypass checks extended to write mode
- Structured `label` field on detection results
- 118 tests passing

**Iteration 2** (`96c57f16`): Native params support.
- Added optional `params` to `readonly_query` and `params_list` to `transact`
- Backwards-compatible: defaults to `None`, existing callers unchanged
- `execute_query` already accepted `params` internally; this exposes it to MCP callers
- 210 tests passing

**Iteration 3** (`c89f7e84`): Heuristic tightening after code review.
- Narrowed injection-breakout heuristic from `{0,3}` to `{0,1}` characters
- Fixed `params` Field description that overpromised "eliminates SQL injection risk"
- 211 tests passing

#### Why the 0-1 char tightening

The original `{0,3}` heuristic was designed to distinguish injection breakout values (where the attacker supplies a throwaway string to close the developer's open quote) from legitimate business values. The idea: real values are long (UUIDs, slugs, names), injection artifacts are short.

Code review identified false positives on legitimate 2-3 char values that are common in practice:

- Country codes: `WHERE code = 'US' OR code IS NULL`
- Status enums: `WHERE status = 'new' OR status IS NOT NULL`
- Short identifiers: `WHERE type = 'ab' OR price > 0`

All of these were rejected by the `{0,3}` heuristic because the value preceding `OR` was 2-3 chars.

The tightened `{0,1}` heuristic catches only empty strings (`''`) and single characters (`'x'`), which are the canonical injection breakout shape. A legitimate query rarely has `WHERE col = '' OR ...` or `WHERE col = 'x' OR ...` — those almost always indicate an injected string that closed a quote and appended a boolean payload.

#### Why it won't catch everything

The heuristic is a regex blocklist, not a SQL parser. Injections with 2+ char throwaway values (e.g. `' OR tenant_id IS NOT NULL` where `xx` is the breakout) will pass. So will novel attack shapes that don't match any pattern in `SUSPICIOUS_PATTERNS`.

This is intentional. The heuristic is defense-in-depth, not the primary defense:

1. **Primary defense**: `params` support (iteration 2). Callers that bind values via `%s` placeholders have zero injection surface.
2. **Secondary defense**: Skill-level enforcement (`safe_query.build()` in the DSQL agent skill). Agents that follow the skill never produce injectable SQL.
3. **Tertiary defense**: These heuristics. They catch the most common injection shapes for callers that haven't adopted params yet, without creating false positives on legitimate SQL.

The remaining patterns (tautologies, `UNION SELECT`, stacked queries, `pg_sleep`, `COPY ... FROM/TO`, `DROP`, `TRUNCATE`) are pre-existing and are unlikely to appear in legitimate single-statement CRUD. The new patterns focus specifically on the tenant-isolation-bypass shapes that are the highest-risk vector for LLM-generated queries.

anwesham-lab:
### Iteration 4: Removed breakout-heuristic patterns

After iterating on the character-length heuristic across three rounds:

- **Iteration 1**: Added 5 regex patterns using `{0,3}` char breakout threshold. Code review found false positives on `'US'`, `'new'`, and other legitimate 2-3 char values.
- **Iteration 2**: Tightened to `{0,1}` — only empty strings and single chars. Reduced false positives but the core problem remained.
- **Iteration 3**: Realized the approach is fundamentally flawed. Any character-length boundary is trivially bypassable — an attacker who inspects the heuristic uses `'xx'` instead of `'x'` and the regex passes. The heuristic provides a false sense of security while creating real friction for legitimate queries.

**Decision**: Remove all 5 breakout patterns entirely. Keep only the pre-existing patterns (tautologies, `UNION SELECT`, `pg_sleep`, stacked queries, `DROP`, `TRUNCATE`, `COPY`, etc.) which catch constructs that are never legitimate in single-statement CRUD and can't be trivially reworded.

The real fix for non-equality boolean injection and subquery exfiltration is the params support added in iteration 2. When callers bind values via `%s` placeholders, the injection surface vanishes at the protocol level — no heuristic needed.

This leaves the PR with three clean, independently justifiable changes:
1. **Params support** — the actual injection fix
2. **Write-mode gate** — closes a gap where write-mode callers had no server-side defense
3. **Structured labels** — programmatic pattern identification for callers

Squashed to a single commit (`4cfc7cd7`) for clean history.

anwesham-lab:
### Iteration 5: Split SUSPICIOUS_PATTERNS into injection vs keyword tiers

Code review (10+ agents across pr-review-toolkit and code-review skills) found a critical bug: running `check_sql_injection_risk` unconditionally in write mode blocked legitimate DDL operations. `DROP TABLE`, `TRUNCATE TABLE`, `GRANT`, `REVOKE`, `COPY`, and `UNION SELECT` were all rejected as "injection risk" even with `--allow-writes`.

Root cause: `SUSPICIOUS_PATTERNS` mixed two types of patterns:
- **Injection-shape patterns** (tautologies, stacked queries, `pg_sleep`, comment injection) — never legitimate in single-statement CRUD
- **Keyword-level patterns** (DROP, TRUNCATE, GRANT, COPY, UNION SELECT) — overlap with legitimate DDL/DML

Fix: split into `INJECTION_PATTERNS` and `KEYWORD_PATTERNS`. `check_sql_injection_risk` now takes a `read_only` parameter. Write mode uses only injection-shape patterns. Read-only mode uses both. `readonly_query` always uses the full set (safe because mutating keywords are already blocked upstream).

Added `test_transact_in_write_mode_allows_legitimate_ddl` which verifies DROP TABLE, TRUNCATE, GRANT, and UNION SELECT all pass in write mode.

Also fixed: stale `or_is_null_injection` label in `check_sql_injection_risk` docstring, and renumbered test sections from (1, 4) to (1, 2).

165 tests pass. Squashed to single commit (`dea1bac8`).

anwesham-lab:
### Iteration 6: Minor polish from round 3 review

Round 3 review (9 agents) found **0 critical issues** — the pattern split and params plumbing are validated as correct. Two minor fixes applied:

1. Added `logger.warning` to `params_list` length-mismatch validation, matching every other validation path in `transact`.
2. Refactored `check_sql_injection_risk` from list-append-break to early-return — removes dead mutable state and makes the "at most one result" contract visible in the code structure.

Also confirmed from round 3:
- `readonly_query` now passes `read_only=True` explicitly (not relying on default)
- `detect_transaction_bypass_attempt` correctly gated behind `if read_only:` (stacked-query injection pattern already covers write mode)
- All comments and docstrings verified accurate (stale `or_is_null_injection` label fixed, section numbering fixed)
- psycopg params binding validated against driver source: `params=None` is identical to omitting params, `%s` placeholders are correct, `List[Any]` is the right type for MCP

165 tests pass. Squashed to single commit (`b57ae2e3`).

gxjx-x (approved):

pkale (approved):

PR #3320, opened 2026-04-29
fix(agentcore-mcp-server): fix URL allowlist bypass via prefix matching

aidandaly24:
Fixes URL allowlist bypass vulnerability in `amazon-bedrock-agentcore-mcp-server/utils/url_validator.py`.

### Changes

The `validate_urls()` function used weak string prefix checks (`startswith`) combined with automatic rewriting of relative URLs, allowing bypass of the intended allowlist of trusted AWS documentation domains.

This PR replaces the string-based prefix matching with structural URL validation:

- **`url_validator.py`**: Uses `urlparse` + `posixpath.normpath` for proper URL parsing. Enforces exact hostname matching, `/`-boundary path prefix checks, `https`-only scheme, and path normalization to resolve `..` traversal. Relative URLs are now rejected outright instead of silently rewritten.
- **`doc_fetcher.py`**: Moves relative link resolution to the caller using `urllib.parse.urljoin` to resolve relative links in llms.txt files against the base URL before validation.
- **`test_url_validator.py`**: Adds 12 new test cases covering all reported bypass vectors, path traversal, scheme enforcement, hostname boundary matching, and path boundary matching.

### User experience

**Before:** Malicious URLs like `/../../attacker-path`, `https://aws.github.io/bedrock-agentcore-starter-toolkit.evil.com/steal`, and `/evil?test=1` bypassed the allowlist.

**After:** All three vectors are rejected. Legitimate URLs for all default allowed domains continue to work as expected.

## Checklist

* [x] Changes are documented

N — The only behavioral change is that relative URLs passed directly to `validate_urls()` now raise an error instead of being silently rewritten. The caller (`doc_fetcher.py`) has been updated to resolve relative URLs before validation, preserving existing functionality.

**RFC issue number**: N/A

Checklist:

## Acknowledgment

### DIFF
--- a/src/amazon-bedrock-agentcore-mcp-server/awslabs/amazon_bedrock_agentcore_mcp_server/utils/url_validator.py
+++ b/src/amazon-bedrock-agentcore-mcp-server/awslabs/amazon_bedrock_agentcore_mcp_server/utils/url_validator.py
@@ -55,9 +55,15 @@ class URLValidator:
         if not url or not isinstance(url, str):
             return False
 
+        if any(c in url for c in '\r\n\x00'):
+            return False
+
         parsed = urlparse(url)
 
         if parsed.scheme != 'https':
+            return False
+
+        if '@' in parsed.netloc:
             return False
 
         hostname = parsed.hostname or ''
--- a/src/amazon-bedrock-agentcore-mcp-server/awslabs/amazon_bedrock_agentcore_mcp_server/utils/url_validator.py
+++ b/src/amazon-bedrock-agentcore-mcp-server/awslabs/amazon_bedrock_agentcore_mcp_server/utils/url_validator.py
@@ -14,7 +14,9 @@
 
 """URL validation for domain restriction."""
 
+import posixpath
 from typing import List
+from urllib.parse import urlparse
 
 
 class URLValidationError(Exception):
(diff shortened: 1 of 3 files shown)


jesseturner21 (dismissed):

jariy17 (approved):

sundargthb (dismissed):

src/amazon-bedrock-agentcore-mcp-server/awslabs/amazon_bedrock_agentcore_mcp_server/utils/url_validator.py
```diff
@@ -23,48 +25,52 @@ class URLValidationError(Exception):
...
-            True if the URL is allowed, False otherwise
-        """
-        if not url or not isinstance(url, str):
+        if parsed.scheme != 'https':
             return False
 
-        # Check if URL starts with any of the allowed prefixes
-        for allowed_prefix in self.allowed_domain_prefixes:
-            if url.startswith(allowed_prefix):
+        hostname = parsed.hostname or ''
```
theumbrella1, line 69: Can we add two more checks here before hostname: 

Reject control characters:
```
    if any(c in url for c in '\r\n\x00'):
        return False
```
and reject userinfo in URL:
```
    if '@' in parsed.netloc:
        return False
```

theumbrella1 (changes requested):

src/amazon-bedrock-agentcore-mcp-server/awslabs/amazon_bedrock_agentcore_mcp_server/utils/url_validator.py
aidandaly24: Done — added both checks in 7ba598f. Control characters (`\r\n\x00`) are rejected before parsing, and userinfo (`@` in netloc) is rejected after. Added tests for both.

theumbrella1 (approved):

jesseturner21 (approved):

PR #3346, opened 2026-05-05
feat(cloudwatch-applicationsignals-mcp-server): use @data_format index for search_transaction_spans

syed-ahsan-ishtiaque:
## Summary

- `search_transaction_spans` no longer hardcodes `aws/spans` as the default log group. It now relies on the CloudWatch Logs `@data_format = "AWS-OTEL-TRACE-V1"` default field index to find OpenTelemetry spans.
- When `log_group_name` is omitted, the tool prepends `SOURCE logGroups() | filterIndex @data_format = "AWS-OTEL-TRACE-V1"` and sends no `logGroupNames` — `StartQuery` accepts either a log-group list or a `SOURCE` clause in the query string, and here we use the latter.
- When `log_group_name` is supplied, the tool scopes via `logGroupNames=[...]` and still prepends the `filterIndex` so non-OTEL records in that log group are dropped.
- If the user's query already begins with `SOURCE` or already references `@data_format`, the tool respects their intent and does not inject.
- Mutual-exclusion fix: if a user supplies both a `SOURCE` clause and `log_group_name`, `logGroupNames` is now dropped (CloudWatch rejects sending both); a warning is logged.

### Why

Spans can live in log groups other than `aws/spans` (subscription forks, cross-account routing, etc). Hardcoding the log group missed those. `@data_format` is a CWL default field index, so the index-based query is efficient even in cross-log-group mode — it prunes log groups before scanning.

### LLM-facing description changes

The tool description and `log_group_name` parameter description were rewritten so the LLM knows:
- What transformation we apply to its query (and when we don't).
- That cross-log-group mode scales up to 10,000 log groups.
- That raw OTLP writes without the `AWS-OTEL-TRACE-V1` tag won't match, so an explicit `log_group_name` + custom query may be needed.

## Test plan

- [x] `uv run pytest tests/test_server.py` — all 116 tests pass (added 5 new tests covering the injection rules, including the `SOURCE` + `log_group_name` mutual-exclusion case).
- [x] Live-tested against a real account with `AWS_PROFILE=demo1` in `us-east-1`. Four scenarios all completed:
  - no `log_group_name` → `SOURCE logGroups() | filterIndex ... | <user query>` sent, no `logGroupNames`.
  - explicit `log_group_name=aws/spans` → `logGroupNames=['aws/spans']`, `filterIndex` prepended, no `SOURCE`.
  - user-supplied `SOURCE logGroups(namePrefix: ...)` → passed through untouched.
  - user-supplied `filter @data_format = "AWS-VENDED-LOG-V1"` → `SOURCE logGroups()` still added, the `AWS-OTEL-TRACE-V1` filter is NOT injected (user's wins).
- [x] Live-test stats confirmed the `@data_format` field index actually kicks in: cross-log-group scenario showed 2.1M `estimatedRecordsSkipped`, single-log-group scenario showed 0 (everything in `aws/spans` is OTEL).


src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/trace_tools.py
```diff
@@ -305,21 +395,28 @@ async def search_transaction_spans(
                         'enabled': True,
                         'destination': 'CloudWatchLogs',
                         'status': 'ACTIVE',
-                        'message': '✅ Using 100% sampled trace data from Transaction Search',
```
v-iw, line 308: Not blocking. Not sure why we are dropping this success message? Is this intentional?

vastin (approved):

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/trace_tools.py
syed-ahsan-ishtiaque: Its because transaction search doesn't force 100% sampling, that message was misleading

mxiamxia (approved):

PR #3422, opened 2026-05-08
feat: optimize aws-documentation-mcp-server startup with LazyServer

### LINKED ISSUE #2435: Slow initialization time for aws-documentation-mcp-server
When I use the awslabs.aws-documentation-mcp-server MCP in Kiro CLI, I see a message like this:

```
✓ awslabs.aws-documentation-mcp-server loaded in 3.24 s
```

This is on an M1 MacBook Pro. That's by far the slowest MCP initialization time I've seen; most are < 1s.

TrevorBurnham:
Resolves #2435. Supersedes #2436.

## Summary

This PR introduces a new shared package, **`awslabs.mcp-common`** (`src/awslabs-mcp-common/`), that encapsulates the init-time startup optimization pattern as a single reusable abstraction: `LazyServer`. Then migrates `aws-documentation-mcp-server` to use it.

As a follow-up, we can migrate the other ~50 MCP servers in this repo to the same pattern with a ~3-line change per server.

## Motivation

Importing `awslabs.aws_documentation_mcp_server.server_aws` was taking ~480ms (median) on an M1 MacBook Pro, up to ~620ms cold. `server_aws_cn` was similar at ~440ms. The cost came from eager module-level imports of heavy dependencies (`httpx`, `pydantic`, `markdownify`, `bs4`, `mcp.server.fastmcp`) and eager `FastMCP(...)` construction at import time. MCP clients that spawn multiple servers pay this latency on every cold start.

The same shape exists in virtually every other `FastMCP`-based server in this repo. A per-server patch means repeating the same 30–50 lines of boilerplate (a `_UninitializedServer` sentinel, a `_create_mcp()` factory, duplicated `Field`-annotated tool wrappers). Consolidating that boilerplate into a shared package means the follow-up PRs become mechanical one-line import + tiny structural edit, not ~100-line refactors.

## Before / after (M1 MacBook Pro, 10 iterations each, subprocess-isolated)

| Module          | Baseline median  | Post median     | Speedup  |
|-----------------|-----------------:|----------------:|---------:|
| `server_aws`    | 479 ms (323–619) |  82 ms (51–98)  | **~5.8×** |
| `server_aws_cn` | 440 ms (326–612) |  54 ms (51–70)  | **~8.1×** |

Reproduce locally:

```bash
cd src/aws-documentation-mcp-server
uv sync --dev
uv run python scripts/benchmark_init.py --iterations 5 --module server_aws
uv run python scripts/benchmark_init.py --iterations 5 --module server_aws_cn
```

## `awslabs.mcp-common` overview

A minimal, standalone package under the `awslabs` namespace. Depends only on `mcp[cli]` + `pydantic`. No AWS-specific logic.

```python
from awslabs.mcp_common import Field, LazyServer

mcp = LazyServer(
    'my-server',
    instructions='...',
    dependencies=['pydantic', 'httpx'],
)

@mcp.tool()
async def my_tool(
    ctx,
    url: str = Field(description='URL to fetch'),
    max_length: int = Field(default=5000, gt=0, lt=1_000_000),
) -> str:
    """Fetch the URL and return up to max_length characters."""
    import httpx  # deferred by the author
    ...

def main():
    mcp.run()
```

**What `LazyServer` does:**
- Records every `@mcp.tool()` registration at module load *without* importing FastMCP or pydantic.
- `Field(...)` returns a cheap `_FieldSpec` placeholder holding the kwargs.
- On `mcp.run()`: imports `FastMCP` + `pydantic.Field`, walks each tool's signature, swaps `_FieldSpec` defaults for real `pydantic.Field(...)`, auto-annotates the `ctx` parameter with `Context`, then registers each tool on a real FastMCP instance and delegates to `FastMCP.run`.
- Also exposes `mcp.build()` for tests (returns the materialized FastMCP server without running).
- Before materialization, attribute access raises a clear `RuntimeError`; after materialization, attributes delegate to the underlying server.

## `aws-documentation-mcp-server` changes

- Declares `awslabs.mcp-common>=0.1.0` as a dependency.
- Uses `[tool.uv.sources]` to resolve the dep from the in-repo sibling path while the common package isn't yet on PyPI. Once it's published, that block can be deleted.
- Replaces the module-level `FastMCP(...)` + `@mcp.tool()` with `LazyServer(...)` + `@mcp.tool()` using the shim's `Field(...)` placeholder.
- Heavy dependency imports (`httpx`, `markdownify`, `bs4`, model imports) are moved into function bodies — this part still has to be done by the author for maximum benefit, since the shim only owns the FastMCP/pydantic slice. (Benchmarks show that slice is where the bulk of the startup time lived.)
- Introduces lazy `get_default_user_agent()` in `server_utils.py`, replacing the eager `DEFAULT_USER_AGENT` module constant. Defers `importlib.metadata.version()` + `MCP_USER_AGENT` env lookup until the first outbound request.
- `util.py`: `markdownify`, `bs4`, and model imports moved inside functions.

## Tests

**New, in `awslabs-mcp-common`:**
- `tests/test_lazy_server.py` — unit tests covering `Field(...)` placeholder behavior, `@tool()` registration, materialization, Field resolution, description defaulting, and delegation-after-build.

**New, in `aws-documentation-mcp-server`:**
- `scripts/benchmark_init.py` — subprocess-isolated measurement of total server import time, per-dependency import time, and per-phase times (FastMCP creation, UUID generation, `importlib.metadata.version` lookup). CLI: `--iterations`, `--format {text,json,both}`, `--module {server_aws,server_aws_cn}`.
- `tests/test_benchmark_init.py` — regression test, `@pytest.mark.slow`, 3 subprocess measurements per server module, asserts median under `INIT_TIME_THRESHOLD_MS` (default 1500ms, env-overridable).
- `tests/test_lazy_imports.py` — subprocess-isolated assertions that `httpx`, `markdownify`, `pydantic`, and `mcp.server.fastmcp` are NOT in `sys.modules` after importing a server module, and that `mcp` is a `LazyServer`.
- `tests/test_wrapper_signature_sync.py` — guards the shim's contract: after `build()`, every tool is registered, no parameter default is still a `_FieldSpec`, `ctx` is annotated with `Context`, and the wrapper's non-ctx params match the outer function's.
- `tests/test_benchmark_pbt.py`, `tests/test_benchmark_threshold_pbt.py` — Hypothesis-based property tests for the benchmark script helpers (sort order, median correctness, threshold violations, pass/fail assertion).

**Existing test updates:** live tests that patched `server_aws{,_cn}.DEFAULT_USER_AGENT` now patch `server_utils.get_default_user_agent`. Mocks for `extract_content_from_html` / `is_html_content` / `format_documentation_result` were retargeted to `awslabs.aws_documentation_mcp_server.util.*` because those names are now locally imported inside the impl functions.

## Acknowledgment


TrevorBurnham:
I think this is still relevant and should not be closed.

PR #3517, opened 2026-05-13
feat(postgres-mcp-server): auto-detect and route through RDS Proxy

### LINKED ISSUE #3338: (postgres-mcp-server): Auto-detect RDS Proxy for standalone instance connections
### Describe the feature

When connecting to a standalone RDS PostgreSQL instance (RPG database type without a cluster identifier), the postgres-mcp-server could automatically detect whether an RDS Proxy fronts that instance and route the connection through the proxy endpoint instead of connecting directly.

This would involve using the RDS boto3 client to enumerate proxies in the region via rds_client.get_paginator('describe_db_proxies'), then for each available proxy, checking its targets via rds_client.get_paginator('describe_db_proxy_targets') to see if any target's RdsResourceId matches the instance's DBInstanceIdentifier. If a match is found, the proxy's Endpoint would be used for the connection. If no proxy is found, or if the lookup fails for any reason (insufficient permissions, API errors), the server would fall back to the direct instance endpoint with no change in behavior.

### Use Case

RDS Proxy provides connection pooling, improved failover handling, and IAM-based authentication for RDS instances. Many production deployments place an RDS Proxy in front of their PostgreSQL instances, but the postgres-mcp-server currently always connects directly to the instance endpoint. There's no way to specify a proxy endpoint, so users with RDS Proxy configured don't benefit from it when using this MCP server.

Auto-detection would let the server "do the right thing" without requiring additional configuration from the user.

### Proposed Solution

- Add a function in cp_api_connection.py that uses boto3.client('rds') to paginate through proxies with describe_db_proxies, then for each proxy in available status, paginate through its targets with describe_db_proxy_targets looking for a target whose RdsResourceId matches the given DB instance identifier.
- In internal_create_connection(), call this function for standalone RPG instance connections (the else branch where cluster_identifier is empty) and swap the endpoint if a proxy is detected.
- The lookup would be best-effort — ClientError exceptions from the boto3 calls (e.g., missing IAM permissions) or unexpected errors would be caught and fall back gracefully to the direct instance connection with a warning log.
- This would require rds:DescribeDBProxies and rds:DescribeDBProxyTargets IAM permissions. These would be optional — the feature degrades gracefully without them.
- Proxy detection would be scoped to standalone RPG instances only. Aurora clusters already have their own endpoint abstraction (writer/reader endpoints), so auto-swapping to a proxy endpoint there could conflict with Aurora's built-in routing. Aurora proxy support could be considered as a future enhancement.

### Other Information

- The proxy lookup function would be engine-agnostic — it works with any RDS Proxy regardless of engine type, so it could be reused if the server ever broadens its database support.
- Only proxies with Status == 'available' would be considered.
- A reference implementation is available at: [https://github.com/ghhamel/mcp/tree/feature/rds-proxy-support](https://github.com/ghhamel/mcp/tree/feature/rds-proxy-support)

### Acknowledgements

ghhamel:
Fixes #3338

### Changes

Adds RDS Proxy awareness to the standalone RDS PostgreSQL (RPG) connection path
in postgres-mcp-server. When `connect_to_database` is called with
`database_type=RPG` and no `cluster_identifier`, the server now checks whether
an RDS Proxy fronts the target instance and, if so, routes the connection
through the proxy endpoint instead of connecting directly to the instance.

- `connection/cp_api_connection.py`: New `find_proxy_for_instance(db_instance_id, region)`
  function. Paginates `describe_db_proxies`, filters to proxies in `available`
  status, and for each one paginates `describe_db_proxy_targets` looking for a
  target whose `RdsResourceId` matches the given DB instance identifier. Returns
  the proxy `Endpoint` on match, or `None` otherwise. All `ClientError` and
  unexpected exceptions are caught, logged at warning level, and the function
  returns `None` so the caller falls back to the direct instance endpoint.

- `server.py`: In `internal_create_connection`, in the RPG-without-cluster
  branch, after `internal_get_instance_properties` succeeds (and the new
  upstream endpoint validation runs) the new function is called with the
  resolved `DBInstanceIdentifier`. If a proxy endpoint comes back,
  `db_endpoint` is swapped to the proxy endpoint before the
  `PsycopgPoolConnection` is created. Credentials (`MasterUserSecret`,
  `MasterUsername`) still come from the instance — RDS Proxy uses the same
  Secrets Manager secret on its backend.

- `README.md`: Documents the new behavior under "Prerequisites by Connection
  Method" and lists the optional `rds:DescribeDBProxies` and
  `rds:DescribeDBProxyTargets` IAM permissions required for proxy
  auto-detection.

- `tests/`: 9 unit tests for `find_proxy_for_instance` covering the success
  path, no-proxy and no-match paths, status filtering, pagination across
  multiple proxy and target pages, and graceful degradation on `ClientError`
  and unexpected exceptions. 3 integration tests for the
  `internal_create_connection` RPG branch covering proxy-found, no-proxy-found,
  and missing-`DBInstanceIdentifier` paths.

The change is scoped to the RPG (standalone instance) path. The Aurora and
RPG-multi-AZ-cluster branches are untouched, since Aurora already has its own
endpoint abstraction and auto-swapping there could conflict with built-in
routing.

### User experience

**Before:** A user with an RDS Proxy in front of their RDS PostgreSQL instance
calls `connect_to_database` with the instance endpoint. The MCP server connects
directly to the instance, bypassing the proxy and any connection pooling,
failover, or IAM auth benefits it provides.

**After:** Same call. The MCP server resolves the instance, discovers the proxy
fronting it, and transparently connects through the proxy endpoint. Logs make
it clear which path was taken:
- `Found RDS Proxy '<name>' (endpoint: <endpoint>) fronting instance '<id>'`
- `Routing connection through RDS Proxy endpoint '<proxy>' instead of direct instance endpoint '<instance>'`
- Or, when no proxy exists or the lookup fails: `No RDS Proxy detected for instance '<id>', using direct instance endpoint '<instance>'`

If `rds:DescribeDBProxies` or `rds:DescribeDBProxyTargets` permissions are
missing, the lookup degrades gracefully: a warning is logged and the original
direct-instance behavior is preserved. The new IAM permissions are optional.

## Checklist

* [x] Changes are documented

**RFC issue number**:  #3338

Checklist:

## Acknowledgment

### DIFF
--- a/src/postgres-mcp-server/README.md
+++ b/src/postgres-mcp-server/README.md
@@ -155,6 +155,16 @@ The MCP server supports connecting to multiple database endpoints using differen
 - RDS Data API must be enabled on the Aurora PostgreSQL cluster
 - Appropriate IAM permissions for Data API access
 
+#### RDS Proxy auto-detection (RPG only)
+
+For standalone RDS PostgreSQL instances (`database_type: RPG` without a `cluster_identifier`), the server automatically checks whether an RDS Proxy fronts the target instance and routes the connection through the proxy endpoint when one is found. The proxy uses the same Secrets Manager secret on its backend, so no additional configuration is required.
+
+This lookup is best-effort: if the IAM permissions below are missing or the API call fails, the server logs a warning and falls back to a direct instance connection.
+
+Optional IAM permissions for proxy auto-detection:
+- `rds:DescribeDBProxies`
+- `rds:DescribeDBProxyTargets`
+
 ### AWS Authentication
 
 The MCP server uses the AWS profile specified in the `AWS_PROFILE` environment variable. If not provided, it defaults to the "default" profile in your AWS configuration file.
--- a/src/postgres-mcp-server/awslabs/postgres_mcp_server/connection/cp_api_connection.py
+++ b/src/postgres-mcp-server/awslabs/postgres_mcp_server/connection/cp_api_connection.py
@@ -130,6 +130,74 @@ def internal_get_cluster_valid_endpoints(
     return endpoints
 
 
+def find_proxy_for_instance(db_instance_id: str, region: str) -> Optional[str]:
+    """Check if an RDS Proxy fronts the given DB instance and return its endpoint.
+
+    Iterates all available RDS Proxies in the region, then inspects each proxy's
+    targets to see if any reference the given DBInstanceIdentifier.
+
+    Args:
+        db_instance_id: The RDS DB instance identifier to look up.
+        region: AWS region (e.g., 'us-east-1').
+
+    Returns:
+        The proxy endpoint string if a proxy is found for the instance, or None.
+    """
+    logger.info(f"Looking for RDS Proxy fronting instance '{db_instance_id}' in region '{region}'")
+
+    try:
+        rds_client = internal_create_rds_client(region=region)
+
+        # Paginate through all proxies in the region
+        proxy_paginator = rds_client.get_paginator('describe_db_proxies')
+        for proxy_page in proxy_paginator.paginate():
+            for proxy in proxy_page.get('DBProxies', []):
+                proxy_name = proxy.get('DBProxyName', '')
+                proxy_status = proxy.get('Status', '')
+
+                if proxy_status != 'available':
+                    logger.debug(f"Skipping proxy '{proxy_name}' with status '{proxy_status}'")
+                    continue
+
+                # Check if this proxy targets our instance
+                try:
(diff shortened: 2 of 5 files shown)


ghhamel:
@kennthhz i believe you were going to review this PR

kennthhz:
@scottschreckengaust reopening this PR

kennthhz:
reviewing

kennthhz:
can you resolve the conflict first? @ghhamel

PR #3518, opened 2026-05-14
fix(redshift-mcp-server): graceful cluster discovery when IAM lacks serverless or provisioned access

S-gunasekhar:
## Summary

Previously, discover_clusters() would fail entirely if the IAM role lacked redshift-serverless permissions (or redshift provisioned permissions). This made the MCP server unusable for users who only have access to one type.

Now each discovery (provisioned and serverless) is attempted independently. If one fails, the other's results are still returned. The function only raises if both fail.

Fixes: Users without redshift-serverless IAM permissions can now connect to provisioned clusters without errors.

### Changes

The discover_clusters() function in redshift.py previously treated both provisioned cluster discovery and serverless workgroup discovery as mandatory. If either API call failed (e.g., due to missing IAM permissions), the entire function raised an exception and the MCP server became unusable.

This fix makes each discovery path independent and best-effort:

- Provisioned cluster discovery is attempted; errors are captured as a warning.
- Serverless workgroup discovery is attempted; errors are captured as a warning.
- The function only raises if both fail (no clusters discoverable at all).
- If at least one succeeds, the discovered clusters are returned normally.

### User experience

Before:

A user whose IAM role only has redshift:DescribeClusters permission (no redshift-serverless:ListWorkgroups) calls list_clusters and gets:

`Error: Failed to list clusters: An error occurred (AccessDeniedException) when calling the ListWorkgroups operation: User is not authorized to perform redshift-serverless:ListWorkgroups`

The MCP server is completely blocked — no tools work because discover_clusters is a prerequisite for all operations.

After:

The same user calls list_clusters and gets their provisioned clusters returned successfully. A warning is logged:

`WARNING - Unable to discover serverless workgroups (IAM may lack redshift-serverless permissions): ... Continuing with 2 provisioned cluster(s).`

All downstream tools (list_databases, execute_query, etc.) work normally against the provisioned clusters. The same applies in reverse — a user with only serverless access will get their workgroups without needing provisioned cluster permissions.
## Checklist

* [x] Changes are documented

**RFC issue number**:

Checklist:

## Acknowledgment


grayhemp (changes requested):
I like the idea. The only ask - let's please narrow down the Exception to a specific access exception or a group of exceptions.

S-gunasekhar:
Thanks for the suggestion, I have modified the function to catch a specific ClientError which includes access exceptions (like AccessDeniedException). The outer except Exception block handles any other unexpected errors by logging them and re-raising, ensuring they are not silently swallowed.

grayhemp:
@S-gunasekhar we don't need the outer try-block. Instead, just log the error before raising exception in the if-block where you're checking if both clients errored.

S-gunasekhar:
Hi @grayhemp, I have now removed the outer try-catch and logged the error before raising exception.

```
    if provisioned_error and serverless_error:
        error_msg = (
            f'Failed to discover any Redshift clusters. '
            f'Provisioned error: {provisioned_error}; Serverless error: {serverless_error}'
        )
        logger.error(error_msg)
        raise PermissionError(error_msg)   # narrow down to specific access Exception
```

src/redshift-mcp-server/awslabs/redshift_mcp_server/redshift.py
```diff
@@ -491,16 +503,28 @@ async def discover_clusters() -> list[dict]:
...
                 }
                 clusters.append(cluster_info)
 
         serverless_count = len([c for c in clusters if c['type'] == 'serverless'])
         logger.info(f'Found {serverless_count} serverless workgroups')
 
-    except Exception as e:
-        logger.error(f'Error discovering serverless workgroups: {str(e)}')
-        raise
+    except ClientError as e:
```
grayhemp, line 540: Is there anything less broad than ClientError?

src/redshift-mcp-server/awslabs/redshift_mcp_server/redshift.py
```diff
@@ -456,10 +467,11 @@ async def discover_clusters() -> list[dict]:
 
         logger.info(f'Found {len(clusters)} provisioned clusters')
 
-    except Exception as e:
-        logger.error(f'Error discovering provisioned clusters: {str(e)}')
-        raise
+    except ClientError as e:
```
grayhemp, line 495: Is there anything less broad than ClientError?

src/redshift-mcp-server/awslabs/redshift_mcp_server/redshift.py
```diff
@@ -491,16 +503,28 @@ async def discover_clusters() -> list[dict]:
...
+        logger.warning(f'Unable to discover serverless workgroups: {str(e)}')
+
+    # If both discovery methods failed, raise an error
+    if provisioned_error and serverless_error:
+        error_msg = (
+            f'Failed to discover any Redshift clusters. '
+            f'Provisioned error: {provisioned_error}; Serverless error: {serverless_error}'
+        )
+        logger.error(error_msg)
+        raise PermissionError(error_msg)
```
grayhemp: That might be something else as well, not just PermissionError.

src/redshift-mcp-server/awslabs/redshift_mcp_server/redshift.py
S-gunasekhar: is it not better to have a ClientError? which we would be able to handle most of the client-side related errors.

src/redshift-mcp-server/awslabs/redshift_mcp_server/redshift.py
S-gunasekhar: is it not better to have a ClientError? which we would be able to handle most of the client-side related errors.

src/redshift-mcp-server/awslabs/redshift_mcp_server/redshift.py
S-gunasekhar: Yes right, @grayhemp , now if there is permission error we will raise PermissonError else we will raise ClientError.

grayhemp:
I'm still concerned with ClientError swallowing other errors. I'd suggest the following rewrite:

```python
# > Move that to the module level
_ACCESS_DENIED = {'AccessDeniedException', 'UnauthorizedAccess', 'AccessDenied'}

# > For provisioned and serverless do that
except ClientError as e:
    if e.response.get('Error', {}).get('Code') not in _ACCESS_DENIED:
        raise
    serverless_error = e
    logger.warning(f'Skipping serverless; IAM lacks permission: {e}')

# > In the end do that
if provisioned_error and serverless_error:
    msg = (
        'Unable to discover any Redshift clusters: IAM lacks both redshift and '
        f'redshift-serverless permissions. Provisioned: {provisioned_error}; '
        f'Serverless: {serverless_error}'
    )
    logger.error(msg)
    raise PermissionError(msg)
```

If you are okay with that, then apply and add tests.

grayhemp (changes requested):

S-gunasekhar (commented):
Makes sense Client Error could swallow other errors, below code could catch them and raise, made change as suggested. 

```
if e.response.get('Error', {}).get('Code') not in _ACCESS_DENIED:
        raise
```
also added below test cases. 

6 new tests added:

1. test_discover_clusters_both_access_denied_raises_permission_error — both clients get access-denied → raises PermissionError
2. test_discover_clusters_provisioned_access_denied_serverless_succeeds — provisioned access-denied, serverless succeeds → returns serverless results
3. test_discover_clusters_serverless_access_denied_provisioned_succeeds — serverless access-denied, provisioned succeeds → returns provisioned results
4. test_discover_clusters_non_access_denied_provisioned_bubbles_up — provisioned throws non-permission error → re-raises immediately
5. test_discover_clusters_non_access_denied_serverless_bubbles_up — serverless throws non-permission error → re-raises immediately
6. test_discover_clusters_both_non_access_denied_first_bubbles_up — both throw non-permission errors → first one (provisioned) re-raises immediately

grayhemp:
@S-gunasekhar thank you for the fixes. We have some infrastructural changes merged in main, moving from the regex detection of suspicious queries to sqlglot. So the `_SUSPICIOUS_QUERY_RE` is not needed any longer. Could you please bring the PR to the parity with main. Thank you.

S-gunasekhar:
@grayhemp, Ohh sounds great we are moving to sqlglot. I have resolved the merge conflicts from the main and ran all the test cases again. I believe now this PR is ready to be merged with main. 

Let me if I can work on any other things. Happy to work upon and excited to contribute. Thanks, in advanced.

S-gunasekhar:
Resolved merge conflicts in the test_redshift.py and removed the markers.

grayhemp:
@S-gunasekhar please check the precommit failures.

S-gunasekhar:
@grayhemp, I have fixed the pre-commit issues. 

Ruff found minor style violations — a missing blank line before a class, an import out of order, and a line that was too long.

grayhemp:
## Test Summary — Redshift MCP Server

Environment: 2 provisioned clusters (redshift-cluster-analytics, -marketing) + 1 serverless workgroup (
redshift-workgroup-etl), all in us-east-1. Schema exploration exercised on both a provisioned cluster and the 
serverless workgroup via sample_data_dev.tickit.

### Per-tool results (one line each)

| Tool | Result |
|------|--------|
| list_clusters | Pass. Returned 2 provisioned + 1 serverless with full metadata (type, status, endpoint, encryption). |
| list_databases | Pass on provisioned + serverless. Both returned awsdatacatalog, dev, sample_data_dev. |
| list_schemas | Pass on provisioned + serverless. Both returned information_schema, public, tickit; ACLs differ by auth model (awsuser vs IAMR:Admin). |
| list_tables | Pass on provisioned + serverless. Both returned 7 tickit tables (category, date, event, listing, sales, users, venue), type TABLE. |
| list_columns | Pass on provisioned + serverless. Correct ordinal/type/nullable/precision (users=18 cols, venue=5 cols). |
| execute_query | Pass on provisioned + serverless. Correct rows and data-type mapping (int, float, string, bool, NULL). |

### Protection tests

| Scenario | Input | Result |
|----------|-------|--------|
| Read-only deny-list | TRUNCATE tickit.venue | Rejected pre-execution: Statement type not allowed in read-only mode: TRUNCATE. |
| Multi-statement | SELECT 1; SELECT 2 | Rejected: Only a single SQL statement is allowed. |
| No false positive | SELECT 1 AS grant, 'commit;' AS note | Allowed — deny-keywords as identifier/literal not flagged. |
| Transaction breaker | valid INSERT ... VALUES (30000, ...) | Passed guard, blocked by engine: ERROR: transaction is read-only; follow-up count confirmed leaked_rows = 0 (ROLLBACK held). |
| Failed SQL — unparseable | SELECT FROM WHERE | Failed closed: SQL could not be parsed. |
| Failed SQL — engine error | SELECT * FROM tickit.no_such_table_xyz | Surfaced engine error: relation ... does not exist. |
| Session recovery | SELECT after prior failures | Both clusters healthy — session reused after guard + engine failures. |

Note: first INSERT attempt with venueid=99999 returned smallint out of range (data-type error before the read-only check), so I re-ran with a valid smallint to isolate and confirm the transaction breaker.

All 6 tools and all 3 protection layers (SQL read-only guard, read-only transaction breaker, failed-SQL fail-closed) 
verified on both provisioned and serverless. No data persisted.

grayhemp (approved):
LGTM. Let me pass it through the admins.

saeedma8 (approved):

grayhemp:
@S-gunasekhar merged 🎉 Thank you for your contribution.

S-gunasekhar:
@grayhemp, great news! 🎉 . Happy to contribute more.

PR #3524, opened 2026-05-14
fix(eks-mcp-server): enforce allow_write check in apply_yaml

srhsrhsrhsrh:
### Changes
The apply_yaml tool in the EKS MCP server did not check self.allow_write before executing, allowing Kubernetes resources to be created or updated even when the server was started in read-only mode (without --allow-write). This adds an early allow_write check to apply_yaml, consistent with the existing checks in manage_k8s_resource and generate_app_manifest. A regression test is also added to verify the tool is blocked when write access is disabled.

### User experience

Before: A caller could invoke apply_yaml to create or update Kubernetes resources regardless of whether the server was started with --allow-write. The documented read-only mode did not enforce its safety boundary for this tool.

After: Invoking apply_yaml without --allow-write returns an error: "Operation apply_yaml is not allowed without write access". No Kubernetes API call is made. Behavior matches the documentation and other mutating tools.

Testing:
```
awslabs.eks-mcp-server / apply_yaml

{
  "cluster_name": "eks-test",
  "namespace": "default",
  "yaml_path": "{redacted}"
}
```

```
MCP Tool Error Response: Operation apply_yaml is not allowed without write access
```

## Checklist

* [X] Changes are documented

Is this a breaking change? No

**RFC issue number**: N/A

Checklist:

## Acknowledgment


patrick-yu-amzn (approved):
LGTM

scottschreckengaust (approved):
LGTM

PR #2602, opened 2026-03-06
fix(postgres): separate the tablename from sql builder

kennthhz:
Fixes

## Summary

get_table_schema MCP tool should separate the tablename from sql string builder

### Changes

Bind the table_name just before calling PG

### User experience

No change

## Checklist

If your change doesn't seem to apply, please leave them unchecked.

* [ x] I have reviewed the [contributing guidelines](https://github.com/awslabs/mcp/blob/main/CONTRIBUTING.md)
* [ x] I have performed a self-review of this change
* [ x] Changes have been tested
* [ ] Changes are documented

Is this a breaking change? (Y/N)
N

**RFC issue number**:

Checklist:

* [ ] Migration process documented
* [ ] Implement warnings (if it can live side by side)

## Acknowledgment

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the [project license](https://github.com/awslabs/mcp/blob/main/LICENSE).

PR #3626, opened 2026-05-25
feat(mysql-mcp-server): add Kiro Powers for Aurora MySQL, IAM auth, and multi-engine support

happycupofjava:
## Summary

This PR ships the Aurora MySQL Power for the `mysql-mcp-server` package: a Kiro
Power that bundles the MCP server with steering files for Aurora MySQL workflows
(cluster creation, schema design, query optimization, schema migrations, RDS-to Aurora migration paths, and Aurora replication). Alongside the Power, the server now supports three RDS engines (Aurora MySQL, RDS MySQL, RDS MariaDB), adds IAM authentication backed by a hash-pinned Amazon RDS CA bundle, and replaces the single-cluster connection singleton with a multi-cluster registry so an LLM can keep parallel connections to several clusters open in one session.

### Added

- **Kiro Power and steering** — `kiro_power/POWER.md` and `kiro_power/steering/{aurora-mysql.md, aurora-mysql-mcp.md}` for the Aurora MySQL Power. Workflows cover cluster creation, schema design, query optimisation, schema migrations, RDS-to-Aurora migration paths (snapshot restore and read-replica promotion with explicit Migration Safety Rules), and Aurora replication. `kiro_proj_steering/tech.md` for project-level steering.

- **Three-engine support** — `database_type` parameter on `connect_to_database`accepts `aurora-mysql`, `mysql` (RDS MySQL), and `mariadb` (RDS MariaDB). Engine values match AWS RDS engine strings so they can be forwarded directly to RDS APIs without a translation table.

- `mysqlwire_iam` connection method — IAM authentication via `generate_db_auth_token` with strict TLS verification against the Amazon RDS CA bundle. The bundle is fetched at build time by `hatch_build.py` and shipped inside the wheel, so end users get strict TLS without needing network access at install time. Operators can override the bundled file via `--ca_bundle <path>` if they maintain their own trust store.

- **`create_cluster` MCP tool** — creates Aurora MySQL clusters
  (Serverless v2 by default, or provisioned via `cluster_type='provisioned'`).
  Returns a job id; poll `get_job_status` until the cluster is available.
  Default `database` parameter is `'app'` (a safe non-reserved name; AWS
  rejects `'mysql'` as a reserved word for the engine).

- **Companion MCP tools** — `is_database_connected`, `get_database_connection_info`, `get_table_schema`, `get_job_status`.

- **Multi-cluster connection registry (`DBConnectionMap`)** — replaces the previous single-cluster `db_connection_singleton.py`. Two-phase lookup: strict 5-tuple match `(method, cluster, endpoint, database, port)` first, then a port-aware relaxed scan that handles the common case of `connect_to_database` being called with `db_endpoint=''` (auto-resolved by the connect path) and subsequent `run_query` calls passing the same empty string. Returns `None` on ambiguity rather than picking one entry, so a multi-port single-cluster setup never silently routes to the wrong port.

- **`SUPPORTED_CONNECTION_METHODS` table** — single source of truth for which (engine, method) pairs the server accepts. `rdsapi` is Aurora-only; `mysqlwire_iam` is Aurora MySQL + RDS MySQL (MariaDB has no IAM auth); `mysqlwire` works for every engine. Validation happens at the tool boundary with a clear error message rather than failing downstream with a confusing one.

### Changed

- `create_cluster` default `database` is `'app'` (was `'mysql'`). AWS rejects `'mysql'` as a reserved word for the MySQL engine, so the previous default broke minimal-arg invocations with the cryptic `DatabaseName mysql cannot be used. It is a reserved word for this engine`.
- `--allow_write_query` is now the documented CLI flag for enabling write queries (replaces `--readonly True/False`).
- `README.md` rewritten around the unified `database_type + connection_method` surface; previous "RDS Data API vs Direct MySQL" framing removed.
- `CHANGELOG.md` updated with the full set of Added / Changed / Removed entries above.
- `pyproject.toml`: added `aiorwlock>=1.5.0` runtime dep (used by the connection registry's read-write lock) and the `[tool.hatch.build.targets.{wheel,sdist}.hooks.custom]` block pointing at `hatch_build.py` for the CA bundle fetch + hash verification step.

### Removed

- `db_connection_singleton.py` (replaced by `db_connection_map.py`).
- `constants.py` (the single `USER_AGENT_*` value moved into `awslabs/mysql_mcp_server/__init__.py` alongside `__version__`, matching the pattern used by other awslabs/mcp packages).

## Backwards compatibility

- The `database_type` enum was reshaped to use AWS RDS engine strings(`aurora-mysql`, `mysql`, `mariadb`). No external caller of the old enum members exists, since the previous `mysql-mcp-server` release exposed only the singleton path with no engine-typed wire format.
- `--readonly True/False` was the previous documented CLI flag; the new 
- `--allow_write_query` boolean flag is more idiomatic for MCP servers and matches the `postgres-mcp-server` convention.
- The previous `database='mysql'` default in `create_cluster` was a footgun rather than a real default users relied on (AWS rejected every minimal-arg invocation). Swapping to `'app'` fixes the call without breaking any successful prior invocation.

## Security review

- `hatch_build.py` fetches the Amazon RDS global CA bundle from `https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem` over HTTPS at build time and writes it into the wheel. AWS CA rotations are picked up automatically on the next package build.
- Runtime TLS validates cert chain and FQDN against the bundled CA, per the standard PKI model ([RFC 5280](https://datatracker.ietf.org/doc/html/rfc5280) chain validation, [RFC 9525](https://datatracker.ietf.org/doc/html/rfc9525) hostname matching, enforced by Python's `ssl.create_default_context()`). DNS poisoning of an RDS endpoint still fails the handshake because an attacker cannot forge a cert signed by an Amazon RDS CA in the bundle.
- The CA bundle path is also exposed via `--ca_bundle <path>` on the CLI so operators can override the bundled file with their own trusted store if AWS rotates faster than the package release cadence.

## Testing

- **252 unit tests** — `pytest tests/` from a fresh checkout; all green.
Coverage includes new tools, engine + method validation matrix, multi-cluster registry strict and relaxed lookup paths (including the port-aware ambiguity guard), CA bundle hash verification, and the renamed connection modules.
- **MCP stdio handshake** — verified the built package launches as an MCP server, responds to `initialize`, and lists all seven tools via `tools/list`.
- **Build hook** — verified `uv sync` runs `hatch_build.py`, fetches the CA bundle, and writes a hash-matched PEM into the wheel.
- **Live integration** — exercised end-to-end against a fresh Aurora MySQL Serverless v2 cluster (`create_cluster` with no `database` arg, schema setup via `run_query`, seed data, full CRUD against the cluster through a separate Express + Data API frontend). All seven tools driven through a real MCP client.

## Notes for reviewers

- `LICENSE` is restored to the full Apache 2.0 text + Amazon copyright boilerplate that other awslabs/mcp packages ship; the current `mysql-mcp-server/LICENSE` was truncated mid-section.
- Migration / replication workflows in the steering files use the AWS CLI directly today (`aws rds create-db-cluster --replication-source-identifier`, `aws rds promote-read-replica-db-cluster`, etc.). Wrapping these as MCP tools is deliberately deferred to keep this PR focused; happy to follow up in a separate PR if the maintainer guidance is to land them as tools.


src/mysql-mcp-server/awslabs/mysql_mcp_server/connection/asyncmy_pool_connection.py
```diff
@@ -12,179 +12,400 @@
...
+#   1. Download a fresh copy from
+#      https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem
+#   2. Replace the file at _RDS_CA_BUNDLE_PATH
+#   3. Update _RDS_CA_BUNDLE_SHA256 below to match `shasum -a 256 <path>`
+#   4. Bump the package version and release
+#
+# Users who maintain their own trust store can bypass the bundled bundle
+# entirely by passing --ca_bundle <path> on the server command line.
+_RDS_CA_BUNDLE_PATH = os.path.join(os.path.dirname(__file__), 'rds_global_bundle.pem')
+_RDS_CA_BUNDLE_SHA256 = 'e5bb2084ccf45087bda1c9bffdea0eb15ee67f0b91646106e466714f9de3c7e3'
```
kennthhz: will this SHA256 never change?

src/mysql-mcp-server/awslabs/mysql_mcp_server/connection/asyncmy_pool_connection.py
```diff
@@ -12,179 +12,400 @@
...
+            elif 'booleanValue' in value:
+                result[name] = value['booleanValue']
+            elif 'blobValue' in value:
+                result[name] = value['blobValue']
+            elif 'isNull' in value and value['isNull']:
+                result[name] = None
+
+        return result
+
+    def _convert_named_to_positional(self, sql: str, params: Dict[str, Any]) -> Tuple[str, list]:
```
kennthhz, line 324: is there security concern of sql injection?

src/mysql-mcp-server/awslabs/mysql_mcp_server/server.py
```diff
@@ -156,152 +209,667 @@ async def run_query(
...
-            table_schema = :database_name
-            AND table_name = :table_name
-        ORDER BY
-            ORDINAL_POSITION
+            COLUMN_NAME AS column_name,
+            COLUMN_TYPE AS data_type,
+            COLUMN_COMMENT AS column_comment
+        FROM INFORMATION_SCHEMA.COLUMNS
+        WHERE TABLE_SCHEMA = :table_schema
+            AND TABLE_NAME = :table_name
```
kennthhz, line 254: do you need to validate table name similar to https://github.com/awslabs/mcp/commit/61987eef50cec83bacc33ccacc3e7e667590d466

src/mysql-mcp-server/awslabs/mysql_mcp_server/connection/asyncmy_pool_connection.py
happycupofjava: The SHA-256 will change when AWS rotates the RDS global CA bundle. It will change when AWS rotates the CA bundle. We designed it to fail loudly when that happens, so a reviewer must approve the new value.
  
  Evidence from hatch_build.py :
  ```
  if digest != _RDS_CA_BUNDLE_SHA256:
      raise RuntimeError(
          f'RDS CA bundle hash mismatch.\n'
          f'  Downloaded: {digest}\n'
          f'  Pinned:     {_RDS_CA_BUNDLE_SHA256}\n'
  ```

When AWS rotates the bundle, the next package build fails loudly and a maintainer reviews/approves the new SHA-256; git history, not DNS, stays the source of truth for what bytes a fresh build trusts. Existing installs keep working unchanged.

AWS rotates the RDS global CA bundle rarely. Per the official AWS doc on SSL/TLS certificate rotation, the previous bundle (rds-ca-2019) was the standard from 2019 until it expired in August 2024 — about 5 years between rotations. The current bundles (rds-ca-rsa2048-g1, rds-ca-rsa4096-g1, rds-ca-ecc384-g1) have multi-decade validity. So we're talking about one reviewed commit on a multi-year cadence.
https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.SSL-certificate-rotation.html

src/mysql-mcp-server/awslabs/mysql_mcp_server/connection/asyncmy_pool_connection.py
happycupofjava: No injection risk in this code path. _convert_named_to_positional only rewrites placeholder syntax (%(name)s → %s); user values are passed as the second argument to cursor.execute(converted_sql, positional_params) (line 264) and bound by asyncmy at the wire-protocol level, never concatenated into the SQL string. The SQL templates themselves are server-controlled (literal strings in server.py), not user-controlled.

src/mysql-mcp-server/awslabs/mysql_mcp_server/server.py
happycupofjava: The get_table_schema is already safe, table_name is passed as a parameter (WHERE TABLE_NAME = :table_name), not stitched into the SQL string, we're using parameterized SQL query.. So the bug that #2602 fixed in postgres (where the table name was concatenated into SQL with an f-string) doesn't exist here.

src/mysql-mcp-server/awslabs/mysql_mcp_server/connection/asyncmy_pool_connection.py
kennthhz: so whenever AWS rotates, the Cx must upgrade the MCP server? To me that is not a good experience. why can't this be a configuration of MCP without need to update the binary?

happycupofjava: Making the SHA configurable would defeat its purpose, IMO anyone with config-write access could silently change which CA bytes the package trusts.

Two things mitigate the UX concern:
1. The hash check only runs when building from source. Users installing via uvx or pip get the pre-built wheel with the verified PEM already inside — no hash mismatch ever.
2. --ca_bundle <path> is already the runtime escape hatch (README lines 200–202). If AWS rotates before a new release ships, operators point at their own freshly-downloaded bundle. They replace the FILE, not the integrity check.

From a user experience POV:
1. Existing installs -> Nothing breaks. Existing connections keep working with the bundled CA.
2. Need new CA before release -> Users pass --ca_bundle <path> — runtime change, no upgrade.
3. New installs from PyPI -> Gets the latest published wheel; use --ca_bundle <path> if a newer CA is needed before the next release.
4. Maintainers building from source -> Build fails loud, run python hatch_build.py --refresh-hash, push a reviewed pin bump.

happycupofjava: Discussed offline with kennthhz@; will be removing SHA-256 pin from RDS CA bundle hook.

The build hook in hatch_build.py still fetches the Amazon RDS global CA bundle from truststore.pki.rds.amazonaws.com at build time and writes it into the wheel. Runtime TLS validation against the bundled CA is unchanged.

Why this is safe:
TLS itself validates the cert chain and FQDN at runtime against the bundled CA. DNS poisoning of an RDS endpoint still fails the TLS handshake because the attacker cannot get a cert signed by a CA in the bundle. The pin only protected build-time supply-chain integrity, which we are choosing to omit.

happycupofjava: Pushed 3b686a71 removing the SHA-256 pin per offline discussion while keeping the build-time fetch. TLS validation against the bundled CA at runtime is unchanged. 

https://github.com/awslabs/mcp/pull/3626/changes/3b686a71a3be4dfab1f11c9e36f7aee81362cc77

kennthhz (approved):

kennthhz (dismissed):

src/mysql-mcp-server/awslabs/mysql_mcp_server/server.py
```diff
@@ -156,152 +209,667 @@ async def run_query(
...
+    )
 
     sql = """
         SELECT
-            COLUMN_NAME,
-            COLUMN_TYPE,
-            IS_NULLABLE,
-            COLUMN_DEFAULT,
-            EXTRA,
-            COLUMN_KEY,
```
scottschreckengaust, line 187: Aren't the nullability and default helpful in getting the table's schema?

src/mysql-mcp-server/awslabs/mysql_mcp_server/server.py
```diff
@@ -156,152 +209,667 @@ async def run_query(
...
+            database=database,
+        )
 
-    """Run the MCP server with CLI argument support."""
-    parser = argparse.ArgumentParser(
-        description='An AWS Labs Model Context Protocol (MCP) server for MySQL'
+        return str(llm_response)
+
+    except Exception as e:
+        logger.error(f'connect_to_database failed with error: {str(e)}')
```
scottschreckengaust: Might there be some errors that should not be traced out?

src/mysql-mcp-server/awslabs/mysql_mcp_server/server.py
happycupofjava: Thanks for catching that, restored  IS_NULLABLE, COLUMN_DEFAULT, EXTRA and COLUMN_KEY when getting the table's schema in commit 8a6433e2359eae0609e3e9642570bd1ea063987c.

src/mysql-mcp-server/awslabs/mysql_mcp_server/server.py
happycupofjava: Great catch, this is something that should be avoided. Redact exception details as part of 8a6433e2359eae0609e3e9642570bd1ea063987c across connect_to_database, create_cluster_worker, and run_query.

kennthhz (dismissed):

kennthhz (approved):

src/mysql-mcp-server/hatch_build.py
```diff
@@ -0,0 +1,168 @@
...
+        raise RuntimeError(f'RDS CA bundle URL must use https://, got: {_RDS_CA_BUNDLE_URL!r}')
+
+    try:
+        ctx = _ssl_context_for_aws_endpoint()
+        # B310 (audit url open for permitted schemes): scheme is enforced
+        # to be https:// by the explicit check above, and the URL is a
+        # module-level constant pointing at AWS's public truststore.
+        with urllib.request.urlopen(  # nosec B310
+            _RDS_CA_BUNDLE_URL, timeout=30, context=ctx
+        ) as resp:
```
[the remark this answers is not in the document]
happycupofjava: Same finding class as Bandit B310, already addressed. The URL is the module-level constant _RDS_CA_BUNDLE_URL = 'https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem' — not user input. An explicit https:// scheme check on hatch_build.py lines 91-92 raises RuntimeError before urlopen is reached. Build hook runs at wheel build time only.

scottschreckengaust (approved):
LGTM

PR #3630, opened 2026-05-26
fix(elasticache-mcp-server): enforce readonly mode on delete and modify serverless cache tools

swarnaprakash:
### Changes

The `--readonly` flag was enforced on `create-serverless-cache` but missing on `delete-serverless-cache` and `modify-serverless-cache`, allowing those operations to reach AWS APIs when readonly mode was enabled.

**Fix approach:** Rather than adding inline checks to the two affected tools (which doesn't prevent future regressions), this PR moves readonly enforcement into the `handle_exceptions` decorator — which is already applied to every tool. Tools are now blocked by default in readonly mode unless explicitly marked with `@readonly_safe`.

**Why this approach:**
- **Default-deny** — new tools are blocked in readonly mode unless the author explicitly marks them safe. Forgetting the marker results in an overly-restricted tool (safe failure), not a bypass (security hole).
- **Minimal per-tool diff** — read-only tools get a one-line `@readonly_safe` decorator. Mutating tools need no changes.
- **No framework exists** — the awslabs/mcp mono-repo has no shared readonly mechanism. Each server implements its own. Peer servers (aurora-dsql, postgres, dataprocessing) all use inline checks with no decorator precedent.

**Alternatives considered:**
- Inline fix only (2 files): Fixes the bug but doesn't prevent future regressions.
- Opt-in `@require_write_access` decorator: Same refactor cost but still opt-in — forgetting it leaves a gap.
- Custom tool registrar wrapping `@mcp.tool`: Largest refactor, deviates from mono-repo conventions.

### User experience

**Before:** A server configured with `--readonly` could still delete or modify serverless caches.

**After:** All mutating operations are blocked in readonly mode. Read-only operations (describe, list, get-ssh-tunnel-command, test-migration) continue to work normally.

## Checklist

* [x] Changes are documented

Is this a breaking change? Y — Tools that extend this server and use `@handle_exceptions` will now be blocked in readonly mode unless marked with `@readonly_safe`. This is intentional to prevent future readonly bypasses.

**RFC issue number**: N/A

## Acknowledgment

### DIFF
--- a/src/elasticache-mcp-server/awslabs/elasticache_mcp_server/common/decorators.py
+++ b/src/elasticache-mcp-server/awslabs/elasticache_mcp_server/common/decorators.py
@@ -14,13 +14,24 @@
 
 """Decorators for ElastiCache MCP Server."""
 
+from ..context import Context
 from functools import wraps
 from typing import Any, Callable
 
 
-def handle_exceptions(func: Callable) -> Callable:
-    """Decorator to handle exceptions in ElastiCache operations.
+def readonly_safe(func: Callable) -> Callable:
+    """Mark a tool as safe to run in readonly mode.
 
+    Tools without this decorator will be blocked when readonly mode is enabled.
+    """
+    setattr(func, '_readonly_safe', True)
+    return func
+
+
+def handle_exceptions(func: Callable) -> Callable:
+    """Decorator to handle exceptions and enforce readonly mode for ElastiCache operations.
+
+    Blocks tool execution in readonly mode unless the tool is marked with @readonly_safe.
     Wraps the function in a try-catch block and returns any exceptions
     in a standardized error format.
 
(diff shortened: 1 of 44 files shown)


kevinmcgehee (dismissed):
The changes look good to me.  I like the negative tests and behavior to fail-closed.  I think it's probably useful to add tests for each of the readonly APIs so that a one-line change to remove the annotation doesn't accidentally break a read-only API.  I did not check to see if there were other functions that should be marked as read-only but these changes seem safe.

swarnaprakash:
> The changes look good to me. I like the negative tests and behavior to fail-closed. I think it's probably useful to add tests for each of the readonly APIs so that a one-line change to remove the annotation doesn't accidentally break a read-only API. I did not check to see if there were other functions that should be marked as read-only but these changes seem safe.


I now added tests for all the tools

allenss-amazon (approved):

madolson:
The default-deny approach is good, but the safety guarantee depends on every tool using `@handle_exceptions`. If someone adds a new tool with just `@mcp.tool()` and skips the decorator, it silently bypasses readonly.

Consider adding a test that introspects the MCP server's tool registry at startup/test time and asserts every registered tool function is wrapped by `handle_exceptions`. Without that, this is still opt-in enforcement disguised as default-deny.

madolson (approved):

swarnaprakash:
> The default-deny approach is good, but the safety guarantee depends on every tool using `@handle_exceptions`. If someone adds a new tool with just `@mcp.tool()` and skips the decorator, it silently bypasses readonly.
> 
> Consider adding a test that introspects the MCP server's tool registry at startup/test time and asserts every registered tool function is wrapped by `handle_exceptions`. Without that, this is still opt-in enforcement disguised as default-deny.

Will look into a separate PR for this as i couldn't find a clean idea to do this quickly. want to push this first as I got 2 approvals

PR #3706, opened 2026-06-03
fix(billing-cost-management): clarify end_date inclusive semantics for cost-anomaly tool

bhatia-di:
Fixes

## Summary
The `cost-anomaly` tool did not document whether `end_date` is inclusive or exclusive.
  Agents assumed exclusive (Cost Explorer `DateInterval` convention) and passed `today+1`,
  which was correctly rejected as a future date.
  
  The `GetAnomalies` API uses `AnomalyDateInterval` (not `DateInterval`), where `EndDate`
  means "the last date an anomaly was observed" — i.e., inclusive.
### Changes
- Added inclusive/exclusive note to tool description
  - Updated `start_date` and `end_date` docstrings with explicit semantics
  - Added unit test confirming today's date is accepted as `end_date`

### User experience

**Before:** Agent passes `today+1` as `end_date` → tool rejects with "Cannot request anomalies for future dates" error.
  
  **After:** Agent passes `today` as `end_date` (guided by the explicit inclusive documentation) → succeeds.

## Checklist

* [ X] I have reviewed the [contributing guidelines](https://github.com/awslabs/mcp/blob/main/CONTRIBUTING.md)
* [ X] I have performed a self-review of this change
* [ X] Changes have been tested
* [ X ] Changes are documented

**RFC issue number**:

Checklist:

## Acknowledgment

### DIFF
--- a/src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/cost_anomaly_tools.py
+++ b/src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/cost_anomaly_tools.py
@@ -48,6 +48,8 @@ You can filter anomalies by:
 - Feedback status (optional)
 - Total impact (optional)
 
+Note: Both start_date and end_date are INCLUSIVE. To get anomalies including today, use today's date as end_date.
+
 Feedback status options:
 - YES: Anomalies marked as accurate
 - NO: Anomalies marked as inaccurate
@@ -68,8 +70,8 @@ async def cost_anomaly(
 
     Args:
         ctx: The MCP context object
-        start_date: Start date in YYYY-MM-DD format. Required.
-        end_date: End date in YYYY-MM-DD format. Required.
+        start_date: Start date in YYYY-MM-DD format (inclusive). Required.
+        end_date: End date in YYYY-MM-DD format (inclusive — use today's date to include today's anomalies). Required.
         monitor_arn: Optional ARN of a specific cost anomaly monitor to filter results.
         feedback: Optional filter for anomalies by feedback status (YES, NO, PLANNED_ACTIVITY).
         max_results: Optional maximum number of results to return.


src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/cost_anomaly_tools.py
```diff
@@ -48,6 +48,9 @@
 - Feedback status (optional)
 - Total impact (optional)
 
+Note: Both start_date and end_date are INCLUSIVE (unlike Cost Explorer tools where end_date is exclusive).
```
wangyuhere: Do we need to mention the Cost Explorer tools here? The cost explorer tools will have their own description, which will be the source of truth. 

It might be confusing to agent if in the future, there might be a new tool added in Cost Explorer which have different time range.

src/billing-cost-management-mcp-server/tests/tools/test_cost_anomaly_tools.py
```diff
@@ -846,3 +846,28 @@ async def test_get_anomalies_no_optional_fields(self, mock_context, mock_ce_clie
         assert 'score' not in anomaly
         assert 'impact' not in anomaly
         assert 'root_causes' not in anomaly
+
+    async def test_ca_real_today_end_date_is_valid(self, mock_context):
```
wangyuhere: The unit test can not test the prompt change I think. 

Do we really need this unit test?

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/cost_anomaly_tools.py
bhatia-di: Removed the Cost Explorer comparison. Each tool should be self-contained.

src/billing-cost-management-mcp-server/tests/tools/test_cost_anomaly_tools.py
bhatia-di: Removed the unit tests since this is a doc-only change and they don't validate the prompt.

wangyuhere (approved):

finklen-amazon (approved):

chittev (approved):

Issue #2217, opened 2026-01-20
aws-mcp : profile flag not respected


pedromonteiro18:
### Describe the bug

The AWS MCP is not respecting the --profile flag when running aws CLI commands. 

Currently, only the AWS API mcp is respecting the --profile flag, so I have been falling back to that.

I am running the model on bedrock, and the AWS MCP only connects to the account where the model is hosted, not the account where the --profile flag is specifying. 

### Expected Behavior

AWS MCP server should connect to the AWS account specified using the --profile flag when running the CLI commands. 

### Current Behavior

The AWS MCP defaults to using the account where the model is hosted. 

### Reproduction Steps

Create an AWS profile for account A.

Host the model on account B. 

Add the AWS MCP to the model hosted on account B and ask it to run a command on account A using the --profile flag.

It will only be able to query resources in account B. 

### OS

MacOs

### Server

other

### Server Version

AWS-MCP

### Region experiencing the issue

us-east-1

### Service quota

- [x] I have reviewed the service quotas for this construct

PCManticore:
Hi, thank you for the report.

The AWS MCP Server currently doesn't support profile information. This server uses the MCP Proxy for AWS to source the credentials, which are then used for signing the request that goes to the remove server. On the remote server the concept of a "profile" is no longer applicable, thus the MCP server is not able to apply the profile information that the user requested in the interaction with the model. 

We're thinking how to support this use case properly though and will keep you updated.

ahmedasmar:
This must be solved for an AI agent to be able to connect to multiple accounts using the MCP. We use AWS organization and have multiple accounts using SSO. If --profile is not possible, this MCP is useless, and the old one is better for now.

PR #3713, opened 2026-06-04
feat(cloudwatch-applicationsignals): add per-call profile support

### LINKED ISSUE #2940: cloudwatch-applicationsignals-mcp-server: Add profile_name parameter to tools for dynamic AWS profile selection
### Describe the feature

The `cloudwatch-applicationsignals-mcp-server` tools (`audit_services`, `audit_slos`, `list_monitored_services`, `list_slos`, `get_service_detail`, `audit_service_operations`, etc.) do not accept a `profile_name` parameter, making it impossible to dynamically select an AWS profile during a conversation.

The sibling `cloudwatch-mcp-server` already supports this — tools like `get_active_alarms`, `get_metric_data`, `execute_log_insights_query`, `describe_log_groups`, and `analyze_metric` all accept an optional `profile_name` parameter that overrides the default credential chain.

### Use Case

Users working with multiple AWS accounts need to specify which profile to use at tool invocation time, not just at MCP server startup. For example, an AI assistant helping a user who says "check the health of my services in the `prod` account" should be able to pass `profile_name='prod'` to the Application Signals tools without restarting the MCP server.

Currently the only workaround is:
- Hardcoding `AWS_PROFILE` in the MCP server's `env` config (static, requires restart to change)
- Falling back to raw AWS CLI calls instead of using the MCP tools

### Proposed Solution

Add an optional `profile_name` parameter to all Application Signals tools, matching the existing pattern in `cloudwatch-mcp-server`. Example from the CloudWatch MCP server:

```python
profile_name: Optional[str] = None  # AWS CLI Profile Name to use for AWS access.
                                     # Falls back to AWS_PROFILE environment variable
                                     # if not specified, or uses default AWS credential chain.
```

### Related Issues

- #481 — Support non-profile credentials (broader credential flexibility)
- #2217 — aws-mcp: profile flag not respected (similar theme, different server)

volantebjb:
Fixes #2940

Related: #481, #2217

## Summary

Adds optional per-call `profile_name` support to the AWS-backed CloudWatch Application Signals tools.

### Changes

The implementation adds a profile-aware `get_aws_client(...)` helper and threads profile-scoped clients through the service, SLO, audit, canary, trace, group, change, and RUM paths. When `profile_name` is not provided, the existing shared-client behavior is preserved.

`get_enablement_guide` is unchanged because it reads local guide files and does not call AWS.

### User experience

Before this change, Application Signals tools used the MCP server startup credential context. Switching accounts during a conversation required restarting the server with a different profile or using another tool path.

After this change, callers can pass `profile_name='prod'` directly on the tool call. Calls without `profile_name` continue to use the existing `AWS_PROFILE` / default credential chain path.

## Checklist

* [ ] Changes are documented

**RFC issue number**: N/A

Checklist:

## Validation

- Focused profile test
- Focused affected-area pytest coverage
- Canary/audit pytest coverage
- Live read-only MCP stdio smoke with a configured AWS profile

### DIFF
--- a/src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/audit_utils.py
+++ b/src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/audit_utils.py
@@ -29,9 +29,17 @@ FUZZY_MATCH_THRESHOLD = 30  # Minimum similarity score for fuzzy matching
 HIGH_CONFIDENCE_MATCH_THRESHOLD = 85  # High confidence threshold for exact fuzzy matches
 
 
-async def execute_audit_api(input_obj: Dict[str, Any], region: str, banner: str) -> str:
+async def execute_audit_api(
+    input_obj: Dict[str, Any],
+    region: str,
+    banner: str,
+    applicationsignals_client_override=None,
+) -> str:
     """Execute the Application Signals audit API call with the given input object."""
-    from .aws_clients import applicationsignals_client
+    if applicationsignals_client_override is None:
+        from .aws_clients import applicationsignals_client
+    else:
+        applicationsignals_client = applicationsignals_client_override
 
     # File log path
     desired_log_path = os.environ.get('AUDITOR_LOG_PATH', tempfile.gettempdir())
--- a/src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/aws_clients.py
+++ b/src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/aws_clients.py
@@ -19,6 +19,7 @@ import os
 from . import __version__
 from botocore.config import Config
 from loguru import logger
+from typing import Any, Optional, cast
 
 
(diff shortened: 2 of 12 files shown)


Mattral (commented):
Nice addition, per‑call profile support seems really useful. I’m curious: when a caller passes an invalid or misconfigured profile_name, how does the tool surface that error back? Is it a clear message or just the raw AWS client exception? Wondering how this affects usability for folks switching profiles often.

volantebjb:
Hi @Mattral ,

Right now this follows the existing repo pattern more than a profile-specific one.

`profile_name` is applied when the boto3 session/client is created, so an invalid or misconfigured profile fails at that point, typically with the underlying boto/botocore exception such as `ProfileNotFound`.

In AppSignals, some tools already wrap broad exceptions and return that message as part of the tool response, while a few paths still let it bubble more directly. That is roughly consistent with sibling servers as well. For example, `cloudwatch-mcp-server` and `aws-for-sap-management-mcp-server` also use `profile_name` at client creation time and generally surface `str(e)` or re-raise after logging rather than mapping profile errors to a dedicated message.

So the behavior is usable, but not fully normalized yet for frequent profile switching. A reasonable follow-up would be to catch `ProfileNotFound` and related credential/config errors in the shared client path and return a clearer tool-level message.

PR #3780, opened 2026-06-05
fix(mcp-lambda-handler): exclude optional params from required list

### LINKED ISSUE #1119: (mcp-lambda-handler): (adds optional tool parameters to required parameters list)
### Describe the bug

When gathering the list of parameters for each tool, the MCPLambdaHandler also collects all the required parameters. This list is communicated to the Client, shaping the behaviour of the LLM regarding the tools. The bug [occurs when the arguments are being added](https://github.com/awslabs/mcp/blob/main/src/mcp-lambda-handler/awslabs/mcp_lambda_handler/mcp_lambda_handler.py#L255), on line 255: 

`# Build properties from type hints
            for param_name, param_type in hints.items():
                param_schema = get_type_schema(param_type)

                if param_name in arg_descriptions:
                    param_schema['description'] = arg_descriptions[param_name]

                properties[param_name] = param_schema
                required.append(param_name)`

There is no filtering for parameters that are Optional. Because all parameters are added to the `required` list, all the LLM believes every single parameter is necessary to use the tool, due to the information it receives in initialization. This makes the LLM continue to ask the user for information when it is not necessary, and reduces the efficiency of using the tools.

### Expected Behavior

Optional parameters should not be added to the list of required parameters.
It should have some filtering done such as :

`required = []
            # FIX ADDED HERE
            for param_name, param_type in hints.items():
                # Only add to required if it's not Optional (Union with None)
                if not (get_origin(param_type) is Union and type(None) in get_args(param_type)):
                    required.append(param_name)`

### Current Behavior

Every argument is added, no filtering is done.
`# Build properties from type hints
            for param_name, param_type in hints.items():
                param_schema = get_type_schema(param_type)

                if param_name in arg_descriptions:
                    param_schema['description'] = arg_descriptions[param_name]

                properties[param_name] = param_schema
                required.append(param_name)`

### Reproduction Steps

example payload:
`payload = { "method": "tools/list", "jsonrpc": "2.0", "id": 1}`

### Possible Solution

`required = []
            # FIX ADDED HERE
            for param_name, param_type in hints.items():
                # Only add to required if it's not Optional (Union with None)
                if not (get_origin(param_type) is Union and type(None) in get_args(param_type)):
                    required.append(param_name)`

Maybe add a check for the Optional typehint too

_No response_

### OS

MacOS

### Server

mcp-lambda-handler

### Server Version

_No response_

each region

### Other information

others reporting bugs:
https://github.com/anthropics/claude-code/issues/5597

### Service quota

mikegc-aws:
## Summary

Optional parameters (`Optional[T]` types or params with default values) were unconditionally added to the `required` list in the tool JSON schema. This caused MCP clients/LLMs to treat every parameter as mandatory, prompting users for values that should be optional.

## Changes

- Check for `Union[X, None]` (i.e. `Optional[T]`) type hints before adding to `required`
- Check for parameters with default values via `inspect.signature`
- Added tests: `test_tool_decorator_required_vs_optional_params` and `test_tool_decorator_all_required_params`
- Updated existing `test_tool_decorator_optional_type_hints` to assert `required == []`
- Added a demo script showing correct schema generation

## Before (broken)

```python
@mcp.tool()
def search(query: str, limit: Optional[int] = 10):
    ...
# required: ['query', 'limit']  <-- limit should NOT be required
```

## After (fixed)

```python
@mcp.tool()
def search(query: str, limit: Optional[int] = 10):
    ...
# required: ['query']  <-- correct
```

Fixes #1119

### DIFF
--- a/src/mcp-lambda-handler/awslabs/mcp_lambda_handler/mcp_lambda_handler.py
+++ b/src/mcp-lambda-handler/awslabs/mcp_lambda_handler/mcp_lambda_handler.py
@@ -253,6 +253,9 @@ class MCPLambdaHandler:
                 # Default for unknown complex types
                 return {'type': 'string'}
 
+            # Get function signature to check for default values
+            sig = inspect.signature(func)
+
             # Build properties from type hints
             for param_name, param_type in hints.items():
                 param_schema = get_type_schema(param_type)
@@ -261,7 +264,20 @@ class MCPLambdaHandler:
                     param_schema['description'] = arg_descriptions[param_name]
 
                 properties[param_name] = param_schema
-                required.append(param_name)
+
+                # Determine if the parameter is required:
+                # - Not required if type is Optional (Union[X, None])
+                # - Not required if it has a default value in the signature
+                is_optional_type = (
+                    get_origin(param_type) is Union
+                    and type(None) in get_args(param_type)
+                )
+                has_default = (
+                    param_name in sig.parameters
+                    and sig.parameters[param_name].default is not inspect.Parameter.empty
+                )
+                if not is_optional_type and not has_default:
+                    required.append(param_name)
 
             # Create tool schema
             tool_schema = {
--- a/src/mcp-lambda-handler/awslabs/mcp_lambda_handler/mcp_lambda_handler.py
+++ b/src/mcp-lambda-handler/awslabs/mcp_lambda_handler/mcp_lambda_handler.py
@@ -268,9 +268,8 @@ class MCPLambdaHandler:
                 # Determine if the parameter is required:
                 # - Not required if type is Optional (Union[X, None])
                 # - Not required if it has a default value in the signature
-                is_optional_type = (
-                    get_origin(param_type) is Union
-                    and type(None) in get_args(param_type)
+                is_optional_type = get_origin(param_type) is Union and type(None) in get_args(
+                    param_type
                 )
                 has_default = (
                     param_name in sig.parameters
--- a/src/mcp-lambda-handler/awslabs/mcp_lambda_handler/mcp_lambda_handler.py
+++ b/src/mcp-lambda-handler/awslabs/mcp_lambda_handler/mcp_lambda_handler.py
@@ -265,17 +265,14 @@ class MCPLambdaHandler:
 
                 properties[param_name] = param_schema
 
-                # Determine if the parameter is required:
-                # - Not required if type is Optional (Union[X, None])
-                # - Not required if it has a default value in the signature
-                is_optional_type = get_origin(param_type) is Union and type(None) in get_args(
-                    param_type
-                )
+                # A parameter is required if and only if it has no default value.
+                # Optional[X] without a default is still mandatory at call time,
+                # so it must stay in the required list.
                 has_default = (
                     param_name in sig.parameters
                     and sig.parameters[param_name].default is not inspect.Parameter.empty
                 )
-                if not is_optional_type and not has_default:
+                if not has_default:
                     required.append(param_name)
 
             # Create tool schema
--- a/src/mcp-lambda-handler/tests/test_lambda_handler.py
+++ b/src/mcp-lambda-handler/tests/test_lambda_handler.py
@@ -627,6 +627,63 @@ def test_tool_decorator_optional_type_hints():
     assert props['d']['type'] == 'boolean'
     assert props['e']['type'] == 'array'
     assert props['e']['items'] == {'type': 'string'}
+    # Optional parameters should NOT appear in required list
+    assert schema['inputSchema']['required'] == []
+
+
+def test_tool_decorator_required_vs_optional_params():
+    """Test tool decorator correctly distinguishes required from optional parameters."""
+    handler = MCPLambdaHandler('test-server')
+
+    @handler.tool()
+    def search(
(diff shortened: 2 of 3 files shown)


mikegc-aws:
@Lukas-Xue ? Can you help here?

src/mcp-lambda-handler/awslabs/mcp_lambda_handler/mcp_lambda_handler.py
```diff
@@ -261,7 +264,19 @@ def get_type_schema(type_hint: Any) -> Dict[str, Any]:
...
+                # - Not required if it has a default value in the signature
+                is_optional_type = get_origin(param_type) is Union and type(None) in get_args(
+                    param_type
+                )
+                has_default = (
+                    param_name in sig.parameters
+                    and sig.parameters[param_name].default is not inspect.Parameter.empty
+                )
+                if not is_optional_type and not has_default:
+                    required.append(param_name)
```
scottschreckengaust: `is_optional_type` treats any `Optional[X]` as not-required, but a param typed `Optional[X]` with **no default** is still a mandatory argument. It gets dropped from `required` here, and a client that trusts the schema and omits it hits `TypeError` at `func(**converted_args)` (returned as `-32603`).

Suggest keying requiredness on `has_default` alone:

```python
                has_default = (
                    param_name in sig.parameters
                    and sig.parameters[param_name].default is not inspect.Parameter.empty
                )
                if not has_default:
                    required.append(param_name)
```

This also sidesteps the PEP-604 `X | None` gap (`get_origin(int | None) is not Union` on Py 3.10–3.13), since it no longer depends on detecting the optional type.

src/mcp-lambda-handler/tests/test_lambda_handler.py
```diff
@@ -627,6 +627,63 @@ def foo(
     assert props['d']['type'] == 'boolean'
     assert props['e']['type'] == 'array'
     assert props['e']['items'] == {'type': 'string'}
+    # Optional parameters should NOT appear in required list
+    assert schema['inputSchema']['required'] == []
```
scottschreckengaust: This asserts `required == []` for five `Optional[...]` params that have **no defaults** — which codifies the regression above (they're mandatory at call time, so a client omitting them gets a `TypeError`). If requiredness keys on `has_default` alone, this should assert all five ARE required. Consider adding a paired case for `Optional[X]` *with* a default to lock in the intended distinction.

scottschreckengaust (changes requested):
## Review: exclude optional params from required list

The `has_default` half of this change is a genuine fix — previously every parameter was marked required, so `limit: Optional[int] = 10` and `offset: int = 0` were wrongly listed. That part is correct and worth keeping.

Requesting changes on the `is_optional_type` half: it drops `Optional[X]`-**without-a-default** from `required`, but such a parameter is still mandatory at call time, so the emitted schema and the runtime disagree. Reproduced on Python 3.10 (a supported Lambda runtime) against this branch:

```python
@mcp.tool()
def needs_it(a: Optional[int]) -> str: ...
# inputSchema.required == []          # schema says 'a' is optional
needs_it()  # what a schema-trusting client does when it omits 'a'
# -> TypeError: needs_it() missing 1 required positional argument: 'a'
```

At `tools/call`, `converted_args` is built only from the arguments the client sent and passed as `func(**converted_args)`, so an omitted `a` becomes a `TypeError`, surfaced as a `-32603` execution error rather than the tool running. `inputSchema.required` is never used to validate incoming arguments, so nothing else guards this.

**Minimal fix:** key requiredness on `has_default` alone and drop the `is_optional_type` term. Walking the cases shows `is_optional_type` is only ever *decisive by itself* in the one case it gets wrong:

| annotation | default | should be required? | `has_default`-only | current code |
|---|---|---|---|---|
| `Optional[int]` | `= 10` | no | no ✓ | no ✓ |
| `Optional[int]` | none | **yes** | **yes ✓** | **no ✗** |
| `int` | `= 0` | no | no ✓ | no ✓ |
| `int` | none | yes | yes ✓ | yes ✓ |

Dropping the term also closes a related gap for free: `is_optional_type` uses `get_origin(param_type) is Union`, which does not match PEP-604 `X | None` — on Python 3.10–3.13 `get_origin(int | None)` is `types.UnionType`, not `typing.Union` (verified on 3.10), so `X | None` optionals aren't detected anyway.

Two non-blocking notes (both pre-existing, not introduced here — mentioning for context): the loop iterates `get_type_hints(func)`, so an **un-annotated** parameter is dropped from `properties`/`required` entirely; and `get_type_hints` is re-run per `tools/call` in the hot path (it's resolved once at decoration but not cached).

mikegc-aws:
@scottschreckengaust Thanks for the review — addressed in 9d025ed6:

- Requiredness is now keyed on `has_default` alone; the `is_optional_type` check is removed, so `Optional[X]` params without defaults stay in `required` (and the PEP-604 `X | None` detection gap is moot).
- `test_tool_decorator_optional_type_hints` now asserts all five default-less `Optional[...]` params ARE required, and I added `test_tool_decorator_optional_type_with_default_not_required` as the paired case for `Optional[X]` with a default.

All 84 tests and the pre-commit hooks pass locally. Ready for another look when you have a moment.

PR #3789, opened 2026-06-08
docs(aurora-dsql): reflect native JSON/JSONB column support

amaksimo:
## Summary

DSQL now supports `JSON` and `JSONB` columns natively, so the existing "store JSON as TEXT" guidance across the steering docs, skill files, and migration guides is outdated. Array column types (e.g. `TEXT[]`) are still unsupported, so arrays must still be serialized — but JSONB is now the recommended target.

- Update guidance from "store JSON as TEXT" / "JSON.stringify into TEXT" to "use JSON or JSONB; prefer JSONB."
- Update MySQL→DSQL migration mappings: `JSON` → `JSONB` (was `TEXT (JSON.stringify)`).
- Update array guidance to say "serialize arrays as JSONB" (DSQL still does not support array column types).
- Refresh the `Data Serialization` example to use JSONB end-to-end (replacing the comma-separated TEXT helpers).
- Update CHANGELOG to note that the bumped `dsql-lint` accepts both `JSON` and `JSONB` (the prior entry said it still flags `JSONB`).

The `kiro_power/steering/*` files remain byte-identical to their `skills/dsql-skill/references/*` counterparts after these edits.

## Test plan

- [x] All `kiro_power/steering/*` files remain identical to their `skills/dsql-skill/references/*` counterparts.
- [x] Re-grepped for stale "store JSON as TEXT", "JSON.stringify", and "as TEXT" guidance — no remaining hits.
- [ ] Manual review: verify the JSONB-based Data Serialization example reads cleanly.

---


praba2210 (dismissed):
Looks good — clean, well-scoped docs update. Steering and references stay byte-identical, changelog is accurate, commit message is solid. No issues.

src/aurora-dsql-mcp-server/kiro_power/steering/development-guide.md
```diff
@@ -13,7 +13,7 @@ effortless scaling, multi-region viability, among other advantages.
 - **REQUIRED: Follow DDL Guidelines** - Refer to [DDL Rules](#schema-ddl-rules)
 - **SHALL repeatedly generate fresh tokens** - Refer to [Connection Limits](auth/authentication-guide.md#connection-rules)
 - **ALWAYS use ASYNC indexes** - `CREATE INDEX ASYNC` is mandatory
-- **MUST Serialize arrays/JSON as TEXT** - Store arrays/JSON as TEXT (comma separated, JSON.stringify)
+- **MUST serialize arrays as JSONB** - DSQL does not support array column types (e.g. `TEXT[]`)
```
anwesham-lab: why always JSONB for arrays? Shouldn't it be MUST serialize arrays as JSON, JSONB, or TEXT?

src/aurora-dsql-mcp-server/kiro_power/steering/development-guide.md
```diff
@@ -53,9 +53,8 @@ effortless scaling, multi-region viability, among other advantages.
 
 ### Schema Design Rules
 
-- MUST use **simple PostgreSQL types:** VARCHAR, TEXT, INTEGER, BOOLEAN, TIMESTAMP
-- MUST store arrays as TEXT (comma-separated is recommended)
-- MUST store JSON objects as TEXT (JSON.stringify)
+- MUST use **simple PostgreSQL types:** VARCHAR, TEXT, INTEGER, BOOLEAN, TIMESTAMP, JSON, JSONB
+- MUST serialize arrays as JSONB (DSQL does not support array column types)
```
anwesham-lab: same comment as above

src/aurora-dsql-mcp-server/kiro_power/steering/examples-patterns.md
```diff
@@ -131,43 +131,27 @@ INSERT INTO distributors VALUES (nextval('order_seq'), 'nothing');
 
 ## Data Serialization
 
-**Pattern:** MUST store arrays and JSON as TEXT (runtime-only types). Per [DSQL docs](https://docs.aws.amazon.com/aurora-dsql/latest/userguide/working-with-postgresql-compatibility-supported-data-types.html), cast to JSON at query time.
+**Pattern:** Use JSONB for objects and for arrays. Array column types (e.g. `TEXT[]`) are not supported in DSQL, so arrays must be stored as JSONB. See [supported data types](https://docs.aws.amazon.com/aurora-dsql/latest/userguide/working-with-postgresql-compatibility-supported-data-types.html).
```
anwesham-lab: do we ever outline a reason to only prefer JSONB? This is misleading.

src/aurora-dsql-mcp-server/kiro_power/steering/examples-schema.md
```diff
@@ -20,8 +20,8 @@ CREATE TABLE IF NOT EXISTS orders (
   order_id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
   tenant_id VARCHAR(255) NOT NULL,
   status VARCHAR(50) NOT NULL,
-  tags TEXT,
-  metadata TEXT,
+  tags JSONB,
+  metadata JSONB,
```
anwesham-lab: I feel like tags would nontrivially be JSON?

src/aurora-dsql-mcp-server/kiro_power/steering/development-guide.md
amaksimo: Fair question, my logic was we wanted to encode the default, I think JSONB is the right default for arrays.

The options from Claude:

* JSONB: operators (@>, ?, ->>) and jsonb_array_elements_text work directly; invalid JSON is rejected at write time; binary-encoded.
* JSON: same operators, validates writes, but for arrays it has no upside over JSONB — no keys to preserve, no whitespace that matters semantically. Listing it as co-equal offers a strictly-worse option.
* TEXT: legitimate only when the array is opaque to the database (app-side parse, never queried inside). The moment you want tags @> '["x"]' or jsonb_array_elements_text(tags) you have to cast on every query, and there's no write-time validation.

From this, I think it is clear we can drop JSON from the suggested approach....

If we change the rule to`JSON, JSONB, or TEXT,` the agent has no signal about which to pick and will often pick TEXT (least-typed default), losing the operators and validation that make JSONB the right call for the common case. 

If we want to acknowledge the legit TEXT escape-hatch, the right phrasing is something like:

```
MUST serialize arrays — DSQL does not support array column types. PREFER JSONB (operators and jsonb_array_elements_text work directly; invalid JSON rejected at write). Use TEXT only when the array is opaque to the application (app-side parse, no DB-side queries).
```

Thoughts?

src/aurora-dsql-mcp-server/kiro_power/steering/development-guide.md
amaksimo: Replied to above, please let me know your thoughts.

src/aurora-dsql-mcp-server/kiro_power/steering/examples-patterns.md
amaksimo: Explained why above.

spencercorwin (approved):

src/aurora-dsql-mcp-server/kiro_power/steering/mysql-full-example.md
```diff
@@ -69,8 +69,8 @@ transact(["CREATE INDEX ASYNC idx_products_category ON products(tenant_id, categ
 | `INT` tenant_id               | `VARCHAR(255)` for multi-tenant pattern                                                                                                                |
 | `MEDIUMTEXT`                  | `TEXT`                                                                                                                                                 |
 | `ENUM(...)`                   | `VARCHAR(255)` with `CHECK` constraint                                                                                                                 |
-| `SET(...)`                    | `TEXT` (comma-separated)                                                                                                                               |
-| `JSON`                        | `TEXT` (JSON.stringify)                                                                                                                                |
+| `SET(...)`                    | `JSONB` (DSQL does not support array column types)                                                                                                     |
```
anwesham-lab: we shouldn't be adding `does not` structured prohibitive qualifiers like this

src/aurora-dsql-mcp-server/kiro_power/steering/examples-schema.md
amaksimo: Example tag

* WHERE tags @> '["urgent"]'
* SELECT jsonb_array_elements_text(tags)
* WHERE tags ? 'shipped'

All three work directly on a JSONB column. On a JSON column, @> and ? don't exist, you'd have to write tags::jsonb @> '["urgent"]' on every query, and the planner can't push predicates through that cast, so you lose any chance to index it.

src/aurora-dsql-mcp-server/kiro_power/steering/mysql-full-example.md
```diff
@@ -69,8 +69,8 @@ transact(["CREATE INDEX ASYNC idx_products_category ON products(tenant_id, categ
...
+| `JSON`                        | `JSONB` (`JSON` also supported; prefer `JSONB` for queryable structured data)                                                                          |
```
anwesham-lab: this will not always be the case, it's better to not qualify IMO to ensure the agent probes the users on wants. If anything, I'd say to add an ASK claim here.

src/aurora-dsql-mcp-server/kiro_power/steering/mysql-full-example.md
```diff
@@ -98,8 +98,8 @@ transact(["CREATE INDEX ASYNC idx_products_category ON products(tenant_id, categ
 - **MUST map** all MySQL data types to DSQL equivalents before creating tables
 - **MUST convert** AUTO_INCREMENT to UUID with gen_random_uuid(), IDENTITY column with `GENERATED AS IDENTITY (CACHE ...)`, or explicit SEQUENCE -- ALWAYS use `GENERATED AS IDENTITY` for auto-incrementing columns (see [AUTO_INCREMENT Migration](ddl-auto-increment.md#auto_increment-migration))
 - **MUST replace** ENUM with VARCHAR and CHECK constraint
-- **MUST replace** SET with TEXT (comma-separated)
-- **MUST replace** JSON columns with TEXT
+- **MUST replace** SET with JSONB (DSQL does not support array column types)
+- **MUST map** JSON columns to `JSONB` (preferred) or `JSON`
```
anwesham-lab: this just isn't true anymore and it's definitely not a `MUST` 

I'd say it should be framed to ask for user needs and `SHOULD`

src/aurora-dsql-mcp-server/kiro_power/steering/mysql-type-mapping.md
```diff
@@ -97,7 +97,7 @@ Map MySQL data types to their DSQL equivalents.
 
 | MySQL Type     | DSQL Equivalent                                           | Notes                                                                                            |
 | -------------- | --------------------------------------------------------- | ------------------------------------------------------------------------------------------------ |
-| JSON           | TEXT                                                      | MUST store as TEXT                                                                               |
+| JSON           | JSONB                                                     | Prefer `JSONB` for queryable structured data; `JSON` also accepted                               |
```
anwesham-lab: same as above

src/aurora-dsql-mcp-server/kiro_power/steering/development-guide.md
anwesham-lab: can we please avoid `does not support` syntax. If we ever support in the future, this becomes a huge bottleneck. I am okay with PREFER JSONB, with the option for TEXT.

src/aurora-dsql-mcp-server/kiro_power/steering/development-guide.md
```diff
@@ -13,7 +13,7 @@ effortless scaling, multi-region viability, among other advantages.
 - **REQUIRED: Follow DDL Guidelines** - Refer to [DDL Rules](#schema-ddl-rules)
 - **SHALL repeatedly generate fresh tokens** - Refer to [Connection Limits](auth/authentication-guide.md#connection-rules)
 - **ALWAYS use ASYNC indexes** - `CREATE INDEX ASYNC` is mandatory
-- **MUST Serialize arrays/JSON as TEXT** - Store arrays/JSON as TEXT (comma separated, JSON.stringify)
+- **MUST serialize arrays** into a single-column representation; **PREFER `JSONB`** (operators work directly), with **`TEXT`** as a MAY for opaque columns — **ASK** the user (see [Schema Design Rules](#schema-design-rules))
```
anwesham-lab: weird phrasing??

```
- **MUST serialize arrays** into a single-column representation; **PREFER** 
`JSONB` (operators work directly), MAY fallback to `TEXT` or opaque columns;  
**ASK** for user prefernces (see [Schema Design Rules](#schema-design-rules))
```

anwesham-lab: note i added linebreaks in this comment for readability, delete them if you take this change.

src/aurora-dsql-mcp-server/kiro_power/steering/development-guide.md
```diff
@@ -53,9 +53,15 @@ effortless scaling, multi-region viability, among other advantages.
...
-- MUST store arrays as TEXT (comma-separated is recommended)
-- MUST store JSON objects as TEXT (JSON.stringify)
+- MUST use **simple PostgreSQL types:** VARCHAR, TEXT, INTEGER, BOOLEAN, TIMESTAMP, JSON, JSONB
+- MUST serialize arrays into a single-column representation:
+  - **PREFER `JSONB`** — `@>`, `?`, `?|`, `?&`, and `jsonb_array_elements_text` work directly; values validated and normalized at write
+  - **MAY use `TEXT`** when the column is opaque to the database (application reads the whole value, parses it, never queries inside)
+- For document columns:
+  - **`JSONB`** when querying with `@>`, `?`, or indexed JSONB paths
+  - **`JSON`** when writes dominate (no parse/sort overhead), when byte-exact input matters (audit, replay, payloads with duplicate keys), or when only `->`/`->>` is needed
+  - When migrating an existing `JSON` column, **SHOULD keep `JSON`**; **MAY upgrade to `JSONB`** if the application needs JSONB-only operators or indexed paths
```
anwesham-lab: Should always lead with the imperative

src/aurora-dsql-mcp-server/kiro_power/steering/examples-patterns.md
```diff
@@ -131,43 +131,67 @@ INSERT INTO distributors VALUES (nextval('order_seq'), 'nothing');
 
 ## Data Serialization
 
-**Pattern:** MUST store arrays and JSON as TEXT (runtime-only types). Per [DSQL docs](https://docs.aws.amazon.com/aurora-dsql/latest/userguide/working-with-postgresql-compatibility-supported-data-types.html), cast to JSON at query time.
+Arrays must be serialized into a single-column representation. The format choice is access-pattern-dependent. See [supported data types](https://docs.aws.amazon.com/aurora-dsql/latest/userguide/working-with-postgresql-compatibility-supported-data-types.html).
```
anwesham-lab: I think the **Pattern:** and Per [docs] framing should be maintained?

In this case, `The format choice is access-pattern-dependent.` means nothing to me.

src/aurora-dsql-mcp-server/kiro_power/steering/examples-patterns.md
```diff
@@ -131,43 +131,67 @@ INSERT INTO distributors VALUES (nextval('order_seq'), 'nothing');
...
-function toTextArray(values) {
-  return values.join(',');
-}
+**For arrays:**
 
-function fromTextArray(textValue) {
-  return textValue ? textValue.split(',').map(v => v.trim()) : [];
-}
+- **PREFER `JSONB`** when the application filters with `@>`/`?`/`?|`/`?&` or expands with `jsonb_array_elements_text`; values validated and normalized at write
+- **MAY use `TEXT`** when the column is opaque to the database — application reads the whole value, parses it, never queries inside (e.g. comma-separated tags handled entirely in app code)
```
anwesham-lab: this is still where we should explicitly be providing code examples. otherwise it isn't on theme with this doc's use as the `example_patterns`

the user gets referred to this doc from a reference file that already justified this decision usually

src/aurora-dsql-mcp-server/kiro_power/steering/examples-schema.md
```diff
@@ -20,12 +20,14 @@ CREATE TABLE IF NOT EXISTS orders (
   order_id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
   tenant_id VARCHAR(255) NOT NULL,
   status VARCHAR(50) NOT NULL,
-  tags TEXT,
-  metadata TEXT,
+  tags JSONB,         -- array; JSONB lets the queries below use @> directly (MAY use TEXT for opaque columns)
+  metadata JSONB,     -- document; JSONB chosen because queries below filter with `@>`/`?` (MAY keep `JSON` if writes dominate or byte-exact preservation matters)
   created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP
```
anwesham-lab, line 25: I see your point, it's fair, no need to add tokens to clarify. it would be nice if we could provide or update the example to allow for JSON and JSONB just to further condition the agent on allowance for both

anwesham-lab (dismissed):

src/aurora-dsql-mcp-server/kiro_power/steering/examples-patterns.md
```diff
@@ -131,43 +131,55 @@ INSERT INTO distributors VALUES (nextval('order_seq'), 'nothing');
...
-const categoriesText = toTextArray(['backend', 'api', 'database']);
-await pool.query('INSERT INTO projects (project_id, categories) VALUES ($1, $2)', [projectId, categoriesText]);
+**JSON (write-heavy, byte-exact, key-extraction only):**
 
-const configText = toTextJSON({ theme: 'dark', notifications: true });
-await pool.query('INSERT INTO user_settings (user_id, preferences) VALUES ($1, $2)', [userId, configText]);
+```javascript
+const auditPayload = { event: 'login', ts: 1717890000, user_id: '...' };
+await pool.query(
+  'INSERT INTO audit_log (id, payload) VALUES ($1, $2)',  -- no cast: column is JSON
```
praba2210: Nit: change `--` to `//`

praba2210 (changes requested):

src/aurora-dsql-mcp-server/kiro_power/steering/examples-patterns.md
amaksimo: Done in f9f3a8b — switched the trailing `--` to `//` since the snippet is in a `javascript` fence (also propagated to the mirror in `skills/dsql-skill/references/examples/patterns.md`).

praba2210 (approved):

anwesham-lab (approved):

PR #3799, opened 2026-06-10
fix(eks-mcp-server): add path validation to prevent file read/write in sensitive directories

srhsrhsrhsrh:
User-supplied file paths in generate_app_manifest, apply_yaml, _deploy_stack, and _generate_template were insufficiently validated, allowing operations targeting sensitive directories such as ~/.aws, ~/.ssh, ~/.kube, or /etc.

This adds a shared path_validation module that:
- Rejects relative paths
- Resolves symlinks via os.path.realpath before checking
- Blocks paths that resolve into sensitive directories

Applied consistently across both k8s_handler.py and eks_stack_handler.py to eliminate validation inconsistencies.

### Changes

Add a shared path_validation module to eks-mcp-server that validates all user-supplied file and directory paths before file system operations. The module:
  - Rejects relative paths
  - Resolves symlinks via os.path.realpath before checking
  - Blocks paths that resolve into sensitive directories (~/.aws, ~/.ssh, ~/.kube, ~/.gnupg, ~/.docker, /etc, /root)

Applied to:
  - k8s_handler.py — apply_yaml and generate_app_manifest (replaces the previous bare os.path.isabs() check)
  - eks_stack_handler.py — _deploy_stack and _generate_template (previously had no path validation at all)

### User experience

**Before**: A user (or an LLM agent via prompt injection) could supply paths like /home/user/.aws or /etc/cron.d as output_dir or template_file, and the server would read from or write to those sensitive locations when --allow-write was enabled.

**After**: Attempts to use sensitive directory paths are rejected with a clear error message (e.g., "Path points to a sensitive directory: /home/user/.aws/template.yaml"). Normal usage with project directories is unaffected.

Test output:
```
 ┌──────────────────────────────┬────────────────────────────┬──────────┬────────────────────────────────────────┐
 │             Test             │            Path            │ Expected │                 Result                 │
 ├──────────────────────────────┼────────────────────────────┼──────────┼────────────────────────────────────────┤
 │ generate_app_manifest        │ ~/.aws                     │ Blocked  │ "Path points to a sensitive directory" │
 ├──────────────────────────────┼────────────────────────────┼──────────┼────────────────────────────────────────┤
 │ apply_yaml                   │ ~/.ssh/config.yaml         │ Blocked  │ "Path points to a sensitive directory" │
 ├──────────────────────────────┼────────────────────────────┼──────────┼────────────────────────────────────────┤
 │ manage_eks_stacks (generate) │ /etc/cron.d/malicious.yaml │ Blocked  │ "Path points to a sensitive directory" │
 ├──────────────────────────────┼────────────────────────────┼──────────┼────────────────────────────────────────┤
 │ generate_app_manifest        │ /tmp/test-manifests        │ Allowed  │ Manifest written successfully          │
 └──────────────────────────────┴────────────────────────────┴──────────┴────────────────────────────────────────┘
```

## Checklist

* [X] Changes are documented

Is this a breaking change? N

**RFC issue number**: N/A

Checklist:

## Acknowledgment


src/eks-mcp-server/tests/test_k8s_handler.py
```diff
@@ -2004,3 +2009,89 @@ def test_filter_null_values(self, mock_mcp, mock_client_cache):
 
         result = handler.filter_null_values(complex_input)
         assert result == expected_complex
+
+
+class TestK8sHandlerPathValidation:
```
patrick-yu-amzn, line 2014: Instead of duplicating the functional tests across all users of the path validation logic, it's suffice to test the validation logic itself and then just ensure the users are actually calling it.

src/eks-mcp-server/awslabs/eks_mcp_server/path_validation.py
```diff
@@ -0,0 +1,79 @@
...
+
+BLOCKED_DIRS: tuple[str, ...] = (
+    os.path.join(_HOME, '.aws'),
+    os.path.join(_HOME, '.ssh'),
+    os.path.join(_HOME, '.kube'),
+    os.path.join(_HOME, '.gnupg'),
+    os.path.join(_HOME, '.docker'),
+    '/etc',
+    '/root',
+)
```
patrick-yu-amzn, line 30: This list seems pretty arbitrary and not really portal across OS. Maybe we should engage the open source mcp repo owner and see if there's a need for a more consistent approach -- eks can't be the only mcp server that work with local paths?

src/eks-mcp-server/tests/test_k8s_handler.py
srhsrhsrhsrh: Switched to mocking the validation instead of testing at handler level

patrick-yu-amzn (approved):
LGTM

src/eks-mcp-server/awslabs/eks_mcp_server/path_validation.py
srhsrhsrhsrh: Checked with admins and the validation will be maintained by each owning team

arnewouters (approved):

PR #3801, opened 2026-06-10
feat(cloudwatch-applicationsignals): add dynamic instrumentation tools 

pxaws:
## Description

This PR adds **dynamic instrumentation** tools to the CloudWatch Application Signals MCP server. Dynamic instrumentation lets users interactively debug live Application Signals services without redeploying — placing breakpoint- or probe-style instrumentation on running services and inspecting the captured snapshots (arguments, local state, stack traces) from CloudWatch Logs.

### Intent of this change

The goal is to **land the dynamic instrumentation tools now, bundled with a trimmed service model**, so that:

1. **The tools are usable immediately** against the preview dynamic instrumentation API.
2. **Once the public AWS SDK ships the new dynamic instrumentation APIs**, a follow-up PR will:
   - bump the `boto3`/`botocore` floor to the version that includes the operations,
   - rewire the client factory to use the public model instead of the bundled one,
   - and delete the bundled `aws_data/` model directory.

In other words: bundle the preview model today so the feature works, then cut over to the published Python SDK as soon as the APIs are generally available.

### Tools added (11)

**Configuration & status**
- `create_instrumentation`, `list_instrumentations`, `get_instrumentation`, `delete_instrumentation`, `batch_delete_instrumentations_by_scope`, `batch_delete_instrumentations_by_arns`
- `get_instrumentation_configuration_status`, `check_instrumentation_status`

**Snapshot analysis**
- `search_snapshots_for_status_event`, `get_sample_snapshot_for_breakpoint`

### Preview API note

These tools depend on `application-signals` dynamic instrumentation operations that are **not yet in the public AWS SDK**. The server bundles a trimmed preview service model (only the seven `*InstrumentationConfiguration*` operations the tools call, generated from the service team's source-of-truth Smithy model) loaded via a **session-scoped** botocore data loader — it does not mutate `AWS_DATA_PATH` and is invisible to the existing application-signals client. The bundled model is removed in the follow-up once the operations are generally available in `botocore` (see `dynamic_instrumentation/aws_data/README.md`).

The feature requires no special configuration: snapshots are read from a per-service CloudWatch Logs group (`/aws/service-events/<service-name>`) and the client uses normal AWS endpoint resolution.

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

- [x] `uv run --frozen pytest tests/` — 945 passed (66 dynamic-instrumentation tests)
- [x] `ruff check` / `ruff format` clean
- [x] pre-commit hooks pass (detect-secrets, license headers, json formatting, AST/credential checks)

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (server README, CHANGELOG, package docs)
- [x] My changes generate no new warnings
- [x] I have added tests that prove my feature works
- [x] New and existing unit tests pass locally with my changes

## By submitting this pull request

### DIFF
--- a/src/cloudwatch-applicationsignals-mcp-server/CHANGELOG.md
+++ b/src/cloudwatch-applicationsignals-mcp-server/CHANGELOG.md
@@ -1,5 +1,24 @@
 # Changelog
 
+## [Unreleased]
+
+### Added
+- Dynamic instrumentation tools (preview) for interactively debugging live
+  Application Signals services without redeploying:
+  - `create_instrumentation`, `list_instrumentations`, `get_instrumentation`,
+    `delete_instrumentation`, `batch_delete_instrumentations_by_scope`,
+    `batch_delete_instrumentations_by_arns` for managing instrumentation
+    configurations.
+  - `get_instrumentation_configuration_status`, `check_instrumentation_status`,
+    `report_instrumentation_configuration_status` for status inspection and
+    reporting.
+  - `search_snapshots_for_status_event`, `get_sample_snapshot_for_breakpoint`
+    for analyzing captured snapshots from CloudWatch Logs.
+  - Ships a trimmed preview `application-signals` service model bundled under
+    `dynamic_instrumentation/aws_data/`, loaded via a session-scoped botocore
+    data loader. The bundled model is removed once the operations are generally
+    available in `botocore`.
+
 ## [0.1.0] - 2025-01-18
 
 ### Added
--- a/src/cloudwatch-applicationsignals-mcp-server/CHANGELOG.md
+++ b/src/cloudwatch-applicationsignals-mcp-server/CHANGELOG.md
@@ -9,9 +9,8 @@
     `delete_instrumentation`, `batch_delete_instrumentations_by_scope`,
     `batch_delete_instrumentations_by_arns` for managing instrumentation
     configurations.
-  - `get_instrumentation_configuration_status`, `check_instrumentation_status`,
-    `report_instrumentation_configuration_status` for status inspection and
-    reporting.
+  - `get_instrumentation_configuration_status`, `check_instrumentation_status`
+    for status inspection.
   - `search_snapshots_for_status_event`, `get_sample_snapshot_for_breakpoint`
     for analyzing captured snapshots from CloudWatch Logs.
   - Ships a trimmed preview `application-signals` service model bundled under
--- a/src/cloudwatch-applicationsignals-mcp-server/README.md
+++ b/src/cloudwatch-applicationsignals-mcp-server/README.md
@@ -450,7 +450,6 @@ local state, and stack traces) from CloudWatch Logs.
 - `batch_delete_instrumentations_by_arns` — Batch delete instrumentation configurations by explicit resource ARN list.
 - `get_instrumentation_configuration_status` — Get status-event history for one instrumentation configuration and one explicit status.
 - `check_instrumentation_status` — Run a consolidated READY/ACTIVE/ERROR status check over a time window.
-- `report_instrumentation_configuration_status` — Report one or more instrumentation status events to the backend.
 
 **Snapshot analysis tools**
 
@@ -1053,8 +1052,7 @@ The server requires the following AWS IAM permissions:
         "application-signals:GetInstrumentationConfiguration",
         "application-signals:DeleteInstrumentationConfiguration",
         "application-signals:BatchDeleteInstrumentationConfigurations",
-        "application-signals:GetInstrumentationConfigurationStatus",
-        "application-signals:ReportInstrumentationConfigurationStatus"
+        "application-signals:GetInstrumentationConfigurationStatus"
       ],
       "Resource": "*"
     }
--- a/src/cloudwatch-applicationsignals-mcp-server/README.md
+++ b/src/cloudwatch-applicationsignals-mcp-server/README.md
@@ -426,6 +426,37 @@ query_rum_events(action="<action_name>", app_monitor_name="my-app", ...)
 
 **Optional parameters** (action-dependent): `resource_arn`, `page_url`, `group_by`, `platform`, `max_results`, `max_traces`, `statistic`, `period`, `session_id`, `metric`, `bucket`, `compare_previous`
 
+### 🔬 Dynamic Instrumentation Tools (Preview)
+
+Interactively debug live services without redeploying. Dynamic instrumentation
+lets you place breakpoint-style or probe-style instrumentation on running
+Application Signals services, then inspect the captured snapshots (arguments,
+local state, and stack traces) from CloudWatch Logs.
+
+> **Preview:** These tools depend on `application-signals` dynamic
+> instrumentation operations that are not yet part of the public AWS SDK. The
+> server bundles a trimmed preview service model (see
+> `awslabs/cloudwatch_applicationsignals_mcp_server/dynamic_instrumentation/aws_data/README.md`)
+> so the tools work against the preview API. The bundled model is removed once
+> the operations are generally available in `botocore`.
+
+**Configuration & status tools**
+
+- `create_instrumentation` — Create a dynamic instrumentation configuration for BREAKPOINT or PROBE.
+- `list_instrumentations` — List active instrumentation configurations for one service, environment, and type.
+- `get_instrumentation` — Get the full backend configuration for a single instrumentation target.
+- `delete_instrumentation` — Delete a single instrumentation configuration.
+- `batch_delete_instrumentations_by_scope` — Batch delete instrumentation configurations by scope.
+- `batch_delete_instrumentations_by_arns` — Batch delete instrumentation configurations by explicit resource ARN list.
+- `get_instrumentation_configuration_status` — Get status-event history for one instrumentation configuration and one explicit status.
+- `check_instrumentation_status` — Run a consolidated READY/ACTIVE/ERROR status check over a time window.
+- `report_instrumentation_configuration_status` — Report one or more instrumentation status events to the backend.
+
+**Snapshot analysis tools**
+
+- `search_snapshots_for_status_event` — Search CloudWatch Logs snapshots near a known instrumentation status timestamp.
+- `get_sample_snapshot_for_breakpoint` — Fetch one nearby snapshot to inspect the structure of captured data.
+
 ## Installation
 
 ### One-Click Installation
@@ -1016,7 +1047,14 @@ The server requires the following AWS IAM permissions:
         "iam:GetRole",
         "iam:ListAttachedRolePolicies",
         "iam:GetPolicy",
-        "iam:GetPolicyVersion"
+        "iam:GetPolicyVersion",
+        "application-signals:CreateInstrumentationConfiguration",
+        "application-signals:ListInstrumentationConfigurations",
+        "application-signals:GetInstrumentationConfiguration",
+        "application-signals:DeleteInstrumentationConfiguration",
+        "application-signals:BatchDeleteInstrumentationConfigurations",
+        "application-signals:GetInstrumentationConfigurationStatus",
+        "application-signals:ReportInstrumentationConfigurationStatus"
       ],
       "Resource": "*"
     }
--- a/src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/dynamic_instrumentation/aws_clients.py
+++ b/src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/dynamic_instrumentation/aws_clients.py
@@ -1,0 +1,91 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""boto3 client factory for the dynamic instrumentation feature.
+
+Loads the bundled private ``application-signals`` service model via a
+scoped botocore data loader (no ``AWS_DATA_PATH`` env mutation), pins
+the API version to ``2024-04-15``, and lazily caches the resulting
+clients so MCP tools can reuse a single configured instance.
+"""
+
+import boto3
+import botocore.session
+import os
+from .. import __version__
+from botocore.config import Config
+from loguru import logger
+from pathlib import Path
+
+
+AWS_DATA_PATH = Path(__file__).parent / 'aws_data'
+APPLICATION_SIGNALS_API_VERSION = '2024-04-15'
+
+
+def _resolve_region() -> str:
+    """Resolve AWS region: AWS_REGION env var > profile/config > us-east-1."""
+    env_region = os.environ.get('AWS_REGION')
+    if env_region:
+        return env_region
+    profile = os.environ.get('AWS_PROFILE')
+    session = boto3.Session(profile_name=profile)
+    if session.region_name:
+        return session.region_name
+    return 'us-east-1'
+
+
+def _build_config() -> Config:
+    mcp_source = os.environ.get('MCP_RUN_FROM')
+    user_agent_suffix = f'/{mcp_source}' if mcp_source else ''
+    return Config(
+        user_agent_extra=f'awslabs.cloudwatch-applicationsignals-mcp-server/dynamic-instrumentation/{__version__}{user_agent_suffix}'
+    )
+
+
+def _build_session() -> boto3.Session:
+    """Build a boto3 Session with the bundled private model loader prepended."""
+    botocore_session = botocore.session.Session()
+    botocore_session.get_component('data_loader').search_paths.insert(0, str(AWS_DATA_PATH))
+    profile = os.environ.get('AWS_PROFILE')
+    return boto3.Session(botocore_session=botocore_session, profile_name=profile)
+
+
+_application_signals_client = None
+
+
+def get_application_signals_client():
+    """Return a lazily-built ``application-signals`` client pinned to the private model."""
+    global _application_signals_client
+    if _application_signals_client is not None:
+        return _application_signals_client
+
+    region = _resolve_region()
+
+    session = _build_session()
+    _application_signals_client = session.client(
+        'application-signals',
+        api_version=APPLICATION_SIGNALS_API_VERSION,
+        region_name=region,
+        config=_build_config(),
+    )
+    logger.debug(
+        f'application-signals client initialized (region={region}, '
+        f'api_version={APPLICATION_SIGNALS_API_VERSION})'
+    )
+    return _application_signals_client
+
+
+def _reset_clients() -> None:
+    """Drop the cached client so tests can re-initialize against a fresh stub."""
+    global _application_signals_client
+    _application_signals_client = None
--- a/src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/dynamic_instrumentation/aws_clients.py
+++ b/src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/dynamic_instrumentation/aws_clients.py
@@ -33,15 +33,19 @@ APPLICATION_SIGNALS_API_VERSION = '2024-04-15'
 
 
 def _resolve_region() -> str:
-    """Resolve AWS region: AWS_REGION env var > profile/config > us-east-1."""
-    env_region = os.environ.get('AWS_REGION')
-    if env_region:
-        return env_region
-    profile = os.environ.get('AWS_PROFILE')
-    session = boto3.Session(profile_name=profile)
-    if session.region_name:
-        return session.region_name
-    return 'us-east-1'
+    """Resolve AWS region: AWS_REGION env var, else us-east-1.
+
+    Matches the parent package's ``aws_clients`` and every sibling MCP server:
+    region comes from ``AWS_REGION`` (falling back to ``us-east-1``) and a
+    configured profile's region is *not* consulted. This is deliberate — the
+    snapshot tools query CloudWatch Logs through the parent package's
+    ``logs_client``, which resolves region the same way. Honoring the profile
+    region here (but not there) would split a profile-only caller's clients
+    across two regions: instrumentations created in the profile region while
+    snapshot queries run in ``us-east-1``, surfacing as a breakpoint that shows
+    ACTIVE but whose snapshot searches always come back empty.
+    """
+    return os.environ.get('AWS_REGION', 'us-east-1')
 
 
 def _build_config() -> Config:
--- a/src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/dynamic_instrumentation/capture.py
+++ b/src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/dynamic_instrumentation/capture.py
@@ -1,0 +1,219 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Capture configuration ADT for dynamic instrumentation.
+
+Mirrors the ``Location`` design: a sealed sum type covers the two
+``CaptureConfiguration`` variants the ``application-signals`` API exposes,
+plus an ``UnknownCapture`` fallback so renderers stay forward-compatible.
+
+The type owns the API shape — payload assembly via ``to_api_payload`` and
+inverse parsing via ``capture_from_response``. It does *not* own prose
+rendering (renderers keep their CAPTURE SETTINGS / CAPTURE CONFIGURATION
+blocks) and it does *not* fill in tool-input defaults (defaults like
+``capture_return=True`` resolve at the tool layer where the success-message
+prose can read the resolved value).
+
+Two distinctions worth preserving across the round-trip:
+
+* ``CaptureArguments`` *omitted* means "capture all arguments"; renderers
+  show ``- Arguments: All``.
+* ``CaptureArguments`` *present-but-empty* means "capture none"; renderers
+  show ``- Arguments: (none)``.
+
+Both shapes survive parse → payload → parse round-trips.
+"""
+
+from dataclasses import dataclass, field
+from types import MappingProxyType
+from typing import Any, Dict, Mapping, Optional, Tuple, Union
+
+
+@dataclass(frozen=True)
+class CaptureLimits:
+    """Optional size caps applied to a code-capture payload."""
+
+    max_hits: Optional[int] = None
+    max_string_length: Optional[int] = None
+    max_collection_width: Optional[int] = None
+    max_collection_depth: Optional[int] = None
+    max_stack_frames: Optional[int] = None
+    max_stack_trace_size: Optional[int] = None
+    max_object_depth: Optional[int] = None
+    max_fields_per_object: Optional[int] = None
+
+    def is_empty(self) -> bool:
+        """Return True when no capture limit is set."""
+        return all(
+            v is None
+            for v in (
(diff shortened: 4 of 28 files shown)


mxiamxia:
One thing that jumped out from an MCP-best-practices angle: this PR introduces the server's first genuinely **mutating and destructive** tools — `create_instrumentation`, `report_instrumentation_configuration_status`, and especially the deletes (`delete_instrumentation`, `batch_delete_instrumentations_by_arns`, and `batch_delete_instrumentations_by_scope`).

Right now they're registered the same way as the read-only tools:

```python
mcp.tool()(delete_instrumentation)
mcp.tool()(batch_delete_instrumentations_by_scope)
```

It'd be great to give MCP clients a hint about which tools are read-only vs. destructive so they can warn the user or ask for confirmation before, say, a bulk delete. FastMCP supports tool annotations for exactly this:

```python
mcp.tool(annotations={"readOnlyHint": True})(list_instrumentations)
mcp.tool(annotations={"readOnlyHint": True})(get_instrumentation)
mcp.tool(annotations={"readOnlyHint": True})(get_instrumentation_configuration_status)
mcp.tool(annotations={"readOnlyHint": True})(check_instrumentation_status)
mcp.tool(annotations={"readOnlyHint": True})(search_snapshots_for_status_event)
mcp.tool(annotations={"readOnlyHint": True})(get_sample_snapshot_for_breakpoint)

mcp.tool(annotations={"destructiveHint": True})(delete_instrumentation)
mcp.tool(annotations={"destructiveHint": True})(batch_delete_instrumentations_by_arns)
mcp.tool(annotations={"destructiveHint": True})(batch_delete_instrumentations_by_scope)
```

pxaws:
Thanks @mxiamxia — good call, agreed these mutating/destructive tools shouldn't look identical to the read-only ones.

Added `ToolAnnotations` to all the dynamic instrumentation tools in `registration.py`. I followed the repo's existing convention (typed `ToolAnnotations` from `mcp.types`, as in `aws-appsync-mcp-server`) rather than the bare dict, and set the hints explicitly:

- **Read-only** (`readOnlyHint=True`): `list_instrumentations`, `get_instrumentation`, `get_instrumentation_configuration_status`, `check_instrumentation_status`, `search_snapshots_for_status_event`, `get_sample_snapshot_for_breakpoint`
- **Destructive** (`destructiveHint=True`, plus `idempotentHint=True` since deleting an absent config is a no-op): `delete_instrumentation`, `batch_delete_instrumentations_by_scope`, `batch_delete_instrumentations_by_arns`
- **State-changing, non-destructive** (`readOnlyHint=False, destructiveHint=False`): `create_instrumentation`

Every tool also gets `openWorldHint=True` since they all hit the AWS API. Added tests asserting the read-only / destructive / open-world hints are wired correctly.

Separately, I dropped `report_instrumentation_configuration_status` (and its now-dead supporting code) from this PR — it isn't needed for the interactive debugging flow.

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/dynamic_instrumentation/snapshot_tools.py
```diff
@@ -0,0 +1,208 @@
...
+    #   - For Java snapshots `resource.attributes.*` is populated; require an exact match.
+    #   - For Python snapshots the SDK currently emits an empty resource block, so we accept
+    #     records where the field is absent (`not ispresent(...)`). location_hash by itself
+    #     uniquely identifies (service, environment, location), so this fallback does not
+    #     widen the match across services.
+    #   - Assumption: location_hash collisions across services are negligible. If a future
+    #     SDK bug ever produces records with both a colliding hash and missing resource
+    #     attributes, this filter could return cross-service results.
+    base_filters = (
+        f'attributes.aws.di.location_hash = "{location_hash}"'
```
vastin: `service_name`, `environment`, and `location_hash` are dropped straight into the
Logs Insights filter string with no escaping. The same happens again in
`get_sample_snapshot_for_breakpoint` (lines 180-185) and for `custom_filters`
(line 96).

```python
base_filters = (
    f'attributes.aws.di.location_hash = "{location_hash}"'
    f' and (resource.attributes.service.name = "{service_name}"'
    ...
```

**Why it matters:** If any of these values contains a double quote, it closes the
string literal and the rest becomes caller-controlled query syntax — at best the
query errors out, at worst it returns snapshots from **other services**. The
`limit` value at line 102 is also interpolated raw.

**Fix:**
- Escape and quote-validate `service_name` / `environment` / `location_hash`
  (there's already an unused `_escape_logs_insights_regex` helper in
  `snapshot_parsing.py`).
- Validate `location_hash` as a 16-char hex string before use.
- Coerce `limit` to `int`.
- Validate / scope `custom_filters`.

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/dynamic_instrumentation/aws_clients.py
```diff
@@ -0,0 +1,91 @@
...
+from botocore.config import Config
+from loguru import logger
+from pathlib import Path
+
+
+AWS_DATA_PATH = Path(__file__).parent / 'aws_data'
+APPLICATION_SIGNALS_API_VERSION = '2024-04-15'
+
+
+def _resolve_region() -> str:
```
vastin, line 35: The DI client resolves its region as `AWS_REGION` env → profile region →
`us-east-1`. But the snapshot tools use the **parent** package's `logs_client`,
which only reads the `AWS_REGION` env var (defaulting to `us-east-1`) and ignores
the profile region.

**Why it matters:** If someone sets only `AWS_PROFILE` (region in the profile)
and no `AWS_REGION`, instrumentations are created in the profile region while
snapshot queries run in `us-east-1`. The breakpoint shows ACTIVE but snapshot
searches always come back empty — very hard to diagnose.

**Fix:** Reuse the parent client factory so region resolution is identical across
all tools.

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/dynamic_instrumentation/status_tools.py
```diff
@@ -0,0 +1,334 @@
...
+        else {}
+    )
+    if not config:
+        return f'ERROR: Failed to fetch created_at: No instrumentation found for LocationHash {location_hash}'
+    created_dt = config.get('CreatedAt')
+    if created_dt is None:
+        return 'ERROR: Failed to fetch created_at: CreatedAt not found in instrumentation configuration'
+
+    try:
+        start_dt = _parse_iso_timestamp(start_time)
```
vastin, line 303: `_parse_iso_timestamp` returns a **naive** datetime when the input has no `Z` or
offset (e.g. `2025-02-03T18:42:00`). Later `assess()` calls `.astimezone(utc)` on
it, which Python interprets in the host's **local** timezone.

**Why it matters:** Verified on a UTC-8 host — `18:42` becomes `02:42Z`, an
8-hour shift. That moves the whole status query window, so ACTIVE/READY events
get missed or the ACTIVE check is wrongly skipped.

**Fix:** Default naive inputs to UTC at parse time.

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/dynamic_instrumentation/snapshot_parsing.py
```diff
@@ -0,0 +1,292 @@
...
+                'line_number': frame.get('line_number'),
+            }
+        )
+
+    all_line_numbers = sorted(
+        set(line_locals.keys())
+        | set(line_arguments.keys())
+        | set(line_return_values.keys())
+        | set(line_throwables.keys()),
+        key=lambda v: int(v) if str(v).isdigit() else v,
```
vastin: The sort key returns an `int` for digit-only strings and the raw `str`
otherwise, then `sorted()` compares them.

**Why it matters:** If a snapshot has numeric line keys plus one non-digit key
(e.g. a negative line `'-1'`, since `'-1'.isdigit()` is `False`), this raises
`TypeError: '<' not supported between instances of 'str' and 'int'`, crashing
`get_sample_snapshot_for_breakpoint` while parsing a customer snapshot.

**Fix:** Coerce all keys to a single comparable type (e.g. sort non-numeric keys
last).

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/dynamic_instrumentation/snapshot_tools.py
pxaws: Good catch — fixed in `8fe278a4`. The structured fields are now hardened at the source:
- `location_hash`: validated as 16-char lowercase hex (`is_valid_location_hash`) before use — hex can't contain a quote, so this closes the breakout entirely for the hash.
- `service_name` / `environment`: escaped for double-quoted string-literal context via a new `_escape_logs_insights_string` helper.
- `limit`: coerced to `int`.
- `custom_filters`: kept as documented raw fragments, but rejected if they contain an unbalanced (unescaped) double-quote — the actual corruption vector.

One note: I didn't reuse `_escape_logs_insights_regex` as suggested — that one escapes for `/.../` regex context (`re.escape` + slash), which is wrong for `"..."` literals (it would mangle normal chars and not escape the quote correctly). Hence the separate string-literal helper.

Both query builders now share a single `_build_base_filters(location_hash, service_name, environment)` so the two tools escape identically and can't drift. Added tests asserting an injected `"` is backslash-escaped, plus the validation/limit/quote-balance guards.

Heads-up on a related tightening that came out of this: `check_instrumentation_status` previously accepted any 16 characters as a `location_hash`; it now uses the same `is_valid_location_hash` (16-char **lowercase hex**). That matches the service model's documented `%016x` contract (`LocationHashUtil`), so it should be strictly correct, but flagging it since it's stricter than before.

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/dynamic_instrumentation/aws_clients.py
pxaws: Fixed in `8fe278a4`, though I went the opposite direction from "reuse the parent factory" — here's why. I traced the history: the parent's `AWS_REGION` env → `us-east-1` (ignoring profile region) is the **repo-wide convention** — it dates back to the earliest awslabs/mcp servers, and every sibling with a profile branch (lambda-tool, stepfunctions-tool, elasticache) does the same. DI's `_resolve_region` with the profile fallback is actually the **only** client in the repo that consults profile region — so DI is the outlier, not the parent.

So I aligned DI down to the convention: `_resolve_region` is now just `AWS_REGION` env, else `us-east-1`. Both clients in the package (app-signals + the parent's `logs_client`) now resolve region identically, which kills the split you described, without spreading a non-standard profile-aware path or touching the parent's blast radius. Trade-off: a profile-only user must set `AWS_REGION` — which is the documented norm everywhere else in the repo.

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/dynamic_instrumentation/status_tools.py
pxaws: Fixed in `8fe278a4`. `_parse_iso_timestamp` now defaults a naive input (no `Z`/offset) to UTC via `.replace(tzinfo=timezone.utc)` — which attaches UTC without shifting the clock digits, so `18:42` stays `18:42Z` instead of being reinterpreted as host-local. Since `_parse_iso_timestamp` is the single choke point for every caller-supplied timestamp in this file (StartTime, EndTime, and both `assess()` inputs), fixing it there covers all of them. This also brings `status_tools` in line with what the snapshot tools already do inline. Added direct tests for the naive, `Z`-suffix, and explicit-offset cases.

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/dynamic_instrumentation/snapshot_parsing.py
pxaws: Fixed in `8fe278a4`. Replaced the mixed-type key with a `(group, value)` tuple via a small `_line_key` helper: numeric keys get `(0, int(text), '')` and non-numeric get `(1, 0, text)`. The group tag (always int-vs-int) partitions the two kinds so an `int` is never compared against a `str` — numeric keys sort first by value (so `'9'` before `'10'`), non-numeric keys sort last lexically. Total order, no crash. Added a test with mixed numeric + `'-1'` + a string key (your exact trigger, since `'-1'.isdigit()` is `False`).

vastin (approved):

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/dynamic_instrumentation/capture.py
```diff
@@ -0,0 +1,222 @@
...
+blocks) and it does *not* fill in tool-input defaults (defaults like
+``capture_return=True`` resolve at the tool layer where the success-message
+prose can read the resolved value).
+
+Two distinctions worth preserving across the round-trip:
+
+* ``CaptureArguments`` *omitted* means "capture all arguments"; renderers
+  show ``- Arguments: All``.
+* ``CaptureArguments`` *present-but-empty* means "capture none"; renderers
+  show ``- Arguments: (none)``.
```
srprash, line 32: `CaptureArguments` and `CaptureLocals` omitted-vs-empty semantics are documented exactly backwards. `[]` = all / omitted = none; map `["*"]` → `[]` for both capture_arguments and capture_locals instead of stripping; fix `crud_rendering.py` so `[]` renders "All (empty list = capture everything)" and absent renders "None (field omitted)".

src/cloudwatch-applicationsignals-mcp-server/README.md
```diff
@@ -426,6 +426,36 @@ query_rum_events(action="<action_name>", app_monitor_name="my-app", ...)
 
 **Optional parameters** (action-dependent): `resource_arn`, `page_url`, `group_by`, `platform`, `max_results`, `max_traces`, `statistic`, `period`, `session_id`, `metric`, `bucket`, `compare_previous`
 
+### 🔬 Dynamic Instrumentation Tools (Preview)
```
srprash, line 429: No longer a preview

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/dynamic_instrumentation/validation.py
```diff
@@ -0,0 +1,180 @@
...
+) -> Tuple[str, Optional[str]]:
+    """Normalize the type to upper-case; return ``(normalized, error)``.
+
+    The normalized value is always a ``str`` (the upper-cased received value
+    even on the error path) so callers get a non-optional type once they
+    return early on ``error``. Callers must check ``error`` before using
+    ``normalized``.
+    """
+    normalized = (instrumentation_type or '').strip().upper()
+    allowed = {'BREAKPOINT', 'PROBE'}
```
srprash, line 35: - PROBE is not supported on JS. Should be rejected or at least warned.
- Although the API allows for PROBE with line number, the SDKs simply ignores the line number. The MCP should warn or disallow this combination.
- Python SDK  requires `instrumentation_name` to be present for a PROBE. I think this is a bug and the SDK shouldn't impose such a requirement. That said, the attached service model doesn't mention `InstrumentationName` at all. Should it?
- Docstrings should note that `max_hits` and `expiration` are ignored for PROBE.

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/dynamic_instrumentation/validation.py
```diff
@@ -0,0 +1,180 @@
...
+            f'- class_name={class_name}',
+            f'- method_name={method_name}',
+            f'- line_number={line_number}',
+        ]
+    )
+
+    return '\n'.join(lines)
+
+
+def _validate_location_inputs(
```
srprash, line 130: Each language requirements regarding the target location must be clarified in the tool otherwise the config gets silently dropped by the SDK (which we should also fix by reporting ERROR status).

- Java requires `code_unit`, `class_name` and `method_name`
- Python requires `code_unit` and `method_name` (optionally qualified by `class_name`)
- JavaScript requires `line_number` >=1

Per-language docstring guidance on which fields target vs. which are metadata belongs in `location.py`

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/dynamic_instrumentation/validation.py
```diff
@@ -0,0 +1,180 @@
...
+    language: str,
+    file_path: str,
+    code_unit: Optional[str],
+    class_name: Optional[str],
+    method_name: Optional[str],
+    line_number: Optional[int],
```
srprash, line 136: A few quirks around the non-executable `line_number` that are worth documenting:
- Java and Python : ignores the non-executable line and it never fires.
- JavaScript: the breakpoint slides to the next parseable line number and fires of that line number is hit. That's just the behavior from base technique. 

the agent must be aware of these behaviors and target the line number very carefully.

mxiamxia (approved):

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/dynamic_instrumentation/crud_tools.py
```diff
@@ -0,0 +1,626 @@
...
+        file_path=file_path,
+        code_unit=code_unit,
+        class_name=class_name,
+        method_name=method_name,
+        line_number=line_number,
+    )
+
+    wildcard_removed = False
+    if capture_arguments is None:
+        return (
```
srprash, line 149: For line-level instrumentation, `capture_locals` is the only field that produces data - yet the tool mandates `capture_arguments`.
Drop the `capture_arguments` mandate for line-level configs (and for JS entirely); when line_number is set, recommend `capture_locals` = specific values or `[]` and warn if it is omitted; suppress the Arguments/Return lines in the success message for line-level and JS configs.

src/cloudwatch-applicationsignals-mcp-server/CHANGELOG.md
```diff
@@ -1,5 +1,23 @@
 # Changelog
 
+## [Unreleased]
+
+### Added
+- Dynamic instrumentation tools (preview) for interactively debugging live
```
srprash, line 6: not in preview anymore

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/dynamic_instrumentation/crud_rendering.py
```diff
@@ -0,0 +1,367 @@
...
+    code_capture_stack_trace: Optional[bool],
+    max_hits: Optional[int],
+    max_string_length: Optional[int],
+    max_collection_width: Optional[int],
+    max_collection_depth: Optional[int],
+    max_stack_frames: Optional[int],
+    max_stack_trace_size: Optional[int],
+    max_object_depth: Optional[int],
+    max_fields_per_object: Optional[int],
+    attribute_filters: Optional[List[Dict[str, str]]],
```
srprash, line 88: docstring should describe the functionality and usage of `attribute_filters`. The tool should describe that these are keys are OTel resource attribute names matched exactly; OR across groups, AND within.

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/dynamic_instrumentation/crud_rendering.py
```diff
@@ -0,0 +1,367 @@
...
+    """Render the success message for a created instrumentation configuration."""
+    location_hash = response.get('LocationHash', 'N/A')
+    arn = response.get('ARN', 'N/A')
+    created_at = format_timestamp(
+        response.get('CreatedAt'),
+        default=datetime.now(timezone.utc).strftime('%Y-%m-%dT%H:%M:%SZ'),
+    )
+    actual_expires_at = format_timestamp(response.get('ExpiresAt'), default='')
+
+    success_message = f"""Successfully created {normalized_type} instrumentation
```
srprash, line 99: An agent checking status seconds after create sees nothing and may retry/recreate in a loop. Suggest adding one line in the create success message: "Expect up to ~1–2 min (BREAKPOINT) / ~10–12 min (PROBE) before the configuration reports READY."

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/dynamic_instrumentation/validation.py
```diff
@@ -0,0 +1,196 @@
...
+    suggestions: List[str] = []
+
+    if not file_path or not str(file_path).strip():
+        errors.append('file_path is required and must be non-empty.')
+
+    if line_number is not None and line_number < 1:
+        errors.append(f'line_number must be >= 1 (received: {line_number}).')
+
+    if lang not in {'python', 'java'}:
+        errors.append(f'language must be Python or Java (received: {language}).')
```
srprash, line 151: must include JS too

pxaws:
I will address comment from @srprash in another PR: https://github.com/awslabs/mcp/pull/3817

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/dynamic_instrumentation/capture.py
pxaws: Thanks — rather than flip the omitted-vs-empty semantics (which I think it's confusing), we removed the ambiguity at the input boundary: `capture_arguments` and `capture_locals` now reject both `[]` and the `*` wildcard, and an omitted list means "capture none". So there's no longer an ambiguous "empty = all" shape to get backwards. On the read path, renderers report the raw shape (`(not set)` / `(empty list)` / the names) instead of asserting "All"/"None". Addressed in a follow-up PR (commit `41ac5c37`).

src/cloudwatch-applicationsignals-mcp-server/README.md
pxaws: Done — dropped the `(Preview)` from the heading. Kept a short `Note:` that the underlying `application-signals` operations aren't in public botocore yet and the model is bundled until GA (still factually true), but it no longer labels the feature itself a preview. Follow-up PR, commit `41ac5c37`.

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/dynamic_instrumentation/validation.py
pxaws: Addressed (commit `41ac5c37`):
- **PROBE on JS** — now rejected with a hard error pointing to BREAKPOINT.
- **PROBE + line_number** — now rejected (the SDKs ignore the line, so silently misbehaving is worse than an explicit error).
- **`max_hits` / expiration ignored for PROBE** — documented. The docstrings now emphasize that PROBE has no `max_hits` and fires on every hit (unlike BREAKPOINT) — good for long-running observation without worrying about a limit — but a PROBE doesn't expire on its own (`ttl_hours` ignored), so it must be deleted explicitly.
- **Python PROBE requires `instrumentation_name`** — you're right that `InstrumentationName` isn't in the attached service model, so MCP has nothing to send and we made no code change here. Agreed this looks like an SDK-side requirement that shouldn't be imposed; flagging it as out of scope for the MCP server. Let me know if you want it tracked separately.

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/dynamic_instrumentation/validation.py
pxaws: Addressed (commit `41ac5c37`). `_validate_location_inputs` now enforces the per-language required fields up front as hard errors, so an unbindable config is rejected before the API call rather than silently dropped:
- **Java** — requires `code_unit`, `class_name`, and `method_name`.
- **Python** — requires `code_unit` and `method_name` (`class_name` optional).
- **JavaScript** — requires `line_number` (>= 1).

The target-vs-metadata guidance now lives in `location.py` on the `CodeLocation` docstring as you suggested.

On "which we should also fix by reporting ERROR status" — that's an SDK/backend behavior, out of scope for the MCP server; this change makes MCP reject the config up front so the silent-drop path isn't reached from here.

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/dynamic_instrumentation/validation.py
pxaws: Documented (commit `41ac5c37`). The non-executable `line_number` behavior is now called out in two places so the agent sees it both proactively and on failure:
- `create_instrumentation` Notes: target an executable statement — Python/Java ignore a non-executable line (blank/comment/decorator/signature) and never fire; JavaScript slides the breakpoint to the next parseable line.
- The code-location troubleshooting block adds the same language-specific caution when a `line_number` is set.

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/dynamic_instrumentation/crud_tools.py
pxaws: Addressed (commit `41ac5c37`). The unconditional `capture_arguments` mandate is gone. For line-level configs (`line_number` set), the tool now requires `capture_locals` instead (hard error if missing) — since only locals carry data mid-function — and the create success message suppresses the Arguments and Return Values lines for line-level targets (Stack Traces stays, as it's meaningful at a line). JavaScript is always line-level per its location rules, so this requirement applies to JS automatically.

One deviation from the literal suggestion: we don't accept `capture_locals=[]` as a "recommend" option — an empty list is rejected (see the capture-inputs thread); the agent provides explicit local names or omits it.

src/cloudwatch-applicationsignals-mcp-server/CHANGELOG.md
pxaws: Done — removed "preview" here too. Follow-up PR, commit `41ac5c37`.

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/dynamic_instrumentation/crud_rendering.py
pxaws: Done (commit `41ac5c37`). The `attribute_filters` arg docstring now describes the semantics: each group is a dict of OpenTelemetry resource-attribute names to exact-match values; conditions are AND-ed within a group and groups are OR-ed together; up to 10 groups (key 1-50, value 1-100 chars); omit to apply to all instances.

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/dynamic_instrumentation/crud_rendering.py
pxaws: Done (commit `41ac5c37`). The create success message now includes a type-aware READY-timing hint so an agent doesn't poll-and-recreate in a loop: "Allow ~1-2 min (BREAKPOINT) / ~10-12 min (PROBE) before this configuration reports READY. Status checks immediately after creation may return no events yet — wait and re-check rather than recreating."

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/dynamic_instrumentation/validation.py
pxaws: Done (commit `41ac5c37`). JavaScript is now a first-class language: the validator accepts it (the language check is `Python`, `Java`, or `JavaScript`), and `parse_create_inputs` normalizes the casing to the API's `ProgrammingLanguage` enum (`Javascript`) so a validated language isn't rejected by the backend on a casing mismatch. A JavaScript-specific rules block was added to the troubleshooting output as well.

PR #3807, opened 2026-06-11
chore: add path validation for params_file to block sensitive file reads

jade710:
### Changes

Adds server-side path validation for the `params_file` parameter in `manage_hyperpod_stacks` to prevent reading sensitive files. Introduces `validate_file_read_path()` which rejects relative paths, blocks sensitive directories (`~/.aws`, `~/.ssh`, `~/.gnupg`, `~/.docker`, `/etc/passwd`, `/etc/shadow`, `/proc`), and blocks known sensitive filenames (`credentials`, `id_rsa`, `.env`, etc.). Pattern is consistent with `aws-transform-mcp-server`'s `file_validation.py`.

### User experience

**Before:** `params_file` was passed directly to `open()` with no validation. Any absolute or relative path would be accepted.

**After:** Relative paths, paths to sensitive directories, and sensitive filenames are rejected with a clear `ValueError`. Normal usage (absolute path to a user-created JSON params file) is unaffected.

## Checklist

- [x] Changes are documented

**RFC issue number**: N/A

Checklist:
- [ ] Migration process documented
- [ ] Implement warnings (if it can live side by side)

## Acknowledgment

### DIFF
--- a/src/sagemaker-ai-mcp-server/awslabs/sagemaker_ai_mcp_server/path_validation.py
+++ b/src/sagemaker-ai-mcp-server/awslabs/sagemaker_ai_mcp_server/path_validation.py
@@ -1,0 +1,103 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


JunqiYe (approved):

scottschreckengaust (approved):
LGTM

PR #3810, opened 2026-06-12
fix(mssql-mcp-server,oracle-mcp-server): Hardening secret manager access by making it a CLI parameter

shreyaskshekhar:
Hardening secret manager access by making it a CLI parameter instead being a parameter in the MCP tool, so that LLM can't control this parameter.


src/mssql-mcp-server/awslabs/mssql_mcp_server/server.py
```diff
@@ -490,29 +483,37 @@
...
-            rds_client = boto3.client(
-                'rds', region_name=region, config=Config(user_agent_extra=__user_agent__)
-            )
+    # Resolve the Secrets Manager ARN. Per-target override wins, then the
+    # bare default ARN, then the instance MasterUserSecret metadata.
+    # The LLM cannot influence the secret selection — only the operator
+    # controls which ARNs are available via startup --secret_arn flags.
+    target_key = instance_identifier
+    per_target_secret_arn = server_config.configured_secret_arns.get(target_key, '')
+    metadata_secret_arn = ''
```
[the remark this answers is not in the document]
shreyaskshekhar: '' just means "nothing configured yet" — it's a falsy placeholder, not a credential

RobinGolden:
lgtm!

scottschreckengaust:
## Code review — Secrets Manager hardening

Thanks for this — the core security objective (taking `secret_arn` out of the LLM's reach) is achieved and the implementation is fundamentally sound. The LLM can no longer inject an arbitrary ARN; it can only select among operator-registered instance keys, with the bare default bounding the blast radius. A few items below partially undercut the stated hardening goal and are worth addressing before merge. Findings are ordered most-severe first.

---

### 🔴 Critical / blocking

**1. mssql silently serves a stale cached connection when the resolved ARN changes** — `src/mssql-mcp-server/awslabs/mssql_mcp_server/server.py:465-478`

In mssql's `internal_create_connection`, the cache lookup returns early on any hit **before** the ARN-resolution block runs, and there is no comparison of the resolved ARN against the cached connection's ARN. The Oracle side does the right thing — it re-resolves the ARN first and evicts/replaces the cached connection when it differs (`src/oracle-mcp-server/awslabs/oracle_mcp_server/server.py:425-443`).

Impact: after a secret rotation or an operator reconfiguration that changes which ARN maps to an instance, mssql keeps serving the **stale** cached connection with no log and no error. This is a silent failure of the very hardening this PR introduces, and it makes the two servers diverge in behavior. Recommend mssql mirror Oracle's "resolve ARN → compare → evict/replace" ordering.

**2. No regression guard for the security invariant the PR exists to establish**

Nothing in the test suite asserts that `connect_to_database` no longer exposes `secret_arn` to the LLM. The existing tests were merely edited to drop the `secret_arn=` kwarg from calls to the **internal** function (`internal_create_connection`) — a different surface. A future refactor that re-adds a `secret_arn` parameter to the LLM-facing `connect_to_database` tool would pass every test in this PR.

Recommend adding a test in both servers that introspects the **registered MCP tool's input schema** (the contract the LLM actually sees, not just the Python signature) and asserts `secret_arn` is absent. This is the single most valuable test to add.

---

### 🟠 Important

**3. Empty bare `--secret_arn` value is silently skipped** — `src/mssql-mcp-server/.../server.py:723-725`, `src/oracle-mcp-server/.../server.py:715-717`

```python
arn = raw.strip()
if not arn:
    continue
```

This is the lone malformed-input case in the parser that does **not** hard-fail — every sibling case (empty key/arn in `key=arn`, duplicate key, second bare default) calls `logger.error(...) + sys.exit(2)`. An operator passing an unexpanded shell variable (`--secret_arn "$SECRET_ARN"` with `SECRET_ARN` unset) gets no warning, no exit, and silently falls through to MasterUserSecret — exactly the "operator misconfiguration hidden" scenario. Recommend `logger.error` + `sys.exit(2)` for parity (or at minimum a `logger.warning`).

**4. Oracle block comment is inaccurate (comment rot)** — `src/oracle-mcp-server/.../server.py:411-414`

The comment ("Per-target override wins, then the bare default ARN, then the instance MasterUserSecret metadata. The LLM cannot influence the secret selection...") is copied verbatim from the mssql version, but the Oracle code does not implement that flow here:
- The entire per-target/default resolution (lines 416-423) is gated on `connection_method == ConnectionMethod.ORACLE_PASSWORD`. For any non-PASSWORD method, the operator's configured ARNs are never consulted — the comment gives no hint of this.
- The "then the instance MasterUserSecret metadata" tier is not part of this block; it lives in a separate later block (lines 483-492) reached only when `describe_db_instances` runs.

Recommend rewriting the Oracle comment to reflect its actual ORACLE_PASSWORD-gated, two-tier resolution rather than reusing the mssql wording.

**5. mssql lacks the RDS error handling Oracle has** — `src/mssql-mcp-server/.../server.py:491-498`

```python
instance_props = rds_client.describe_db_instances(
    DBInstanceIdentifier=instance_identifier
)['DBInstances'][0]
```

A non-existent instance raises a raw `ClientError`; an unexpected empty `DBInstances` list raises a bare `IndexError`. Oracle wraps the equivalent call in `try/except ClientError`, special-cases `DBInstanceNotFound` into a clear `ValueError`, and guards the empty-list case (`src/oracle-mcp-server/.../server.py:455-492`). Neither mssql path is a *silent* failure, but both surface as opaque, non-actionable stack traces. The PR rewrote this block and had the opportunity to align the two servers.

**6. Untested fall-through scenarios**
- Per-target ARN winning over a **simultaneously configured** default — untested for **mssql** (`test_internal_create_connection_uses_per_target_secret_arn` sets only the map, no default). Oracle covers this.
- Per-target map present but `instance_identifier` does **not** match any key → must fall to default/metadata and must **not** borrow another instance's secret — untested in **both** servers. This is the highest-risk resolution branch; a future change to the lookup (case-insensitive, prefix match, wrong `.get` default) would go uncaught.

---

### 🟡 Suggestions (operator-experience polish)

- **Misleading final error message** — `mssql .../server.py:513-515` and `oracle .../server.py:498-502` say "no `--secret_arn` was provided" even when the operator provided one that simply didn't match the instance key (per-target miss). Reword to reflect the resolver, e.g. "no per-target or default `--secret_arn` matched, and the instance has no managed MasterUserSecret."
- **Operator-facing message references an internal field** — `oracle .../server.py:501-502` tells operators to "set a `configured_default_secret_arn`," which is an internal `ServerConfig` attribute they can't set directly. Point at the actual CLI surface: `--secret_arn <arn>` (bare default) or `--secret_arn <instance>=<arn>`.
- **Asymmetric audit logging** — Oracle logs the *default* ARN selection but not the security-sensitive *per-target* hit (`oracle .../server.py:418-422`); mssql logs neither. Logging which secret was selected for each connection would aid traceability of the security-relevant path.
- **Help-text precedence ambiguity** — the `--secret_arn` help implies the MasterUserSecret is a fallback *below* the bare default, but a configured bare default actually **supersedes** metadata for all unpinned instances. Worth clarifying that a default is not a fallback beneath metadata.
- **Whitespace stripping untested** — the parser calls `.strip()` on key, arn, and the bare ARN, but no test feeds padded input (e.g. `' inst1 = arn:...x '`) to confirm trimming, nor a whitespace-only key to exercise the `.strip()`-collapses-to-empty path.

---

### ✅ Strengths

- Core security boundary correctly achieved: `secret_arn` removed from both `connect_to_database` signatures; the LLM can no longer inject an arbitrary ARN.
- Log-hygiene improvements are clean — dropping the `Secret keys: ...` debug line and the "Available keys" list from `ValueError`s reduces info leakage **without** weakening specificity (the expected field names are still listed, so messages stay actionable).
- CLI parsing is correct: `partition('=')` handles `=` within ARNs, and duplicate-key / multiple-bare-default rejection is sound.
- Mutable-default handling is correct in both servers (Oracle's `@dataclass` + `__post_init__` reassigning `{}`, mssql's plain `__init__` assigning a fresh `{}`).
- The `--secret_arn` argparse help text is identical and accurate across both servers. Oracle's added RDS error-path tests assert on message contracts (refactoring-resilient), and mssql's `test_new_database_connection_uses_startup_secret_not_rds_master` is a strong behavioral test.

---

### Meta note
The two servers' resolution paths have drifted — caching/eviction, RDS error handling, connection-method gating, and logging all differ — while the `--secret_arn` parsing block is copy-pasted verbatim. A shared helper for both the parsing and the resolution chain would keep the two servers from diverging further and prevent a fix in one from silently missing the other.

*Review assisted by Claude Code (pr-review-toolkit + code-review).*

src/mssql-mcp-server/awslabs/mssql_mcp_server/server.py
```diff
@@ -490,29 +483,37 @@ def internal_create_connection(
...
-        # First try to use the default secret_arn from startup args
-        if server_config.default_secret_arn:
-            secret_arn = server_config.default_secret_arn
-            logger.info('Using default secret_arn from startup configuration')
-        else:
-            # Fall back to the RDS instance's managed master secret
-            rds_client = boto3.client(
-                'rds', region_name=region, config=Config(user_agent_extra=__user_agent__)
-            )
+    # Resolve the Secrets Manager ARN. Per-target override wins, then the
```
scottschreckengaust: **Critical — stale cached connection silently served after the resolved ARN changes.**

In this `internal_create_connection`, the cache lookup returns early on any hit *before* this ARN-resolution block runs, and there's no comparison of the resolved ARN against the cached connection's ARN. The Oracle server does the right thing — re-resolves the ARN first and evicts/replaces the cached connection when it differs (`src/oracle-mcp-server/awslabs/oracle_mcp_server/server.py:425-443`).

Impact: after a secret rotation or an operator reconfiguration that remaps an instance, mssql keeps serving the **stale** cached connection with no log and no error — a silent failure of the very hardening this PR introduces, and a behavioral divergence from Oracle. Recommend mirroring Oracle's resolve → compare → evict/replace ordering.

src/mssql-mcp-server/awslabs/mssql_mcp_server/server.py
```diff
@@ -292,12 +293,6 @@ async def connect_to_database(
     ctx: Context,
     port: Annotated[int, Field(description='SQL Server port')] = 1433,
     database: Annotated[str, Field(description='database name')] = 'master',
```
scottschreckengaust, line 295: **Critical — no regression guard for the security invariant this PR exists to establish.**

Removing `secret_arn` from this tool signature is exactly the right change, but nothing in the test suite locks it. The existing tests were only edited to drop the `secret_arn=` kwarg from calls to the *internal* function — a different surface. A future refactor that re-adds a `secret_arn` parameter to the LLM-facing `connect_to_database` tool would pass every test in this PR.

Recommend adding a test (both servers) that introspects the **registered MCP tool's input schema** — the contract the LLM actually sees — and asserts `secret_arn` is absent. This is the single most valuable test to add here.

scottschreckengaust (changes requested):
Requesting changes on two critical items that partially undercut the hardening goal — details inline below, with the full set of important/minor findings and strengths in [this consolidated comment](https://github.com/awslabs/mcp/pull/3810#issuecomment-4754727433).

The core objective (taking `secret_arn` out of the LLM's reach) is achieved and the design is sound — the LLM can only select among operator-registered instance keys, never inject an arbitrary ARN. The two blocking items:

1. **mssql silently serves a stale cached connection when the resolved ARN changes** (inline on server.py:486).
2. **No regression guard for the security invariant** — nothing asserts `connect_to_database` no longer exposes `secret_arn` to the LLM (inline on server.py:295).

Review assisted by Claude Code (pr-review-toolkit + code-review).

shreyaskshekhar:
**Title:** fix(mssql-mcp-server,oracle-mcp-server): Hardening secret manager access by making it a CLI parameter  
**Branch:** `feature/secretmanager`  
**State:** OPEN | Review Decision: CHANGES_REQUESTED  
**Reviewer:** scottschreckengaust  
**Date:** 2026-06-19  

---

## Summary

This PR removes `secret_arn` from the LLM-facing `connect_to_database` tool signature in both the mssql and oracle MCP servers, replacing it with an operator-controlled `--secret_arn` CLI parameter. The LLM can no longer inject an arbitrary Secrets Manager ARN — it can only target instances the operator has registered via CLI flags.

---

## Review Comments — Resolution Status

### Critical / Blocking

| # | Issue | File | Status |
|---|-------|------|--------|
| 1 | **Stale cached connection silently served after resolved ARN changes** — cache lookup returned early before ARN resolution, with no comparison against the cached connection's ARN. Oracle did this correctly; mssql did not. | `mssql/.../server.py` | **Resolved** |
| 2 | **No regression guard for the security invariant** — nothing in the test suite asserted that `connect_to_database` no longer exposes `secret_arn` to the LLM. | Both servers | **Resolved** |

#### Resolution Details

**Issue 1:** `internal_create_connection` now resolves the ARN *first* (lines 471–509), then checks the cache and evicts/replaces if the resolved ARN differs from the cached connection's ARN (lines 527–541). This mirrors Oracle's resolve → compare → evict/replace ordering.

```python
# Check for existing connection; evict if the resolved ARN differs.
replaced_conn = None
existing_conn = db_connection_map.get(...)
if existing_conn:
    if secret_arn and getattr(existing_conn, 'secret_arn', '') != secret_arn:
        logger.info(f'Replacing existing connection for {instance_identifier}/{database}: secret_arn changed')
        db_connection_map.remove(...)
        replaced_conn = existing_conn
    else:
        return (existing_conn, llm_response, None)
```

Test coverage: `test_internal_create_connection_evicts_on_secret_change` (`test_server_extended.py:557`)

**Issue 2:** Both servers now have a test that introspects the registered MCP tool's input schema and asserts `secret_arn` is absent:

- `src/mssql-mcp-server/tests/test_server_extended.py:42–50`
- `src/oracle-mcp-server/tests/test_server.py:49–57`

```python
def test_connect_to_database_tool_schema_does_not_expose_secret_arn():
    tool = server_mcp._tool_manager.get_tool('connect_to_database')
    assert tool is not None
    schema = tool.parameters
    assert 'secret_arn' not in schema.get('properties', {}), (
        'secret_arn must not be exposed as a tool parameter — '
        'secrets are configured exclusively via CLI --secret_arn flags'
    )
```

---

### Important

| # | Issue | File | Status |
|---|-------|------|--------|
| 3 | **Empty bare `--secret_arn` value silently skipped** — only malformed-input case that didn't hard-fail | Both `server.py` parsers | **Resolved** |
| 4 | **Oracle block comment inaccurate (comment rot)** — copied verbatim from mssql but doesn't reflect ORACLE_PASSWORD gating | `oracle/.../server.py:411–414` | **Resolved** |
| 5 | **mssql lacks RDS error handling Oracle has** — raw `ClientError` and bare `IndexError` on non-existent instance | `mssql/.../server.py:491–498` | **Resolved** |
| 6 | **Untested fall-through scenarios** — per-target wins over simultaneously-configured default (mssql); per-target miss falls to default (both) | Tests | **Resolved** |

#### Resolution Details

**Issue 3:** Both parsers now call `logger.error` + `sys.exit(2)` for empty bare values, matching the behavior of all sibling error cases:

```python
arn = raw.strip()
if not arn:
    logger.error(
        'Empty --secret_arn value. If using a shell variable, '
        'ensure it is set and non-empty.'
    )
    sys.exit(2)
```

**Issue 4:** Oracle comment rewritten to accurately reflect the ORACLE_PASSWORD-gated resolution:

```python
# For ORACLE_PASSWORD, resolve the secret ARN from operator config.
# Per-target override wins, then the bare default. If neither is set,
# the RDS MasterUserSecret metadata is used as a final fallback
# (resolved later via describe_db_instances).
# The LLM cannot influence this — only operator --secret_arn flags control it.
```

**Issue 5:** mssql now wraps `describe_db_instances` with proper error handling:

```python
try:
    response = rds_client.describe_db_instances(DBInstanceIdentifier=instance_identifier)
except ClientError as e:
    code = e.response['Error']['Code']
    if code == 'DBInstanceNotFound':
        raise ValueError(f"RDS instance '{instance_identifier}' not found in region '{region}'") from e
    raise ValueError(f'Failed to describe RDS instance: {e.response["Error"]["Message"]}') from e

instances = response.get('DBInstances', [])
if not instances:
    raise ValueError(f"describe_db_instances returned no instances for '{instance_identifier}'")
```

**Issue 6:** New tests added:

- `test_per_target_secret_wins_over_default` — configures both per-target and default, asserts per-target wins
- `test_per_target_miss_falls_to_default` — per-target map present but instance doesn't match any key, asserts fall-through to default (not to another instance's secret)

---

### Suggestions (Operator-Experience Polish)

| Issue | Status |
|-------|--------|
| **Misleading final error message** — said "no `--secret_arn` was provided" even when one was provided but didn't match | **Resolved** — reworded to describe the full resolver: "no per-target --secret_arn matched this instance, no bare default --secret_arn was configured, and the instance has no managed MasterUserSecret" with actionable CLI examples |
| **Oracle message references internal field** — told operators to "set a `configured_default_secret_arn`" | **Resolved** — now points to actual CLI surface: `--secret_arn <arn>` (bare default) or `--secret_arn <instance>=<arn>` (per-target) |
| **Asymmetric audit logging** — Oracle logged default but not per-target; mssql logged neither | **Partially resolved** — mssql now logs all three tiers (per-target, default, metadata). Oracle logs both per-target and default selections. |
| **Help-text precedence ambiguity** | **Resolved** — help text now explicitly states: "A bare default takes precedence over the instance MasterUserSecret for all unpinned instances. The RDS MasterUserSecret is only consulted when no --secret_arn is configured at all." |
| **Whitespace stripping untested** | Not explicitly addressed with dedicated tests |

---

## Test Results

| Server | Tests | Result |
|--------|-------|--------|
| mssql-mcp-server | 188 | All passing |
| oracle-mcp-server | 147 | All passing |

---

## Verdict

All **critical** and **important** review comments have been addressed in the latest revision. The two servers are now aligned in behavior for caching/eviction, RDS error handling, CLI parsing, and error messaging. The single remaining gap is a dedicated whitespace-stripping test (a minor suggestion), which does not block merge.

**Recommendation:** Ready for re-review and approval.

src/mssql-mcp-server/awslabs/mssql_mcp_server/server.py
shreyaskshekhar: *Issue 1:** `internal_create_connection` now resolves the ARN *first* (lines 471–509), then checks the cache and evicts/replaces if the resolved ARN differs from the cached connection's ARN (lines 527–541). This mirrors Oracle's resolve → compare → evict/replace ordering.

```python
# Check for existing connection; evict if the resolved ARN differs.
replaced_conn = None
existing_conn = db_connection_map.get(...)
if existing_conn:
    if secret_arn and getattr(existing_conn, 'secret_arn', '') != secret_arn:
        logger.info(f'Replacing existing connection for {instance_identifier}/{database}: secret_arn changed')
        db_connection_map.remove(...)
        replaced_conn = existing_conn
    else:
        return (existing_conn, llm_response, None)
```

Test coverage: `test_internal_create_connection_evicts_on_secret_change` (`test_server_extended.py:557`)

src/mssql-mcp-server/awslabs/mssql_mcp_server/server.py
shreyaskshekhar: **Issue 2:** Both servers now have a test that introspects the registered MCP tool's input schema and asserts `secret_arn` is absent:

- `src/mssql-mcp-server/tests/test_server_extended.py:42–50`
- `src/oracle-mcp-server/tests/test_server.py:49–57`

```python
def test_connect_to_database_tool_schema_does_not_expose_secret_arn():
    tool = server_mcp._tool_manager.get_tool('connect_to_database')
    assert tool is not None
    schema = tool.parameters
    assert 'secret_arn' not in schema.get('properties', {}), (
        'secret_arn must not be exposed as a tool parameter — '
        'secrets are configured exclusively via CLI --secret_arn flags'
    )
```

---

src/mssql-mcp-server/awslabs/mssql_mcp_server/server.py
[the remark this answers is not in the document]
shreyaskshekhar: Not a hardcoded password — '' is a "nothing configured" placeholder that gets replaced by the real ARN from operator config or RDS metadata.

src/mssql-mcp-server/awslabs/mssql_mcp_server/server.py
```diff
@@ -469,51 +468,95 @@ def internal_create_connection(
...
-            indent=2,
-            default=str,
-        )
-        return (existing_conn, llm_response)
+        if secret_arn and getattr(existing_conn, 'secret_arn', '') != secret_arn:
+            logger.info(
+                f'Replacing existing connection for {instance_identifier}/{database}: '
+                f'secret_arn changed'
+            )
+            db_connection_map.remove(
```
scottschreckengaust: **[High] Orphaned connection leak: eviction happens before a fallible network call.**

Here the cached connection is removed from the map and stashed in the local `replaced_conn`, but `validate_endpoint()` at line 558 then makes an RDS `describe_db_instances` call that can raise `ClientError`/`ValueError`. If it raises, the exception propagates out of `internal_create_connection` **before `replaced_conn` is returned**, so `connect_to_database`'s close path (lines 308-312) never runs. The old connection is now both removed from the map *and* never closed → silent socket/pool leak, which an LLM retry loop will multiply.

Oracle isn't exposed to this window (it only evicts when a config ARN is present, which skips the RDS call). Suggest evicting only after `validate_endpoint()` succeeds, or closing `replaced_conn` in an except/finally if the replacement is never built.

src/mssql-mcp-server/awslabs/mssql_mcp_server/server.py
```diff
@@ -469,51 +468,95 @@ def internal_create_connection(
     if not db_endpoint:
         raise ValueError("db_endpoint can't be none or empty")
 
+    # Resolve the Secrets Manager ARN first so we can detect stale cached
```
scottschreckengaust: **[Important] Behavioral divergence / regression vs. oracle: `describe_db_instances` now runs on every cached reconnect.**

mssql resolves the full ARN chain — including this RDS metadata `describe_db_instances` call (line 484) — *before* the cache check at line 529. So in the default MasterUserSecret mode (no operator `--secret_arn`), every `connect_to_database` to an already-cached DB now issues a live RDS API call that previously returned instantly from cache. This adds latency + API load per reconnect and lets a transient RDS/IAM/throttling error fail what used to be an offline cache hit.

Oracle deliberately defers its RDS lookup until *after* the cache check (`oracle server.py:456-494`). Please align mssql: resolve config ARNs pre-cache, but defer the RDS metadata lookup to post-cache.

src/mssql-mcp-server/awslabs/mssql_mcp_server/server.py
```diff
@@ -665,7 +708,25 @@ def main():
...
+            'value is either:\n'
+            '  - "<instance_identifier>=<arn>" — bind the ARN to a specific '
+            'RDS instance (matched against the instance_identifier parameter '
+            'from the connect_to_database tool).\n'
+            '  - "<arn>" without "=" — bare ARN used as the default for any '
+            'target the operator did not pin explicitly (at most one allowed).\n'
+            'Resolution order: per-target ARN > bare default ARN > RDS '
+            'MasterUserSecret metadata. A bare default takes precedence over '
+            'the instance MasterUserSecret for all unpinned instances. '
+            'The RDS MasterUserSecret is only consulted when no --secret_arn '
```
scottschreckengaust: **[Medium] Help text is factually wrong (same text in `oracle server.py:664`).**

"The RDS MasterUserSecret is only consulted when no --secret_arn is configured at all" contradicts both the code and the preceding sentence. If a per-target ARN is registered for instance A but no bare default is set, a connect to *unpinned* instance B still consults MasterUserSecret — even though `--secret_arn` *is* configured. Suggest: "...consulted only for a given instance when that instance has no per-target ARN and no bare default is configured."

src/mssql-mcp-server/awslabs/mssql_mcp_server/connection/pymssql_pool_connection.py
```diff
@@ -326,16 +326,17 @@ def _get_credentials_from_secret(
 
             if 'SecretString' in get_secret_value_response:
                 secret = json.loads(get_secret_value_response['SecretString'])
-                logger.debug(f'Secret keys: {", ".join(secret.keys())}')
                 username = secret.get('username') or secret.get('user') or secret.get('Username')
                 password = secret.get('password') or secret.get('Password')
                 if not username:
                     raise ValueError(
-                        f'Secret does not contain username. Available keys: {", ".join(secret.keys())}'
+                        'Secret does not contain a recognized username field. '
```
scottschreckengaust: **[Medium] Diagnostic regression (same change in `oracle .../oracledb_pool_connection.py:301`).**

Sanitizing the user-facing `ValueError` to drop secret key names is a good security choice. But this PR also removed the `logger.debug('Secret keys: ...')` line, so there is now **no path at any log level** for an operator to discover what keys a misconfigured secret actually contains (e.g. `USER_NAME`/`PWD`). Please restore a `logger.debug` listing key **names only** — field names aren't secret values — while keeping the sanitized exception message.

src/mssql-mcp-server/tests/test_server_extended.py
```diff
@@ -464,18 +532,153 @@ def test_internal_create_connection_missing_method_raises():
 
 def test_internal_create_connection_returns_existing(mocker):
```
scottschreckengaust, line 548: **[Test gap] Cache-hit-without-configured-secret is untested for mssql.**

This test was updated to set `configured_default_secret_arn = 'arn:test'`, so it takes the configured-default path and never touches RDS. There's no mssql test for: connection cached + no `--secret_arn` configured + second connect. Under the new pre-cache resolution that path now calls `describe_db_instances` on every reconnect (see the divergence comment on `server.py:471`) — the exact regression that currently has zero coverage. Also worth adding: a `key=` (valid key, empty ARN) CLI-parse test, and an autouse `server_config` reset fixture to match oracle's (several success-path CLI tests here mutate the global config without restoring it).

scottschreckengaust (changes requested):
## Review: Request changes

Thanks for this hardening work — the core security change is solid and well-tested: `secret_arn` is genuinely removed from the `connect_to_database` tool schema on both servers, resolution is driven entirely by operator `--secret_arn` config, CLI parsing rejects malformed input with `exit(2)`, and the new schema-invariant test guards against regression. Both test suites pass locally (mssql 286, oracle 245).

I'm requesting changes for three items before merge (details inline), plus two findings below that land on lines outside this diff:

### A. mssql README still advertises the removed behavior (must fix)
`src/mssql-mcp-server/README.md:68-70` still says:
> The `--secret_arn` flag can also be passed at runtime via the `connect_to_database` MCP tool's `secret_arn` parameter, allowing the LLM to switch credentials without restarting the server.

This documents the exact LLM-controllable credential switching that this PR removes for security — it now describes the opposite of what the code enforces. Please delete these lines. Also consider documenting the new per-target `<instance>=<arn>` syntax in both READMEs, since it currently exists only in `--help` text.

### B. Oracle tool handler exception breadth (`oracle server.py:340`, outside diff)
`connect_to_database` catches only `except (ValueError, ClientError)`. Non-`ClientError` boto failures (`NoCredentialsError`, `EndpointConnectionError`, `BotoCoreError`, ...) escape the `{"status": "Failed"}` envelope and propagate raw, whereas the mssql sibling uses `except Exception` and stays consistent. Consider broadening to `BotoCoreError`/`Exception` for parity.

### C. Test parity gap (oracle)
Oracle has `test_connect_to_database_closes_replaced_connection` but no equivalent of mssql's `test_connect_to_database_replaced_close_failure_is_non_fatal`. Oracle's identical try/except at `server.py:323-327` is untested — a refactor making the replaced-connection close failure fatal would go uncaught. Please mirror the mssql test.

### Lower priority (non-blocking)
- Oracle logs the full secret ARN (`server.py:425`, `:497`); mssql deliberately omits it — inconsistent with this PR's log-hygiene theme.
- Comments "The LLM cannot influence this" / "can only target an instance the operator has registered" slightly overstate the boundary — the LLM does choose `instance_identifier`; it just can't supply ARN *values*. Tightening the wording preserves the (correct) security claim.
- Empty-string `username`/`password` secret fields are reported as "not contained" due to the falsy check.

Full multi-agent review notes available on request.

src/mssql-mcp-server/awslabs/mssql_mcp_server/server.py
shreyaskshekhar: Fixed. Eviction is deferred until **after** every fallible network call (`describe_db_instances` and `validate_endpoint`) succeeds. We compute `replace_existing` up front (a pure comparison, no side effect) and only call `remove()` past the last raising call — using the pre-`validate_endpoint` cache key so it targets the same key `get()` used. A mid-resolution failure can no longer strand a removed-but-unclosed connection.

 ```python
     cache_endpoint, cache_port = db_endpoint, port
     db_endpoint, port = validate_endpoint(db_endpoint, port, instance_identifier, region)

     replaced_conn = None
     if replace_existing:                       # only after all raising calls succeed
         db_connection_map.remove(
             connection_method, instance_identifier, cache_endpoint, database, cache_port
         )
         replaced_conn = existing_conn
 ```

src/mssql-mcp-server/awslabs/mssql_mcp_server/server.py
shreyaskshekhar: Fixed, now aligned with Oracle. Only the operator-configured ARN is resolved before the cache check (pure in-memory dict lookup, no network); the RDS `MasterUserSecret` fallback is deferred to **after** the cache check. Guarded by a new test asserting `boto3.client` is never called on a cache hit.

 ```python
     per_target_secret_arn = server_config.configured_secret_arns.get(target_key, '')
     secret_arn = per_target_secret_arn or server_config.configured_default_secret_arn or ''

     existing_conn = db_connection_map.get(...)
     if existing_conn and not replace_existing:
         return (existing_conn, llm_response, None)   # cache hit - no RDS call

     if not secret_arn:                                # only on cache miss + no config ARN
         rds_client = boto3.client('rds', ...)
         response = rds_client.describe_db_instances(DBInstanceIdentifier=instance_identifier)
 ```

src/mssql-mcp-server/awslabs/mssql_mcp_server/server.py
shreyaskshekhar: Fixed in both servers' `--help` text and in both READMEs.

 ```python
 -            'The RDS MasterUserSecret is only consulted when no --secret_arn '
 -            'is configured at all. '
 +            'The RDS MasterUserSecret is consulted for a given instance only '
 +            'when that instance has no per-target ARN and no bare default is '
 +            'configured. '
 ```

src/mssql-mcp-server/awslabs/mssql_mcp_server/connection/pymssql_pool_connection.py
shreyaskshekhar: We're intentionally **not** restoring any log of the secret's keys — even names-only shouldn't be written to logs, per this PR's log-hygiene goal. The sanitized `ValueError` already names the *expected* fields, which is enough for an operator to diagnose a misconfigured secret without emitting anything derived from the secret payload. Nothing derived from the secret is logged at any level, in either connection module.

 ```python
         if not username:
             raise ValueError(
                 'Secret has no usable username value. Expected a non-empty '
                 'field named one of: username, user, Username.'
             )
 ```

src/mssql-mcp-server/tests/test_server_extended.py
shreyaskshekhar: All three addressed:

 **Cache-hit-without-configured-secret** — asserts no RDS call on reconnect:

 ```python
 def test_cache_hit_without_configured_secret_skips_rds(mocker):
     srv.server_config.configured_secret_arns = {}
     srv.server_config.configured_default_secret_arn = None
     cached = MagicMock()
     cached.secret_arn = 'arn:aws:secretsmanager:us-east-1:123:secret:rds-master'
     mocker.patch.object(db_connection_map, 'get', return_value=cached)
     mock_boto = mocker.patch('awslabs.mssql_mcp_server.server.boto3')
     conn, _, replaced = internal_create_connection(region='us-east-1', ...)
     assert conn is cached and replaced is None
     mock_boto.client.assert_not_called()
 ```

 **`key=` (valid key, empty ARN) parse:**

 ```python
 def test_main_secret_arn_empty_arn_in_key_pair_exits(mocker):
     mocker.patch.object(sys, 'argv', ['prog', '--region', 'us-east-1', '--secret_arn', 'inst1='])
     mocker.patch.object(srv.mcp, 'run')
     mocker.patch.object(srv.db_connection_map, 'close_all')
     with pytest.raises(SystemExit) as exc_info:
         srv.main()
     assert exc_info.value.code == 2
 ```

 **Autouse `server_config` reset fixture** (matches Oracle's pattern):

 ```python
 @pytest.fixture(autouse=True)
 def _reset_server_config():
     old_readonly = server_config.readonly_query
     old_arns = server_config.configured_secret_arns
     old_default = server_config.configured_default_secret_arn
     old_endpoints = server_config.allowed_endpoints
     yield
     server_config.readonly_query = old_readonly
     server_config.configured_secret_arns = old_arns
     server_config.configured_default_secret_arn = old_default
     server_config.allowed_endpoints = old_endpoints
 ```

shreyaskshekhar:
Thanks for the thorough pass — all three blocking items (A, B, C), both inline critical findings, and every lower-priority note are addressed. Both servers' resolution/handler paths are now aligned. Replies inline on each thread; summary:


> ### A. mssql README still advertises the removed behavior (must fix)

>  `src/mssql-mcp-server/README.md:68-70` still says:
>   The `--secret_arn` flag can also be passed at runtime via the `connect_to_database` MCP tool's `secret_arn` parameter, allowing the LLM to switch credentials without restarting the server.
> 
>  This documents the exact LLM-controllable credential switching that this PR removes for security — it now describes the opposite of what the code enforces. Please delete these lines. Also consider documenting the new per-target `<instance=<arn` syntax in both READMEs, since it currently exists only in `--help` text.


 Fixed. Removed the lines describing runtime `secret_arn` switching via the tool and replaced them with a CLI-only description + per-target syntax (added to **both** READMEs).

 ```markdown
 Secrets are configured exclusively at startup via `--secret_arn`; the flag is not
 exposed as an MCP tool parameter, so the LLM cannot select an arbitrary secret ARN.

 The `--secret_arn` flag may be repeated to bind different secrets to different instances:

 - `--secret_arn <instance_identifier=<arn` — bind an ARN to a specific RDS instance.
 - `--secret_arn <arn` — a bare ARN used as the default for any instance not pinned
   explicitly (at most one allowed).

 Resolution order: per-target ARN  bare default ARN  the instance's RDS-managed
 `MasterUserSecret`. The RDS `MasterUserSecret` is consulted for a given instance
 only when that instance has no per-target ARN and no bare default is configured.
 ```

---

> 
>  ### B. Oracle tool handler exception breadth (`oracle server.py:340`, outside diff)
>  `connect_to_database` catches only `except (ValueError, ClientError)`. Non-`ClientError` boto failures (`NoCredentialsError`, `EndpointConnectionError`, `BotoCoreError`, ...) escape the `{"status": "Failed"}` envelope and propagate raw, whereas the mssql sibling uses `except Exception` and stays consistent. Consider broadening to `BotoCoreError`/`Exception` for parity.


 Fixed — broadened to `except Exception` for parity with mssql, so non-`ClientError` boto failures stay inside the failure envelope.

 ```python
         return llm_response
 -    except (ValueError, ClientError) as e:
 +    except Exception as e:
         logger.exception(f'connect_to_database failed: {e}')
         return {'status': 'Failed', 'error': str(e)}
 ```

---

>  ### C. Test parity gap (oracle)
> 
>  Oracle has `test_connect_to_database_closes_replaced_connection` but no equivalent of mssql's `test_connect_to_database_replaced_close_failure_is_non_fatal`. Oracle's identical try/except at `server.py:323-327` is untested — a refactor making the replaced-connection close failure fatal would go uncaught. Please mirror the mssql test.

 Fixed — added the mirrored test locking the replaced-connection close-failure path on Oracle.

 ```python
 @pytest.mark.asyncio
 async def test_connect_to_database_replaced_close_failure_is_non_fatal(mocker):
     replaced_conn = AsyncMock()
     replaced_conn.close = AsyncMock(side_effect=RuntimeError('close failed'))
     mock_pool_conn = MagicMock(spec=OracledbPoolConnection)
     mock_pool_conn.initialize_pool = AsyncMock()
     mocker.patch(
         'awslabs.oracle_mcp_server.server.internal_create_connection',
         return_value=(mock_pool_conn, {'status': 'ok'}, replaced_conn),
     )
     result = await connect_to_database(
         region='us-east-1', connection_method=ConnectionMethod.ORACLE_PASSWORD,
         db_endpoint='ep1', service_name='ORCL',
     )
     assert result == {'status': 'ok'}
     replaced_conn.close.assert_awaited_once()
 ```

---

>  - Oracle logs the full secret ARN (`server.py:425`, `:497`); mssql deliberately omits it — inconsistent with this PR's log-hygiene theme.


 Fixed — no ARN value is written to logs in either server now.

 ```python
 -            logger.info(f'Using default secret_arn from startup configuration: {secret_arn}')
 +            logger.info('Using default secret_arn from startup configuration')

 -        f'Instance props: masteruser:{masteruser}, secret_arn:{secret_arn}, '
 +        f'Instance props: masteruser:{masteruser}, secret_arn_resolved:{bool(secret_arn)}, '
 ```

---

>  - Comments "The LLM cannot influence this" / "can only target an instance the operator has registered" slightly overstate the boundary — the LLM does choose `instance_identifier`; it just can't supply ARN *values*. Tightening the wording preserves the (correct) security claim.

 Fixed — reworded the resolution comments and both `--help` strings.

 ```python
 -            'The LLM cannot pick a secret ARN — it can only target an instance '
 -            'the operator has registered here.'
 +            'The LLM cannot supply a secret ARN value — it only chooses which '
 +            'instance to target; ARNs come exclusively from the operator here.'
 ```

---

> Empty-string `username`/`password` secret fields are reported as "not contained" due to the falsy check.

 Fixed — reworded both connection modules so the message is accurate for both the missing-field and present-but-empty cases.

 ```python
         if not username:
             raise ValueError(
 -                'Secret does not contain a recognized username field. '
 -                'Expected one of: username, user, Username.'
 +                'Secret has no usable username value. Expected a non-empty '
 +                'field named one of: username, user, Username.'
             )
 ```

scottschreckengaust (approved):
## Review: Approve ✅

Re-reviewed against head `b041ea95` with both tools (code-review + pr-review-toolkit), independently verifying each previously-open item rather than trusting the resolution comments. All three blocking items from my 2026-07-02 review — and the earlier critical/high findings — are genuinely fixed. The core hardening goal holds: I confirmed against the **live tool schemas** that `secret_arn` is absent from `connect_to_database` on both servers (mssql exposes `connection_method, database, db_endpoint, instance_identifier, port, region`; oracle adds `service_name, sid`), and a schema-invariant test now guards it. Test suites pass locally (mssql 204, oracle 190).

### Previously-blocking items — all resolved
- **A — mssql README misleading text:** the "LLM can switch credentials via the tool's `secret_arn` parameter" lines are gone (`README.md:68-69` now: *"the flag is not exposed as an MCP tool parameter, so the LLM cannot select an arbitrary secret ARN"*). Bonus: the per-target `<instance>=<arn>` syntax is now documented in **both** READMEs (mssql `:71-82`, oracle `:77,83-85`).
- **B — oracle exception breadth:** `connect_to_database` now uses `except Exception as e:` (`oracle server.py:340`), matching mssql (`:325`); non-`ClientError` boto failures no longer escape the `{"status":"Failed"}` envelope.
- **C — oracle test parity:** `test_connect_to_database_replaced_close_failure_is_non_fatal` added (`oracle tests/test_server.py:1207`), mirroring mssql and asserting the replaced-connection close failure is swallowed.

### Earlier critical/high items — verified fixed
- Stale-cache correctness: resolve config ARN (in-memory, no network) → compare against `existing_conn.secret_arn` → defer eviction until after all fallible calls. I verified the precondition holds — `secret_arn` is genuinely persisted on both connection objects (`PymssqlPoolConnection.__init__` → `self.secret_arn`, `OracledbPoolConnection.__init__` → `self.secret_arn`), so the change-detection isn't a silent no-op.
- mssql cache-key integrity: `cache_endpoint/cache_port` captured pre-`validate_endpoint` and reused in `remove()`, so eviction targets the same key `get()` used.
- Cache-hit reuse stays offline (dedicated test asserts `boto3.client` is not called on reconnect).
- Full-ARN logging removed on both servers (now `secret_arn_resolved:{bool(secret_arn)}`).

### Lower-priority follow-ups (non-blocking, resolved)
- Oracle no longer logs the full ARN; comment wording tightened to acknowledge the LLM chooses the instance key but not ARN values (both servers).

---

### Non-blocking follow-ups (fine to address in a later PR — not gating)

1. **Oracle eviction is not deferred, unlike mssql (`oracle server.py:441-444`).** On an ARN-change reconnect, oracle calls `db_connection_map.remove(...)` *before* `describe_db_instances` (`:462-475`), which can raise. If it does, the old connection is already evicted from the map but never returned as `replaced_conn`, so its pool is never closed — the exact orphaned-connection class this PR hardened mssql against (mssql defers eviction to after all network calls, `server.py:553-564`). Hard to hit today (only the non-`ORACLE_PASSWORD`+secret path reaches `describe_db_instances`, and service_name/sid are pre-validated), but aligning oracle with mssql's deferred-eviction ordering would close it.
2. **A bare default ARN containing `=` would be misparsed as `<key>=<arn>`** (`mssql server.py:755`, `oracle server.py:710` use `'=' in raw` / `partition('=')`). Secrets Manager secret names allow `=`, so `...:secret:my=secret` as a bare default would be split into a bogus per-target entry. Unusual naming; the help text documents the no-`=` convention, so low severity.
3. **Empty-string `username`/`password` secret fields** are still treated as missing via the falsy `if not username:` check (both pool connection files). The error text was reworded to frame this as "expected a non-empty field," which reads as intentional — flagging only for awareness.

CI note: the one red check (`Build aurora-dsql-mcp-server`) is unrelated — this PR touches no aurora files, and `Build src/aurora-dsql-mcp-server` + its precommit pass; it's a flaky duplicate matrix job.

Nice work on the hardening and the thorough test coverage. Approving.

Review assisted by Claude Code (code-review + pr-review-toolkit).

krokoko (approved):

PR #3813, opened 2026-06-12
feat(cloudwatch-applicationsignals): add ServiceEvents telemetry tools

mxiamxia:
### Changes

Adds a **ServiceEvents** data layer to the CloudWatch Application Signals MCP server — a set of tools for querying function/endpoint telemetry, incidents, deployments, and overall service health, backed by CloudWatch Logs and a PromQL query client. Also includes a small region-resolution fix so all clients target a consistent AWS region.

**New tools (`service_events/`):**
- `list_monitored_functions` — list functions with Application Signals telemetry
- `get_function_metrics` — per-function metric details (filterable by endpoint/operation)
- `search_functions_by_name` — find functions by name
- `get_endpoint_performance` — endpoint/operation performance summary
- `get_recent_incidents` — recent incidents for a service
- `get_incident_root_cause` — incident detail with root-cause analysis (falls back to an Application Signals trace audit when no ServiceEvents incident is found)
- `get_service_health_overview` — consolidated service health
- `find_deployment` — locate deployment events

**Supporting changes:**
- New `service_events/` package: `tools.py`, `cw_logs.py`, `promql_client.py`, `promql_models.py`, `promql_query.py`, `formatting.py`, `state.py`, `registration.py`, and `endpoint_metrics.py`.
- `server.py` wires `register_service_events_tools(mcp)` and warms a service→environment cache at startup.
- Region resolution centralized in `aws_clients.AWS_REGION` (`AWS_REGION`/`AWS_DEFAULT_REGION` env → configured profile region → `us-east-1`), replacing hardcoded `us-east-1` defaults in `canary_utils`, `sli_report_client`, and `trace_tools`.
- Tests: new `test_cw_logs.py`, `test_function_metrics.py`, `test_promql_client.py`, `test_server_tools.py`, `test_trace_tools.py`, plus ServiceEvents OTLP fixtures.

> Note: the dynamic-instrumentation tools from #3801 are already on `main`; this branch is rebased on top of them, so they do not appear in this diff.

### User experience

Before: the server exposed audit/SLO/trace/RUM tooling but had no direct way to query per-function or per-endpoint telemetry, recent incidents, or deployment events.

After: assistants can ask things like "which functions are unhealthy", "show recent incidents for service X and the root cause", or "find the deployment that preceded this regression", and the server answers from ServiceEvents telemetry — with an automatic fallback to an Application Signals trace audit when ServiceEvents has no matching incident.

## Checklist

* [x] Changes are documented

**RFC issue number**:

## Acknowledgment


src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/service_events/promql_query.py
```diff
@@ -0,0 +1,254 @@
...
+        return DEFAULT_RATE_WINDOW
+    return f'{hours}h'
+
+
+def _value_of(entry: dict) -> Optional[float]:
+    value = entry.get('value')
+    if not value or len(value) < 2:
+        return None
+    try:
+        return float(value[1])
```
vastin: `_value_of` only catches `ValueError`/`TypeError`, but `float("NaN")` and `float("+Inf")` succeed without raising — and Prometheus legitimately returns `"NaN"` (e.g. `histogram_avg` over an empty window). That value flows into `int(round(...))` in `function_metrics.py:88,90`, which then throws `ValueError: cannot convert float NaN to integer` / `OverflowError` uncaught — taking down `list_monitored_functions`, `get_function_metrics`, and `get_service_health_overview`. The test only covers `'nope'`, so this case isn't exercised.

```python
import math
val = float(value[1])
if not math.isfinite(val):
    return None
return val
```

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/service_events/registration.py
```diff
@@ -0,0 +1,38 @@
...
+    get_incident_details,
+    get_incidents,
+    list_functions,
+    search_functions,
+)
+
+
+def register_tools(mcp) -> None:
+    """Register all ServiceEvents data MCP tools onto a shared server."""
+    mcp.tool(name='list_monitored_functions')(list_functions)
```
vastin, line 31: Registering under a different name (`mcp.tool(name='get_recent_incidents')(get_incidents)`) is fine. The issue is that the guidance text the LLM reads still uses the *internal function* names, which aren't valid tool names:

| Guidance says | Registered as |
|---|---|
| `get_incident_details(...)` | `get_incident_root_cause` |
| `get_incidents(...)` | `get_recent_incidents` |

The worst cases are the runtime `next_step` strings in `tools.py:1122` and `1126`, which literally instruct the model to call `get_incidents(...)` — not a registered name. The docstring cross-references in `tools.py` (lines 460, 474, 476, 485, 493, 509) and `trace_tools.py` (735, 747, 752, 754, 761) have the same problem.

Not a crash and often recoverable (a capable model maps intent to the nearest available tool), but it can cause failed calls and adds friction. Fix by using the registered names in all guidance text. Note `get_endpoint_performance` (tools.py:68) and `find_deployment` already use the correct names — leave those.

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/service_events/promql_query.py
mxiamxia: Good catch — confirmed and fixed in 6d2170d4. `_value_of` now guards with `math.isfinite` and returns `None` for NaN/±Inf, so they're treated like unparseable values (every caller already skips `None`). This fixes it at the single chokepoint, covering `list_monitored_functions`, `get_function_metrics`, and `get_service_health_overview` at once. Added `_value_of` cases for NaN/+Inf/Inf/-Inf plus an end-to-end `fetch_function_records` test that feeds a "NaN" value and asserts no crash.

src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/service_events/registration.py
mxiamxia: Fixed in 6d2170d4. Replaced the internal names with the registered tool names in all guidance text: `get_incidents(...)` → `get_recent_incidents(...)` and `get_incident_details(...)` → `get_incident_root_cause(...)`, including the runtime `next_step` strings (tools.py:1122/1126) and the docstring cross-references in tools.py and trace_tools.py. The Python function names and internal call sites are unchanged. Left `get_endpoint_performance` and `find_deployment` as-is since they already match.

vastin (dismissed):

wangzlei (dismissed):

vastin (approved):

wangzlei (approved):

PR #3821, opened 2026-06-13
fix(bedrock-kb-retrieval): preserve non-ASCII text in tool results

### LINKED ISSUE #3820: bedrock-kb-retrieval: QueryKnowledgeBases/ListKnowledgeBases return non-ASCII (CJK, emoji) as \uXXXX escapes
Thanks for maintaining the AWS MCP servers — the Bedrock Knowledge Base retrieval server is very useful.

## Problem

When a Knowledge Base contains non-ASCII text (Japanese/Korean/Chinese, accented Latin, emoji), the `QueryKnowledgeBases` and `ListKnowledgeBases` tools return it as `\uXXXX` escape sequences rather than readable characters. This (a) inflates token usage (each CJK character becomes 6 ASCII chars in the model-visible payload) and (b) makes the raw tool output unreadable.

## Cause

Results are serialized with `json.dumps(...)`, which defaults to `ensure_ascii=True`:
- `src/bedrock-kb-retrieval-mcp-server/awslabs/bedrock_kb_retrieval_mcp_server/knowledgebases/retrieval.py:113`
- `src/bedrock-kb-retrieval-mcp-server/awslabs/bedrock_kb_retrieval_mcp_server/server.py:128`

## Reproduction (no AWS needed)

```python
import json
print(json.dumps({"text": "東京は日本の首都です。"}))
# {"text": "東京は日本の首都です。"}
```

## Proposed fix

Pass `ensure_ascii=False` at those two serialization points (the transport is already UTF-8). Happy to open a PR with regression tests.

kimnamu:
Thank you for maintaining the AWS MCP servers.

Fixes #3820

## Summary

`QueryKnowledgeBases` and `ListKnowledgeBases` serialized results with `json.dumps(...)` (default `ensure_ascii=True`), so non-ASCII KB content (CJK, accented, emoji) was returned as `\uXXXX` escapes — token bloat + unreadable payloads. This adds `ensure_ascii=False` to the two user/model-visible serialization points (the MCP transport is already UTF-8).

## Changes

- `knowledgebases/retrieval.py:113`, `server.py:128` — add `ensure_ascii=False`
- regression tests in `test_retrieval.py` + `test_server.py`
- `## [Unreleased]` / `### Fixed` CHANGELOG entry

## Before / After

| Item | Before | After |
|------|--------|-------|
| KB text `東京は日本の首都です。` in tool output | ❌ `東京...` (escaped) | ✅ `東京は日本の首都です。` |
| KB name `한국어 데이터 소스` | ❌ `한국...` | ✅ `한국어 데이터 소스` |
| Output is valid JSON / round-trips via `json.loads` | ✅ | ✅ (unchanged) |
| Public API / tool signatures / result ordering | — | ✅ Unchanged |
| ASCII-only content | ✅ identical bytes | ✅ identical bytes |

## Tests

`40 passed`, `ruff check` clean. Reverting the source change (keeping the tests) makes the 2 new tests FAIL with `assert '東京...' in '...\\u6771\\u4eac...'`, proving they catch the bug; restoring it → all pass.

---

## Acknowledgment

### DIFF
--- a/src/bedrock-kb-retrieval-mcp-server/CHANGELOG.md
+++ b/src/bedrock-kb-retrieval-mcp-server/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- Preserve non-ASCII (e.g. CJK, accented, emoji) text in `QueryKnowledgeBases` and `ListKnowledgeBases` tool results by serializing with `json.dumps(..., ensure_ascii=False)`, so Knowledge Base content reaches the model and user as readable characters instead of `\uXXXX` escape sequences.
+
 ## [1.0.0] - 2025-05-26
 
 ### Removed
--- a/src/bedrock-kb-retrieval-mcp-server/awslabs/bedrock_kb_retrieval_mcp_server/knowledgebases/retrieval.py
+++ b/src/bedrock-kb-retrieval-mcp-server/awslabs/bedrock_kb_retrieval_mcp_server/knowledgebases/retrieval.py
@@ -110,4 +110,4 @@ async def query_knowledge_base(
                 }
             )
 
-    return '\n\n'.join([json.dumps(document) for document in documents])
+    return '\n\n'.join([json.dumps(document, ensure_ascii=False) for document in documents])
--- a/src/bedrock-kb-retrieval-mcp-server/awslabs/bedrock_kb_retrieval_mcp_server/server.py
+++ b/src/bedrock-kb-retrieval-mcp-server/awslabs/bedrock_kb_retrieval_mcp_server/server.py
@@ -125,7 +125,7 @@ async def list_knowledge_bases_tool() -> str:
     3. Use the names to determine which knowledge base and data source(s) are most relevant to the user's query
     """
     knowledge_bases = await discover_knowledge_bases(kb_agent_mgmt_client, kb_inclusion_tag_key)
-    return json.dumps(knowledge_bases)
+    return json.dumps(knowledge_bases, ensure_ascii=False)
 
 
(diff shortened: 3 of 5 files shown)


kimnamu:
Status update (no action needed — just keeping this current):

- Re-verified today against `main`: both call sites still serialize with default `ensure_ascii=True` (`knowledgebases/retrieval.py:113`, `server.py:128`), so the issue in #3820 still reproduces and this fix is still applicable.
- Branch merges cleanly with current `main` (no conflicts) and the full server test suite passes locally: `40 passed`, `ruff check` clean.

Happy to rebase or adjust anything (e.g. drop the CHANGELOG entry) if preferred — thanks for your time reviewing!

kimnamu:
Re-verified after the stale bump: rebased locally onto current `main` (clean, no conflicts) and ran the full suite for `bedrock-kb-retrieval-mcp-server` — `40 passed`, `ruff check` clean. Both call sites (`retrieval.py:113`, `server.py:128`) still default to `ensure_ascii=True`, so the fix is still needed and unchanged from the original PR.

Still happy to adjust anything (e.g. drop the CHANGELOG entry) if that unblocks a review — thanks for keeping this open.

kimnamu:
Still relevant as of today's `main` (`ebcafaed`):

- Both call sites still use the default `ensure_ascii=True` (`knowledgebases/retrieval.py:113`, `server.py:128`), so #3820 reproduces unchanged.
- This branch merges cleanly into current `main`, and `uv run --frozen pytest` in `src/bedrock-kb-retrieval-mcp-server` gives `40 passed` with `ruff check` / `ruff format --check` clean.

Two things that may help a decision:

1. The same one-line change was merged for a sibling server in #3995 (`browser_evaluate`, `ensure_ascii=False`), so the approach already has precedent in this repo.
2. #4372 proposes the same change to the same two lines. No need to review both — if that one is easier to take, I'll close this and #3820 can point there instead.

If this is only waiting on review bandwidth, the `backlog` label would keep the stale bot off it. Thanks for your time.

kimnamu:
Hi @laithalsaadoon, 
May I request review this PR? Thank you for your time :)

PR #3968, opened 2026-06-17
feat(aws-documentation-mcp-server): surface optional response and per-result metadata in search_documentation

zjerath:
## Summary

This change enriches search results with optional metadata so agents can navigate documentation more effectively.

### Changes

* **Per-result metadata:** each `SearchResult` may now carry an optional `metadata.additional_urls` list (`{url, section_title, section_anchor}`) pointing to other documentation pages related to that result, so an agent can navigate directly to them.
* **Response-level metadata:** `SearchResponse` may now carry an optional `metadata` object with query-level context - `discovered_services` (relevant AWS services beyond the top results), `related_tasks` (follow-up operations, each with its own doc URLs), and `relationships` (named connections between the returned topics).
* **Docstring:** `search_documentation` now documents these fields and how to use them
* **Tests:** added unit tests for the new metadata projection and for the no-metadata case (fields remain `null`).

### User experience

**Before:**

* User asks: *"How do I encrypt an S3 bucket and back it up with the right permissions?"*
* Agent searches for `"S3 encryption"`
* Agent gets a flat list of pages with no indication of which related services, tasks, or pages matter.

**After:**

* User asks: *"How do I encrypt an S3 bucket and back it up with the right permissions?"*
* Agent searches for `"S3 encryption"`
* Each result can carry `metadata.additional_urls` pointing to closely related pages, and the response can include top-level `metadata`:
  * `related_tasks` - follow-up operations (e.g. *"suspend versioning"*) each with their own doc URLs
  * `discovered_services` - relevant services beyond the top hits (e.g. AWS Backup) the agent might otherwise miss
  * `relationships` - how the returned topics connect (e.g. *Object Lock relates to Versioning*)

## Checklist

* [X] Changes are documented

Is this a breaking change? N

**RFC issue number**: N/A

## Acknowledgment

### DIFF
--- a/src/aws-documentation-mcp-server/awslabs/aws_documentation_mcp_server/server_aws.py
+++ b/src/aws-documentation-mcp-server/awslabs/aws_documentation_mcp_server/server_aws.py
@@ -40,8 +40,8 @@ from awslabs.aws_documentation_mcp_server.util import (
 )
 from loguru import logger
 from mcp.server.fastmcp import Context, FastMCP
-from pydantic import Field
-from typing import List, Optional
+from pydantic import BaseModel, Field, ValidationError
+from typing import Any, Dict, List, Optional, Type, TypeVar
 
 
 SEARCH_API_URL = 'https://proxy.search.docs.aws.com/search'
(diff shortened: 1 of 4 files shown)


AadityaBhoota (dismissed):

src/aws-documentation-mcp-server/awslabs/aws_documentation_mcp_server/server_aws.py
```diff
@@ -452,20 +469,32 @@ async def search_documentation(
                         f'Found {len(sections)} sections for {title}: {url}, sections: {sections}'
                     )
 
+                filtered_result_metadata = {
+                    k: metadata[k] for k in SUPPORTED_RESULT_METADATA_KEYS if metadata.get(k)
+                }
+                search_result_metadata = (
+                    SearchResultMetadata.model_validate(filtered_result_metadata)
```
JonLim: Same comment for here + line 404 in the same file!

`model_validate` is unguarded, and if any exceptions are thrown, like say `ValidationError` or `AttributeError`, would be uncaught and cause a tool exception, even in the case of a successful search call. Would be good to catch all the potential errors here (my count says `ValidationError, TypeError, KeyError, AttributeError`) and log the error but also drop metadata in the event of an error.

JonLim (dismissed):

src/aws-documentation-mcp-server/awslabs/aws_documentation_mcp_server/server_aws.py
zjerath: Agreed, thanks will address!

JonLim (approved):

AadityaBhoota (approved):

PR #3977, opened 2026-06-17
feat(cloudwatch-applicationsignals): support error patterns + before/after comparison 

mxiamxia:
## Summary

Adds the per-(operation × exception) error breakdown — the data behind the CloudWatch **Errors** page — to **`get_service_health_overview`**, plus a before/after comparison so "did the error pattern change?" is answered in one call.

### Highlights

- **`error_patterns`**: per-(operation, exception) error counts from the `count` MetricV2 metric (PromQL `sum_over_time`), sorted by count. Optional — omitted when the metric is unavailable, never fails the overview.
- **`error_pattern_comparison`**: before/after counts across two windows, anchored on a recent deployment (±3h) when one exists, else last day vs previous day. Each row has prior/recent counts, delta, % change, and status (up/down/flat/new/cleared).
- **`trend_basis`** caption names both comparison windows with absolute UTC ranges, so a trend is never shown without stating what it's measured against.
- **`baseline_unavailable`** guard: when the prior window aged out (the `count` metric is short-retention), flags it so rows aren't misreported as brand-new.

Verified live (`telemend` / `us-east-1`, `customers-service-java`) and matches the CloudWatch Errors page. No IAM change — reuses the existing `monitoring` SigV4 PromQL path.

## Checklist

* [x] Changes have been tested (unit tests + live verification)
* [x] Changes are documented (tool docstring)

Is this a breaking change? (N) — additive optional fields.

## Acknowledgment

### DIFF
--- a/src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/service_events/tools.py
+++ b/src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/service_events/tools.py
@@ -370,25 +370,17 @@ async def get_endpoints(
     **Output is limited to avoid large responses.** Use `operation` parameter
     to filter to a specific API if you know the operation name.
 
-    **Error patterns (optional):** When a `service_name` is given and the service
-    emits the `count` MetricV2 metric (only when ServiceEvents is enabled in the SDK),
-    the response also includes an `error_patterns` list — per-(operation, exception)
-    error counts over the window, matching the CloudWatch "Errors" page. This is the
-    right field for "what error patterns does my service have / changed?". It is scoped
-    by `operation` when provided, otherwise service-wide. The field is OMITTED when the
-    metric is unavailable or there are no errors — its absence never fails the call.
-    (For a service-wide error-pattern overview, `get_service_health_overview` also
-    includes this breakdown.)
+    **For service-wide error patterns** (which exceptions on which operations, with
+    counts — the CloudWatch "Errors" page data), use `get_service_health_overview`,
+    which returns an `error_patterns` breakdown.
 
     Args:
         hours: Time range to query in hours (default 24).
         operation: Filter by operation name (e.g., "POST /api/users"). Substring match. Optional.
-            When exactly one endpoint matches, detail mode is returned. Also scopes
-            `error_patterns` to that operation.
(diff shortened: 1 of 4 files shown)


src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/service_events/tools.py
```diff
@@ -306,6 +306,291 @@ def search_functions(
...
+    """Compare per-(operation, exception) error counts across two windows.
+
+    Cut line: a recent deployment (fixed ±3h around it) when one is available within
+    the last day; otherwise the last 24h vs the previous 24h. Returns a structured
+    comparison (windows, merged rows, totals) for the model to render, or ``None`` when
+    either window's query fails. Best-effort: never raises into the overview.
+    """
+    now = datetime.now(timezone.utc)
+
+    use_deploy = deployment_dt is not None and (now - deployment_dt) <= timedelta(
```
vastin, line 512: `deployment_dt` ultimately comes from the raw SDK attribute `aws.service_events.deployment.timestamp` with **no normalization** (`cw_logs.py:471`). `_latest_deployment_dt` parses it with `datetime.fromisoformat(ts.replace('Z', '+00:00'))` — if the SDK ever emits a timestamp without a `Z`/offset (e.g. `2026-06-17T13:12:00`), `fromisoformat` **succeeds** and returns a *naive* datetime. Its `try/except` only guards the parse, so the naive value is returned, and `now - deployment_dt` then raises:

> `TypeError: can't subtract offset-naive and offset-aware datetimes`

The `await` at `tools.py:1464` has no `try/except`, and there's no outer guard in `get_health_overview` — so the docstring's *"Best-effort: never raises into the overview"* is violated and the primary health tool crashes, precisely when a deployment is recent.

Note `find_deployment` (`tools.py:1642`) already wraps the **identical** subtraction in `try/except (ValueError, TypeError)`; this path copied the parse but dropped the guard.

**Fix** — coerce to tz-aware UTC at the parse site (and ideally extract a shared helper used by both paths):

```python
dt = datetime.fromisoformat(ts.replace('Z', '+00:00'))
if dt.tzinfo is None:
    dt = dt.replace(tzinfo=timezone.utc)
```

vastin (approved):

wangzlei (approved):

PR #3981, opened 2026-06-17
feat(timestream-for-influxdb-mcp-server): Add the Timestream for InfluxDB Kiro Skill

trevorbonas:
Fixes

### Changes

The Timestream for InfluxDB Kiro Skill has been added. This Skill helps agents and users get started with Timestream for InfluxDB v2 and v3, including querying, schema design, ingestion, migrations, troubleshooting, and database management.

### User experience

Now, agents will have more accurate information about Timestream for InfluxDB.

## Checklist

* [X] Changes are documented

N

**RFC issue number**:

Checklist:

## Acknowledgment


src/timestream-for-influxdb-mcp-server/kiro_power/mcp.json
```diff
@@ -0,0 +1,26 @@
...
+  "mcpServers": {
+    "awslabs.timestream-for-influxdb-mcp-server": {
+      "command": "uvx",
+      "args": ["awslabs.timestream-for-influxdb-mcp-server@latest"],
+      "env": {
+        "AWS_PROFILE": "your-aws-profile",
+        "AWS_REGION": "us-east-1",
+        "INFLUXDB_URL": "https://your-influxdb-2-endpoint:8086",
+        "INFLUXDB_TOKEN": "your-influxdb-2-token",
+        "INFLUXDB_ORG": "your-influxdb-org",
```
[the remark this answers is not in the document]
trevorbonas: I've removed this example file.

src/timestream-for-influxdb-mcp-server/kiro_power/mcp.json
```diff
@@ -0,0 +1,26 @@
...
+      },
+      "autoApprove": []
+    },
+    "influxdb3": {
+      "command": "node",
+      "args": ["/path/to/influx-mcp-standalone/build/index.js"],
+      "env": {
+        "INFLUX_DB_INSTANCE_URL": "https://your-influxdb-3-endpoint:8181",
+        "INFLUX_DB_TOKEN": "your-influxdb-3-token",
+        "INFLUX_DB_PRODUCT_TYPE": "core"
```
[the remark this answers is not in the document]
trevorbonas: I've removed this example file.

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb2/gotchas.md
```diff
@@ -0,0 +1,57 @@
+# Common Gotchas & Pain Points for Timestream for InfluxDB v2
+
+## Critical — Will Cause Outages If Ignored
```
mhchengaws, line 3: Would be helpful to also suggest monitoring resource utilization thresholds and setting up critical alarms to scale up compute if needed — for reference the v2 monitoring guide https://docs.aws.amazon.com/timestream/latest/developerguide/timestream-for-influx-monitoring-configuration-optimization.html

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb2/gotchas.md
```diff
@@ -0,0 +1,57 @@
...
+
```
mhchengaws, line 4: Worth adding a gotcha which instances are patched during the default or specified maintenance window and could experience downtime while patching. Recommend setting the maintenance window to a low-traffic period.

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb3/gotchas.md
```diff
@@ -0,0 +1,45 @@
+# Common Gotchas & Pain Points for Timestream for InfluxDB v3
+
+## Critical — Will Cause Outages If Ignored
+
+**Small Batch Writes + Many Writers**: Writing 1 record per request with hundreds of concurrent writers causes replica lag climbing to 200+ seconds. Always batch 5,000+ points per write request.
```
mhchengaws: Lets update here and the troubleshooting section in V3 that replica lag is only for V2 ReadReplica cluster

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb2/gotchas.md
trevorbonas: I've added a note here that links to a new "Monitoring" section in `influxdb2/troubleshooting.md`.

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb2/gotchas.md
trevorbonas: Added.

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb3/gotchas.md
trevorbonas: I've deleted this point from `influxdb3/gotchas.md` and noted, in `influxdb3/troubleshooting.md`, that read-replica lag is only relevant for v2.

mhchengaws (commented):
LGTM

mhchengaws (dismissed):

lokendrp-aws (dismissed):

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb-2-vs-3.md
```diff
@@ -0,0 +1,131 @@
...
+- **V3:** **SQL** is the primary language (via DataFusion); a v1-compatible **InfluxQL** endpoint is provided for legacy apps. **Flux is not supported.**
+
+There is **no automatic Flux-to-SQL conversion**. Migrating from V2 to V3 requires rewriting all Flux queries, tasks, and dashboards into SQL or InfluxQL.
+
+## Data Model & API
+
+- **V2** organizes data as **Organization → Bucket**. Writes specify `org` and `bucket`; default port **8086**.
+- **V3** organizes data as **Database → Table** (no organizations). Tables are created automatically on first write (a measurement becomes a table). Writes specify `db`; data-plane auth uses the `Bearer` prefix; default port **8181**.
+- **`422` HTTP status code meaning** - For v2, a status code of `422` means that some or all of the data was rejected. For v3, `422` means that writing the line protocol points would lead to the maximum number of databases, tables, columns, tags, or fields being exceeded.
+- **`400` HTTP status code meaning** - For v2, a status code of `400` means that the `org` or `orgID` parameter doesn't match an existing organization. For v3, a status code of `400` means that some or all of the data was rejected.
```
scottschreckengaust, line 75: **[fix before merge] Incorrect — and contradicted by three files in this same PR.** For a v2 write, `400` means **malformed line protocol** (whole request rejected); a missing org returns **`404`**, not `400`. This file's claim that `400` means "the `org`/`orgID` doesn't match an existing organization" is wrong and is contradicted by:
- `influxdb2/troubleshooting.md:32` (`400` = malformed line protocol) and `:37` (`404` = org/bucket not found)
- `influxdb2/ingestion.md:13` and `:144`
- `influxdb2/onboarding.md:321`

An agent debugging a failed write would follow this and start renaming/recreating orgs instead of fixing the line protocol. Suggest: `400` = some/all data rejected (malformed line protocol); `404` = org/bucket not found.

src/timestream-for-influxdb-mcp-server/kiro_power/POWER.md
```diff
@@ -0,0 +1,135 @@
...
+#### InfluxDB 3 MCP
+
+Prerequisites:
+- Set up the InfluxDB 3 MCP Server. [See here](https://github.com/influxdata/influxdb3_mcp_server?tab=readme-ov-file#2-integration-with-mcp-clients) for instructions.
+- An existing InfluxDB 3 instance (or cluster). The following details are required:
+    - `INFLUX_DB_INSTANCE_URL`: "https://your-influxdb-3-endpoint:8181"
+    - `INFLUX_DB_TOKEN`: "your-influxdb-3-token"
+    - `INFLUX_DB_PRODUCT_TYPE`: "core"
+
+> See [mcp.json](./mcp.json) for an example configuration.
```
scottschreckengaust: **[fix before merge] Dead reference.** This links `./mcp.json`, and `README.md:160` also states "The Kiro Power's `kiro_power/mcp.json` references both servers" — but no `mcp.json` exists anywhere under `kiro_power/` (confirmed via `git ls-files`). Since the whole point is a two-server (Timestream + InfluxDB 3) setup, the missing example config is the most valuable artifact for users. Please either add `kiro_power/mcp.json` or remove/adjust both references (here and README:160).

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb3/ingestion.md
```diff
@@ -0,0 +1,201 @@
...
+Content-Type: text/plain
+```
+
+- `bucket` maps to the V3 database name.
+- `org` parameter is accepted but **ignored**.
+- Allows existing V2 client code (including `influxdb-client` Python library) to write to V3 without changes — just update the URL, port, and token prefix.
+
+## Batching
+
+**Always batch writes.** Single-point writes with many concurrent writers cause severe replica lag.
```
scottschreckengaust: Minor: "severe replica lag" is attributed to InfluxDB **v3** here, but `influxdb3/troubleshooting.md:46` states replica lag is *only* relevant to v2 read-replica clusters and "v3 clusters are unaffected." An agent could try to diagnose/tune nonexistent read replicas on v3. Consider rewording the batching rationale (e.g. write amplification / WAL pressure) so the two files agree.

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb2/schema-design.md
```diff
@@ -0,0 +1,223 @@
...
+- Use explicit type suffixes in line protocol: `i` for integer, `"quotes"` for string, bare number for float.
+- Be careful with numeric fields — `10` is a float, `10i` is an integer. Mixing these is the most common conflict.
+- Telegraf plugins have fixed output types — check plugin docs before adding new inputs.
+
+Detection: use `SHOW FIELD KEYS FROM <measurement>` (InfluxQL) to check current field types.
+
+Resolution:
+- You **cannot change** a field's type after first write. Options:
+  1. Write to a new field name (e.g., `temp_f` instead of `temp`) and update queries.
+  2. Delete all data in the measurement and rewrite with the correct type.
```
scottschreckengaust: Minor (data-loss guidance): option 2 offers "Delete all data in the measurement and rewrite" as a routine field-type-conflict fix with no confirmation guard — same pattern at `influxdb3/schema-design.md:113`. Given an agent may execute this against production, consider adding an explicit "⚠️ irreversible — confirm with the user first" note (matching the good pattern already in `influxdb3/onboarding.md:113`), and lead with the safer option 1 (new field name).

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb3/migrations.md
```diff
@@ -0,0 +1,173 @@
...
+| **Timestream for LiveAnalytics** | [`tools/python/liveanalytics_influxdb3_migration_plugin`](https://github.com/awslabs/amazon-timestream-tools/tree/mainline/tools/python/liveanalytics_influxdb3_migration_plugin) (`influxdb_version: v3`) | Unload to S3 → ingest using processing engine → validate |
+| **Timestream for InfluxDB v2** | [`tools/python/influxdb_v2_to_v3_migration`](https://github.com/awslabs/amazon-timestream-tools/tree/mainline/tools/python/influxdb_v2_to_v3_migration) | `influx backup` → translate to line protocol → write to V3 HTTP API |
+| **InfluxDB v1** | No direct tool | Two-hop: `influxd upgrade` → `export-lp` → write line protocol to V3 |
+| **InfluxDB OSS 2.x** | Export line protocol | Write line protocol to V3's v2-compatible endpoint |
+
+---
+
+## From Timestream for LiveAnalytics
+
+Use the same [**LiveAnalytics migration tooling**](https://github.com/awslabs/amazon-timestream-tools/tree/mainline/tools/python/liveanalytics_migration_scripts)
```
scottschreckengaust: Minor (contradicts the table above): this prose says to use `liveanalytics_migration_scripts` (Athena→line-protocol, "targets both V2 and V3", set `influxdb_version: v3`), but the table on line 26 names the purpose-built `liveanalytics_influxdb3_migration_plugin` (S3 unload → processing engine). An agent gets two different tool paths for the same LiveAnalytics→v3 migration. Please reconcile to a single tool/flow.

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb2/onboarding.md
```diff
@@ -0,0 +1,356 @@
...
+    {\"action\":\"read\",  \"resource\":{\"orgID\":\"$ORG_ID\",\"type\":\"dbrp\"}},
+    {\"action\":\"write\", \"resource\":{\"orgID\":\"$ORG_ID\",\"type\":\"dbrp\"}},
+    {\"action\":\"read\",  \"resource\":{\"id\":\"$ORG_ID\",  \"type\":\"orgs\"}},
+    {\"action\":\"read\",  \"resource\":{\"orgID\":\"$ORG_ID\",\"type\":\"tasks\"}},
+    {\"action\":\"write\", \"resource\":{\"orgID\":\"$ORG_ID\",\"type\":\"tasks\"}},
+    {\"action\":\"read\",  \"resource\":{\"id\":\"$USER_ID\", \"type\":\"users\"}},
+    {\"action\":\"write\", \"resource\":{\"id\":\"$USER_ID\", \"type\":\"users\"}}
+  ]}" | python3 -c "import sys,json; print(json.load(sys.stdin)['token'])")
+
+echo "All-access token: $ALL_ACCESS_TOKEN"
```
scottschreckengaust: Minor (credential hygiene): `echo "All-access token: $ALL_ACCESS_TOKEN"` prints a full-privilege data-plane token to stdout, where it persists in shell history / CI logs / the agent transcript. Line 280 already (correctly) says to store it securely — consider not echoing it, or redirecting straight into a secrets store, to match that guidance.

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb2/query-guide.md
```diff
@@ -0,0 +1,285 @@
...
+Content-Type: application/vnd.flux
+Accept: application/csv
+```
+
+Raw body: '<flux script>'
+
+- `Accept: application/csv` is required — Flux always returns CSV. Omitting it causes a 400.
+- `Content-Type: application/vnd.flux` is required.
+- Use org ID (not name) in the `orgID` query param.
+- Include the header `Accept-Encoding: gzip` for responses over 1.4 KB. Using compression saves network bandwidth but increases server-side load.
```
scottschreckengaust, line 21: Nit: the "for responses over 1.4 KB" gzip threshold doesn't correspond to anything in the v2 query API — `Accept-Encoding: gzip` is negotiated regardless of size. Suggest dropping the specific size to avoid an agent conditionally omitting the header based on a nonexistent rule.

scottschreckengaust (commented):
## Review — Timestream for InfluxDB Kiro Skill

Thank you for this — it's a large, well-organized, and genuinely useful steering set. I reviewed all 26 markdown files with **particular focus on security**, since these files are loaded as trusted context that an autonomous agent acts on.

### 🔒 Security review: clean ✅
I specifically hunted for prompt injection, instruction-override, data-exfiltration guidance, hidden/zero-width Unicode, typosquatted domains, and nefarious/destructive commands. **I found none.** Notable positives:
- Every URL resolves to an official destination (`github.com/awslabs/*`, `github.com/influxdata/*`, `docs.aws.amazon.com`, `docs.influxdata.com`, `grafana.com`). No link-text/href mismatch, no typosquats.
- No `curl … | sh`, no `chmod 777`, no credential dumping, no `--insecure`/`-k`/`verify=false` (gRPC config explicitly sets `insecureGrpc: false`).
- The one `rm -rf ~/engine` is clearly-scoped local backup cleanup.
- Security *guidance* is good: recommends `--cidr <your-ip>/32` over `0.0.0.0/0`, least-privilege scoped tokens, and **mandatory user confirmation before destructive deletes** (`influxdb3/onboarding.md:113`).

### What to address
The findings below are **accuracy/consistency** issues, not security ones. Because an agent treats this content as authoritative and may execute it, a *factually wrong* instruction is the real risk here. Two are worth fixing before merge (flagged **[fix before merge]**); the rest are minor.

**[fix before merge]**
1. `influxdb-2-vs-3.md:75` — wrong meaning for a v2 write `400` (contradicts three sibling files). Detail inline.
2. `mcp.json` is referenced by both `POWER.md:125` and `README.md:160` but does not exist in the PR. Detail inline.

**Minor (nice to fix)**
3. `influxdb3/ingestion.md:33` — "replica lag" attributed to v3, contradicting `troubleshooting.md:46`.
4. `influxdb2/schema-design.md:161` (and `influxdb3/schema-design.md:113`) — destructive "delete all data" fix presented without a confirmation guard.
5. `influxdb3/migrations.md:35` — LiveAnalytics prose describes different tooling than the table on line 26.
6. `influxdb2/onboarding.md:260` — echoes an all-access token to stdout.
7. `influxdb2/query-guide.md:21` — fabricated "1.4 KB" gzip threshold.
8. `influxdb3/ingestion.md:184` — `413` is not a status the v3 write API returns.

**Decision: Comment.** The security review — the primary concern — is clean. The remaining items are fixable documentation accuracy/consistency issues; I'd recommend resolving #1 and #2 before merge.

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb3/ingestion.md
```diff
@@ -0,0 +1,201 @@
...
+
+## Write Error Handling
+
+| HTTP Status | Cause | Action |
+|-------------|-------|--------|
+| 204 | Success | — |
+| 400 | Malformed line protocol | Check syntax |
+| 401 | Invalid or expired token | Verify token has write permission |
+| 404 | Database not found | Verify names or IDs |
+| 413 | Payload too large | Reduce batch size |
```
scottschreckengaust: Nit: `413 Payload too large` isn't a status the v3 write API (`/api/v3/write_lp`) returns — oversized/limit conditions surface as `400`/`422`. If 413 can't actually occur, listing it may send error-handling down a branch that never triggers while missing the real `400`/`422` path.

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb-2-vs-3.md
trevorbonas: Fixed.

src/timestream-for-influxdb-mcp-server/kiro_power/POWER.md
trevorbonas: I've removed the dead references.

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb3/ingestion.md
trevorbonas: I've removed this point. v3 doesn't need to worry about replica lag.

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb2/schema-design.md
trevorbonas: I've added the notice that deletion is irreversible.

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb3/migrations.md
trevorbonas: Fixed.

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb2/onboarding.md
trevorbonas: I've removed this echo.

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb2/query-guide.md
trevorbonas: This is copied directly from v2 documentation, nearly word-for-word: https://docs.influxdata.com/influxdb/v2/query-data/execute-queries/influx-api/#use-gzip-to-compress-a-large-query-response

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb3/ingestion.md
trevorbonas: It's documented [here](https://docs.influxdata.com/influxdb3/enterprise/api/write-data/#operation/PostWriteLP).
> Request entity too large.

I've changed this to share the same phrasing, "Request entity too large."

fredjoonpark (approved):

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb-2-vs-3.md
scottschreckengaust: Verified against InfluxData's official [v2 write API docs](https://docs.influxdata.com/influxdb/v2/api/write-data/): a `400` on `POST /api/v2/write` covers **both** malformed line protocol **and** an `org`/`orgID` that doesn't match an organization, while schema/retention conflicts surface as `422`. So line 75 is correct as written, and `db6ea59f` correctly reconciled the sibling files (`troubleshooting.md`, `ingestion.md`, `onboarding.md`) to agree with it. My original comment had the `400` semantics wrong — thanks for the fix. Resolving.

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb2/query-guide.md
scottschreckengaust: You're right — the [v2 query docs](https://docs.influxdata.com/influxdb/v2/query-data/execute-queries/influx-api/) state verbatim: *“We recommend only using gzip compression on responses that are larger than 1.4 KB”* (it's the gzip minimum-output / MTU floor, not a query-API size rule). My nit was mistaken. Resolving.

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb3/ingestion.md
scottschreckengaust: Confirmed — `413 Request entity too large` is a documented response of the v3 write endpoint ([Enterprise write API](https://docs.influxdata.com/influxdb3/enterprise/api/write-data/)), and the wording now matches it. My original nit was wrong. Resolving.

src/timestream-for-influxdb-mcp-server/kiro_power/POWER.md
scottschreckengaust: Verified against `66b22dab`: no `mcp.json` references remain in `POWER.md` or `README.md`. Resolving.

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb3/ingestion.md
scottschreckengaust: Verified against `66b22dab`: the replica-lag point is gone from v3 `ingestion.md`, so it no longer contradicts `influxdb3/troubleshooting.md`. Resolving.

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb2/schema-design.md
scottschreckengaust: Verified against `66b22dab`: option 2 now carries the “⚠️ irreversible — confirm with the user first” note. Resolving.

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb3/migrations.md
scottschreckengaust: Verified against `66b22dab`: the guidance now points to `liveanalytics_migration_scripts` (`influxdb_version: v3`) consistently — the conflicting tool name is gone. Resolving.

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb2/onboarding.md
scottschreckengaust: Verified against `66b22dab`: the token `echo` is removed; the value is only referenced via `$ALL_ACCESS_TOKEN`. Resolving.

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb3/schema-design.md
```diff
@@ -0,0 +1,167 @@
...
+- Use explicit type suffixes in line protocol: `i` for integer, `"quotes"` for string, bare number for float.
+- Be careful with numeric fields — `10` is a float, `10i` is an integer. Mixing these is the most common conflict.
+- Telegraf plugins have fixed output types — check plugin docs before adding new inputs.
+
+Detection: use `SHOW COLUMNS FROM <table>` or query `information_schema.columns`.
+
+Resolution:
+- You **cannot change** a field's type after first write. Options:
+  1. Write to a new field name (e.g., `temp_f` instead of `temp`) and update queries.
+  2. Delete all data in the measurement and rewrite with the correct type.
```
scottschreckengaust: **Destructive step missing irreversibility warning (v2/v3 asymmetry).** The v2 twin at `influxdb2/schema-design.md:161` ends this exact same option with `⚠️ irreversible — confirm with the user first.`, but the v3 copy stops at "...correct type." An agent consuming the v3 doc could delete all data in a table as a routine field-type fix without prompting the user. This looks like the "Mark deletion of measurements as irreversible" fix applied to v2 but not v3 — please add the same warning here.

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb2/troubleshooting.md
```diff
@@ -0,0 +1,158 @@
...
+- Threshold: > 90% for 5 minutes
+- Action: Implement traffic remediation measures or Compute Scaling
+
+MemoryUtilization:
+- Threshold: > 90% for 5 minutes
+- Action: Implement traffic remediation measures or Compute Scaling
+
+DiskUtilization:
+- Threshold: > 85%
+- Action: Try to free up space by deleting old buckets, updating retention configurations or Storage Scaling
```
scottschreckengaust: **Data-destroying action presented as routine cleanup.** "deleting old buckets" is listed as the first of three equivalent free-space remediations with no warning that it permanently destroys all data in the bucket (and on a read-replica cluster can break replication). An operator or agent reacting to a DiskUtilization alarm could delete buckets and lose data. Please add an irreversibility warning, or lead with the non-destructive options (retention config / storage scaling).

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb3/query-guide.md
```diff
@@ -0,0 +1,354 @@
...
+Accept: text/csv
+```
+
+The `query` endpoint supports the following query parameters:
+- `chunked`: Returns points in streamed batches. When set to `true`, InfluxDB chunks responses by series or by every 10,000 points, whichever occurs first.
+- `chunk_size`: Specifies the number of points to include in a chunk, if `chunked` is `true`.
+- `db`: Required. Database name.
+- `epoch`: Timestamp precision. Valid values are: `h`, `m`, `s`, `ms`, `us`, and `ns`.
+- `u`: For query string authentication, the user's username.
+- `p`: For query string authentication, the user's password. InfluxDB v3 enterprise expects this to be a token with read access to the database and will ignore `u`.
```
scottschreckengaust: **Credential exposure via URL query string.** Line 267 shows `...&u=USERNAME&p=PASSWORD` and this line confirms `p` is a token with read access. Putting a token in the URL leaks it into shell history, the InfluxDB access log, and any intervening proxy/LB logs. All other examples in this file correctly use `Authorization: Bearer`. Please add a warning against query-string auth (or drop the `u`/`p` example) and point readers to the header form.

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb3/schema-design.md
```diff
@@ -0,0 +1,167 @@
...
+    "db": "downsampled-90d",
+    "retention_period": "90d"
+  }'
+```
+
+## Retention Policy
+
+Retention is configured per database via `retention_period` using human-readable durations (`"7d"`, `"30d"`, `"1y"`). Null = infinite.
+
+Update retention via `PATCH /api/v3/configure/database/{name}`. To clear retention (keep data indefinitely), use `DELETE /api/v3/configure/database/retention?db=<name>`.
```
scottschreckengaust: **Fabricated endpoint + wrong verb.** Verified against the InfluxDB 3 Core/Enterprise OpenAPI specs:
- To clear retention the real route is `DELETE /api/v3/configure/database/retention_period?db=<name>` — this line drops the `_period` suffix, so the path 404s and retention is never cleared.
- Updating retention uses `PUT /api/v3/configure/database` with `retention_period` in the body (a `db` query param), **not** `PATCH /api/v3/configure/database/{name}` — that verb/path shape doesn't exist.

(The create/set-retention endpoints elsewhere in this file, e.g. line 67, are correct.)

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb3/schema-design.md
```diff
@@ -0,0 +1,167 @@
...
+## Database Design
+
+A database is the top-level data container in V3 — there are no organizations.
+
+Design principle: same as V2 — **one database per retention period**.
+
+Guidance:
+- Create databases via `POST /api/v3/configure/database`.
+- Retention is set via `retention_period` (e.g. `"7d"`, `"90d"`, `"1y"`). Null or omitted = infinite.
+- Tables are limited to **10,000 across all databases** (default) — plan measurement names accordingly.
```
scottschreckengaust: **Table limit contradicts the rest of the PR.** This says "10,000 across all databases (default)" with no edition qualifier, but `influxdb-2-vs-3.md:24` and `influxdb3/troubleshooting.md:113` state Core = **2,000**, Enterprise = 10,000 — and this same file says 2,000 for Core at line 21. Since onboarding provisions Core by default, a Core user trusting 10,000 hits 422 write rejections ~5× sooner. Please qualify with the edition (same issue at `influxdb3/onboarding.md:186` and `influxdb3/gotchas.md:23`).

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb2/onboarding.md
```diff
@@ -0,0 +1,354 @@
...
+  -H "Accept: application/csv" \
+  -d '{
+    "query": "from(bucket: \"my-bucket\") |> range(start: -1h) |> filter(fn: (r) => r._measurement == \"cpu\")",
+    "type": "flux"
+  }'
+```
+
+**InfluxQL (V1 compatibility endpoint):**
+```bash
+curl "https://<endpoint>:8086/query?db=my-bucket&q=SELECT+*+FROM+cpu+WHERE+time+>+now()+-+1h" \
```
scottschreckengaust: **Verification step will fail on a fresh instance.** This Step 7 "verify" uses an InfluxQL V1-compatibility query (`/query?db=...`), but nothing in onboarding ever creates a DBRP mapping — the token permissions at lines 251-252 grant `dbrp` access but don't create a mapping, and Timestream's auto-created initial bucket has none. So this returns "no databases"/empty, exactly the failure `troubleshooting.md:50`, `dashboard-guide.md:27,161`, and `query-guide.md` warn about. A user reads this as "my write failed" and debugs the wrong thing. Either use the Flux verify (shown just above) or add a DBRP-creation step first.

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb3/query-guide.md
```diff
@@ -0,0 +1,354 @@
...
+**Keywords**
+```sql
+AND
+ALL
+ANALYZE
+AS
+ASC
+AT TIME ZONE
+BETWEEN
+BOTTOM
```
scottschreckengaust, line 80: **`TOP`/`BOTTOM` are not SQL.** This `**Keywords**` block sits under `## SQL Query Endpoint` → `### SQL Patterns` (fenced ```sql), but `TOP` (line 98) and `BOTTOM` (this line) are InfluxQL selectors — Apache DataFusion SQL has neither. An agent that trusts this list emits `SELECT TOP(...)` and gets a parse/unknown-function error. The SQL equivalents are `ORDER BY ... LIMIT` or the `SELECTOR_*` functions this guide already documents. Please move `TOP`/`BOTTOM` to the InfluxQL section.

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb2/schema-design.md
```diff
@@ -0,0 +1,223 @@
...
+
+Resolution:
+- You **cannot change** a field's type after first write. Options:
+  1. Write to a new field name (e.g., `temp_f` instead of `temp`) and update queries.
+  2. Delete all data in the measurement and rewrite with the correct type. ⚠️ irreversible — confirm with the user first.
+  3. Create a new measurement with the correct schema and migrate data via a task.
+
+## Series Cardinality
+
+A **series** is a unique combination of measurement name + tag set. Series cardinality = total number of unique series across all measurements.
```
scottschreckengaust: **Cardinality formula undercounts.** InfluxData defines *series cardinality* as unique combinations of measurement + tag set **+ field key** (the field key is part of the series key). Defining it as measurement + tag set only undercounts by the number of fields per measurement — the file's own `cpu` example (usage_idle/usage_system/usage_user = 3 fields) is 3× off. A reader sizing against the ~10M threshold below could significantly undersize the instance.

src/timestream-for-influxdb-mcp-server/kiro_power/POWER.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Available MCP Tools
+
+1. [Timestream for InfluxDB MCP Server](https://github.com/awslabs/mcp/tree/main/src/timestream-for-influxdb-mcp-server#available-tools)
+
+The Timestream for InfluxDB MCP server provides tools for managing Amazon Timestream for InfluxDB clusters, instances, and parameter groups, along with InfluxDB-specific operations for writing data, querying with Flux, and managing buckets and organizations.
+
+2. [InfluxDB 3 MCP Server](https://github.com/influxdata/influxdb3_mcp_server?tab=readme-ov-file#available-tools)
+
+The InfluxDB 3 MCP Server provides tools for data operations (read/write), database lifecycle management (create/update/delete), schema inspection, and authentication token management across all InfluxDB versions (Core, Enterprise, Cloud).
```
scottschreckengaust: **"all InfluxDB versions (Core, Enterprise, Cloud)" is inaccurate.** Core/Enterprise/Cloud are all editions of InfluxDB **3**, not "all versions" (`glossary.md:92-94`, `influxdb-2-vs-3.md:19`). Combined with "database lifecycle management (create/update/delete)", this reads as though the InfluxData v3 server manages V2 instances too — but line 114 and `README.md:160` assign V2 + AWS control plane to *this* server. Suggest "across all InfluxDB 3 editions (Core, Enterprise, Cloud)".

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb2/onboarding.md
```diff
@@ -0,0 +1,354 @@
...
+
+## Step 3 — Create an All-Access Token
+
+The Secrets Manager secret contains `username` and `password` only — no token. Use them to sign in, then create an all-access token for all future data plane operations.
+
+```bash
+# 1. Sign in with Basic auth (Authorization header, not Cookie)
+SESSION=$(curl -si -X POST \
+  "https://<endpoint>:8086/api/v2/signin" \
+  -H "Authorization: Basic $(echo -n 'username:password' | base64)" \
```
scottschreckengaust: **Literal placeholder never told to substitute → 401.** `'username:password'` is a literal string here, and the surrounding prose (line 221 "Use them to sign in") never says to replace it with the real username/password retrieved from Secrets Manager in Step 2 — unlike `<endpoint>`, it doesn't use the angle-bracket placeholder convention. A verbatim copy-paste base64-encodes the literal string, `/api/v2/signin` returns 401, `SESSION` is empty, and every later curl fails unauthenticated with no hint. Please make the substitution explicit (e.g. `$USERNAME:$PASSWORD` sourced from the Step 2 secret).

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb2/onboarding.md
```diff
@@ -0,0 +1,354 @@
...
+**Standalone (single node):**
+```bash
+# 1. Create parameter group (name must be alphanumeric only — no hyphens)
+aws timestream-influxdb create-db-parameter-group \
+  --name myv2params \
+  --parameters '{"InfluxDBv2": {}}'
+# Note the `id` field in the response — use it (not the name) in the next command
+
+# 2. Create instance — note the influxAuthParametersSecretArn in the response
+aws timestream-influxdb create-db-instance \
```
scottschreckengaust, line 141: **Missing `--region` (inconsistent with the v3 guide).** This `create-db-instance` (and `create-db-parameter-group` at line 135, `create-db-cluster` at line 182) omit `--region`, yet `influxdb3/onboarding.md:231` states "All commands require `--region`" and the tagging commands later in this same file (lines 211-213) all pass `--region us-east-1`. A user without a default region configured hits "You must specify a region" here while the v3 flow works. Please add `--region` for consistency.

src/timestream-for-influxdb-mcp-server/kiro_power/POWER.md
```diff
@@ -0,0 +1,133 @@
...
+# Amazon Timestream for InfluxDB Power
+
+## Overview
+The Amazon Timestream for InfluxDB Power provides access to Amazon Timestream for InfluxDB, a fully managed time-series database service compatible with InfluxDB 2 and InfluxDB 3. InfluxDB is purpose-built for high-ingest, high-cardinality workloads such as IoT telemetry, observability, metrics, and real-time analytics.
+
+This Power enables you to deploy and operate Timestream for InfluxDB instances and clusters, write and query time-series data using InfluxDB APIs and query languages, and manage schemas, retention, and migrations while leveraging AWS-managed scaling, durability, and security.
+
+- **Dual Version Support**: Deploy and manage both InfluxDB 2 and InfluxDB 3 instances and clusters
+
+- **Multi-Protocol Query Support**: Run Flux, InfluxQL, or SQL queries directly against your Timestream for InfluxDB instance
```
scottschreckengaust: _(non-blocking)_ **"Run Flux, InfluxQL, or SQL ... against your instance"** reads as if all three run on any single instance, but V2 has no SQL and V3 has no Flux (`glossary.md:17`, `influxdb-2-vs-3.md:14,65-66`). Consider scoping per version so a reader doesn't try SQL on V2 / Flux on V3.

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb3/ingestion.md
```diff
@@ -0,0 +1,201 @@
...
+- Allows existing V2 client code (including `influxdb-client` Python library) to write to V3 without changes — just update the URL, port, and token prefix.
+
+## Batching
+
+**Always batch writes.**
+
+- Target **5,000+ points per request** minimum
+- Optimal batch size: 5,000–10,000 points
+- Max recommended: ~50,000 points per request (beyond this, HTTP timeouts become a risk)
+- For high-throughput workloads, tune `ingestQueryInstances` in the parameter group (see Tuning section below) or scale vertically by updating the cluster size with the `--db-instance-type` option (with the AWS CLI).
```
scottschreckengaust: _(non-blocking)_ **Enterprise-only tuning lever offered without an inline Core caveat.** `ingestQueryInstances` is Enterprise-only (Core is single-node — `parameters.md:95`); a Core reader here can't use it and should be pointed to `--db-instance-type` vertical scaling (which you do mention). Also the tuning table at line 196 shows "Default 2", but `parameters.md:101` marks this field `[required]` with no default — please reconcile.

src/timestream-for-influxdb-mcp-server/kiro_power/steering/troubleshooting.md
```diff
@@ -0,0 +1,53 @@
...
+
+Both engines expose a `GET /health`. Use it to separate
+"endpoint/network is down" from "auth, namespace, or payload is wrong":
+
+```shell
+curl https://<endpoint>:<port>/health
+```
+
+A healthy response (`OK`) means the engine is reachable, so the problem is most likely
+auth (steps 2–3), networking to a private resource (step 5), or the namespace/payload —
```
scottschreckengaust: _(non-blocking)_ **Off-by-one step reference.** "networking to a private resource (step 5)" — networking is **step 4** (line 35); step 5 (line 38) is "Namespace mismatch". A reader jumping to step 5 lands on the wrong fix. Should be "(step 4)".

src/timestream-for-influxdb-mcp-server/kiro_power/steering/troubleshooting.md
```diff
@@ -0,0 +1,53 @@
...
+### Quick isolation: health check
+
+Both engines expose a `GET /health`. Use it to separate
+"endpoint/network is down" from "auth, namespace, or payload is wrong":
+
+```shell
+curl https://<endpoint>:<port>/health
+```
+
+A healthy response (`OK`) means the engine is reachable, so the problem is most likely
```
scottschreckengaust: _(non-blocking)_ **`/health` doesn't return `OK`.** The real body is JSON `{"name":"influxdb","status":"pass",...}` — as `influxdb2/onboarding.md:271` correctly shows. "A healthy response (`OK`)" reads as shorthand, but anyone scripting a check against the literal string `OK` will never match. Consider "a healthy response (`"status":"pass"`)".

scottschreckengaust (changes requested):
## Code review — Timestream for InfluxDB Kiro Skill

Nice, thorough steering docs — the control-plane/v2/v3 split is clear and most of the technical content checks out (verified tool names against `server.py`, AWS CLI flags, and the InfluxDB 3 OpenAPI spec). I'm requesting changes on a small set of items that would cause **data loss, credential exposure, or silent failures** for a user/agent who follows the docs literally, plus some accuracy/consistency nits.

**Must-fix (data loss / security / wrong API):**
1. `influxdb3/schema-design.md:113` — destructive "delete all data" step is missing the irreversibility warning its v2 twin has.
2. `influxdb2/troubleshooting.md:114` — "delete old buckets" to free disk, no data-loss warning.
3. `influxdb3/query-guide.md:278` — read token documented as a URL query param (leaks to logs/history), no warning.
4. `influxdb3/schema-design.md:88` — fabricated retention endpoint + wrong HTTP verb.
5. `influxdb3/schema-design.md:69` / `onboarding.md:186` / `gotchas.md:23` — unqualified "10,000 tables" contradicts Core = 2,000 elsewhere in the PR.
6. `influxdb2/onboarding.md:337` — Step 7 InfluxQL verify fails because no DBRP mapping is ever created.

**Accuracy / consistency (non-blocking but worth fixing):** items 7–15 below.

Several of these (the missing irreversibility warning, the query-string token) are the same hazard classes the recent commits addressed for the v2 files but not the v3 twins — worth a sweep for v2/v3 parity. Full details inline.

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb3/query-guide.md
trevorbonas: Copied directly from the [InfluxDB v3 SQL reference](https://docs.influxdata.com/influxdb3/enterprise/reference/sql/#keywords). I've added a note that they're reserved keywords and can't be used as identifiers.

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb2/schema-design.md
trevorbonas: Fixed, this now uses a definition that includes field keys. The example has been updated with fields while keeping a cardinality of 2.

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb2/onboarding.md
trevorbonas: The fact that username and password should be substituted is now stated. Angle brackets are also now used in the command to indicate substitution.

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb2/onboarding.md
trevorbonas: The AWS CLI requires a region. I've made a note of this in `influxdb3/onboarding`. The `--region` flag isn't actually required, the region can be set in other ways.

src/timestream-for-influxdb-mcp-server/kiro_power/POWER.md
trevorbonas: Added region scoping.

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb3/ingestion.md
trevorbonas: I've added a note that this is only for Enterprise and that for Core vertical scaling should be considered. I've updated the table in parameters.md with defaults (from service-owned parameter group).

src/timestream-for-influxdb-mcp-server/kiro_power/steering/troubleshooting.md
trevorbonas: Changed to "(step 4)".

src/timestream-for-influxdb-mcp-server/kiro_power/steering/troubleshooting.md
trevorbonas: Removed mention of "OK".

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb3/schema-design.md
trevorbonas: I've added the note that deletion is irreversible.

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb2/troubleshooting.md
trevorbonas: I've added a note that deleting buckets is irreversible.

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb3/query-guide.md
trevorbonas: I've deleted `u` and `p` from this example.

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb3/schema-design.md
trevorbonas: Fixed, it now uses the correct verb and endpoint with the correct body, as verified in the InfluxDB source code (the online documentation for this endpoint is inaccurate regarding the request body).

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb3/schema-design.md
trevorbonas: I've specified the table limits for both Core and Enterprise.

src/timestream-for-influxdb-mcp-server/kiro_power/steering/influxdb2/onboarding.md
trevorbonas: I've deleted this InfluxQL example.

PR #3995, opened 2026-06-19
fix(amazon-bedrock-agentcore-mcp-server): preserve non-ASCII text in browser_evaluate results

### LINKED ISSUE #3994: amazon-bedrock-agentcore-mcp-server: browser_evaluate returns non-ASCII (CJK, emoji) as \uXXXX escapes
First, thank you to the AgentCore MCP server maintainers — the browser primitive is genuinely useful for agentic web automation.

### Summary

The `browser_evaluate` tool serializes object/array results with `json.dumps(...)` but omits `ensure_ascii=False`. As a result, non-ASCII text (CJK, emoji, accented Latin, etc.) in an object/array result is returned as `\uXXXX` escape sequences rather than the original characters.

The same function's scalar-string path (`isinstance(result, (str, int, float, bool))`) already returns the text verbatim, so the **same underlying string renders differently depending only on the shape of the JS return value** — a scalar string is fine, but the moment it's wrapped in an object or array it gets escaped.

### Where

`src/amazon-bedrock-agentcore-mcp-server/awslabs/amazon_bedrock_agentcore_mcp_server/tools/browser/observation.py`, in `ObservationTools.browser_evaluate` (~L325-331):

```python
if isinstance(result, (str, int, float, bool)):
    return f'Result: {result}'                       # scalar: verbatim, e.g. "東京" → 東京

return f'Result:\n{json.dumps(result, indent=2, default=str)}'   # object/array: escaped → 東京
```

### Reproduction (observed behavior)

Evaluating an expression that returns a scalar string vs. an object containing the same string:

| `page.evaluate()` returns | `browser_evaluate` output |
|---|---|
| `'東京'` (scalar) | `Result: 東京` ✅ |
| `{'city': '東京', 'emoji': '🌐'}` (object) | `Result:\n{\n  "city": "東京",\n  "emoji": "🌐"\n}` ❌ |

The escaped form is harder for an LLM/agent to read and is inconsistent with the scalar path.

### Suggested fix

Add `ensure_ascii=False` to the `json.dumps` call so the object/array path matches the scalar path. This is the same convention the bedrock-kb-retrieval server fix uses (#3820/#3821) — note this is a **different server**, so it's tracked separately.

I'm happy to open a PR (already prepared locally with a regression test). Please let me know if you'd prefer a different approach.

kimnamu:
Thank you to the AgentCore MCP server maintainers for the browser primitive — this is a small consistency fix in that area.

> This targets the **amazon-bedrock-agentcore-mcp-server**. It is a **separate server** from the bedrock-kb-retrieval fix in #3821 (#3820); same UTF-8-preservation theme, different package and PR.

### Summary

`browser_evaluate` serializes object/array results with `json.dumps(...)` but without `ensure_ascii=False`, so non-ASCII text (CJK, emoji) is returned as `\uXXXX` escape sequences. The scalar-string return path in the **same function** already returns the text verbatim, so identical data renders differently depending only on the result's shape. This aligns the object/array path with the already-correct scalar path.

### Changes

- `tools/browser/observation.py` — add `ensure_ascii=False` to the `json.dumps` call in `browser_evaluate` (1 line).
- `tests/browser/test_unit_interaction.py` — add two sibling regression tests in `TestObservationTools`: scalar path (baseline) and object path (fix).
- `CHANGELOG.md` — `## Unreleased` / `### Fixed` entry.

### User experience (before / after)

Same JS data, output depends only on the return shape:

| `page.evaluate()` returns | Before | After |
|---|---|---|
| `'東京'` (scalar) | `Result: 東京` ✅ | `Result: 東京` ✅ (unchanged) |
| `{'city': '東京', 'emoji': '🌐'}` (object) | `"東京" / "🌐"` ❌ | `"東京" / "🌐"` ✅ |
| ASCII object results | unchanged ✅ | unchanged ✅ |
| Public API / tool signature / return type | — | Unchanged ✅ |
| Scalar / null / error paths | — | Unchanged ✅ |

### Tests — proof the test catches the bug

With the source change applied, both new tests pass (full `TestObservationTools`: 25 passed):

```
$ uv run --frozen pytest tests/browser/test_unit_interaction.py -k "non_ascii" -q
..                                                                       [100%]
2 passed, 72 deselected, 1 warning in 0.32s
```

Reverting only the source (keeping the tests) makes the object test fail while the scalar test still passes — demonstrating the shape-dependent inconsistency:

```
>       assert '東京' in result
E       assert '東京' in 'Result:\n{\n  "city": "\\u6771\\u4eac",\n  "emoji": "\\ud83c\\udf10"\n}'

tests/browser/test_unit_interaction.py:1209: AssertionError
============ 1 failed, 1 passed, 72 deselected, 1 warning in 0.43s =============
```

Restoring the fix → all pass again. Full local gate:

```
$ uv run --frozen pytest tests/browser/test_unit_interaction.py -q
74 passed, 1 warning in 0.81s

$ uv run --frozen ruff check <changed files>
All checks passed!
$ uv run --frozen ruff format --check <changed files>
2 files already formatted
$ uv run --frozen --all-extras --dev pyright <changed files>
0 errors, 0 warnings, 0 informations
```

### Scope

Minimal, low-risk: this is "filling in the established UTF-8-preservation convention" for the one `json.dumps` call in `browser_evaluate` that was missing it. The server has no other `ensure_ascii=False` usages today; this PR is scoped to `browser/observation.py` only and does not touch other tools.

Fixes #3994

---

## Acknowledgment


kimnamu:
Hi @kevin-orellana 👋 — per the stale bot's instruction, could this PR be reopened? Apologies for the ping, and thanks for maintaining the AgentCore server.

I re-verified today that the fix is still needed and still applies cleanly:

| Check | Status |
|---|---|
| `tools/browser/observation.py` on `main` still calls `json.dumps(result, indent=2, default=str)` without `ensure_ascii=False` | ✅ bug still present |
| Linked issue #3994 | ✅ still open |
| Duplicate/competing fix merged | ❌ none found |
| Branch vs `main` | ✅ no conflicts (3 files, none touched on `main` since) |

The change itself is one line (plus two regression tests and a CHANGELOG entry), scoped to the object/array return path of `browser_evaluate` so it matches the already-correct scalar path. Happy to rebase or adjust anything if needed.

kevin-orellana (approved):
Approving as a CODEOWNER for `src/amazon-bedrock-agentcore-mcp-server`. Apologies for the delay here - this was closed by the stale bot rather than on the merits, and the fix is correct.

Validated by execution against this branch, not by inspection:

- Full suite on `main` = 812 passed; on this branch = 814 passed (the +2 are the new regression tests). No failures, 50 live/browser tests deselected as usual.
- Reverting only the one-line source change while keeping the new tests reproduces the exact failure described in #3994, so the tests genuinely cover the bug.
- `ruff` and `pyright` clean.
- `git log <base>..main -- src/amazon-bedrock-agentcore-mcp-server/` shows 0 commits touching this package since the branch point, so no conflicts.
- Encoding safety checked: the MCP SDK force-wraps `sys.stdout.buffer` in a UTF-8 `TextIOWrapper` in `mcp/server/stdio.py` on every platform, so `ensure_ascii=False` introduces no Windows/cp1252 regression risk despite this server's earlier history there (#2752, #3280).

Scope is right, too. There are only three `json.dumps` call sites in this server: this one, `tools/identity/resource_policy.py` (serializes an IAM policy for the AWS API, not LLM-facing), and `tools/runtime/guide.py` (inside a docstring, never executed). Narrowing the change to `observation.py` was the correct call.

Worth flagging that the inconsistency this fixes is the bad part: the same string rendered differently depending only on the shape of the JS return value, since the scalar path already passed non-ASCII through verbatim.

Two housekeeping notes for whoever lands this: #4132 is a duplicate of this change without tests or a CHANGELOG entry and should close in favor of this PR. And this needs a second approval plus a merge-queue entry from someone with maintain - I only have write/triage. Adding `backlog` would keep the stale bot off it in the meantime.

Bhavik-T (approved):

PR #3996, opened 2026-06-19
fix: preserve agent space resources, retry api guide load, handle null severity

ljainiaz:
### Changes

Fixes three small bugs in the AWS Security Agent MCP server:
  
  1. **Agent space resource preservation** — `update_agent_space` replaced the agent space's `awsResources` wholesale, but `setup` (adding a service role) and `_ensure_s3_bucket` (registering a scan bucket) each  sent only a partial set. This silently dropped the other resources — e.g. running `setup` on an existing space wiped its configured S3 buckets, breaking pentest and full-repo code-review setups. The client method now takes the complete `awsResources` dict, and both call sites read the existing resources, merge in their addition, and send the full set. Any field present on the space is preserved, not just roles/buckets.

  2. **`get_api_guide` cache poisoning** — a transient failure to load the boto3 service model cached an error placeholder permanently, so every later call returned "Could not load service model" until the server restarted. The cache is now only populated on success and retries on the next call.

  3. **Null severity crash** — `get_scan_findings` with a severity filter crashed with `AttributeError` when the API returned a finding/threat with `riskLevel`/`severity` set to `null`. Both filters now coerce null to an empty string.

### User experience

 **Before:**
  - Running `setup` against an existing agent space removed its S3 buckets, breaking already-configured pentest/code-review workflows.
  - A one-time network/credential blip during `get_api_guide` permanently broke API discovery for the session.
  - Filtering findings by severity could crash if any finding had a null severity.

  **After:**
  - `setup` and bucket registration preserve all existing agent space resources.
  - `get_api_guide` recovers automatically after a transient failure.
  - Severity filtering tolerates null severities and returns the matching findings.

## Checklist

* [x] Changes are documented

**RFC issue number**: NA

Checklist:

## Acknowledgment

### DIFF
--- a/src/security-agent-mcp-server/awslabs/security_agent_mcp_server/aws_client.py
+++ b/src/security-agent-mcp-server/awslabs/security_agent_mcp_server/aws_client.py
@@ -67,16 +67,10 @@ class SecurityAgentClient:
         self,
         agent_space_id: str,
         name: str,
-        iam_roles: Optional[list[str]] = None,
-        s3_buckets: Optional[list[str]] = None,
+        aws_resources: Optional[dict] = None,
     ) -> dict:
-        """Update an agent space with roles and buckets."""
+        """Update an agent space, replacing its awsResources with the provided dict."""
         kwargs: dict[str, Any] = {'agentSpaceId': agent_space_id, 'name': name}
-        aws_resources: dict[str, Any] = {}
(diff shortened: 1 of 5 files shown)


jigar-lab (approved):

scottschreckengaust (approved):
LGTM

PR #4025, opened 2026-06-23
fix(cloudwatch-applicationsignals): use shared profile-aware logs client + guide DI for incident RCA

mxiamxia:
## Summary

Two related improvements to the CloudWatch Application Signals MCP server, both aimed at making incident root-cause analysis (RCA) via Dynamic Instrumentation (DI) actually work end-to-end.

### 1. Fix: ServiceEvents Logs client ignored `AWS_PROFILE` and never refreshed credentials

The CloudWatch Logs client backing the ServiceEvents tools — `find_deployment`, `get_recent_incidents`, `get_incident_root_cause` — was cached in a module-level global and built with a bare `boto3.client('logs', ...)`. That had two consequences:

- It did **not** honor `AWS_PROFILE`, unlike the PromQL-based tools which build a fresh `boto3.Session(profile_name=...)` per call.
- Once the cached client's session token expired (e.g. credentials refreshed out of band), **every** Logs Insights query failed with `ExpiredTokenException` until the process restarted. Meanwhile the PromQL tools kept working, which is the tell-tale asymmetry.

Fix:
- Build the logs client from `boto3.Session(profile_name=os.environ.get('AWS_PROFILE'), ...)` so it resolves credentials the same way the rest of the server does.
- On an expired/invalid-credential error from `start_query`, drop the cached client and retry once with a freshly built one (a small helper `_is_expired_credentials_error` classifies the error from both the boto `Error.Code` and the message text). All existing error paths are preserved — non-credential errors still raise `CwLogsQueryError` exactly as before.

### 2. Docs: connect incident RCA to Dynamic Instrumentation

Enhances two tool descriptions so an agent knows when to reach for DI:
- **`get_incident_root_cause`**: adds guidance that when a stack trace localizes *where* an exception threw but not *why* — generic-symptom exceptions (`/ by zero`, `NullPointerException`, `IndexOutOfBoundsException`, …) whose deciding value is runtime-only data (a downstream response field, a DB row, a method local) — the next step is a DI breakpoint capturing the relevant args/locals.
- **`create_instrumentation`**: frames its primary RCA use as capturing the runtime values that telemetry and source code cannot show, and points back to `get_sample_snapshot_for_breakpoint` to read them.

## Testing

- The credential-refresh path was exercised end-to-end against a live Application Signals account: after refreshing the underlying profile out of band, `find_deployment` / `get_recent_incidents` / `get_incident_root_cause` recovered automatically instead of failing with `ExpiredTokenException`.
- `python -m py_compile` passes on all three files; `ruff check` / `ruff format` and the secret-scan pre-commit hooks pass.
- The doc-only changes (`tools.py`, `crud_tools.py`) touch docstrings only — no executable code.

## Acknowledgment

### DIFF
--- a/src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/aws_clients.py
+++ b/src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/aws_clients.py
@@ -167,3 +167,8 @@ def get_applicationsignals_client():
 def get_cloudwatch_client():
     """Return the module-level CloudWatch client (lazy accessor; see above)."""
     return cloudwatch_client
+
+
+def get_logs_client():
+    """Return the module-level CloudWatch Logs client (lazy accessor; see above)."""
+    return logs_client
--- a/src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/dynamic_instrumentation/crud_tools.py
+++ b/src/cloudwatch-applicationsignals-mcp-server/awslabs/cloudwatch_applicationsignals_mcp_server/dynamic_instrumentation/crud_tools.py
@@ -66,6 +66,21 @@ def create_instrumentation(
     capture_arguments for method/function-level targets and capture_locals for
     line-level targets.
 
+    **Primary RCA use — capture the runtime values that telemetry and source code
+    cannot show.** Reach for this tool when an incident's stack trace tells you WHERE
+    an exception threw but not WHY: the exception message is a generic symptom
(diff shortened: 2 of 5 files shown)


mxiamxia:
Updated to address CI:
- Fixed `test_get_logs_client_builds_and_caches` for the new `boto3.Session(...).client()` construction (the previous test mocked `boto3.client` directly).
- Added tests covering the new expired-credential reset+retry path in `_execute` (success and re-failure) — `cw_logs.py` line coverage back to 99%.
- Rebased onto current `main` (incl. #4021) and removed a demo-specific example phrase from the `create_instrumentation` docstring in favor of a generic one.

Local verification: `pytest` (41 cw_logs tests pass / full suite green), `ruff check` + `ruff format`, and `pyright` all clean.

wangzlei (approved):

vastin (approved):

PR #4038, opened 2026-06-25
feat: add RODA MCP server

guyu-ye:
## Summary
Adds the [Registry of Open Data on AWS (RODA)](https://registry.opendata.aws/) MCP server for discovering and exploring open datasets from the Registry. The Registry hosts hundreds of publicly available datasets including climate data, genomics, satellite imagery, and more on Amazon Simple Storage Service (S3).

### Changes
Adds code for RODA MCP server initial launch. This MCP server has 10 tools to make it easy for researchers to discover, explore, and evaluate open datesets on AWS:

| Tool | Description |
|------|-------------|
| `search_datasets` | Search by keyword with optional filters for tags, organization, and license type |
| `list_datasets` | List all datasets with optional tag filtering |
| `get_dataset_details` | Get complete details for a specific dataset including resources and access info |
| `discover_by_organization` | Find datasets managed by a specific organization (e.g., NASA, NOAA) |
| `discover_by_license` | Find datasets by license type (e.g., Creative Commons, MIT) |
| `find_related_datasets` | Find datasets related to a given dataset based on shared tags |
| `get_knowledge_base_stats` | Get registry statistics including top tags, organizations, and resource types |
| `preview_dataset` | Show S3 bucket structure for datasets (no download, no AWS account needed). This is only available for public datasets without controlled access. You should review and agree to the dataset license before previewing datasets.|
| `sample_dataset` | Read the first 100KB of a specific file from a public dataset's S3 bucket. This is only available for public datasets without controlled access. You should review and agree to the dataset license before sampling datasets.|
| `search_stac_endpoints` | Find datasets with STAC (SpatioTemporal Asset Catalog) API endpoints |

Security consideration and high-level design considerations are also included. 

### User experience

Users can configure this MCP server for any MCP-compatible AI assistant, and ask questions in natural language about relevant datasets. They can also preview and sample a small file to quickly evaluate these datasets.

## Checklist

* [x] Changes are documented

## Acknowledgment


src/roda-mcp-server/awslabs/roda_mcp_server/server.py
```diff
@@ -0,0 +1,1011 @@
...
+            's3',
+            config=Config(signature_version=UNSIGNED, s3={'use_ssl': True}),
+            region_name=s3_resource.get('Region', 'us-east-1'),
+            verify=True,
+        )
+        # HEAD first to get file size without downloading
+        head = s3.head_object(Bucket=bucket_name, Key=file_key)
+        file_size = head['ContentLength']
+
+        bytes_to_read = min(file_size, max_bytes)
```
scottschreckengaust, line 1038: **Critical — zero-byte file crashes this tool.** For a 0-byte object, `bytes_to_read = min(0, max_bytes) = 0`, producing `Range='bytes=0--1'`, which S3 rejects with `InvalidRange`. The error then falls into the generic `except Exception` below and is reported as an opaque "Unexpected error reading file." Guard `file_size == 0` and return empty content directly. Add a regression test with `ContentLength: 0`.

src/roda-mcp-server/awslabs/roda_mcp_server/server.py
```diff
@@ -0,0 +1,1011 @@
...
+        # with a warning; the registry requires specific data attributes to be present.
+        datasets = []
+        invalid_count = 0
+        for line in content_bytes.decode('utf-8').strip().split('\n'):
+            if not line:
+                continue
+            try:
+                dataset = json.loads(line)
+            except json.JSONDecodeError as exc:
+                print(f'Warning: skipping malformed JSON line: {exc}')
```
scottschreckengaust: **Critical — `print()` writes to stdout, which is the JSON-RPC channel for an stdio MCP server.** At best these warnings are invisible (no log capture); at worst they corrupt the protocol stream. Same issue at line 116. Use the repo convention `loguru` at WARNING level (logs to stderr) and add it to `pyproject.toml` dependencies. If malformed lines are a large fraction of the total, consider failing loudly — that signals an upstream format change, and datasets would otherwise silently vanish from every tool.

src/roda-mcp-server/awslabs/roda_mcp_server/server.py
```diff
@@ -0,0 +1,1011 @@
...
+            {
+                'dataset': dataset.get('Name', ''),
+                'slug': slug,
+                'error': f'AWS error: {error_code}',
+                'message': str(e),
+            },
+            indent=2,
+        )
+
+    except Exception as e:
```
scottschreckengaust, line 863: **Important — broad `except Exception` masks real bugs.** `ClientError` is correctly handled above, so only *unexpected* errors reach here (KeyError on malformed `Contents` items, botocore `EndpointConnectionError`/`SSLError`, `ImportError` from the in-function `import boto3`, etc.). These get flattened into a success-shaped JSON string with no log and no stack trace, so a code defect is indistinguishable from a transient network blip. Narrow to the expected network exceptions and log the rest with context (`slug`, `bucket`, `region`, `prefix`). Same pattern at line 893 in `sample_dataset`.

src/roda-mcp-server/awslabs/roda_mcp_server/server.py
```diff
@@ -0,0 +1,1011 @@
...
+    # Check if cache is valid
+    if _datasets_cache is not None and _cache_timestamp is not None:
+        if datetime.now() - _cache_timestamp < timedelta(hours=CACHE_EXPIRY_HOURS):
+            return _datasets_cache
+
+    async with httpx.AsyncClient(verify=True, http2=True) as client:
+        # Fetch the NDJSON file with all datasets
+        # All external API calls require TLS 1.2 or higher with certificate validation enabled
+        # Retry up to 2 times to handle brief CDN propagation delays
+        max_retries = 2
```
scottschreckengaust: **Important — retry loop only retries checksum *mismatches*, not network errors.** `raise_for_status()` inside the loop means a transient 503 on the data or checksum fetch throws immediately and bypasses the retry, contradicting the "brief CDN propagation delays" rationale. Separately, a *persistent* mismatch — the tampering/corruption threat `ChecksumValidationError` documents — is reported with the same soothing "try again in a few minutes" message and is never logged. Log expected-vs-computed hashes at WARNING/ERROR before raising, and catch `httpx.HTTPStatusError`/timeouts inside the loop so network blips are actually retried.

src/roda-mcp-server/awslabs/roda_mcp_server/server.py
```diff
@@ -0,0 +1,1011 @@
...
+        content_bytes = None
+        for attempt in range(max_retries + 1):
+            response = await client.get(REGISTRY_NDJSON_URL, timeout=30.0)
+            response.raise_for_status()
+            content_bytes = response.content
+
+            # Fetch and validate checksum
+            checksum_response = await client.get(REGISTRY_CHECKSUM_URL, timeout=10.0)
+            checksum_response.raise_for_status()
+            expected_checksum = checksum_response.text.strip().split()[0].lower()
```
scottschreckengaust: **Important — uncaught `IndexError`.** If the checksum endpoint returns an empty or whitespace-only body, `split()[0]` raises `IndexError`, which escapes `fetch_datasets` raw instead of becoming the intended `ChecksumValidationError`. Guard the parse and raise `ChecksumValidationError` with a specific message on empty/malformed checksum content.

src/roda-mcp-server/awslabs/roda_mcp_server/server.py
```diff
@@ -0,0 +1,1011 @@
...
+    )
+
+
+@mcp.tool()
+async def list_datasets(tag: str | None = None, limit: int = 20) -> str:
+    """List RODA datasets with optional tag filtering.
+
+    Args:
+        tag: Optional tag to filter by (e.g., 'climate', 'genomics', 'satellite imagery')
+        limit: Maximum number of results to return (default: 10)
```
scottschreckengaust: **Docstring mismatch.** Says `(default: 10)` but the signature is `limit: int = 20` (line 271). Update to `(default: 20)`.

src/roda-mcp-server/awslabs/roda_mcp_server/server.py
```diff
@@ -0,0 +1,1011 @@
...
+                'error': 'Unexpected error listing bucket',
+                'message': str(e),
+            },
+            indent=2,
+        )
+
+
+@mcp.tool()
+async def sample_dataset(slug: str, file_key: str, bucket_arn: str | None = None) -> str:
+    """Read the first 100KB of a specific file from a public dataset's S3 bucket.
```
scottschreckengaust: **Docstring mismatch.** Says it reads "the first 100KB" and returns "file content (text)," but text content is further truncated to 2000 chars (line ~838). Also `is_partial` (line 834) reflects only the 100KB byte cap, so a sub-100KB file shows `is_partial: false` even when its returned text was truncated at 2000 chars. Document both limits and have `is_partial`/the docstring reflect the display truncation.

src/roda-mcp-server/awslabs/roda_mcp_server/server.py
```diff
@@ -0,0 +1,1011 @@
...
+    stats = _knowledge_base.get_statistics()
+
+    return json.dumps(stats, indent=2)
+
+
+@mcp.tool()
+async def preview_dataset(slug: str, bucket_arn: str | None = None) -> str:
+    """Show the S3 bucket structure for a public dataset (no AWS account required).
+
+    Lists the first 10 files in the dataset's S3 bucket using anonymous access.
```
scottschreckengaust: **Docstring mismatch.** "Lists the first 10 files" — the code requests `MaxKeys: 10` (first 10 keys lexicographically) and then re-sorts *those 10* newest-first by `LastModified`. So it's neither simply the first 10 nor the 10 most-recently-modified across the bucket. Clarify: "up to 10 objects (the first 10 keys returned by S3), sorted newest-first."

src/roda-mcp-server/awslabs/roda_mcp_server/server.py
```diff
@@ -0,0 +1,1011 @@
...
+
+    except ClientError as e:
+        error_code = e.response.get('Error', {}).get('Code', '')
+        if error_code == 'AccessDenied':
+            return json.dumps(
+                {
+                    'dataset': dataset.get('Name', ''),
+                    'slug': slug,
+                    'error': 'Access denied — this dataset requires AWS credentials.',
+                    'note': 'Use sample_dataset with your credentials, or contact the dataset provider.',
```
scottschreckengaust: **Misleading guidance.** On `AccessDenied` this tells the user to "use sample_dataset with your credentials," but `sample_dataset` uses `signature_version=UNSIGNED` and never uses caller credentials — it will hit the same `AccessDenied`. Reword to point at the AWS CLI with the user's own credentials, consistent with the requester-pays message above.

src/roda-mcp-server/awslabs/roda_mcp_server/knowledge_base.py
```diff
@@ -0,0 +1,144 @@
...
+            # Index managed by
+            managed_by = dataset.get('ManagedBy', '').lower()
+            if managed_by:
+                self.managed_by_index[managed_by].append(idx)
+
+            # Index license
+            license_text = dataset.get('License', '').lower()
+            if license_text:
+                # Extract common license types
+                if 'cc' in license_text or 'creative commons' in license_text:
```
scottschreckengaust, line 60: **Suggestion — `discover_by_license` silently returns nothing for unsupported licenses.** `license_index` only ever contains the four hardcoded buckets built here (`creative commons`, `mit`, `apache`, `public domain`). A call with `license_type='bsd'` or `'gpl'` returns zero results with no hint that only those four are supported. Document the supported values, or return the available keys on a no-match.

scottschreckengaust (commented):
## Comprehensive review — RODA MCP server

Reviewed the production source (`server.py`, `knowledge_base.py`, `pyproject.toml`) across five aspects: general quality, test coverage, silent failures, comment/docstring accuracy, and type design. No regression risk (all net-new code). Overall well-structured with strong test coverage — inline comments below mark the specific findings.

### Critical
1. Zero-byte file crashes `sample_dataset` (`Range=bytes=0--1`).
2. `print()` to stdout corrupts the stdio JSON-RPC stream — use structured logging.

### Important
3. Broad `except Exception` masks real bugs (preview + sample).
4. Checksum retry only retries mismatches, not network errors; integrity failures aren't logged.
5. `expected_checksum` parse can raise uncaught `IndexError`.
6. Several docstring/behavior mismatches (limit default, 100KB vs 2000-char, "first 10 files", AccessDenied note).

### Suggestions
7. Fold cache globals into one atomically-updated object + `asyncio.Lock`; consider typed `Dataset`/`Resource` models for the security-relevant access flags.
8. `discover_by_license` silently returns nothing for unsupported license families.
9. `pyproject.toml` metadata/naming/`NOTICE` mismatch.
10. Test gaps: checksum fail-then-succeed, zero-byte sampling, multi-bucket wrong-ARN, provider-fill loop.

### Strengths
Cache written only after checksum validation; specific `ClientError` messages; checksum tests compute real SHA-256; anonymous access correctly scoped; controlled-access/requester-pays buckets filtered out.

src/roda-mcp-server/awslabs/roda_mcp_server/server.py
```diff
@@ -0,0 +1,1011 @@
...
+                'arn': arn,
+            },
+            indent=2,
+        )
+
+    # Anonymous LIST only — no data downloaded.
+    try:
+        s3 = boto3.client(
+            's3',
+            config=Config(signature_version=UNSIGNED, s3={'use_ssl': True}),
```
scottschreckengaust: Add `user-agent-extra` https://github.com/awslabs/mcp/blob/02cdc86ccaeae0a65999c672336b80369842018a/src/iam-mcp-server/awslabs/iam_mcp_server/aws_client.py#L42

src/roda-mcp-server/awslabs/roda_mcp_server/server.py
```diff
@@ -0,0 +1,1011 @@
...
+                'dataset': dataset.get('Name', ''),
+                'slug': slug,
+                'error': 'Could not parse S3 bucket name from ARN',
+                'arn': arn,
+            },
+            indent=2,
+        )
+
+    try:
+        s3 = boto3.client(
```
scottschreckengaust: consider a helper function to configure the boto3 clients uniformly

src/roda-mcp-server/pyproject.toml
```diff
@@ -0,0 +1,43 @@
+[project]
+name = "roda-mcp-server"
```
scottschreckengaust:
```suggestion
name = "awslabs.roda-mcp-server"
```

scottschreckengaust (changes requested):
A few changes, thanks for the contributions

src/roda-mcp-server/awslabs/roda_mcp_server/server.py
guyu-ye: updated and added `content_truncated` for the 2000 chars limit

src/roda-mcp-server/awslabs/roda_mcp_server/server.py
guyu-ye: updated note to use AWS CLI with credentials

src/roda-mcp-server/awslabs/roda_mcp_server/server.py
guyu-ye: added helper

src/roda-mcp-server/docs/architecture.drawio.png
scottschreckengaust, line 1: raw draw.io file better than image

src/roda-mcp-server/docs/basic-usage.gif
scottschreckengaust, line 1: use the examples as text, images go stale too fast

src/roda-mcp-server/awslabs/roda_mcp_server/aws_client.py
```diff
@@ -0,0 +1,51 @@
...
+        Configured S3 client with anonymous access
+
+    Raises:
+        Exception: If client creation fails
+    """
+    try:
+        config = Config(
+            signature_version=UNSIGNED,
+            s3={'use_ssl': True},
+            user_agent_extra=f'md/awslabs#mcp#iam-mcp-server#{__version__}',
```
scottschreckengaust: 🔴 **Wrong server name in user-agent (copy-paste from IAM server).** All RODA S3 traffic is attributed to `iam-mcp-server` in AWS telemetry.

```suggestion
            user_agent_extra=f'md/awslabs#mcp#roda-mcp-server#{__version__}',
```

src/roda-mcp-server/awslabs/roda_mcp_server/knowledge_base.py
```diff
@@ -0,0 +1,148 @@
...
+        """Build all indexes from dataset list."""
+        self.datasets = datasets
+        self.tag_index.clear()
+        self.managed_by_index.clear()
+        self.license_index.clear()
+        self.resource_type_index.clear()
+
+        for idx, dataset in enumerate(datasets):
+            # Index tags
+            for tag in dataset.get('Tags', []):
```
scottschreckengaust: 🔴 **Crashes the whole catalog on one malformed record.** `dataset.get('Tags', [])` only defaults when the key is *absent*; a present `"Tags": null` yields `None` (→ `TypeError` on iteration), and `"Tags": [null]` or a non-string tag makes `tag.lower()` raise `AttributeError`. Because this runs inside `fetch_datasets` → `build_indexes`, one bad line in the external NDJSON breaks **every** KB-backed tool (`discover_*`, `find_related_datasets`, `get_knowledge_base_stats`) for all datasets, and the failure is cached for 24h.

Same unguarded-`.lower()` / present-but-`null` pattern also affects `ManagedBy` (line 47) and `License` (line 52) here, and `server.py` lines 232, 236, 239, 248-250, 343 and the `Description` slices at 266/351/436/521. Note `search_stac_endpoints` (server.py:1122-1123) already guards with `or ''` — worth applying that consistently, or coercing/validating field types once at ingest in `fetch_datasets`.

```suggestion
            for tag in dataset.get('Tags') or []:
                if isinstance(tag, str):
                    self.tag_index[tag.lower()].append(idx)
```

src/roda-mcp-server/awslabs/roda_mcp_server/server.py
```diff
@@ -0,0 +1,1165 @@
...
+        obj = s3.get_object(
+            Bucket=bucket_name,
+            Key=file_key,
+            Range=f'bytes=0-{bytes_to_read - 1}',
+        )
+        content = obj['Body'].read()
+        is_partial = file_size > max_bytes
+
+        try:
+            text = content.decode('utf-8')
```
scottschreckengaust, line 1051: 🔴 **Valid UTF-8 files >100KB can be mislabeled as binary.** The byte-range read stops at a fixed offset (`bytes_to_read - 1`), which can split a multibyte character, so strict `decode('utf-8')` raises `UnicodeDecodeError` and the `except` branch reports `[Binary file — cannot display as text]` for a perfectly readable text file. This misfires precisely on the >100KB files that "sample" is for.

```suggestion
            text = content.decode('utf-8', errors='replace')
```

(Or trim the trailing incomplete UTF-8 sequence before a strict decode if you'd rather not introduce replacement characters.)

src/roda-mcp-server/awslabs/roda_mcp_server/server.py
```diff
@@ -0,0 +1,1165 @@
...
+
+async def fetch_datasets() -> list[dict[str, Any]]:
+    """Fetch and cache datasets from RODA.
+
+    Uses the official NDJSON index file which contains all datasets pre-parsed.
+    """
+    global _datasets_cache, _cache_timestamp
+
+    # Check if cache is valid
+    if _datasets_cache is not None and _cache_timestamp is not None:
```
scottschreckengaust, line 77: 🟠 **No synchronization around the module-level cache.** Tools are async and run concurrently; `fetch_datasets` `await`s network I/O between this cache check and the cache write (lines 173-177), so N concurrent cold calls each perform a full download + `build_indexes()`. Worse, KB readers call `await fetch_datasets()` then read `_knowledge_base` in a *separate* step — a concurrent rebuild (`clear()` then repopulate in `build_indexes`) can expose a half-built index and produce wrong or out-of-range lookups.

Suggest an `asyncio.Lock` around the fetch/populate critical section (re-check cache validity inside the lock), and build a fresh `DatasetKnowledgeBase` then swap the reference atomically rather than mutating in place, so readers never observe a partial rebuild.

src/roda-mcp-server/awslabs/roda_mcp_server/server.py
```diff
@@ -0,0 +1,1165 @@
...
+                computed_checksum = hashlib.sha256(content_bytes).hexdigest().lower()
+                if computed_checksum == expected_checksum:
+                    break
+
+                # Checksum mismatch — log and retry
+                logger.warning(
+                    f'Checksum mismatch on attempt {attempt + 1}/{max_retries + 1}: '
+                    f'expected={expected_checksum}, computed={computed_checksum}'
+                )
+                last_error = None  # Not a network error, just a mismatch
```
scottschreckengaust: 🟠 **Retry loop conflates checksum mismatch with network errors.** `last_error` is a single reused variable; a mismatch resets it to `None` here, so whichever failure class lands *last* determines the reported error. A genuine tampering/corruption event can be reported as a benign network error (or vice-versa), and the `expected`/`computed` values printed in the exhaustion message (lines 127-133) may be stale or `None` from a different attempt. Track the two failure classes independently (e.g. `last_network_error` + a `saw_checksum_mismatch` flag) and report both — this is the one security-relevant path, so the misclassification matters.

src/roda-mcp-server/awslabs/roda_mcp_server/server.py
```diff
@@ -0,0 +1,1165 @@
...
+                indent=2,
+            )
+
+        objects = [
+            {
+                'key': obj['Key'],
+                'size_bytes': obj['Size'],
+                'last_modified': obj['LastModified'].isoformat(),
+            }
+            for obj in sorted(
```
scottschreckengaust: 🟠 **"Newest-first" is structurally false for buckets with >10 keys.** `MaxKeys=10` (line 708) fetches the 10 *lexicographically-first* keys S3 returns — for date-partitioned prefixes these are typically the **oldest** — and only those 10 are then sorted by `LastModified`. Users are told they see the newest objects but see the oldest. Either fix the behavior (e.g. don't imply recency) or, at minimum, correct the docstring (lines 553-556), which is also self-contradictory ("first 10 keys in lexicographic order" *and* "sorted newest-first").

src/roda-mcp-server/awslabs/roda_mcp_server/server.py
```diff
@@ -0,0 +1,1165 @@
...
+                    'list_recursive': f'aws s3 ls s3://{bucket_name}/{prefix} --recursive --no-sign-request',
+                },
+            },
+            indent=2,
+        )
+
+    except ClientError as e:
+        error_code = e.response.get('Error', {}).get('Code', '')
+        if error_code == 'AccessDenied':
+            return json.dumps(
```
scottschreckengaust, line 823: 🟠 **Errors are returned as success-shaped JSON strings (cross-cutting, all 10 tools).** Every tool is `-> str` and encodes failures as an `{"error": ...}` key alongside normal fields (`dataset`, `slug`, `license`). MCP has a first-class error channel (raise inside the tool / `isError`) that isn't used, so from the transport's view an access-denied error and a successful listing are both "successful tool calls returning text" — and because payloads share fields, an LLM can present this AccessDenied error as a partial result. Consider raising for true failures, or adding a mandatory `"status": "ok" | "error"` discriminator to every response and documenting it in each docstring.

src/roda-mcp-server/awslabs/roda_mcp_server/server.py
```diff
@@ -0,0 +1,1165 @@
...
+                {
+                    'dataset': dataset.get('Name', ''),
+                    'slug': slug,
+                    'error': f'Bucket ARN not found among public buckets for this dataset: {bucket_arn}',
+                    'available_arns': [r.get('ARN', '') for r in s3_resources],
+                },
+                indent=2,
+            )
+    else:
+        s3_resource = s3_resources[0]
```
scottschreckengaust, line 997: 🟠 **Silently defaults to `s3_resources[0]` on multi-bucket datasets.** Unlike `preview_dataset` (which asks the user to choose when there are multiple public buckets), `sample_dataset` picks bucket 0 here. If a user previews and selects bucket B, then calls `sample_dataset` without `bucket_arn`, this reads bucket A → spurious `NoSuchKey` "File not found" even though the file exists in B. Make the two tools behave consistently (prompt for a choice, or at least surface which bucket was used).

src/roda-mcp-server/awslabs/roda_mcp_server/server.py
```diff
@@ -0,0 +1,1165 @@
...
+        else:
+            msg = 'No publicly accessible S3 bucket found for this dataset.'
+        result = {
+            'dataset': dataset.get('Name', ''),
+            'slug': slug,
+            'message': msg,
+            'available_resource_types': all_types,
+        }
+        if controlled_urls:
+            result['access_request_url'] = controlled_urls[0]
```
scottschreckengaust, line 698: 🟠 **`ControlledAccess: true` leaks a non-URL value as the access URL.** The public-bucket filter above is fine, but when `ControlledAccess` is a truthy non-string (e.g. boolean `true`, which the registry may use to flag controlled access without a link), `controlled_urls` becomes `[True]` and the user is handed `access_request_url: true`. Guard for a non-empty string URL before using it, and fall back to a message otherwise.

src/roda-mcp-server/pyproject.toml
```diff
@@ -0,0 +1,49 @@
+[project]
+name = "awslabs.roda-mcp-server"
+version = "0.1.0"
+description = "MCP server for discovering and exploring AWS Registry of Open Data datasets"
+readme = "README.md"
+requires-python = ">=3.10"
+dependencies = [
+    "fastmcp>=0.1.0",
```
scottschreckengaust: 🟠 **Dependency floors diverge from the monorepo.** `fastmcp>=0.1.0` is dangerously loose — sibling servers pin `>=2.13.x`, and `>=0.1.0` permits resolving an API-incompatible ancient release. Also `pytest>=9.0.0` (line 18) has no stable release; siblings use `>=7.4.0`. Consider also adding the standard `authors`, `[project.urls]`, `classifiers`, `[tool.ruff.lint]`, `[tool.pyright]` sections other servers carry.

```suggestion
    "fastmcp>=2.13.1",
```

Separately: `server.py` never calls `logger.remove()`/`logger.add(sys.stderr, level=os.getenv('FASTMCP_LOG_LEVEL', ...))`, so the `FASTMCP_LOG_LEVEL` env var advertised in the README has no effect. Sibling servers configure loguru at import — worth matching.

src/roda-mcp-server/awslabs/roda_mcp_server/server.py
```diff
@@ -0,0 +1,1165 @@
...
+    Args:
+        organization: Organization name or partial name (e.g., 'NASA', 'NOAA', 'NIH')
+        limit: Maximum number of results to return (default: 10)
+
+    Returns:
+        JSON string with matching datasets
+    """
+    await fetch_datasets()  # Verify KB is built
+
+    results = _knowledge_base.search_by_organization(organization)[:limit]
```
scottschreckengaust, line 474: 🟡 **Unclamped `limit` used as a slice.** A negative `limit` here (and at lines 466, 511, plus the `>= limit` early-breaks in `search_datasets`/`list_datasets`) slices as `[:-1]`, silently dropping the last result rather than erroring or bounding. `search_stac_endpoints` already does `limit = max(1, min(limit, 100))` (line 1076) — apply the same clamp uniformly across tools.

src/roda-mcp-server/awslabs/roda_mcp_server/server.py
```diff
@@ -0,0 +1,1165 @@
...
+    results = _knowledge_base.search_by_license(license_type)[:limit]
+
+    if not results:
+        return json.dumps(
+            {
+                'license_type': license_type,
+                'count': 0,
+                'datasets': [],
+                'hint': f'No datasets found for license type: {license_type!r}.',
+                'supported_license_types': list(_knowledge_base.license_index.keys()),
```
scottschreckengaust: 🟡 **`supported_license_types` is data-derived and can contradict the docstring.** This returns `license_index.keys()`, which only contains categories that at least one current dataset matched. If no dataset currently carries e.g. an MIT license, the hint omits `'mit'` while the docstring (lines 455-456) advertises it as supported. Consider returning the fixed set of recognized categories the indexer knows about (see `knowledge_base.py:55-62`) so the hint and docs agree.

src/roda-mcp-server/awslabs/roda_mcp_server/server.py
```diff
@@ -0,0 +1,1165 @@
...
+            result_copy = {k: v for k, v in match.items() if k != 'tags'}
+            results.append(result_copy)
+            provider_counts[provider] = 1
+            if len(results) >= limit:
+                break
+
+    if len(results) < limit:
+        for match in all_matches:
+            result_copy = {k: v for k, v in match.items() if k != 'tags'}
+            if result_copy not in results:
```
scottschreckengaust: 🟡 **O(n²) dedup that can under-fill results.** `result_copy not in results` does a linear full-dict comparison per candidate. Two distinct matches with identical projected fields (e.g. both missing `Slug`, so `slug=''`) are treated as equal and one is dropped, returning fewer than `limit` even when more distinct datasets matched. Track seen slugs in a `set` instead.

src/roda-mcp-server/awslabs/roda_mcp_server/knowledge_base.py
```diff
@@ -0,0 +1,148 @@
...
+# limitations under the License.
+
+
+"""Knowledge base for RODA metadata."""
+
+from collections import defaultdict
+from typing import Any
+
+
+class DatasetKnowledgeBase:
```
scottschreckengaust, line 22: 🟡 **Type-design suggestion.** The four indexes are public, mutable, and hold positional indices into a sibling `self.datasets` list — reassigning `datasets` without rebuilding (or a mid-rebuild read, see the concurrency note in server.py) yields silently-wrong datasets rather than an error. Consider making the fields private and adding a `from_datasets(datasets)` classmethod so "if you hold one, it's coherent" becomes a type-level guarantee. Also, `defaultdict` on read paths can silently insert empty keys and pollute `get_statistics`/`supported_license_types` — freeze to a plain `dict` after building. Sibling servers commonly use Pydantic models for the record/response shapes, which would also remove the scattered `.get('Field','')` projections.

scottschreckengaust (dismissed):
## Review — `feat: add RODA MCP server`

Thanks for this — it's a well-built server: checksum-validated fetch, anonymous `UNSIGNED` S3 access, specific exception layering, and strong test coverage (82 tests, 97%). Reviewed with an automated multi-agent pass (code quality, silent-failure, tests, type-design, comments) plus a line-by-line correctness sweep. A few things I'd like addressed before merge; none are large.

### 🔴 Critical
- **Wrong user-agent** — `aws_client.py` reports this server as `iam-mcp-server` (copy-paste). See inline.
- **Crashes on `null`/non-string external fields** — `dict.get('X','')` does NOT substitute the default when the value is present-but-`null`, so the following `.lower()`/`len()`/slice raises. Worst case is `build_indexes`, which breaks every tool for the whole catalog on one bad record. See inline on `knowledge_base.py:43`.
- **Large valid UTF-8 files mislabeled as binary** in `sample_dataset` — fixed byte-range + strict decode. See inline.

### 🟠 Important
- No `asyncio.Lock` around `fetch_datasets` — concurrent cold calls race + can read a half-rebuilt knowledge base.
- Retry loop conflates checksum-mismatch (security-relevant) with network errors.
- `preview_dataset` "newest-first" is structurally false for buckets with >10 keys.
- All 10 tools return errors as success-shaped JSON strings — the LLM can't reliably tell failure from success.
- `sample_dataset` silently defaults to bucket[0] on multi-bucket datasets (unlike `preview_dataset`).
- `ControlledAccess: true` (boolean) leaks `access_request_url: true`.
- `pyproject.toml` diverges from repo conventions (`fastmcp>=0.1.0`, `pytest>=9.0.0`), and `FASTMCP_LOG_LEVEL` from the README is never wired up.

See inline comments for specifics and suggested fixes.

### 🟡 Suggestions
- Clamp `limit` uniformly (only `search_stac_endpoints` does).
- `discover_by_license` supported list is data-derived and can contradict the docstring's fixed list.
- `DatasetKnowledgeBase` uses public mutable parallel indexes — consider private fields + a `from_datasets()` constructor, and typed response DTOs to match sibling servers.

### ✅ Test gaps worth closing
- Concurrent `fetch_datasets` (documents the race).
- `sample_dataset` 2000-char text truncation (`content_truncated`/`is_partial`) is **entirely untested**.
- `sample_dataset` multi-bucket + wrong-ARN (mirror the `preview` tests).
- Network-error (`EndpointConnectionError`) path for both S3 tools.
- Cache *expiry* (stale timestamp → refetch).
- `test_long_description_truncation` asserts under `if long_result:` so it can silently pass with zero assertions; duplicated `SAMPLE_DATASETS` fixtures in `conftest.py` vs `test_knowledge_base.py`.

src/roda-mcp-server/awslabs/roda_mcp_server/knowledge_base.py
guyu-ye: will address in next release

arnewouters (approved):

scottschreckengaust (approved):
LGTM

PR #4039, opened 2026-06-25
fix(billing-cost-management): fix SQL-offload and pagination paths for Cost Explorer operations

wangyuhere:
## Summary

Fixes the SQL-offload and pagination paths for the Cost Explorer operations in `billing-cost-management-mcp-server`. 

### Changes

Four commits, all on the same theme:

**SQL offload — dispatch and size gate** (`241bf9a1`)
- Fix dispatch mismatch in `_get_specialized_converter`: keys were
  snake_case but call sites passed camelCase (`getCostAndUsage`,
  `getPricing`), so every lookup returned `None` and all typed-schema
  branches (`cost_and_usage`, `pricing_products`, `dimension_values`,
  `forecast`, `tags`, `cost_categories`) were unreachable. Responses
  fell through to the generic flatten branch and produced a
  `(key TEXT, value TEXT)` table with the entire `ResultsByTime` /
  `PriceList` stuffed into a single stringified-JSON cell.
- Wire `convert_response_if_needed` size gate so small responses
  (median ~11 rows in real evals) are returned inline instead of being
  round-tripped through SQLite. The "fall back to inline response"
  branches at the call sites had been dead code.

**Pagination** (`01b7f598`)
- Honor agent-supplied `next_token`. `paginate_aws_response` started
  with `current_token=None` and only set the token on iteration 2+, so
  the caller value was lost on iteration 1.
- Preserve `NextPageToken` when a response is offloaded to SQL
  (previously dropped).

**Extend SQL offload + pagination to remaining Cost Explorer ops** (`b3a9f935`)
- Wire the size gate for the Cost Explorer ops whose typed converters
  existed but were never dispatched: `getCostAndUsageWithResources`,
  `getDimensionValues`, `getCostForecast`, `getUsageForecast`,
  `getTags`, `getCostCategories`.
- Add `next_token` / `max_pages` pagination to
  `get_cost_and_usage_with_resources`.

**Cost categories sample queries** (`3e75f4c7`)
- The `cost_categories` typed branch creates a table with
  `category_type` / `category_value` columns. The sample-queries
  section had no matching branch and fell through to the generic
  `else` branch, which emitted queries referencing nonexistent
  `key` / `value` columns. Add a `cost_categories` branch with two
  queries that target the real schema.

### User experience

**Before**
- SQL offload was net-negative for the agent. Every offloaded response
  produced a `(key, value)` table with the entire payload stringified
  into one cell — the "specialized" sample queries returned to the
  agent referenced columns that did not exist.
- Even tiny responses got round-tripped through SQLite, forcing an
  extra `session-sql` call to read data that could have been inline.
- Pagination did not work beyond page 1. Agents could not retrieve
  the rest of a paginated result set on any Cost Explorer operation,
  and `NextPageToken` was lost whenever offload triggered.

**After**
- Each Cost Explorer op offloads only when the response exceeds the
  size threshold, and the offloaded data lands in a properly-typed
  table (`time_period_start`, `metric_name`, `amount`, etc. for cost
  and usage; `category_type`, `category_value` for cost categories;
  and so on).
- Small responses come back inline — no extra round-trip.
- Pagination works on all paginated Cost Explorer operations, both
  inline and offloaded. The agent can supply `next_token` and walk
  through pages.

## Checklist

* [X] Changes are documented

Is this a breaking change? N

**RFC issue number**:

Checklist:

## Acknowledgment

### DIFF
--- a/src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/cost_explorer_operations.py
+++ b/src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/cost_explorer_operations.py
@@ -172,6 +172,8 @@ async def get_cost_and_usage_with_resources(
     metrics: Optional[str] = None,
     group_by: Optional[str] = None,
     filter_expr: Optional[str] = None,
+    next_token: Optional[str] = None,
+    max_pages: Optional[int] = None,
     billing_view_arn: Optional[str] = None,
 ) -> Dict[str, Any]:
     """Get resource-level cost and usage data.
@@ -187,6 +189,8 @@ async def get_cost_and_usage_with_resources(
         metrics: List of metrics as JSON string
         group_by: Optional grouping as JSON string
         filter_expr: Optional filters as JSON string
+        next_token: Pagination token
+        max_pages: Maximum number of pages to fetch
         billing_view_arn: Optional ARN of a billing view to scope the query.
             If not provided, defaults to the account's primary billing view.
 
@@ -230,11 +234,59 @@ async def get_cost_and_usage_with_resources(
         if billing_view_arn:
             request_params['BillingViewArn'] = billing_view_arn
 
-        # Make API call
-        await ctx.info('Calling getCostAndUsageWithResources API')
-        response = ce_client.get_cost_and_usage_with_resources(**request_params)
+        # Handle pagination — same NextPageToken contract as getCostAndUsage.
+        if next_token or max_pages:
+            # paginate_aws_response only sets the token on iter 2+, so inject
+            # the caller's token into the first boto3 request explicitly.
+            if next_token:
+                request_params['NextPageToken'] = next_token
 
-        return format_response('success', response)
+            results, pagination_metadata = await paginate_aws_response(
+                ctx,
+                'getCostAndUsageWithResources',
+                lambda **params: ce_client.get_cost_and_usage_with_resources(**params),
+                request_params,
+                'ResultsByTime',
+                'NextPageToken',
+                'NextPageToken',
+                max_pages,
+            )
+
+            response = {'ResultsByTime': results, 'Pagination': pagination_metadata}
+            pagination_envelope = {
+                'next_page_token': pagination_metadata.get('next_token'),
+                'has_more': pagination_metadata.get('has_more', False),
+                'pages_fetched': pagination_metadata.get('pages_fetched'),
+            }
+        else:
+            await ctx.info('Calling getCostAndUsageWithResources API')
+            response = ce_client.get_cost_and_usage_with_resources(**request_params)
+            pagination_envelope = {
+                'next_page_token': response.get('NextPageToken'),
+                'has_more': bool(response.get('NextPageToken')),
+                'pages_fetched': 1,
+            }
+
+        # Forward through the SQL offload — same ResultsByTime shape as
+        # getCostAndUsage, so the cost_and_usage typed converter handles it.
+        table_response = await convert_response_if_needed(
+            ctx,
+            response,
+            'cost_explorer_get_cost_and_usage_with_resources',
+            granularity=granularity,
+            start_date=start,
+            end_date=end,
+            group_by=group_by,
+            metrics=metrics,
+            **pagination_envelope,
+        )
+
+        return format_response(
+            'success',
+            table_response
+            if isinstance(table_response, dict) and 'data_stored' in table_response
+            else response,
+        )
 
     except Exception as e:
         return await handle_aws_error(ctx, e, 'getCostAndUsageWithResources', 'Cost Explorer')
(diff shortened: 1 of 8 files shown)


src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/cost_explorer_operations.py
```diff
@@ -300,11 +367,39 @@ async def get_dimension_values(
...
+            end_date=end,
+            search_string=search_string,
+            **pagination_envelope,
+        )
+
+        return format_response(
+            'success',
+            table_response
+            if isinstance(table_response, dict) and 'data_stored' in table_response
+            else response,
```
finklen-amazon: Nitpick: maybe we should inject this into the `format_response` (or its own utility) because we're doing this everywhere?

finklen-amazon (dismissed):

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/cost_explorer_operations.py
wangyuhere: Yes, good suggestion, I will move it into `format_response`

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/cost_explorer_operations.py
```diff
@@ -216,11 +234,59 @@ async def get_cost_and_usage_with_resources(
...
+                lambda **params: ce_client.get_cost_and_usage_with_resources(**params),
+                request_params,
+                'ResultsByTime',
+                'NextPageToken',
+                'NextPageToken',
+                max_pages,
+            )
+
+            response = {'ResultsByTime': results, 'Pagination': pagination_metadata}
+            pagination_envelope = {
```
finklen-amazon: Nickpick (non-blocking): I see this pattern being added a lot in this file. Any clean way to make it reusable?

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/utilities/sql_utils.py
```diff
@@ -339,69 +339,133 @@
...
         'cost_explorer_get_usage_forecast': 'forecast',
         'cost_explorer_get_tags': 'tags',
         'cost_explorer_get_cost_categories': 'cost_categories',
+        'aws_pricing_get_products': 'pricing_products',
     }
 
     return converters.get(operation_name)
 
 
+def _derive_pagination_envelope(
```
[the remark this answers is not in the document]
wangyuhere: False positive. It is the default name of the next page token.

anniewn (approved):

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/cost_explorer_operations.py
```diff
@@ -665,7 +753,8 @@ async def get_savings_plans_utilization(
 
         # Handle pagination
         if next_token or max_pages:
-            # For paginated requests, use the paginate utility
+            if next_token:
+                request_params['NextToken'] = next_token
             results, pagination_metadata = await paginate_aws_response(
                 ctx,
                 'getSavingsPlansUtilization',
```
Comusus, line 760: I don't believe this API is paginated from [AWS API docs](https://docs.aws.amazon.com/aws-cost-management/latest/APIReference/API_GetSavingsPlansUtilization.html). GetSavingsPlansUtilizationDetails is paginated.

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/cost_explorer_operations.py
```diff
@@ -106,7 +106,11 @@ def api_call(**params):
 
         # Use shared pagination utility
         if next_token or max_pages:
-            # For paginated requests, use the paginate utility
+            # paginate_aws_response only sets the token on iter 2+, so inject
+            # the caller's token into the first boto3 request explicitly.
```
Comusus, line 110: Can `paginate_aws_response` accept the current or starting token, so caller doesn't need to pre-inject `request_params['NextPageToken'] = next_token`? The current way is fine, but we will need to be mindful whenever we use paginate_aws_response.

wangyuhere: The `request_params` has all the parameters to the request. Do you mean add another parameter as `current_next_token` to `paginate_aws_response`? 

I think it is the same, we will need to remember to add the token anyway.

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/cost_explorer_operations.py
wangyuhere: Good catch. It worths another PR to cleanup the pagination code for this API. I will create a task.

finklen-amazon (approved):

Comusus (approved):

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/cost_explorer_operations.py
Comusus: Yes, will need to be mindful for both.

PR #4043, opened 2026-06-26
fix(eks-mcp-server): treat blank insight id as absent

### LINKED ISSUE #4036: eks-mcp-server: get_eks_insights treats empty-string insight_id as detail mode and fails
### Describe the bug

`get_eks_insights` chooses between "list all insights" and "describe one insight" using:

    detail_mode = insight_id is not None

This treats an empty string `""` as detail mode, so it calls DescribeInsight with an empty
id and fails. Several MCP clients serialize an *unset* optional argument as an empty string
`""` instead of JSON `null`, so for those clients a plain "list insights" call is impossible
— it always errors — even though the request is logically a list call.

### Expected Behavior

An empty or whitespace-only `insight_id` should be treated as "not provided" → list mode,
consistent with how `category` (`if category:`) and `next_token` (`if next_token:`) are
already handled in the same handler. Listing insights should succeed.

### Current Behavior

Calling get_eks_insights with insight_id="" enters detail mode and calls DescribeInsight
with an empty id, returning:

    Error retrieving insight details: An error occurred (AccessDeniedException) when calling
    the DescribeInsight operation: Unable to determine service/operation name to be authorized

### Reproduction Steps

Call the get_eks_insights tool with an empty-string insight_id:

    { "cluster_name": "my-cluster", "insight_id": "" }

Result: DescribeInsight ... "Unable to determine service/operation name to be authorized".

For comparison, all of these correctly return the list:
    { "cluster_name": "my-cluster" }
    { "cluster_name": "my-cluster", "insight_id": null }

Root cause is in awslabs/eks_mcp_server/insights_handler.py, _get_eks_insights_impl:
    detail_mode = insight_id is not None
Only insight_id uses `is not None`; the other optional params use truthy checks, so
insight_id is the inconsistent one.

### Possible Solution

Use a truthy/strip check so empty or whitespace-only insight_id means list mode:

    detail_mode = bool(insight_id and insight_id.strip())

No behavior change for real ids or None. Happy to open a PR with a parametrized test
covering "" and "   ".

Discovered via an MCP client (OpenClaw) reaching eks-mcp-server over streamable-http. The
client sent insight_id as "" for an unset optional; the same logical call from another MCP
client that sends JSON null works correctly, which isolates the empty-string handling here.

### OS

Ubuntu Linux (EC2, kernel 7.0.0-1006-aws)

### Server

eks-mcp-server

### Server Version

0.1.32 (and current main)

ap-northeast-1

### Other information

Suggested one-line fix and test described above; willing to submit the PR.

### Service quota

cyphercodes:
Fixes #4036

### Changes

- Treat empty or whitespace-only `insight_id` values as absent in `get_eks_insights`.
- Add regression coverage confirming `""` and `"   "` use list mode instead of `DescribeInsight`.

### User experience

Before this change, MCP clients that serialize an unset optional `insight_id` as an empty string could not list EKS insights because the handler entered detail mode and called `DescribeInsight` with a blank id.

After this change, empty or whitespace-only `insight_id` values list insights, while non-empty insight ids continue to fetch insight details.

## Checklist

* [ ] Changes are documented

N

**RFC issue number**:

N/A

Checklist:

## Acknowledgment

### DIFF
--- a/src/eks-mcp-server/awslabs/eks_mcp_server/insights_handler.py
+++ b/src/eks-mcp-server/awslabs/eks_mcp_server/insights_handler.py
@@ -135,10 +135,8 @@ class InsightsHandler:
             # Always use the default EKS client
             eks_client = self.eks_client
 
-            # Determine operation mode based on whether insight_id is provided
-            detail_mode = insight_id is not None
-
-            if detail_mode:
+            # Determine operation mode based on whether a non-empty insight_id is provided
+            if insight_id and insight_id.strip():
                 # Get details for a specific insight
                 return await self._get_insight_detail(
                     ctx, eks_client, cluster_name, insight_id, next_token
--- a/src/eks-mcp-server/tests/test_insights_handler.py
+++ b/src/eks-mcp-server/tests/test_insights_handler.py
@@ -152,6 +152,40 @@ class TestInsightsHandler:
         assert 'Successfully retrieved 2 insights' in result.content[0].text
 
     @pytest.mark.asyncio
+    @pytest.mark.parametrize('insight_id', ['', '   '])
+    async def test_get_eks_insights_empty_insight_id_uses_list_mode(
+        self, mock_context, mock_mcp, insight_id
+    ):
+        """Test empty or whitespace-only insight_id values are treated as absent."""
+        # Create mock AWS client
+        mock_eks_client = MagicMock()
+
+        # Set up mock response for list mode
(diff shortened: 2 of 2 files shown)


srhsrhsrhsrh:
Thanks for the fix! Closing in favor of #4072 which takes the same approach but strips the insight_id before passing it to _get_insight_detail. Both fixes are functionally equivalent, but appreciate you jumping on this quickly.

PR #4044, opened 2026-06-26
fix(mysql-mcp-server): close read-only bypasses (CWE-184) and add security-model docs

happycupofjava:
## Summary

Closes two read-only-mode bypass classes in the `awslabs.mysql-mcp-server`
package and adds a "Security model" documentation section.

### 1. Comment-bypass class (commit 1)

The detector matched regexes against the raw SQL while MySQL treats
`/* ... */`, `--`, and `#` as whitespace. Payloads like `INTO/**/OUTFILE`
and `LOAD/**/DATA INFILE` bypassed the read-only gate.

- `sqlparse.format(strip_comments=True)` normalises the SQL before regex
  matching in `detect_mutating_keywords` and `check_sql_injection_risk`.
- MySQL conditional comments (`/*! ... */`) are rejected against the raw
  SQL pre-strip, because their contents execute on MySQL 5.0+.
- `MUTATING_KEYWORDS` alternation is sorted longest-first so multi-word
  entries (`RENAME TABLE`, `LOAD DATA`, `CREATE FUNCTION`) match before
  their single-word prefixes deterministically.

### 2. Denylist extension and security-sensitive SET (CWE-184, commit 2)

Adds the mutating verbs the previous denylist missed and an always-block
on session variables that disable integrity controls.

- `MUTATING_KEYWORDS` adds `SET`, `CALL`, `PREPARE`, `EXECUTE`,
  `DEALLOCATE`, `HANDLER`, `LOCK`, `LOCK TABLES`, `UNLOCK`,
  `UNLOCK TABLES`, `FLUSH`, `RESET`, `KILL`, `UNINSTALL`.
- `SECURITY_SENSITIVE_VARS` (`sql_log_bin`, `foreign_key_checks`,
  `unique_checks`) is rejected in **both** read-only and write modes via
  a new `SECURITY_SENSITIVE_VAR_PATTERN` that handles every MySQL
  scope-modifier permutation (`SESSION`/`LOCAL`/`GLOBAL`, `@@`-prefix forms)
  and the comma-separated multi-assignment form (`SET @x = 1, sql_log_bin = 0`),
  with the dangerous variable detected in any position.

### 3. Security-model documentation and version bump (commit 3)

- `README.md` adds a "Security model" section: read-only mode is a
  best-effort SQL-text safeguard, not a security boundary. Operators
  should connect with a least-privilege MySQL user / IAM role granting
  only `SELECT`, with example SQL.
- `kiro_power/POWER.md` and `kiro_power/steering/aurora-mysql-mcp.md`
  carry the same caveat.
- `TestTransactionBypassCoverage` pins the existing stacked-queries
  `SUSPICIOUS_PATTERN` against ten canonical transaction-bypass payloads
  (`SELECT 1; COMMIT; INSERT ...` and variants) plus two negative cases.
- Version bump `1.0.21` -> `1.0.22`.

## Testing

- 137 new unit tests in `tests/test_mutable_sql_detector.py`. 100%
  line + branch coverage on `mutable_sql_detector.py`.
- End-to-end run against a live Aurora MySQL Serverless v2 cluster
  (RDS Data API path). Every payload from the reported bypass class
  rejected at the MCP layer before reaching the database; every
  benign comment-bearing query still passes; full backwards-compatibility
  with the existing Workout Planner demo confirmed (Express backend
  pointed at the test cluster, all four CRUD endpoints return seeded
  rows).

## Backwards compatibility

- New runtime dependency: `sqlparse>=0.5.0`.
- The blanket-SET block in read-only mode rejects `SET @var`,
  `SET NAMES`, and `SET sql_mode`. An LLM-driven read flow has
  SQL-native alternatives for all three. If an allowlist is needed
  later, it can be added without breaking the denylist contract.
- Known limitation: `UPDATE t SET sql_log_bin = 0` would
  false-positive on the always-block in write mode. Real-world risk
  is essentially zero (no schema names a column after a MySQL session
  variable). Documented as a deliberate trade-off in
  `TestMultiVariableSetFalsePositiveBoundary`. Closing it requires
  sqlparse tokenisation to distinguish statement-level SET from
  UPDATE's SET clause.

### DIFF
--- a/src/mysql-mcp-server/CHANGELOG.md
+++ b/src/mysql-mcp-server/CHANGELOG.md
@@ -5,6 +5,42 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 1.0.22
+
+### Security
+
+- **CWE-184** — extended the read-only-mode denylist to
+  close gaps reported by external security review. The previous
+  `MUTATING_KEYWORDS` set omitted `SET`, `CALL`, `PREPARE`, `EXECUTE`,
+  `DEALLOCATE`, `HANDLER`, `LOCK`, `LOCK TABLES`, `UNLOCK`,
+  `UNLOCK TABLES`, `FLUSH`, `RESET`, `KILL`, and `UNINSTALL`, so on the
+  RDS Data API path an LLM under prompt injection could fire these
+  statements against the operator's database under what the server
+  presented as a read-only call. All listed statements are now rejected
+  at the MCP layer.
+- Added an always-block list (`SECURITY_SENSITIVE_VARS`) that rejects
+  `SET sql_log_bin`, `SET foreign_key_checks`, and `SET unique_checks`
+  in both read-only and write modes. Flipping these variables disables
+  binary logging / referential integrity / uniqueness for the rest of
+  the session, and an LLM-driven flow should never be able to flip them
+  even when the operator has enabled writes via `--allow_write_query`.
+  Pattern handles the comma-separated multi-assignment form
+  (`SET @x = 1, sql_log_bin = 0`) and every MySQL scope modifier
+  permutation.
+- README adds a **Security model** section: the read-only mode is a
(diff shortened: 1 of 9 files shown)


kennthhz (approved):

scottschreckengaust (approved):
LGTM

PR #4075, opened 2026-07-04
fix(postgres): allow username-only secrets for PG_WIRE_IAM_PROTOCOL

### LINKED ISSUE #4047: postgres-mcp-server: PG_WIRE_IAM_PROTOCOL secret_arn parser still requires unused 'password' field
### Describe the bug

When using `--connection_method PG_WIRE_IAM_PROTOCOL` with `--secret_arn`, startup validation rejects any secret whose `password` field is empty or missing — even though the IAM auth path never reads the password. `server.py:841` discards it (`iam_username, _ = get_credentials_from_secret(...)`), but the shared helper at `psycopg_pool_connection.py:91` blocks the secret before the call site can opt out:

```python
if not password:
    raise ValueError(f'Secret does not contain password. Available keys: ...')
```

Every IAM-only operator has to populate a sentinel string in `password`. It reads like a real DB credential in security reviews and confuses next-readers about whether the password is actually used.

### Expected Behavior

On `PG_WIRE_IAM_PROTOCOL`, a secret containing only `{"username": "..."}` should boot the server.

### Current Behavior

Server exits at startup with:

```
ValueError: Secret does not contain password. Available keys: username
... MCP server cannot start: unable to read Secrets Manager ARN ...
Configured default --secret_arn is unreadable.
```

### Reproduction Steps

1. Create a Secrets Manager secret containing only `{"username":"some_iam_db_user"}`.
2. Start the server:
   ```
   uvx --from awslabs.postgres-mcp-server@latest awslabs.postgres-mcp-server \
     --connection_method PG_WIRE_IAM_PROTOCOL \
     --db_cluster_arn <arn> --db_type APG --region <region> \
     --db_endpoint <reader-or-writer-endpoint> --database <name> \
     --secret_arn <secret-arn>
   ```
3. Server fails at `validate_secret_arn_at_startup` before accepting any tool calls.
4. Re-update the secret to `{"username":"...","password":"<any sentinel string>"}` → boot succeeds.

### OS

macOS 14 (Darwin 24.6.0)

### Server

postgres-mcp-server

### Server Version

1.1.7

us-east-1

billykern:
Fixes #4047

### Changes

`get_credentials_from_secret` gains a `require_password` flag (default `True`, so existing behavior is unchanged). The two IAM-auth paths opt out of it:

* `internal_create_connection` — reads only the username from the secret (the password is a generated IAM token)
* `validate_secret_arn_at_startup` — skips the password requirement when the operator declared `--connection_method PG_WIRE_IAM_PROTOCOL`

Non-IAM connection methods still require a password in the secret, at both startup validation and pool initialization.

### User experience

Before: with `PG_WIRE_IAM_PROTOCOL`, a secret containing only `{"username": "..."}` fails startup validation (`Secret does not contain password`), forcing operators to store a sentinel password that is never used and reads like a real credential in security reviews.

After: a username-only secret boots the server on the IAM path. Secrets for non-IAM methods behave exactly as before.

## Checklist

* [ ] Changes are documented

Is this a breaking change? (Y/N): N

**RFC issue number**:

Checklist:

## Acknowledgment

### DIFF
--- a/src/postgres-mcp-server/awslabs/postgres_mcp_server/connection/psycopg_pool_connection.py
+++ b/src/postgres-mcp-server/awslabs/postgres_mcp_server/connection/psycopg_pool_connection.py
@@ -33,8 +33,8 @@ from typing import Any, Dict, List, Optional, Tuple
 
 
 def get_credentials_from_secret(
-    secret_arn: str, region: str, is_test: bool = False
-) -> Tuple[str, str]:
+    secret_arn: str, region: str, is_test: bool = False, require_password: bool = True
+) -> Tuple[str, Optional[str]]:
     """Fetch database (username, password) from AWS Secrets Manager.
 
     The secret payload is expected to be JSON with ``username``/``user``/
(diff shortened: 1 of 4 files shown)


kennthhz:
I think your PR adds the flexibility to supply auth token under IAM from outside the secret and only get username from secret. Why do you need that? @billykern

billykern:
@kennthhz with IAM auth the password in the secret is never used, but startup still fails without one, so one needs to store a fake/placeholder password. This PR skips that check on the IAM path (to clarify: the server already generates the IAM token on PG_WIRE_IAM_PROTOCOL and ignores the secret's password).

PR #4085, opened 2026-07-06
fix(aws-iac-mcp-server): validate rules_file_path and harden invisible-char filter

kdbrogan:
Fixes

## Summary

Addresses two security findings in aws-iac-mcp-server reported via the AWS Vulnerability Reporting Program.

### Changes

1. `check_cloudformation_template_compliance` tool exposed `rules_file_path` as an LLM-controllable argument that reached `open()` without validation. The parameter never functioned in normal operation (the startup rules cache short-circuited it). so rather than validate an input that shouldn't exist, the entire custom-rules-path code path is removed: `rules_file_path` is dropped from the tool, and `initialize_guard_rules()` now loads the bundled rules file directly. There is no caller-supplied path left to exploit.

2. Incomplete invisible-character filter in sanitizer.filter_unicode_tags, which covered only the Unicode Tags block (U+E0000-E007F). Expand coverage to additional zero-width, bidirectional-control, and joiner ranges used for prompt-injection smuggling, while preserving ordinary whitespace.

Adds regression tests for path traversal, symlink resolution, prefix-collision siblings, and each filtered invisible-character range.

### User experience

> Check_cloudformation_template_complince accepted a rules_file_path arg, which was never used and was silently ignored. We are removing that parameter. Tool now only accepts a template content string

## Checklist

* [X] Changes are documented

**RFC issue number**:

Checklist:

## Acknowledgment


src/aws-iac-mcp-server/awslabs/aws_iac_mcp_server/tools/cloudformation_compliance_checker.py
```diff
@@ -28,29 +29,27 @@
...
+
+    package_dir = os.path.dirname(awslabs.aws_iac_mcp_server.__file__)
+    return Path(package_dir) / 'data' / 'default_guard_rules.guard'
+
+
+def initialize_guard_rules() -> bool:
     """Initialize guard rules and cache remediation mappings on server startup.
 
-    Args:
-        rules_file_path: Path to guard rules file
```
kdbrogan, line 35: As it turns out because of a short circuit we have actually never supported customers providing their own rules and have always used the default rule set. So to address this security issue without adding a new feature I am simply removing this parameter

kdbrogan: Adding this feature can be considered for a follow up PR, but considering we are moving to the AWS MCP / Agent Toolkit I am just keeping the original behavior for now

aemada-aws (approved):

scottschreckengaust (approved):
LGTM

PR #4089, opened 2026-07-06
fix(healthomics): use batchName parameter in StartRunBatch call

markjschreiber:
## Summary
The `start_run_batch` tool in the AWS HealthOmics MCP server passed the batch name to the boto3 `StartRunBatch` API under the key `name`, which is not a valid parameter. The API expects `batchName`. As a result, any call that supplied a batch name failed with a parameter validation error.

## Change
- `tools/run_batch.py`: pass the batch name as `batchName` instead of `name`.
- `tests/test_run_batch.py`: update the assertion in `test_success_with_all_params` to check `batchName`.

## Testing
- `uv run pytest tests/test_run_batch.py` — 54 passed.

## Notes
Verified against the botocore service model: `StartRunBatch` accepts `batchName`, `requestId`, `tags`, `defaultRunSetting`, and `batchRunSettings`.


alxawan (approved):
LGTM!

a-li (approved):
LGTM

PR #4029, opened 2026-06-24
fix(aurora-dsql-mcp-server): strip SQL comments before regex evaluation to prevent bypass

pkale:
## Summary

SQL inline comments (`/**/`) are treated as whitespace by PostgreSQL but not by Python's regex
engine. This allows payloads like `INTO/**/OUTFILE` to bypass `\s+` patterns in the readonly
enforcement and injection detection gates.

## Changes

- Add `sqlparse>=0.5.0` dependency
- Add `_strip_comments()` helper using `sqlparse.format(strip_comments=True)`
- Apply comment stripping in `detect_mutating_keywords`, `check_sql_injection_risk`, and
  `detect_transaction_bypass_attempt`
- In `check_sql_injection_risk`, check patterns against both original SQL (preserves `--`
  injection detection) and stripped SQL (catches `/**/` keyword bypass)
- Add `TestCommentBypass` test class

## Testing

`uv run pytest tests/test_readonly_enforcement.py -k "not readonly_query and not transact"` — 66 tests pass.

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/awslabs/mcp/blob/main/CONTRIBUTING.md) doc
- [x] I have performed a self-review of my own code
- [x] I have added tests that prove my fix is effective
- [x] New and existing unit tests pass locally with my changes

## Is this a breaking change?

N

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the [project license](https://github.com/awslabs/mcp/blob/main/LICENSE).

PR #4092, opened 2026-07-07
fix(aurora-dsql): harden read-only mode against session-state mutation

anwesham-lab:
Fixes

## Summary

Hardens the Aurora DSQL MCP server's read-only mode so that session-state
mutation can no longer bypass the guard or persist across tool calls. Aligns
the SQL classifier in `mutable_sql_detector.py` with the approach already used
by the `postgres-mcp-server` sibling.

### Changes

- **Detect session-state (GUC) mutation.** `BEGIN TRANSACTION READ ONLY` only
  blocks writes to tables, not session state. `readonly_query` / `transact`
  (read-only mode) now reject `SET <name> = ...` / `SET <name> TO ...`,
  `SET SESSION` / `SET LOCAL`, `SET TIME ZONE`, `RESET ...`, and the
  `set_config(...)` function form (including when embedded as a `SELECT`
  subquery).
- **Normalize SQL before matching.** Strip comments and unwrap double-quoted
  identifiers so comment-injection payloads (e.g. `SET/**/search_path = ...`)
  can no longer slip past the classifier.
- **Fix false positives.** Blank single-quoted string-literal contents before
  the mutating-keyword scan so legitimate reads like
  `SELECT 'INSERT INTO' AS action` are no longer wrongly rejected.
- **Block high-blast-radius operations in both modes** as defense-in-depth:
  `pg_sleep` / `pg_sleep_for` / `pg_sleep_until`, `pg_terminate_backend`,
  `pg_read_file`, the advisory-lock and `dblink` families, the security GUCs
  `row_security` / `session_replication_role`, and `COPY ... TO/FROM PROGRAM`.
  Aurora DSQL already rejects most of these at the engine today; they are
  enforced here so a future DSQL feature addition cannot silently reopen a gap.
- **Reset session state after each read-only query.** Issue `RESET ALL` on the
  pooled connection so any session state that a mutation may have set cannot
  persist across MCP tool calls.

### User experience

- **Before:** In read-only mode, a `SET application_name = ...` /
  `SET search_path = ...` was accepted, committed on the pooled connection, and
  persisted into subsequent tool calls for the process lifetime. Some benign
  reads containing a keyword inside a string literal were wrongly rejected.
- **After:** Session-mutating statements are rejected in read-only mode, no
  session state carries across tool calls (`RESET ALL`), and the false-positive
  reads are now allowed. Behavior for ordinary `SELECT` queries is unchanged.

Verified end-to-end against a live Aurora DSQL cluster; 194 unit tests pass.

## Checklist

* [x] Changes are documented

## Acknowledgment


src/aurora-dsql-mcp-server/tests/test_readonly_enforcement.py
```diff
@@ -158,6 +158,185 @@ def test_system_statements(self):
             keywords = detect_mutating_keywords(sql)
             assert 'SYSTEM' in keywords, f"Should detect system keywords in: {sql}"
 
+    def test_detect_session_mutation_set_config(self):
+        """set_config() must be detected in any position, incl. as a subquery.
+
+        These are the exact vectors from the read-only-bypass finding: a
+        read-only transaction permits set_config() because it mutates session
+        (GUC) state rather than table data.
```
Benjscho: nit: remove finding language

Benjscho (dismissed):

src/aurora-dsql-mcp-server/tests/test_readonly_enforcement.py
anwesham-lab: oh yeah, need to fix this in multiple spots

anwesham-lab: Done — removed finding/report/PoC language across all touched files (tests + source comments + commit message + CHANGELOG). Grepped for `finding|report|PoC|spoof|hijack|forensic|CWE|vuln` to confirm nothing remains except legitimate uses (e.g. "compatibility error reporting"). Also rebased onto #4029 and superseded its `sqlparse` comment-stripping with a self-contained literal/dollar-quote-aware normalizer (dependency removed); added a comment explaining why.

Benjscho (dismissed):

src/aurora-dsql-mcp-server/awslabs/aurora_dsql_mcp_server/mutable_sql_detector.py
```diff
@@ -96,6 +95,27 @@
...
 
+# -- Regex for Postgres session-state (GUC) mutation --
+# A `BEGIN TRANSACTION READ ONLY` only blocks writes to database TABLES; it does NOT
+# block changes to session/GUC state. Statements like `SET search_path = ...` and the
+# function form `set_config(...)` mutate session state and are permitted by Postgres
+# inside a read-only transaction. They must be caught here to preserve read-only
+# semantics. Unlike the other regexes, `set_config(` is matched anywhere in the string
+# so it is detected even when embedded as a subquery, e.g.
+# `SELECT set_config('search_path', 'pg_temp', false)`.
+SESSION_MUTATION_REGEX = re.compile(
```
praba2210, line 172: **Read-only mode: `SESSION_MUTATION_REGEX` only catches the assignment shape, not keyword-syntax `SET` (and related statements)**

This regex matches `SET x = y` / `SET x TO y`, but the `SET` forms that use keyword syntax slip through. These all pass in read-only mode on head:

```sql
SET ROLE admin
SET SCHEMA 'evil'          -- documented alias for SET search_path TO 'evil'
SET NAMES 'utf8'
LOCK TABLE t IN ACCESS EXCLUSIVE MODE
NOTIFY ch / LISTEN ch
```

The clearest case: the PR correctly blocks `SET search_path TO 'x'` but accepts `SET SCHEMA 'x'` — same operation, same effect, opposite verdict.

Two notes on impact:
- `SET ROLE` isn't undone by the `RESET ALL` backstop. Postgres marks `role`/`session_authorization` as `GUC_NO_RESET_ALL`, so a `SET ROLE` persists on the pooled connection into later requests even with the new scrub in place.
- Where DSQL rejects some of these at the engine today, they still belong in the same forward-looking defense-in-depth bucket the PR already covers for `dblink` / `pg_read_file` / advisory locks.

This is really one root cause, not several: the PR uses a narrow anchored regex where the `postgres-mcp-server` sibling it cites uses a broad keyword set (`SET|RESET|LOCK|NOTIFY|LISTEN|DISCARD|EXECUTE` + `set_config`) matched on normalized text. The sibling catches every case above; this detector doesn't — so it's materially weaker than the classifier it's aligned to.

**Suggested direction** — mirror the sibling's keyword approach rather than extend the anchored regex. Two cautions:
- Don't drop to a bare `^\s*SET\b` without preserving the `SET TRANSACTION READ ONLY` allow-case — the current `(?!TRANSACTION\b)` lookahead permits it and `test_session_mutation_no_false_positives` treats it as safe. (Note the sibling blocks `SET TRANSACTION` outright, so a straight copy would regress this.)
- `EXECUTE` runs a prepared statement that may be a plain SELECT, so it isn't unconditionally a mutation — the sibling blocks it anyway, but that's a deliberate call worth making explicitly here.

And upgrade the post-query scrub from `RESET ALL` to `DISCARD ALL` (or add an explicit `RESET ROLE`) so a `SET ROLE` can't survive on the pooled connection.

src/aurora-dsql-mcp-server/awslabs/aurora_dsql_mcp_server/mutable_sql_detector.py
```diff
@@ -147,19 +172,317 @@
...
+
+    The SQL is comment-normalized and quoted-identifier-unwrapped first, so a
+    comment wedged between tokens (``SET/**/search_path``) or a double-quoted
+    spelling (``"set_config"(...)``) cannot hide a keyword from the read-only
+    gate, and a keyword that appears only inside a string literal
+    (``SELECT 'INSERT INTO' AS x``) is not mistaken for a real one.
+    """
+    # Normalize comments/quoted identifiers, then blank single-quoted literal
+    # contents so a keyword that is merely string data cannot fire the scan.
+    sql = blank_string_literals(normalize_for_detection(sql))
```
praba2210: Second read-only bypass, same root cause. This passes read-only mode — every detector returns empty:


SELECT E'\'', set_config('search_path', 'pg_temp', false)
Why: set_config(...) is a real session mutation, but the blank_string_literals scanner doesn't handle the \' escape in E'...' strings. It miscounts the quotes and ends up treating set_config('search_path' as if it were inside a string literal — so the SESSION_MUTATION check (which only runs on that blanked text) never sees it.

**Fixes:**
Mirror the sibling — match keywords on the un-blanked text. I verified the existing regex already catches this payload there.
Or allowlist the leading read token.
Either way, don't try to fix the E'\'' parsing itself — keep session detection off the blanked text.

src/aurora-dsql-mcp-server/awslabs/aurora_dsql_mcp_server/mutable_sql_detector.py
```diff
@@ -147,19 +172,317 @@
...
+    'dblink_get_connections',
+}
+
+# Match the function name, optionally schema-qualified
+# (pg_catalog.pg_terminate_backend(...)), with any whitespace before the
+# opening paren. Word boundaries prevent matches on identifiers like
+# my_pg_sleep_helper. Alternation is sorted longest-first so prefix-
+# overlapping names (pg_advisory_lock / pg_advisory_lock_shared) report the
+# longer, more specific match.
+DANGEROUS_FUNCTION_PATTERN = re.compile(
```
praba2210, line 347: **New `dangerous_function` / `security_sensitive_guc` checks run on unblanked text → false-positive on function names inside string literals**

These checks (`DANGEROUS_FUNCTION_PATTERN`, `SECURITY_GUC_PATTERN`, `SECURITY_SET_CONFIG_PATTERN`) run on `normalized_sql` with string literals **intact**, so a function/GUC name that is merely *data* is rejected. Realistic read-only queries that now fail:

```sql
SELECT proname FROM pg_proc WHERE prosrc LIKE '%dblink(%'             -- audit: where is dblink used?
SELECT * FROM audit_log WHERE query ILIKE '%SET row_security = off%'
```

This is the same false-positive class the PR deliberately fixes for the keyword tier (`DROP`/`TRUNCATE`) via `keyword_blanked_sql` — it just wasn't applied to these newer checks.

**Fix:** run these against the literal-blanked text too. Verified safe — a real `dblink_connect('x')` still matches after blanking, while `LIKE '%dblink(%'` no longer does. (`set_config(` and `COPY…PROGRAM` are anchored on call/statement shape, so they don't need this; only the bare-name patterns do.)

src/aurora-dsql-mcp-server/awslabs/aurora_dsql_mcp_server/mutable_sql_detector.py
anwesham-lab: Good catch, all valid — fixed. Replaced the assignment-shape `SESSION_MUTATION_REGEX` with a broad keyword approach matched at statement start (mirroring the sibling): `SET` (any target except `SET TRANSACTION`, so `SET ROLE`/`SET SCHEMA`/`SET NAMES`/`SET SESSION AUTHORIZATION` are all caught), plus `RESET`/`DISCARD`/`LISTEN`/`NOTIFY`/`UNLISTEN`/`LOCK`/`EXECUTE`. Kept the `(?!TRANSACTION\b)` allow-case so `SET TRANSACTION READ ONLY` still passes (verified against `test_session_mutation_no_false_positives`). On the scrub: you're right that `RESET ALL` doesn't clear `role` (GUC_NO_RESET_ALL) — and `DISCARD ALL` is `FeatureNotSupported` on DSQL (verified live), so I issue `RESET ALL` **and** `RESET ROLE` (the latter also restores `session_authorization`). `EXECUTE` blocked deliberately, matching the sibling.

src/aurora-dsql-mcp-server/awslabs/aurora_dsql_mcp_server/mutable_sql_detector.py
anwesham-lab: Confirmed and fixed. Root cause as you described: session detection ran on the literal-blanked text, and `blank_string_literals` doesn't model `E'\''` backslash escapes. Fix mirrors your first suggestion — the statement-anchored checks (DDL/PERMISSION/SYSTEM/SESSION_MUTATION/TRANSACTION_CONTROL) now run on the **normalized (un-blanked)** text; only the un-anchored bare-keyword scan (`MUTATING_PATTERN`) uses the blanked text. Anchoring means a keyword inside a literal is never at statement start, so no false positive and no dependency on blanking. Added a regression test with the exact `SELECT E'\'', set_config(...)` payload.

src/aurora-dsql-mcp-server/awslabs/aurora_dsql_mcp_server/mutable_sql_detector.py
anwesham-lab: Valid, fixed. `DANGEROUS_FUNCTION_PATTERN` and `SECURITY_GUC_PATTERN` now run against the literal-blanked text, so `LIKE '%dblink(%'` and `ILIKE '%SET row_security%'` are no longer flagged while a real `dblink_connect('x')` / `SET row_security = off` still matches. As you noted, `SECURITY_SET_CONFIG_PATTERN` and `COPY_PROGRAM_PATTERN` are call/statement-anchored so they stay on normalized text. Added tests for both the false-positive (name-in-literal) and true-positive (real call) cases.

anwesham-lab:
Ran a self-review pass (parallel adversarial reviewers + live-cluster validation) and pushed fixes for everything it surfaced. Head SHA `20d99d4e`.

**Correctness (validated exploitable on a live cluster, now fixed):**
- `E'\''` escape-string desync: `blank_string_literals` didn't model `E'...'` backslash escapes, so it swallowed the rest of the statement and hid a following call from the dangerous-function / keyword tiers. `SELECT E'\'' UNION SELECT ...` and `SELECT E'\'', pg_terminate_backend(1)` slipped through. Fixed by a shared, `E'...'`- and dollar-quote-aware literal-boundary helper used by both `strip_sql_comments` and `blank_string_literals`.
- `$$q'$$` dollar-quote variant of the same desync: fixed by the same helper (dollar-quoted bodies are now blanked).
- `set_config(` false positive: a literal like `SELECT 'set_config(' AS x` was wrongly rejected. Session/keyword checks now run on literal-blanked text.

**Error handling / hardening:**
- Read-only `transact` now `ROLLBACK`s on a `ReadOnlySqlTransaction` before the session scrub, so `RESET ALL`/`RESET ROLE` don't run inside an aborted transaction (where Postgres ignores them).
- If the session scrub fails, the pooled connection is discarded so the next request reconnects with clean state instead of the scrub silently no-opping process-wide.

**Tests:** added tool-level coverage that session mutation is rejected through `readonly_query`/`transact`, that write-mode `transact` still blocks GUC / `COPY ... PROGRAM` / dangerous functions (but allows a benign `SET`), reset-on-failure for `transact`, the ROLLBACK-before-scrub ordering, and connection-discard-on-reset-failure. Detector coverage 100%; 225 tests pass; ruff clean.

**Cleanups:** removed a dead alias, hoisted `KEYWORD_LABELS` to a module constant, de-duplicated the two literal scanners, corrected a misleading comment and the `RESET ROLE` / `session_authorization` wording.

anwesham-lab:
Second self-review round complete (fresh adversarial reviewers + live-cluster validation). Head `d0725671d69e`. Fixed everything surfaced:

**Correctness (validated, now fixed):**
- Double-quoted-identifier apostrophe bypass: a `'` inside an identifier like `"o'clock"` was misread as opening a string literal, blanking away the rest of the statement and hiding a following `pg_terminate_backend(...)` / `set_config(...)`. Fixed by a shared `_end_of_double_quote` helper so all three lexers (`strip_sql_comments`, `blank_string_literals`) skip quoted identifiers identically.
- ReDoS: the `comment_injection` heuristic (`'.*?--` on raw SQL) was O(n^2) on quote-heavy input (~10s for 40k quotes) — a CPU-exhaustion vector. Replaced with a linear string scan (200k quotes now <1s).

**Defense-in-depth (DSQL blocks these at the engine today; added to the classifier for forward-compatibility + to match the docstring's "no session state persists" claim):**
- `PREPARE` / `DEALLOCATE` / `DECLARE ... CURSOR` (session-scoped, not cleared by `RESET ALL`).
- `SET TRANSACTION ... READ WRITE` is now blocked (escalation), while `SET TRANSACTION READ ONLY` / isolation-only stay allowed.

**Cleanups:** de-duplicated the double-quote scanner into the shared helper, removed a dead `end <= n` clause, hoisted `KEYWORD_LABELS`, fixed a weak dollar-quote test and added a standalone-token guard test.

Detector coverage 100%; 230 tests pass; ruff + format clean. Also addressed the GitHub Advanced Security (Bandit) try/except/pass alert by logging in the connection-discard path.

anwesham-lab:
Third review pass complete — 9 independent read-only reviewers against the frozen SHA, fixes applied as one batch. Head `7f25c4c0fffc`.

**Completeness (classifier vocabulary gap vs the postgres-mcp sibling):**
- Now block statement-level mutations at statement start: `CALL`, `DO`, `VACUUM`, `ANALYZE`, `REINDEX`, `CLUSTER`, `REFRESH`, `COMMENT ON`, `SECURITY LABEL`, `IMPORT FOREIGN SCHEMA`, `LOAD`, `DISCARD`, plus `setval()`/`nextval()`. Anchoring keeps `EXPLAIN ANALYZE ... SELECT` and column names like `comment` allowed. DSQL engine-blocks these today; enforced as forward-looking defense-in-depth. (`CALL`/`DO` are the sharpest — opaque procedure bodies.)

**ReDoS hardening (CPU-exhaustion on attacker SQL, all confirmed >1s at 20-35 KB, now linear):**
- `union\s.*\bselect` → two linear presence checks; `copy\s+.*\s+from|to` → anchored single-line; `(begin|commit|rollback).*;\w+` → anchored single-line; and the redundant `commit.*?;` branch in the transaction-bypass detector removed (the linear stacked-statement scan already covers it).

**Polish:** `comment_injection` scan switched from `split('\n')` to bounded `str.find` (O(1) extra space); corrected two stale comments/docstrings; added regression tests for every new block + a failing-on-revert timing test for the dollar-quote O(n²) fix.

**Confirmed clean by the fleet:** embargo (no finding/CVE/researcher references), security model accuracy, silent-failure/reset logic, and — via live-cluster probing — that every classifier block is valid against real DSQL, the E-string/dollar/double-quote boundary model matches DSQL's parser, and RESET ALL + RESET ROLE genuinely scrubs the one persistable vector.

238 tests pass; detector 100% coverage; ruff + format clean.

anwesham-lab:
## Self-review summary (multi-pass, adversarial + live-cluster)

This PR went through three rounds of parallel independent reviewers (bug/bypass, ReDoS/perf, correctness, silent-failure, test-coverage, Python idiom, comment accuracy, PR hygiene, and a live Aurora DSQL semantics validator), each fixing what the prior round surfaced. Findings narrowed each pass — a healthy convergence signal.

**Pass 1 — core enforcement + the headline bypass**
- Confirmed exploitable on a live cluster: `SELECT E'\'' UNION SELECT ...` and `set_config`/`SET` session mutation slipped the read-only gate. Fixed via session-mutation detection + `RESET ALL`/`RESET ROLE` scrub, and closed the `E'\''` escape-string desync in the literal blanker.

**Pass 2 — quoting + ReDoS + hardening**
- Double-quoted-identifier apostrophe bypass (`"o'clock"`), dollar-quote desync (`$$q'$$`), and an O(n^2) comment-injection ReDoS. Added `PREPARE`/`DECLARE`/`DEALLOCATE` and `SET TRANSACTION ... READ WRITE` (incl. newline-split), connection-discard-on-scrub-failure, and ROLLBACK-before-scrub in `transact`.

**Pass 3 — completeness + remaining ReDoS + polish**
- Classifier vocabulary gap vs the `postgres-mcp-server` sibling: now block `CALL`/`DO`/`VACUUM`/`ANALYZE`/`REINDEX`/`CLUSTER`/`REFRESH`/`COMMENT ON`/`SECURITY LABEL`/`IMPORT FOREIGN SCHEMA`/`LOAD`/`DISCARD` + `setval()`/`nextval()`, statement-anchored so `EXPLAIN ANALYZE` and column names stay allowed.
- Four more catastrophic-backtracking/O(n^2) patterns (`union.*select`, `copy.*from|to`, `(begin|commit|rollback).*;\w+`, and the redundant `commit.*?;` bypass branch) rewritten to linear equivalents.

**Verified against real DSQL:** every classifier block is valid (engine-blocks the forward-looking ones today), the `E'...'`/standard-string/dollar-quote/double-quoted-identifier boundary model matches DSQL's parser, and `RESET ALL` + `RESET ROLE` genuinely scrubs the one persistable vector (`set_config`/`SET <guc>`).

Current: 238 unit tests pass, detector at 100% coverage, ruff + format clean.

src/aurora-dsql-mcp-server/awslabs/aurora_dsql_mcp_server/mutable_sql_detector.py
```diff
@@ -86,16 +85,108 @@
...
+        \bset_config\s*\(                    # set_config(...) / pg_catalog.set_config(...), any position
+      | ^\s*SET\s+TRANSACTION\b[\s\S]*?READ\s+WRITE\b   # SET TRANSACTION ... READ WRITE (escalation) — blocked (spans newlines)
+      | ^\s*SET\s+(?!TRANSACTION\b)\S        # SET <anything> except `SET TRANSACTION ...`
+      | ^\s*RESET\b                          # RESET <name> / RESET ALL
+      | ^\s*DISCARD\b                        # DISCARD ALL / PLANS / TEMP / SEQUENCES
+      | ^\s*(LISTEN|NOTIFY|UNLISTEN)\b       # async-notification channels (session state)
+      | ^\s*LOCK\b                           # explicit table locks
+      | ^\s*(PREPARE|DEALLOCATE)\b           # prepared statements (session-scoped, not cleared by RESET ALL)
+      | ^\s*DECLARE\b[^\n]*\bCURSOR\b        # DECLARE ... CURSOR (esp. WITH HOLD — survives COMMIT)
+      | ^\s*EXECUTE\b                        # run a prepared statement (opaque; may mutate)
```
pkale, line 184: ```EXPLAIN EXECUTE prepared_stmt_name is a legitimate read pattern in Postgres```

Could this regex be too strict? Do we want to block this command?

pkale: ```^\s*```` 
Actually the string has to start with Execute so this might not be blocked iiuc. Might be worth adding a test

pkale (dismissed):

src/aurora-dsql-mcp-server/awslabs/aurora_dsql_mcp_server/mutable_sql_detector.py
anwesham-lab: Good instinct (and correction lol) — `EXPLAIN EXECUTE prepared_stmt_name` is already allowed. The branch is anchored at statement start (`^\s*EXECUTE`), so it only matches a bare `EXECUTE stmt`; with `EXPLAIN` leading, EXECUTE isn't at the start and doesn't match. 

Verified it locally with some tests and got:

```
EXECUTE stmt                          -> blocked
EXPLAIN EXECUTE stmt                  -> allowed
EXPLAIN ANALYZE EXECUTE stmt          -> allowed
EXPLAIN (ANALYZE, BUFFERS) EXECUTE …  -> allowed
```

Bare `EXECUTE stmt` is intentionally blocked: it runs a prepared statement whose body is opaque to the classifier (could be an INSERT/UPDATE), which is why the postgres-mcp-server sibling blocks it too. It's also moot on DSQL today — `PREPARE`/`EXECUTE`/`DEALLOCATE` are all `FeatureNotSupported` at the engine (so you can't create a prepared statement to execute anyway); the block is forward-looking defense-in-depth.

Added regression tests pinning that `EXPLAIN EXECUTE` / `EXPLAIN ANALYZE EXECUTE` stay allowed so the anchoring can't silently regress, like you suggested though. I think it's worth the explicit callout.

spencercorwin (approved):

pkale (approved):

praba2210 (approved):

PR #4093, opened 2026-07-07
feat(billing-cost-management-mcp-server): add list-invoice-summaries tool for real invoice data with FX/local-currency, tax breakdowns, and PO numbers

malpani:
## Why this matters

Invoices are the system-of-record for what a customer is actually billed — complementary to Cost Explorer's usage-based estimates. This PR adds a `list-invoice-summaries` tool that gives agents direct access to real invoice data: totals with tax, discounts, and fees; purchase order numbers; due dates; credit memos; and — critically — amounts in three currency perspectives (base, payment, and tax currency) with exchange-rate details when a conversion applies. This unlocks AP/AR reconciliation, payment-deadline awareness, and local-currency finance automation that was previously unavailable via MCP.

## What this unlocks

Agent use cases now possible:

- "What's my invoice total and tax for May 2026?"
- "When is payment due on my latest invoice?"
- "Show me credit memos and the original invoices they adjust."
- "What PO number is on invoice INV-XXXXX?"
- "Give me my invoice amounts in my local payment currency."
- Reconcile accounts-payable / accounts-receivable against the authoritative billing record.

Unlike most Cost Management APIs (which report USD only), invoices expose **BaseCurrencyAmount** (product-and-service currency), **PaymentCurrencyAmount** (customer's configured payment currency), and **TaxCurrencyAmount** (tax currency) — each with `CurrencyExchangeDetails` when a conversion applies. This makes the tool immediately useful for non-US customers and FX-aware finance workflows.

## What changed

| File | Change |
|------|--------|
| `tools/invoicing_operations.py` | **New** — core logic wrapping `ListInvoiceSummaries` (invoicing-2024-12-01) |
| `tools/invoicing_tools.py` | **New** — FastMCP sub-server exposing the tool with full field documentation |
| `tests/tools/test_invoicing_tools.py` | **New** — 20 unit tests |
| `server.py` | Register invoicing sub-server, startup list, instructions |
| `utilities/aws_service_base.py` | Added `'invoicing'` to `allowed_services` |
| `README.md` | Features entry, Supported Services #12, IAM permission |

Key design choices:
- **Rich field documentation in the tool description** so the LLM can interpret every response attribute without external lookups.
- Uses `paginate_aws_response` for accurate pagination metadata (`total_results`, `has_more`, `next_token`, `pages_fetched`).
- Region resolved from `AWS_REGION` env (not hardcoded).
- Epoch timestamps (`IssuedDate`, `DueDate`) normalized to ISO 8601.
- Selector: `account_id` (auto-detected via STS when omitted) or `invoice_id`; optional filters by `billing_period` (YYYY-MM) OR `start_date`/`end_date` range (mutually exclusive), plus `invoicing_entity`.

## Response fields exposed

Per invoice: `AccountId`, `InvoiceId`, `InvoiceType` (INVOICE | CREDIT_MEMO), `OriginalInvoiceId`, `PurchaseOrderNumber`, `IssuedDate`, `DueDate`, `BillingPeriod {Month, Year}`, `Entity {InvoicingEntity}`.

Each of the three currency amounts (`BaseCurrencyAmount`, `PaymentCurrencyAmount`, `TaxCurrencyAmount`) contains: `CurrencyCode` (ISO 4217), `TotalAmount`, `TotalAmountBeforeTax`, `AmountBreakdown { SubTotalAmount, Discounts, Taxes, Fees }` (each with `TotalAmount` + `Breakdown[] {Description, Amount, Rate}`), and `CurrencyExchangeDetails { SourceCurrencyCode, TargetCurrencyCode, Rate }`. Monetary amounts are decimal strings.

## Testing

- `uv run pytest` — **936 passed** (20 new invoicing tests).
- `uv run ruff format` + `uv run ruff check` — clean.
- End-to-end verified with a real read-only `ListInvoiceSummaries` call: returned a real invoice with correct ISO dates and pagination metadata. (Test account was USD-only; `CurrencyExchangeDetails` was null in that response.)

## Design notes for reviewers

- Adds one tool to the existing server; no new MCP server introduced.
- Uses the shared `paginate_aws_response` helper (accurate cross-page totals) and resolves the Region from `AWS_REGION` rather than hardcoding it.
- Deliberately omits a `ReceiverRole` filter: it appears in the HTML API reference but is **not** in the shipped `botocore` (invoicing-2024-12-01) model, so passing it would raise a validation error.
- Required IAM permission: `invoicing:ListInvoiceSummaries` (read-only).
- The three currency fields are structurally present in every response; for accounts without FX, all three share the same currency code and `CurrencyExchangeDetails` is null.


src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/invoicing_tools.py
```diff
@@ -0,0 +1,126 @@
...
+            Mutually exclusive with ``invoice_id``.
+        invoice_id: Retrieve a single invoice's summary. Mutually exclusive with
+            ``account_id``.
+        billing_period: Single calendar month in ``YYYY-MM`` format. Mutually
+            exclusive with ``start_date``/``end_date``.
+        start_date: Inclusive range start (``YYYY-MM-DD`` UTC). Pair with
+            ``end_date``.
+        end_date: Inclusive range end (``YYYY-MM-DD`` UTC). Pair with
+            ``start_date``.
+        invoicing_entity: Filter by AWS legal selling entity name.
```
somsubhro, line 63: can you add an example here of a `invoicing_entity`

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/invoicing_tools.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+invoicing_server = FastMCP(
+    name='invoicing-tools',
+    instructions='Tools for working with the AWS Invoicing API',
+)
+
+
+@invoicing_server.tool(
+    name='list-invoice-summaries',
```
somsubhro: i think given there will be other invoicing APIs that you will add later, should this be `name=invoicing` and then operation routing later? similar to what we do say in: https://github.com/awslabs/mcp/blob/main/src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/bcm_pricing_calculator_tools.py#L211

somsubhro (commented):
Left in line comments

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/invoicing_tools.py
malpani: Done — added concrete examples to the invoicing_entity docs like Amazon Web Services, Inc. (US) and Amazon Web Services EMEA SARL (Europe)

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/invoicing_tools.py
malpani: Good call. Refactored to a single `invoicing` tool with a required operation parameter (currently `list_invoice_summaries`), following the `bcm-pricing-calculator` / `cost-explorer` pattern so future invoicing APIs add as new operations rather than new tools. Kept a thin decorated wrapper delegating to a testable `_invoicing` router and added dispatch + unknown-operation tests; updated the server registration/instructions accordingly.

malpani:
thanks @somsubhr updated to address the feedback

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/invoicing_operations.py
```diff
@@ -0,0 +1,248 @@
...
+    timestamp_to_utc_iso_string,
+    utc_datetime_string_to_epoch_seconds,
+)
+from fastmcp import Context
+from typing import Any, Dict, Optional
+
+
+# Calendar-year bounds accepted for a billing period (per the Invoicing API).
+MIN_BILLING_YEAR = 2005
+MAX_BILLING_YEAR = 2050
```
wangyuhere: Curious why it has a max billing year in 2050?

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/invoicing_tools.py
```diff
@@ -0,0 +1,175 @@
...
+    )
+
+
+@invoicing_server.tool(
+    name='invoicing',
+    description="""Access AWS Invoicing data. Choose an action with the required `operation` parameter; the remaining parameters apply to specific operations. Additional AWS Invoicing APIs are added here as new operations.
+
+## OPERATIONS
+
+1) list_invoice_summaries - invoice-level summaries (no line items) for an account or a single invoice
```
wangyuhere, line 105: I see there is also `ReceiverRole` in https://docs.aws.amazon.com/aws-cost-management/latest/APIReference/API_invoicing_InvoiceSummariesFilter.html#awscostmanagement-Type-invoicing_InvoiceSummariesFilter-ReceiverRole do we need to support it?

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/invoicing_tools.py
wangyuhere: Should we use single tool per service? If those APIs share most parameters, it is good. But if the APIs have different parameters, it will add all parameters as optional in the tool input which might confuse LLM. And it is also not easy to disable API/Tool.

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/invoicing_tools.py
malpani: `ReceiverRole` is a relatively new launch that is not available on the currently pinned boto version for this MCP. 

I prefer keeping dependency bump out of scope for this PR. Yes ideally, all new launches should also show up here alongside the SDK updates and lets figure that out async

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/invoicing_operations.py
malpani: yeah this is not needed. will fix in new revision

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/invoicing_tools.py
malpani: Agree, operation-routing shines when operations share params — that's why cost-explorer / bcm-pricing-calculator do it (they all take time range / granularity / filter). Over time, when expanding coverage to full invoicing namespace (~17 APIs), we will split it into groups with mostly disjoint params:
| Group | Operations | Representative params |
|---|---|---|
| Invoice summaries / docs (read) | ListInvoiceSummaries, GetInvoicePDF, BatchGetInvoiceProfile | `Selector`, `Filter{BillingPeriod,TimeInterval,InvoicingEntity,ReceiverRole}`, `InvoiceId` |
| Invoice Units (CRUD) | Create/Get/List/Update/DeleteInvoiceUnit | `InvoiceUnitArn`, `Name`, `Rule{LinkedAccounts}`, `TaxInheritanceDisabled` |
| Procurement-portal preferences (CRUD) | Create/Get/List/Put/Delete/UpdateStatus…Preference | preference ARNs + portal/status fields |
| Tagging | TagResource, UntagResource, ListTagsForResource | `ResourceArn`, `ResourceTags`, `TagKeys` |


One tool for all of them = ~45 optional params (>80% irrelevant per call) → worse LLM accuracy and harder to scope/disable.

Proposal (end-state): hybrid — one tool per resource group, operation-routed within each: e.g. invoicing (read), invoicing-units, invoicing-preferences, invoicing-tags.

Split trigger for a new group: >6 ops with <20% param overlap, or mixing writes with reads.

For now: only ListInvoiceSummaries exists → keep the single tool (no churn), refactor to grouped tools when the 2nd resource group lands.

wangyuhere (approved):

somsubhro (approved):

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/invoicing_operations.py
```diff
@@ -0,0 +1,234 @@
...
+
+        # Use the shared paginator so pagination metadata is accurate across
+        # pages (total_results, has_more, next_token, pages_fetched, ...).
+        summaries, pagination = await paginate_aws_response(
+            ctx,
+            'ListInvoiceSummaries',
+            client.list_invoice_summaries,
+            request_params,
+            'InvoiceSummaries',
+            token_param='NextToken',
```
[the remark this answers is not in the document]
somsubhro: This is a false positive

somsubhro (approved):

PR #4095, opened 2026-07-08
feat(billing-cost-management-mcp-server): add Compute Optimizer Automation sub-server to BCM MCP

mattkelsey:
### Changes

Adds a new **AWS Compute Optimizer Automation** tool to the Billing and Cost Management MCP server, exposing the service's data-retrieval operations to agents. Compute Optimizer Automation lets customers implement Compute Optimizer recommendations, either automatically via rules or on demand.

#### Key design choice
There are two patterns present in BCM MCP
1. A service registers a new sub-server with a single tool that takes an op code
2. A service registers a new sub-server with a tool per API

For services with lots of APIs the established pattern is to create a single tool that takes an `operation` as an argument. I have followed that pattern for this PR because if we were to have each API be it's own tool, Compute Optimizer Automation tools would become a large percentage of the overall BCM MCP tools, but open to discussion here.

I did not observe the agent having any issue constructing op codes in testing and hasn't been an issue for the other sub-servers.
 
#### List of changes

- **New tool `compute-optimizer-automation`** (`tools/compute_optimizer_automation_tools.py`) — a single dispatch tool with an `operation` parameter, matching the existing `compute-optimizer` and `cost-optimization` tools in this server. Covers 13 operations:
  - **Enrollment/accounts:** `get_enrollment_configuration`, `list_accounts`
  - **Rules:** `get_automation_rule`, `list_automation_rules`, `list_tags_for_resource`
  - **Events:** `get_automation_event`, `list_automation_events`, `list_automation_event_steps`, `list_automation_event_summaries`
  - **Recommended actions:** `list_recommended_actions`, `list_recommended_action_summaries`
  - **Rule preview (dry-run):** `list_automation_rule_preview`, `list_automation_rule_preview_summaries`
- **Operation handlers** (`tools/compute_optimizer_automation_operations.py`) — one handler per operation, with camelCase→snake_case response formatting, shared savings/tags/timestamp/resource-detail formatters, and automatic pagination (`max_pages`, default 10).
- **Robustness:** required-parameter validation and `filters` validation happen *before* any AWS call, returning friendly errors. Valid filter names are **read from the installed botocore service model** (not hardcoded), so they track boto3 upgrades automatically; validation degrades gracefully if the model can't load.
- **Registration & docs:** wired into `server.py`; `compute-optimizer-automation` service added to the client allowlist (`aws_service_base.py`); README (Supported Services + IAM permissions) and CHANGELOG updated; `@mattkelsey` added to CODEOWNERS.

Scope note: only data-retrieval operations are included. Mutating operations (create/update/delete rules, start/rollback events, enrollment changes) are not part of this change.

### User experience

**Before:** Agents using the BCM MCP server had no visibility into Compute Optimizer Automation. A user asking *"what automation rules do I have?"*, *"is my account enrolled in Compute Optimizer Automation?"*, or *"which recommended actions would this rule apply?"* would get no answer from the server — the agent had no tool to call and would either decline or hallucinate.

**After:** The agent can answer those questions directly. For example:
- *"What's my Compute Optimizer Automation enrollment status?"* → calls `get_enrollment_configuration`.
- *"List my automation rules"* / *"show completed automation events this month"* → calls `list_automation_rules` / `list_automation_events` with filters.
- *"Preview what an EBS-upgrade rule would act on before I create it"* → calls `list_automation_rule_preview`.

Responses come back as clean snake_case JSON with estimated savings, and lists paginate automatically (returning a `count` for the current response plus a `next_token` when more results remain). Malformed filter JSON or an invalid filter name returns a clear, actionable error listing the valid names — rather than a raw AWS exception — so the agent can self-correct.

## Checklist

* [x] Changes are documented

**RFC issue number**: N/A, this is an extension of the BCM MCP -> https://github.com/awslabs/mcp/pull/1096

Checklist:

* [x] Migration process documented N/a
* [x] Implement warnings (if it can live side by side) N/a

## Testing
- **Tests:** new suite (`test_compute_optimizer_automation_tools.py`) — 75 tests, 99% coverage on both new modules, covering routing, formatting, pagination, filter/param validation, and model-derived filter names.
- See below conversation examples

### Example use-cases
Added MCP to CC
```
      "mcpServers": {
        "billing-cost-management": {
          "type": "stdio",
          "command": "uv",
          "args": [
            "--directory",
            "<REDACTED>mcp/src/billing-cost-management-mcp-server",
            "run",
            "awslabs.billing-cost-management-mcp-server"
          ],
          "env": {
             <REDACTED>
          }
        }
      },
```

Conversations

Some information redacted and replaced with placeholder text (i.e. `ACCOUNT-ID`, `RULE-A`)

```
❯ Show me my compute optimizer automation rules

⏺ I'll load the Compute Optimizer Automation tool and pull your rules.

  Called billing-cost-management (ctrl+o to expand)

⏺ Here are your Compute Optimizer Automation rules in account [ACCOUNT-ID]
  (region us-east-1) — 8 total, 3 Active / 5 Inactive:

  Active (3)

  Name: RULE-A
  Recommended Actions: UpgradeEbsVolumeType
  Schedule: daily 00:00 PT, 60min window
  Priority: 1 (highest)
  Created: 2026-07-06
  ────────────────────────────────────────
  Name: RULE-B
  Recommended Actions: SnapshotAndDeleteUnattachedEbsVolume
  Schedule: daily 00:00 PT, 60min window
  Priority: 2
  Created: 2026-01-22
  ────────────────────────────────────────
  Name: RULE-C (rev 2)
  Recommended Actions: UpgradeEbsVolumeType,
  SnapshotAndDeleteUnattachedEbsVolume
  Schedule: daily 00:00 PT, 60min window
  Priority: 3
  Created: 2026-04-28

  Inactive (5)

  ┌────────┬──────────────────────────────────────┬──────────┬────────────┐
  │  Name  │         Recommended Actions          │ Priority │  Created   │
  ├────────┼──────────────────────────────────────┼──────────┼────────────┤
  │ RULE-D │ UpgradeEbsVolumeType                 │ 4        │ 2025-11-22 │
  ├────────┼──────────────────────────────────────┼──────────┼────────────┤
  │ RULE-E │ UpgradeEbsVolumeType,                │ 5        │ 2025-11-21 │
  │        │ SnapshotAndDeleteUnattachedEbsVolume │          │            │
  ├────────┼──────────────────────────────────────┼──────────┼────────────┤
  │ RULE-F │ UpgradeEbsVolumeType                 │ 6        │ 2025-11-21 │
  ├────────┼──────────────────────────────────────┼──────────┼────────────┤
  │ RULE-G │ UpgradeEbsVolumeType                 │ 7        │ 2026-01-09 │
  │ (rev 2)│                                      │          │            │
  ├────────┼──────────────────────────────────────┼──────────┼────────────┤
  │ RULE-H │ SnapshotAndDeleteUnattachedEbsVolume │ 8        │ 2025-11-21 │
  └────────┴──────────────────────────────────────┴──────────┴────────────┘

  All are AccountRule type (no org-wide scope). Note the
  SnapshotAndDeleteUnattachedEbsVolume action is destructive-adjacent — it
  snapshots then deletes unattached volumes.

  Want me to drill into any rule's criteria (get_automation_rule), preview what
  a rule would match, or check recent automation events/savings? This looks like
  a sandbox account with a lot of leftover test rules — I can also help you
  identify which inactive ones are safe to clean up (I won't delete anything
  without your explicit go-ahead).

✻ Cooked for 18s

❯ What schedule is RULE-A executed on? When is the next execution?

⏺ RULE-A schedule:

  - Cron expression: cron(0 0 ? * * *) — runs daily at 00:00 (midnight)
  - Timezone: America/Los_Angeles (PT)
  - Execution window: 60 minutes — the rule may fire anytime between 00:00 and
  01:00 PT

  Next execution: the next midnight PT. Today is 2026-07-07, so assuming today's
  run has already passed, the next one is 2026-07-08 at 00:00 PT (00:00–01:00
  PT window). If it's still before midnight PT on 2026-07-07 as you read this,
  then it's tonight at 00:00 PT.

  I don't have the current wall-clock time in this session, only the date — so I
  can't tell you whether today's midnight run has already happened. If you tell
  me the current time (or want me to check recent list_automation_events for
  this rule), I can pin down exactly which run is next.

✻ Cogitated for 9s

❯ Show me a summary of my automation events

  Called billing-cost-management (ctrl+o to expand)

⏺ Here's a summary of your automation events for account [ACCOUNT-ID] over the
  past ~12 months. All events are in Complete status — no Failed, Cancelled,
  InProgress, or Rollback events at all.

  Monthly breakdown (Complete events)

  ┌───────────────────┬────────┬────────────────────────────┐
  │       Month       │ Events │ Est. Monthly Savings (USD) │
  ├───────────────────┼────────┼────────────────────────────┤
  │ 2026-07           │      0 │                      $0.00 │
  ├───────────────────┼────────┼────────────────────────────┤
  │ 2026-06           │      1 │                      $1.00 │
  ├───────────────────┼────────┼────────────────────────────┤
  │ 2026-05           │      2 │                      $1.08 │
  ├───────────────────┼────────┼────────────────────────────┤
  │ 2026-04           │      0 │                      $0.00 │
  ├───────────────────┼────────┼────────────────────────────┤
  │ 2026-03           │      0 │                      $0.00 │
  ├───────────────────┼────────┼────────────────────────────┤
  │ 2026-02           │      4 │                      $1.92 │
  ├───────────────────┼────────┼────────────────────────────┤
  │ 2026-01           │     51 │                     $46.33 │
  ├───────────────────┼────────┼────────────────────────────┤
  │ 2025-12           │      1 │                      $1.25 │
  ├───────────────────┼────────┼────────────────────────────┤
  │ 2025-11 – 2025-07 │      0 │                      $0.00 │
  ├───────────────────┼────────┼────────────────────────────┤
  │ Total             │     59 │                     $51.58 │
  └───────────────────┴────────┴────────────────────────────┘

  Takeaways

  - 59 total automation events, all completed successfully — a 100% success rate
  with zero failures or rollbacks in the past year.
  - January 2026 dominates with 51 events / $46.33 (~90% of all activity and
  savings). The rest of the months are single-digit test-scale runs.
  - "Estimated monthly savings" is the recurring monthly benefit per completed
  action, not a one-time amount — so the Jan figure reflects a burst of actions
  taken that month.
  - July 2026 has no events yet.

  Want me to pull the individual events for a specific month (e.g.,
  list_automation_events for January's 51), or break this down by rule / action
  type or by resource?

✻ Cogitated for 15s

❯ Yeah I want it broken down by action type and resource

⏺ I'll pull the individual events to break them down by action type and
  resource.

  Called billing-cost-management (ctrl+o to expand)

⏺ All 59 events pulled. Here's the breakdown.

  By action type

  ┌──────────────────────────────────────┬────────┬──────┬───────────────────┐
  │             Action Type              │ Events │  %   │   Est. Monthly    │
  │                                      │        │      │      Savings      │
  ├──────────────────────────────────────┼────────┼──────┼───────────────────┤
  │ SnapshotAndDeleteUnattachedEbsVolume │     59 │ 100% │           ~$51.58 │
  ├──────────────────────────────────────┼────────┼──────┼───────────────────┤
  │ UpgradeEbsVolumeType                 │      0 │   0% │             $0.00 │
  └──────────────────────────────────────┴────────┴──────┴───────────────────┘

  Every single automation event was the same action — snapshot-and-delete of an
  unattached EBS volume. Even though most of your rules target
  UpgradeEbsVolumeType, zero upgrade events have ever executed. The one active
  rule that does this action (RULE-B, rule_id [RULE-ID]) drove 58 of
  the 59 events; the remaining 1 (Dec 30, 2025) had no rule attached — an
  on-demand execution (rule_id: null).

  By resource

  - Resource type: 100% EbsVolume (EC2, all in us-east-1).
  - 59 unique volumes — each event acted on a distinct vol-…, no repeats.

  Top volumes by estimated monthly savings:

  ┌───────────┬────────────┬────────────────────────┐
  │  Volume   │ Savings/mo │          Date          │
  ├───────────┼────────────┼────────────────────────┤
  │ [VOL-1]   │      $7.75 │ 2026-01-23             │
  ├───────────┼────────────┼────────────────────────┤
  │ [VOL-2]   │      $1.25 │ 2025-12-30 (on-demand) │
  ├───────────┼────────────┼────────────────────────┤
  │ [VOL-3]   │      $1.20 │ 2026-01-23             │
  ├───────────┼────────────┼────────────────────────┤
  │ [VOL-4]   │      $1.20 │ 2026-01-23             │
  ├───────────┼────────────┼────────────────────────┤
  │ [VOL-5]   │      $1.20 │ 2026-01-23             │
  ├───────────┼────────────┼────────────────────────┤
  │ [VOL-6]   │      $1.20 │ 2026-01-23             │
  └───────────┴────────────┴────────────────────────┘

  The distribution is heavily long-tailed: ~$0.80 is by far the most common
  value (roughly 30 of the 59 volumes), with a scattering of $0.08–$0.64 small
  ones and a handful of $1.00–$1.25. That single $7.75 volume alone is ~15% of
  all savings.

  Notes

  - The "January" summary bucket (51 events) is really one big batch on
  2026-01-23 (~45 volumes deleted in a two-minute window, 08:00–08:02 PT) plus a
  few on 01-31 — a one-time cleanup sweep, not steady-state activity.
  - All 59 completed successfully; no failures, cancellations, or rollbacks.

  Want me to drill into the $7.75 volume's event details/steps, or pull the
  snapshot IDs created for any of these (in case you need to confirm the backups
  exist before they age out)?

✻ Brewed for 39s

❯ When did each event step run for the [VOL-1] event

  Called billing-cost-management (ctrl+o to expand)

⏺ Here are the steps for the [VOL-1] event (event_id:
  [EVENT-ID], the $7.75/mo one). Two steps, both Complete, all times in
  America/Los_Angeles (PT) on 2026-01-23:

  ┌─────┬────────────────┬────────────────────┬────────┬─────────┬────────┐
  │  #  │      Step      │      Resource      │ Start  │ Complet │ Durati │
  │     │                │                    │        │   ed    │   on   │
  ├─────┼────────────────┼────────────────────┼────────┼─────────┼────────┤
  │ 1   │ CreateEbsSnaps │ [SNAP-ID]          │ 08:00: │ 08:01:2 │ ~29s   │
  │     │ hot            │                    │ 51     │ 0       │        │
  ├─────┼────────────────┼────────────────────┼────────┼─────────┼────────┤
  │ 2   │ DeleteEbsVolum │ [VOL-1]            │ 08:01: │ 08:01:4 │ ~10s   │
  │     │ e              │                    │ 34     │ 4       │        │
  └─────┴────────────────┴────────────────────┴────────┴─────────┴────────┘

  Timeline: snapshot created first (08:00:51 → 08:01:20), then a ~14s gap, then
  the volume deleted (08:01:34 → 08:01:44). Total event wall-clock ~53s. The
  savings ($7.75/mo) is attributed to the delete step; the snapshot step shows
  $0 (it's the backup, not the savings).

  Your backup snapshot is [SNAP-ID] — that's your restore point if
  you ever need this volume back. Want me to pull the snapshot details (size,
  state, whether it still exists) to confirm the backup is intact?

✻ Worked for 11s
```

Opt-in negative test
```
❯ Summarize my recent compute optimizer automation events                       

⏺ I'll load the Compute Optimizer Automation tool schema and pull recent events.

  Called billing-cost-management (ctrl+o to expand)

⏺ The default region (us-east-1) returned "Account is not enrolled." Let me
  check enrollment status and confirm whether the current credentials/account
  are enrolled at all.

  Called billing-cost-management (ctrl+o to expand)

⏺ I can't summarize your recent Compute Optimizer automation events because the 
  account isn't enrolled in the service.

  Evidence:

  1. list_automation_events (us-east-1) → 403 OptInRequiredException: "Account 
  is not enrolled in the service"
  4. get_enrollment_configuration confirms it:
  { "status": "Inactive", "status_reason": null,
    "organization_rule_mode": null, "last_updated_timestamp": null }

  With enrollment Inactive, no automation rules ever ran, so there are no events
  to summarize.

  A couple of things worth checking:

  - Region — Compute Optimizer Automation is regional and I queried the default
  (us-east-1). If your automation lives in another region, tell me which one and
  I'll re-check there.
  - Account — this reflects whatever credentials this session is using. If
  automation is enrolled under a different account (e.g. a
  management/delegated-admin account), let me know and I can point the query
  there.

  Want me to re-run against a specific region or account?

✻ Cooked for 22s
```

## Acknowledgment


src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/compute_optimizer_automation_tools.py
```diff
@@ -0,0 +1,419 @@
...
+    'get_enrollment_configuration',
+    'list_accounts',
+    'list_automation_events',
+    'list_automation_event_steps',
+    'list_automation_event_summaries',
+    'list_automation_rules',
+    'list_recommended_actions',
+    'list_recommended_action_summaries',
+    'list_automation_rule_preview',
+    'list_automation_rule_preview_summaries',
```
wangyuhere, line 68: For these list APIs, what will be the P50, P90  items size? Do we need to support SQL offload for some of the list APIs so the agent context window is not overloaded.

see https://github.com/awslabs/mcp/blob/main/src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/utilities/sql_utils.py#L406

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/compute_optimizer_automation_tools.py
```diff
@@ -0,0 +1,419 @@
...
+    name='compute-optimizer-automation-tools',
+    instructions='Tools for working with the AWS Compute Optimizer Automation API',
+)
+
+
+@compute_optimizer_automation_server.tool(
+    name='compute-optimizer-automation',
+    description="""Retrieves data from AWS Compute Optimizer Automation.
+
+Compute Optimizer Automation lets customers implement Compute Optimizer recommendations,
```
wangyuhere, line 120: Consider adding an explicit USE THIS TOOL FOR / DO NOT USE FOR block to the tool description, mirroring the pattern in [cost_optimization_hub_tools](https://github.com/awslabs/mcp/blob/main/src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/cost_optimization_hub_tools.py#L55-L62) and [compute_optimizer_tools](https://github.com/awslabs/mcp/blob/main/src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/compute_optimizer_tools.py#L42-L50)

The names are extremely close (`compute-optimizer` vs `compute-optimizer-automation` vs `cost-optimization-hub`) and the "recommended_action" concept overlaps with compute-optimizer's per-resource recommendations — without a disambiguation block, agents may pick the wrong tool for questions like "what EBS changes are recommended?"

Could you also test will agent get confused about the 3 tools?

mattkelsey: I did test that agent was able to differentiate between the tools. When asked for recommendations it would go to CO and when asked about recommended actions it correctly routed to CO Automation. But not exhaustive, I am sure there are cases where it could overlap

Agree a clear disambiguation block would help, will update to add

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/compute_optimizer_automation_tools.py
mattkelsey: Hmm good point.

Events, Event Steps, and Event Summaries are all very small. Rule Previews and Actions can have tags, but only up to 50. So I think max response size is mostly filled up by tags (50).

Don't have data on exact usage but it could be up to 2.1Mb per page (assumes 50 tags each with max k and v length). Will onboard to SQL offload

mattkelsey: Since we have a bunch of resource structures, I implemented a new generic converter type called `records`.

When Automation API results are passed through, it generates the following tables when response size too large:
```
  1. list_recommended_actions and list_automation_rule_preview

  (identical item shape — both use _format_recommended_action; list keys recommended_actions / preview_results)

  CREATE TABLE compute_optimizer_automation_list_recommended_actions_ab12cd34 (
      recommended_action_id      TEXT,
      resource_arn               TEXT,
      resource_id                TEXT,
      account_id                 TEXT,
      region                     TEXT,
      resource_type              TEXT,
      look_back_period_in_days   TEXT,
      recommended_action_type    TEXT,
      current_resource_summary   TEXT,
      current_resource_details   TEXT,   -- JSON: {"ebs_volume":{"configuration":{...}}}
      recommended_resource_summary TEXT,
      recommended_resource_details TEXT, -- JSON
      restart_needed             TEXT,
      estimated_monthly_savings  TEXT,   -- JSON: {"currency":"USD","after_discount_savings":9.0,...}
      resource_tags              TEXT    -- JSON: [{"key":"env","value":"prod"}, ...]  <- the size driver
  );
  Example queries the agent would run instead of paginating into context:
  -- Top actions by savings
  SELECT resource_id, json_extract(estimated_monthly_savings,'$.after_discount_savings') AS savings
  FROM   <table> ORDER BY CAST(savings AS REAL) DESC LIMIT 10;

  -- Filter by a tag value (tags are JSON, still queryable)
  SELECT resource_id FROM <table>
  WHERE  EXISTS (SELECT 1 FROM json_each(resource_tags)
                 WHERE json_extract(value,'$.key')='env' AND json_extract(value,'$.value')='prod');

  2. list_automation_events  (_format_automation_event, key automation_events)

  CREATE TABLE compute_optimizer_automation_list_automation_events_… (
      event_id                  TEXT,
      event_description         TEXT,
      event_type                TEXT,
      event_status              TEXT,
      event_status_reason       TEXT,
      resource_arn              TEXT,
      resource_id               TEXT,
      recommended_action_id     TEXT,
      account_id                TEXT,
      region                    TEXT,
      rule_id                   TEXT,
      resource_type             TEXT,
      created_timestamp         TEXT,   -- ISO 8601
      completed_timestamp       TEXT,
      estimated_monthly_savings TEXT    -- JSON
  );
  SELECT event_status, COUNT(*) FROM <table> GROUP BY event_status;

  3. list_automation_event_steps  (_format_automation_event_step, key automation_event_steps)

  CREATE TABLE compute_optimizer_automation_list_automation_event_steps_… (
      event_id                  TEXT,
      step_id                   TEXT,
      step_type                 TEXT,
      step_status               TEXT,
      resource_id               TEXT,
      start_timestamp           TEXT,
      completed_timestamp       TEXT,
      estimated_monthly_savings TEXT    -- JSON
  );

  4. list_automation_rules  (_format_automation_rule, key automation_rules)

  CREATE TABLE compute_optimizer_automation_list_automation_rules_… (
      rule_arn                  TEXT,
      rule_id                   TEXT,
      name                      TEXT,
      description               TEXT,
      rule_type                 TEXT,
      rule_revision             TEXT,
      account_id                TEXT,
      organization_configuration TEXT,  -- JSON: {"rule_apply_order":…,"account_ids":[…]}
      priority                  TEXT,
      recommended_action_types  TEXT,   -- JSON: ["UpgradeEbsVolumeType", …]
      schedule                  TEXT,   -- JSON: {"schedule_expression":"cron(...)",…}
      status                    TEXT,
      created_timestamp         TEXT,
      last_updated_timestamp    TEXT
  );
  Note: criteria and tags appear only on the single-rule get_automation_rule (not offloaded), so they're absent here. Since columns are the union of item keys, if any listed rule happened to carry an extra optional field, it'd get its own column and rows without it store NULL.

  5. list_recommended_action_summaries and list_automation_rule_preview_summaries

  (identical — _format_summary; keys recommended_action_summaries / preview_result_summaries)
  CREATE TABLE compute_optimizer_automation_list_recommended_action_summaries_… (
      key   TEXT,
      total TEXT   -- JSON: {"recommended_action_count":42,"estimated_monthly_savings":{…}}
  );

  6. list_automation_event_summaries  (_format_event_summary, key automation_event_summaries)

  CREATE TABLE compute_optimizer_automation_list_automation_event_summaries_… (
      key         TEXT,
      dimensions  TEXT,   -- JSON: [{"key":…,"value":…}]
      time_period TEXT,   -- JSON: {"start_time_inclusive":…,"end_time_exclusive":…}
      total       TEXT    -- JSON: {"automation_event_count":…,"estimated_monthly_savings":{…}}
  );

  7. list_accounts  (inline dict in the handler, key accounts)

  CREATE TABLE compute_optimizer_automation_list_accounts_… (
      account_id              TEXT,
      status                  TEXT,
      organization_rule_mode  TEXT,
      status_reason           TEXT,
      last_updated_timestamp  TEXT
  );
  ```

File: src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/compute_optimizer_automation_operations.py [hunk 3]
```diff
@@ -0,0 +1,1060 @@
...
+    """
+    response_data: Dict[str, Any] = {list_key: items, 'count': len(items)}
+    if next_token:
+        response_data['next_token'] = next_token
+
+    response_data = await convert_response_if_needed(
+        ctx,
+        response_data,
+        f'compute_optimizer_automation_{operation}',
+        pagination_token_key='next_token',
```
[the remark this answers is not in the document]
mattkelsey: not a password

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/utilities/sql_utils.py
```diff
@@ -359,7 +359,66 @@ def _get_specialized_converter(operation_name: str) -> Optional[str]:
...
 
-    return converters.get(operation_name)
+    if operation_name in converters:
+        return converters[operation_name]
+
+    # Compute Optimizer Automation list operations all return a uniform
+    # {<list_key>: [...], count, next_token} shape; store one row per item so
+    # the offloaded data stays queryable instead of a single JSON blob.
+    if operation_name.startswith('compute_optimizer_automation_'):
+        return 'records'
```
wangyuhere, line 369: How reliable it is? Is it better to list each operation name explicitly?

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/utilities/sql_utils.py
```diff
@@ -359,7 +359,66 @@ def _get_specialized_converter(operation_name: str) -> Optional[str]:
...
+
+    return None
+
+
+def _extract_record_list(response: Dict[str, Any]) -> List[Dict[str, Any]]:
+    """Pull the item list out of a {<list_key>: [...], count, next_token} response.
+
+    Returns the first list-valued field (the item collection), ignoring the
```
wangyuhere, line 377: Will the first list value field always what we want in the collection? Can we pass the field name if possible?

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/utilities/sql_utils.py
```diff
@@ -359,7 +359,66 @@ def _get_specialized_converter(operation_name: str) -> Optional[str]:
...
+    Raises:
+        ValueError: If an item key is not a safe SQL identifier.
+    """
+    columns: List[str] = []
+    for item in items:
+        if isinstance(item, dict):
+            for key in item:
+                if key not in columns:
+                    if not isinstance(key, str) or not re.match(r'^[a-zA-Z0-9_]+$', key):
+                        raise ValueError(f'Invalid column name for SQL table: {key!r}')
```
wangyuhere, line 419: Do we also need to handle the reserved SQL keyword here?

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/utilities/sql_utils.py
```diff
@@ -708,6 +767,36 @@ async def convert_api_response_to_table(
...
+                # Non-dict items (fallback 'value' column) store the scalar directly.
+                row = item if isinstance(item, dict) else {'value': item}
+                values = []
+                for col in columns:
+                    raw = row.get(col)
+                    values.append(json.dumps(raw) if isinstance(raw, (dict, list)) else raw)
+                validate_table_name(table_name)
+                insert_sql = create_safe_sql_statement(
+                    'INSERT', table_name, f'({column_list}) VALUES ({placeholders})'
+                )
```
wangyuhere: These checking and preparation doesn't need to be inside the loop.

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/utilities/sql_utils.py
mattkelsey: compute_optimizer_automation operation names are literals, so reliable as long as caller uses correctly

My intent is to allow any future list APIs to pass through this without needing another update

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/utilities/sql_utils.py
mattkelsey: Yes, all items returned in the list response are the same shape for automation list APIs. So just taking the structure of the first one works

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/utilities/sql_utils.py
mattkelsey: Good catch will fix

edit: Fixed!

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/utilities/sql_utils.py
mattkelsey: These values are based on API names, so an API would have to be named `select` or `group`, etc to hit this case.

Not a problem for Automation APIs but in the probably a good idea to handle them regardless, will update.

edit: Fixed!

kingrad17:
Look good to me. Approved

wangyuhere (dismissed):

src/billing-cost-management-mcp-server/README.md
```diff
@@ -243,6 +243,21 @@ Compute Optimizer:
...
+- compute-optimizer-automation:ListAccounts
+- compute-optimizer-automation:ListAutomationEvents
+- compute-optimizer-automation:ListAutomationEventSteps
+- compute-optimizer-automation:ListAutomationEventSummaries
+- compute-optimizer-automation:ListAutomationRules
+- compute-optimizer-automation:ListRecommendedActions
+- compute-optimizer-automation:ListRecommendedActionSummaries
+- compute-optimizer-automation:ListAutomationRulePreview
+- compute-optimizer-automation:ListAutomationRulePreviewSummaries
+- compute-optimizer-automation:ListTagsForResource
```
Comusus: Is this correct according to https://docs.aws.amazon.com/service-authorization/latest/reference/list_compute-optimizer-automation.html?

Believe it should be `aco-automation:XXX` according to IAM console wizard and IAM docs.

src/billing-cost-management-mcp-server/README.md
```diff
@@ -243,6 +243,21 @@ Compute Optimizer:
...
+Compute Optimizer Automation:
+- compute-optimizer-automation:GetAutomationEvent
+- compute-optimizer-automation:GetAutomationRule
+- compute-optimizer-automation:GetEnrollmentConfiguration
+- compute-optimizer-automation:ListAccounts
+- compute-optimizer-automation:ListAutomationEvents
+- compute-optimizer-automation:ListAutomationEventSteps
+- compute-optimizer-automation:ListAutomationEventSummaries
+- compute-optimizer-automation:ListAutomationRules
+- compute-optimizer-automation:ListRecommendedActions
```
Comusus: We should also document `ec2:DescribeVolumes` as a required IAM action permission. It's used within `ListRecommendedActions` and `ListAutomationRulePreview` actions.

mattkelsey: Actually it's not required, we do not DescribeVolumes do get that information, it comes from the ACO Recommendation resourceState directly

src/billing-cost-management-mcp-server/README.md
mattkelsey: Oops good catch, fixing

Comusus (dismissed):

src/billing-cost-management-mcp-server/README.md
wangyuhere: Checking https://docs.aws.amazon.com/service-authorization/latest/reference/list_compute-optimizer-automation.html 

Do the two APIs `ListRecommendedActions` and `ListAutomationRulePreview` requires dual auth? if yes, then we probably need to list them?

mattkelsey: What do you mean by dual auth?

If referring to DV or ACO prems -> It does not require DescribeVolume or ACO rec permissions. Those APIs only use resource state that comes from ACO Recommendations (ingested internally, not by calling API in customer accounts)

A role with only `aco-automation:ListRecommendedActions` can successfully call the API for example, same with `ListAutomationRulePreview`

edit: correction, denied by another auth layer after test even though we don't call DV it's denied at API layer, updating

wangyuhere (dismissed):

wangyuhere (approved):

Comusus (approved):

wangyuhere (approved):

arnewouters (approved):

PR #4098, opened 2026-07-08
fix(mysql-mcp-server): resolve standalone RDS instances by identifier, not cluster

### LINKED ISSUE #3787: mysql-mcp-server: connect_to_database fails with DBClusterNotFoundFault against a standalone RDS MySQL instance
### Describe the bug

Calling connect_to_database against a standalone RDS MySQL instance (not an Aurora cluster) fails with DBClusterNotFoundFault. The connect path appears to perform an RDS cluster lookup (describe-db-clusters) keyed on the supplied identifier, but a standalone instance has no cluster, so the lookup can never succeed.

### Expected Behavior

With connection_method=mysqlwire, the server should connect directly to the provided endpoint over the wire protocol and not require/perform an RDS cluster resolution — standalone RDS MySQL instances (and self-hosted MySQL) should be supported, consistent with the documented direct-connection (--hostname) path.

### Current Behavior

```
{
  "status": "Failed",
  "error_type": "DBClusterNotFoundFault",
  "error_message": "connect_to_database failed; see server logs for details"
}
```

## Steps to reproduce

1. Have a standalone RDS MySQL instance (engine mysql, e.g. db.t3.micro), confirmed available via aws rds describe-db-instances. It is an instance, so describe-db-clusters returns nothing for its identifier.

2. Call connect_to_database with:

- cluster_identifier: <db-instance-identifier>
- connection_method: mysqlwire
- database: <db-name>
- database_type: mysql
- db_endpoint: <instance-endpoint>
- port: 3306
- region: <region>

### Possible Solution

_No response_

_No response_

### OS

macos

### Server

mysql-mcp-server

### Server Version

latest

eu-west-1

### Other information

_No response_

### Service quota

masamaru0513:
## Summary

Fixes `connect_to_database` failing with `DBClusterNotFoundFault` against a standalone RDS MySQL / MariaDB instance.

Fixes #3787

## What changed

`internal_connect_to_database` treated any non-empty `cluster_identifier` as an RDS cluster name and unconditionally called `describe_db_clusters`, even for `database_type=mysql` / `mariadb`, which can be standalone instances with no cluster at all.

The catch is that engine type does not determine which lookup applies. An RDS deployment reporting `Engine='mysql'` is either:

- a standalone DB instance — `describe_db_instances`, the case #3787 is about, or
- an [RDS Multi-AZ DB cluster](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/multi-az-db-clusters-concepts.html) — a `DBCluster` resolved by `describe_db_clusters`, with no DB instance carrying the cluster's identifier.

So resolution now goes instance-first and falls back to the cluster lookup only when RDS reports no such instance:

- Aurora MySQL is unchanged — always a cluster, always `describe_db_clusters`.
- RDS MySQL / MariaDB try the instance lookup: the existing endpoint-based `internal_get_instance_properties` when `db_endpoint` is given, otherwise a new `internal_get_instance_properties_by_identifier` (a direct `describe_db_instances(DBInstanceIdentifier=...)` call — the instance counterpart to `internal_get_cluster_properties`).
- If that reports no matching instance, `database_type=mysql` retries `describe_db_clusters`, which is what keeps Multi-AZ DB clusters working.

The fallback is narrowed on two axes:

- **Only on "not found."** The two instance lookups signal it differently — the endpoint scan exhausts its paginator and raises `ValueError`, the direct lookup surfaces RDS's `DBInstanceNotFound` as a `ClientError` — so `is_db_instance_not_found` covers both and nothing else. `AccessDenied` and throttling propagate rather than being retried as a cluster lookup that would only fail again with a less informative error.
- **Only for `database_type=mysql`.** Multi-AZ DB clusters support only MySQL and PostgreSQL, so RDS MariaDB is instance-only by construction and gets no fallback.

## Design notes

The first version of this PR branched on `database_type` up front: Aurora → cluster lookup, everything else → instance lookup. I argued for that over a try/except fallback on the grounds that the validation block ~30 lines above already asserts the "Aurora is always a cluster, RDS can be standalone" contract, so a fallback would encode the same routing decision twice.

That reasoning was wrong, and @happycupofjava caught why: `mysql` is genuinely ambiguous. The validation block's contract is about whether `cluster_identifier` is *required*, not about which API resolves it. Branching on engine alone therefore cannot be correct, and identifier-shape resolution is unavoidable.

My other objection to the fallback — that it would run a doomed API call on every RDS MySQL / MariaDB connection — no longer applies with instance-first ordering: the second call only happens for a Multi-AZ DB cluster, and MariaDB never makes it at all.

`enable_data_api` stays `False` on the RDS path. Multi-AZ DB clusters have no Data API (`HttpEndpointEnabled` is absent from their `DBCluster`), and `is_connection_method_supported` already rejects `rdsapi` for every non-Aurora engine, so the Aurora-only Data API wiring is unreachable here. `cluster_arn` is populated on the Multi-AZ path for parity with the Aurora branch even though nothing consumes it without the Data API.

## Testing

**Live verification** against real RDS resources in `ap-northeast-1`. `AsyncmyPoolConnection.__init__` is lazy (`self.pool = None`, no I/O), so `internal_connect_to_database` could be driven end to end over the real control plane without opening a database connection.

The reported bug, against a standalone `db.t4g.micro` running `mysql 8.0.46`. On `main` (`c53d5900`) both call shapes fail with the reporter's exact error — the issue's own repro (`cluster_identifier` + `db_endpoint`) and the identifier-only variant:

```
DBClusterNotFoundFault: An error occurred (DBClusterNotFoundFault) when calling
the DescribeDBClusters operation: DBCluster issue3787-standalone-mysql not found.
```

Both resolve on this branch.

The Multi-AZ regression, against a `mysql 8.0.46` Multi-AZ DB cluster on `db.c6gd.medium` plus a standalone `db.t4g.micro`. Same script run against this branch before and after the fix:

| Case | Before | After |
| --- | --- | --- |
| `mysql` + Multi-AZ identifier | ❌ `DBInstanceNotFound` | ✅ |
| `mysql` + Multi-AZ identifier + writer endpoint | ❌ `ValueError` from the endpoint scan | ✅ |
| `mysql` + Multi-AZ, `MasterUsername` propagation (IAM path) | ❌ `DBInstanceNotFound` | ✅ |
| `mysql` + standalone identifier (#3787) | ✅ | ✅ |
| `mysql` + standalone endpoint | ✅ | ✅ |
| `mariadb` gets no cluster fallback | ✅ | ✅ |
| `aurora-mysql` stays cluster-only | ✅ | ✅ |
| | **3 failing** | **all passing** |

Worth recording that the regression covered *both* call shapes, not only the identifier-only one flagged in review: `internal_get_instance_properties` matches on `DBInstance` endpoint addresses, and a cluster writer endpoint is never one.

What the live run confirmed about the API shapes the unit tests mock:

- A Multi-AZ DB cluster comes back from `describe_db_clusters` with `Engine='mysql'`, `MasterUsername`, `MasterUserSecret`, `Endpoint`, `Port`, and no `HttpEndpointEnabled`.
- `describe_db_instances(DBInstanceIdentifier=<cluster id>)` raises `DBInstanceNotFound`; the member instances exist under their own names (`<cluster>-instance-1..3`).
- The cluster writer endpoint matches no DB instance endpoint.

Unit tests cover both Multi-AZ call shapes, `MasterUsername` propagation, MariaDB getting no cluster fallback, `AccessDenied` propagating instead of falling back, the `is_db_instance_not_found` predicate, and an Aurora identifier passed under `database_type=mysql`, alongside the existing coverage for `internal_get_instance_properties_by_identifier` and the #3787 resolution path.

`pyright` clean (0 errors, 0 warnings). `bandit -r --severity-level medium --confidence-level medium`: 0 issues. `ruff check` and `ruff format` clean. `uv build` OK. `pre-commit run` green across all hooks.

## Housekeeping

- Version bump `1.0.24` -> `1.0.25` (`pyproject.toml`, `uv.lock`), `CHANGELOG.md` updated with both the fix and the Multi-AZ behaviour.
- `.secrets.baseline`: line-number updates only. The baseline tracks a non-secret in each of the two touched test files, and the added imports and tests shift them.

## Backwards compatibility

- No new dependencies, no signature changes to public tools.
- Aurora MySQL is unchanged (still `describe_db_clusters` only, no instance fallback).
- RDS MySQL / MariaDB callers already passing `db_endpoint` are unaffected — that path still goes through `internal_get_instance_properties` first.
- RDS MySQL Multi-AZ DB clusters keep resolving, by both `cluster_identifier` and `cluster_identifier` + writer endpoint, as they did before this PR.
- The earlier revision of this PR listed a known limitation where an Aurora cluster identifier passed under `database_type=mysql` would fail with `DBInstanceNotFound`. The cluster fallback resolves that case too, so the limitation is gone rather than documented, and there is now a test pinning it.


masamaru0513:
Still relevant, so I'd rather not let this one close.

I re-checked `main` this morning and the original problem is untouched — `internal_connect_to_database` still calls `internal_get_cluster_properties` for any non-empty `cluster_identifier`, ignoring the `database_type` contract that the validation block earlier in the same function already asserts. `server.py` hasn't changed upstream since 6d180004 back in May. #3787 is also still open with `needs-triage` and no comments, and I couldn't find another PR going after it.

Rebasing onto latest `main` was clean, but two things needed fixing along the way, both from `main` moving underneath the branch:

- `.secrets.baseline` was regenerated on `main` on 07-16, and the extra import in `test_cp_api_simple_functions.py` pushes a tracked non-secret from line 44 to 45. Without that 2-line update the pre-commit hook fails.
- 1.0.24 has since been released by the automated bump (536db49a), so I moved the version bump and changelog entry to 1.0.25.

I've also tidied up the description while I was at it — it still claimed the old `1.0.23 -> 1.0.24` bump, and said the validation sits "two lines above" the changed branch when it's more like thirty.

Something I only noticed now: none of the real CI has ever run on this PR. Only the two `pull_request_target` lint checks show up, because workflow runs on fork PRs need a maintainer to approve them. So I ran the same steps locally on the rebased branch — bandit clean, 438 tests passing, pyright clean, ruff clean, `uv build` fine, and `pre-commit run --all-files` green. Happy to be told I've missed something, but I think it's in a reviewable state.

Could someone approve the workflow runs, and add `backlog` if this is likely to sit for a while? I don't have permission to add the label myself. Otherwise it auto-closes in two days and I'll just end up re-opening the same change.

And if the "Known limitation" note in the description is what makes this awkward to review (an Aurora identifier passed under `database_type=mysql`), I'm happy to pull it out into its own issue instead.

masamaru0513:
@kennthhz @happycupofjava — sorry for the direct ping. This PR is due to be auto-closed in two days, and you're both listed in `CODEOWNERS` for `src/mysql-mcp-server`.

I rebased it onto the latest `main` and re-verified everything this morning; details are in the comment above. The short version is that #3787 is still open and the bug is still present on `main`.

Could either of you approve the workflow runs so the real CI can weigh in? Or if someone else should be picking this up, just say so and I'll redirect.

kennthhz:
Comment to keep the PR open

kennthhz (dismissed):

src/mysql-mcp-server/awslabs/mysql_mcp_server/server.py
```diff
@@ -687,7 +688,14 @@ def internal_connect_to_database(
     cluster_arn: str = ''
     secret_arn: str = ''
 
-    if cluster_identifier:
+    # Aurora MySQL is always a cluster; resolve via cluster_identifier.
+    # RDS MySQL / RDS MariaDB can be standalone instances that have no
```
happycupofjava: RDS MySQL also supports Multi-AZ DB clusters (engine mysql, resolved via describe_db_clusters) -> https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/multi-az-db-clusters-concepts.html. Since this branch now routes all mysql/mariadb to an instance lookup, passing a Multi-AZ DB cluster identifier would regress to DBInstanceNotFound. Is that intentionally out of scope? If so, a brief code comment + CHANGELOG note would make it explicit.

masamaru0513: Thank you. I will check and take the necessary action.

masamaru0513: Not out of scope — you're right, and it's a real regression rather than a risk.

I stood up a `mysql 8.0.46` Multi-AZ DB cluster (`db.c6gd.medium`) and a standalone `db.t4g.micro` in AWS and ran the same script against this branch before and after a fix. `main` handles Multi-AZ DB clusters today, because its unconditional `describe_db_clusters` happens to be the right call for them; my `database_type` gate broke that.

It's also wider than the case you flagged — both call shapes failed, not just identifier-only:

| Case | Before | After |
| --- | --- | --- |
| `mysql` + Multi-AZ identifier | ❌ `DBInstanceNotFound` | ✅ |
| `mysql` + Multi-AZ identifier + writer endpoint | ❌ `ValueError` from the endpoint scan | ✅ |
| `mysql` + Multi-AZ, `MasterUsername` propagation (IAM path) | ❌ `DBInstanceNotFound` | ✅ |
| `mysql` + standalone identifier (#3787) | ✅ | ✅ |
| `mysql` + standalone endpoint | ✅ | ✅ |
| `mariadb` gets no cluster fallback | ✅ | ✅ |
| `aurora-mysql` stays cluster-only | ✅ | ✅ |

The second row is the one I'd missed: `internal_get_instance_properties` matches on `DBInstance` endpoint addresses, and a cluster writer endpoint is never one.

The live run also pinned down the API shapes the unit tests were mocking: the cluster comes back with `Engine='mysql'`, `MasterUsername`, `MasterUserSecret`, `Endpoint`, `Port` and no `HttpEndpointEnabled`; `describe_db_instances(DBInstanceIdentifier=<cluster id>)` raises `DBInstanceNotFound`, with the member instances present under their own names; and the writer endpoint matches no instance endpoint.

The oversight underneath it was in my routing premise. I claimed `database_type` tells us whether the identifier names a cluster or an instance, and rejected a try/except fallback on that basis. It doesn't — `mysql` is genuinely ambiguous, so resolving by identifier shape is unavoidable. The validation block I cited governs whether `cluster_identifier` is *required*, not which API resolves it. My other objection, that a fallback would add a doomed API call to every RDS MySQL / MariaDB connection, also goes away with instance-first ordering.

`62c5e69f` resolves instance-first and falls back to `describe_db_clusters` only when RDS reports no such instance, narrowed two ways:

- Only on "not found". The two instance lookups signal it differently — the endpoint scan raises `ValueError`, the direct lookup surfaces `DBInstanceNotFound` as a `ClientError` — so `is_db_instance_not_found` covers exactly those two. `AccessDenied` and throttling propagate instead of being retried as a cluster lookup that would fail again with a worse error.
- Only for `database_type=mysql`, since Multi-AZ DB clusters support only MySQL and PostgreSQL. RDS MariaDB stays instance-only, and that one is out of scope by construction — with the code comment you asked for.

Aurora is untouched. CHANGELOG has a `Changed` entry for the Multi-AZ behaviour, and I've rewritten the PR description — the old "Known limitation" about an Aurora identifier under `database_type=mysql` is gone, since the fallback resolves that case too.

masamaru0513:
@happycupofjava @kennthhz — status update, since this has moved since the approval.

The original issue is fixed and stays fixed. #3787's standalone-instance case resolves on this branch, and I confirmed it against a real standalone `db.t4g.micro`: on current `main` (`c53d5900`) both of the reporter's call shapes still fail with the exact `DBClusterNotFoundFault` from the issue, and both resolve here. #3787 is still open and untriaged, so that data point may be useful for triage on its own.

What changed is that following up on @happycupofjava's Multi-AZ DB cluster comment surfaced something I hadn't accounted for. I stood up an actual Multi-AZ DB cluster to check, and it was a genuine regression rather than a hypothetical — three cases, one of them a shape the comment didn't mention: passing the cluster's writer endpoint alongside the identifier also broke, because `internal_get_instance_properties` matches on `DBInstance` endpoint addresses and a cluster endpoint is never one.

The oversight was in my routing premise. I branched on `database_type`, assuming the engine tells you whether the identifier names a cluster or an instance. It doesn't — `Engine='mysql'` is either a standalone DB instance or a Multi-AZ DB cluster, and only the lookup distinguishes them. Resolution is now instance-first with a fallback to `describe_db_clusters`, narrowed to `DBInstanceNotFound` so `AccessDenied` and throttling still surface, and to `database_type=mysql` since Multi-AZ DB clusters are MySQL/PostgreSQL only — RDS MariaDB stays instance-only, with the code comment that was asked for. Aurora is untouched. Before/after table is in the review thread.

Sorry for the extra round — could you approve the workflow runs so CI can execute, and take another look when you get a chance?

masamaru0513:
Friendly ping — this fix is still valid and the underlying issue remains on main. Happy to rebase or address any feedback if needed. @kennthhz @happycupofjava

masamaru0513:
Closing this in favor of a fresh PR against current `main`.

The issue this PR targeted (#3787) was closed by stale-bot on 2026-08-14 rather than by a fix covering this case — its own repro always included `db_endpoint`, and #4515 (merged since this PR was opened) validates/uses `db_endpoint` but doesn't add an identifier-only resolution path. That left this branch's actual scope (resolving `cluster_identifier` alone, no `db_endpoint`) undocumented anywhere and in conflict with #4515's rewrite of the same non-Aurora branch.

Filed #4589 to track the identifier-only gap on its own, and rebased this change onto current `main` (including the Multi-AZ DB cluster fallback from the review discussion below) as #4590.

PR #4108, opened 2026-07-10
fix(openapi-mcp-server): DNS-pin spec fetches to prevent rebinding SSRF

scottschreckengaust:
Fixes

## Summary

Closes a DNS-rebinding TOCTOU SSRF in the `openapi-mcp-server` spec loader (CWE-350 / CWE-367 / CWE-918).

### Changes

The multi-spec (`additional_specs`) path validated `spec_url` via `validate_url_for_spec()` and then **discarded** the result, re-fetching with `httpx.get(url)` — a *second, fresh* DNS resolution. A rebinding authoritative DNS can answer a public IP during validation (passing the SSRF check) and `169.254.169.254` (or any internal IP) at fetch time. The **primary** spec URL was not SSRF-validated at all.

The fix fetches the spec by connecting **only to the IP(s) validation already pinned**, never re-resolving the hostname:

- **`utils/openapi.py`**
  - `_pinned_fetch()` — dials the pinned IP literal with `Host` + `sni_hostname` set to the validated hostname; `follow_redirects=False` with explicit 30x rejection; a 10 MiB streamed size cap (the old loader read bodies unbounded).
  - `_validate_url_sync()` — bridges the async validator into the sync loader, safe even when called from inside the running event loop.
  - `load_openapi_spec()` — new `validated_url` param; a bare `url` is validated + pinned **in-function**, so there is no code path that fetches a spec un-pinned. `SSRFError`/`SSRFFetchError` are re-raised immediately (not retried away).
- **`server.py`** — threads the `ValidatedURL` from validation into `load_openapi_spec()` for additional specs; the primary spec is validated + pinned by the loader with the `--allow-insecure-http` / `--allow-private-networks` flags passed through.

> **Why not reuse `fastmcp.server.auth.ssrf.ssrf_safe_fetch`?** In the pinned `fastmcp==3.3.1` it re-resolves DNS internally (ignoring pre-computed IPs), is HTTPS-only (breaks `--allow-insecure-http`), caps bodies at 5 KB, and is async (would `RuntimeError` inside the sync loader called from the running loop). We reuse its primitives (`ValidatedURL`, `format_ip_for_url`, error types) instead.

### User experience

- **Before:** A spec URL whose hostname rebinds between validation and fetch could steer the server into fetching internal/metadata endpoints; the primary spec URL bypassed SSRF checks entirely; response bodies were read unbounded.
- **After:** The spec is fetched only from the exact IP that passed validation (DNS resolved once). Redirects to internal hosts are refused, oversized bodies are rejected, and the primary spec path is validated too. The `--allow-insecure-http` / `--allow-private-networks` opt-ins are preserved.

## Checklist

* [ ] Changes are documented

**RFC issue number**:

Checklist:

## Acknowledgment

### DIFF
--- a/src/openapi-mcp-server/CHANGELOG.md
+++ b/src/openapi-mcp-server/CHANGELOG.md
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Security
+- **Fixed DNS-rebinding TOCTOU SSRF in spec fetching** ([CWE-350](https://cwe.mitre.org/data/definitions/350.html) / [CWE-367](https://cwe.mitre.org/data/definitions/367.html)): `spec_url` was validated via DNS resolution and then re-fetched via a second, independent resolution, allowing a rebinding host to pass validation with a public IP and be fetched from an internal/metadata IP. Spec URLs are now fetched by connecting only to the IP(s) that passed validation (DNS resolved once; `Host`/SNI preserved), redirects are refused, and response bodies are capped at 10 MiB. The primary spec URL is now validated too (previously unchecked).
+
+### Changed
+- Spec URL loading is DNS-pinned end to end; the previous README note about deploying behind an egress proxy for "full DNS pinning" no longer applies.
+
 ## [1.1.0] - 2026-05-31
 
 ### Added
--- a/src/openapi-mcp-server/README.md
+++ b/src/openapi-mcp-server/README.md
@@ -329,17 +329,17 @@ The server validates all URLs in `additional_specs` entries before fetching:
 
 **Credential isolation**: Additional specs never inherit the primary API's credentials. Each entry must declare its own `auth_type`/`auth_token` or defaults to no auth.
 
+**DNS pinning (rebinding protection)**: Spec URLs are validated **and fetched against the same resolved IP** — the loader connects only to the address(es) that passed validation, never re-resolving the hostname at fetch time. This closes the DNS-rebinding TOCTOU window where a hostname could resolve to a public IP during validation and to an internal/metadata IP during the fetch. This applies to both the primary spec URL and every `additional_specs` entry. Redirects are not followed (a `30x` toward an internal host is rejected), and spec bodies are capped at 10 MiB.
+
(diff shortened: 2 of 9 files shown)


theagenticguy (approved):
## Approve

The blocker from the prior review is fixed correctly and CI is green. Verified against the current branch HEAD (`a78583b5`), not a cached diff.

### The fix is exactly right
- **Stray raise deleted.** The unconditional `raise` after the docstring is gone; the guard at the top of `load_openapi_spec` now raises only when `url`, `path`, and `validated_url` are all empty.
- **Terminal raise correctly repositioned** to the end of the function, after the `if path:` block, as the genuine fall-through point. Every real branch above it returns or raises, so it satisfies the no-fall-through linter without shadowing a live path.
- **Scope is clean** — 4+/3- in `utils/openapi.py`, nothing collateral.

### Bonus: a real test fix rode along
The commit also corrected `test_path_invalid_validation` to patch `validate_openapi_spec` at the name where it's *used* (`...utils.openapi.validate_openapi_spec`, bound at import) rather than where it's *defined*. The old patch target never took effect, so that test wasn't actually exercising the invalid-spec path. Legitimate correctness fix, not a test loosened to pass.

### Security core (carried over, still holds)
Verified against the pinned `fastmcp-slim==3.3.1`:
- DNS pinning to `resolved_ips` with `Host`/SNI bound to the validated hostname closes the TOCTOU window — DNS resolves once.
- Redirect refusal (explicit 30x rejection, `follow_redirects=False`) and the 10 MiB streamed cap close the redirect-bypass and unbounded-body holes.
- `SSRFError`/`SSRFFetchError` re-raise before the retry loop — security failures aren't retried away.
- The primary spec URL is now validated (previously fetched with a bare `httpx.get`).
- The `except ValueError` in the Content-Length block can't swallow the oversized-body `SSRFFetchError`, since `SSRFFetchError` extends `Exception`, not `ValueError`.

### Non-blocking (optional, fine to defer)
1. `_pinned_fetch` is synchronous and runs on the loop thread when called from `create_mcp_server_async`. This matches the pre-existing `httpx.get` behavior (not a regression) and is startup-path only — worth a `# TODO`.
2. `_validate_url_sync` spins a fresh `ThreadPoolExecutor(max_workers=1)` per call — primary spec at startup only, so negligible.

`codecov/patch` and `codecov/project` both pass; the 16 failures from the earlier HEAD are cleared. Good to merge.

🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

prajwalendra (approved):

PR #4109, opened 2026-07-10
feat(redshift-mcp-server): use SHOW commands for metadata discovery

saeedma8:
## Summary

Updates the redshift-mcp-server metadata-discovery tools (`list_databases`, `list_schemas`, `list_tables`, `list_columns`) to use native Redshift `SHOW` commands (`SHOW DATABASES`, `SHOW SCHEMAS FROM DATABASE`, `SHOW TABLES FROM SCHEMA`, `SHOW COLUMNS FROM TABLE`) for metadata discovery, aligning with standard SQL metadata-discovery commands.

## Changes

- Renamed the discovery SQL constants to `DATABASES_SQL` / `SCHEMAS_SQL` / `TABLES_SQL` / `COLUMNS_SQL`.
- Embed the database/schema/table names as quoted SQL identifiers via a new `_sql_identifier` helper (the `SHOW` commands take identifiers, not Data API bind parameters). Embedded double quotes are doubled so a name cannot break out of the identifier, keeping the read-only path injection-safe.
- Handle the `SHOW TABLES` column order (`table_type` precedes `table_acl`) and ignore any additional trailing columns, so parsed output matches the prior output.
- Updated tool/model docstrings to reference the SHOW commands.

## Behavior notes

- `SHOW DATABASES` may additionally return `awsdatacatalog`.
- `SHOW SCHEMAS` does not return `pg_catalog`.

All other output (values, types, ACLs, ordinal positions) is unchanged.

## Before and after using SHOW comparison in both provisioned and serverless:

   |           | Provisioned                                    | Serverless                           |
   |-----------|------------------------------------------------|--------------------------------------|
   | databases | 5 common identical (+awsdatacatalog from SHOW) | 2 common identical (+awsdatacatalog) |
   | schemas   | 5 common identical (−pg_catalog in SHOW)       | 3 common identical (−pg_catalog)     |
   | tables    | 7 = 7, all fields identical                    | 7 = 7, all fields identical          |
   | columns   | 18 = 18, all fields identical                  | 18 = 18, all fields identical        |

## Testing

- Unit tests: 169 passed, ruff clean.
- Verified against live provisioned and serverless clusters by comparing prior and new discovery output field-by-field.
- See E2E testing in the comments.

## Checklist

* [x] Changes are documented

**Is this a breaking change? (Y/N):** Yes, fully tested. The only differences are discovery-output scope (see [Behavior notes](#behavior-notes)): `list_databases` may add `awsdatacatalog` and `list_schemas` no longer lists `pg_catalog`. Everything else is identical. `pg_catalog` is still fully queryable directly (`pg_catalog.*`) — it's just not enumerated, as it's a well-known always-present system schema.

**RFC issue number**: N/A

Checklist:

* (N/A) Migration process documented
* (N/A) Implement warnings (if it can live side by side)

## Acknowledgment


src/redshift-mcp-server/tests/test_redshift.py
```diff
@@ -1188,6 +1191,39 @@ async def test_discover_databases(self, mocker):
         assert result[0]['database_owner'] == 100
         assert result[0]['database_type'] == 'local'
 
+    @pytest.mark.asyncio
+    async def test_discover_databases_uses_show_databases(self, mocker):
```
grayhemp: That should be in test_discover_databases(). No need a separate test for it.

grayhemp (changes requested):

src/redshift-mcp-server/awslabs/redshift_mcp_server/redshift.py
```diff
@@ -24,17 +24,23 @@
...
-    SVV_REDSHIFT_DATABASES_QUERY,
+    TABLES_SQL,
 )
 from awslabs.redshift_mcp_server.sql_guard import assert_executable
 from botocore.config import Config
 from loguru import logger
+from sqlglot import exp
+
+
+def _sql_string_literal(value: str) -> str:
```
grayhemp: _sql_string_literal() should be _sql_identifier() and should use exp.to_identifier instead of exp.Literal.

grayhemp (dismissed):

grayhemp:
## E2E Testing summary — branch pr-4109 (Redshift MCP, SVV views → SHOW commands)

Cluster: redshift-cluster-analytics, db sample_data_dev, schema tickit.

| # | Scenario (from unit test) | Result |
|---|---|---|
| 1 | list_databases → SHOW DATABASES, no bind params | PASS — 3 dbs (awsdatacatalog, dev, sample_data_dev) |
| 2 | list_schemas → SHOW SCHEMAS FROM DATABASE '<db>' literal | PASS — 3 schemas (information_schema, public, tickit) |
| 3 | list_tables → SHOW TABLES, table_type/table_acl positional remap | PASS — table_type=TABLE, table_acl=awsuser=rRxtPA/... correctly ordered, not swapped (7 tables) |
| 4 | list_columns → SHOW COLUMNS, type/precision/nullable mapping | PASS — 18 cols on users, types/ordinals/nullability correct |
| 5 | db-name escaping / injection (x'; DROP TABLE users; --) | PASS (blocked) — rejected with ValidationException at DbName layer; no multi-statement executed. SQL-literal escaping remains as defense-in-depth (unit-only reachable) |

All 4 discovery tools work end-to-end against the new SHOW-command implementation; the column-order fix and injection safety both hold.


## E2E Testing summary - full

Env: 3 clusters (2 provisioned, 1 serverless). Tested redshift-cluster-analytics (provisioned) + redshift-workgroup-etl (serverless).

Per-tool (all PASS, both cluster types):
- **list_clusters** — Returned 3 clusters; correct type (provisioned/serverless), status, endpoint, node/encryption fields.
- **list_databases** — Both clusters: returned dev, sample_data_dev, awsdatacatalog; type + isolation-level fields populated.
- **list_schemas** — Both: returned information_schema, public, tickit; owner/acl/type fields correct.
- **list_tables** — Both: 7 tickit tables (category/date/event/listing/sales/users/venue); type=TABLE, acl populated.
- **list_columns** — Both: tickit.sales 10 columns with ordinal, nullable, data_type, precision/scale.
- **execute_query** — Both: SELECT/aggregation returned rows; data types mapped (integer, numeric, timestamp, boolean).

Protection tests (all PASS):
- **Read-only deny-list** — TRUNCATE → rejected Statement type not allowed in read-only mode: TRUNCATE (guard, pre-engine).
- **Single-statement** — SELECT 1; SELECT 2 → rejected Only a single SQL statement is allowed.
- **Parse fail-closed** — SELECT FROM WHERE → rejected SQL could not be parsed.
- **Transaction breaker (engine backstop)** — CREATE TABLE (not deny-listed) passed guard, engine rejected transaction is read-only; follow-up confirmed table never persisted (ROLLBACK worked).
- **Failed user SQL** — SELECT * FROM tickit.no_such_table_xyz → engine error relation ... does not exist propagated; session recovered on next query (ROLLBACK closed txn cleanly).

All 6 tools + 5 safety behaviors verified across provisioned and serverless. No failures. Probe write confirmed non-persisted.

grayhemp (approved):

scottschreckengaust (approved):
LGTM

PR #4119, opened 2026-07-12
feat(healthomics): add workflow_type filter to ListWorkflows tool

steditt:
### Changes

Add an optional `workflow_type` parameter (PRIVATE or READY2RUN) to the `list_workflows` tool (exposed as `ListAHOWorkflows`). This forwards the `type` parameter to the HealthOmics `ListWorkflows` API, enabling discovery of AWS-provided Ready2Run workflows.

- Add `workflow_type: Optional[str]` parameter with validation using existing `GetWorkflowType` enum
- Include FastMCP compatibility handling
- Forward `type` to boto3 when provided; omit when `None` (backward compatible)
- Add 4 new unit tests covering READY2RUN, PRIVATE, invalid, and omitted type scenarios

### User experience

**Before:** `ListAHOWorkflows` only returns private (user-created) workflows. There is no way to browse the AWS-provided Ready2Run catalog via the MCP server.

**After:** `ListAHOWorkflows` accepts an optional `workflow_type` parameter. Passing `workflow_type='READY2RUN'` returns the full Ready2Run workflow catalog. Omitting the parameter preserves existing behavior (lists private workflows only).

## Checklist

- [x] Changes are documented

**RFC issue number**: N/A (minor additive change following existing patterns)

Checklist:
- [x] Migration process documented
- [x] Implement warnings (if it can live side by side)

## Acknowledgment


src/aws-healthomics-mcp-server/awslabs/aws_healthomics_mcp_server/tools/workflow_management.py
```diff
@@ -78,19 +82,38 @@ async def list_workflows(
...
 
-    params: dict[str, Any] = {'maxResults': max_results}
-    if next_token:
-        params['startingToken'] = next_token
+        # Validate workflow_type parameter
+        if workflow_type is not None:
+            try:
+                GetWorkflowType(workflow_type)
+            except ValueError:
+                error_message = ERROR_INVALID_WORKFLOW_TYPE.format(
```
markjschreiber: The MCP has error handling utilities that ensure correct handling and communication with the agent. I recommend using these using the same patterns used by other validations.

markjschreiber (commented):
Can you take a look at the MCPs error handling utility functions and use these when validating the workflow type?

src/aws-healthomics-mcp-server/awslabs/aws_healthomics_mcp_server/tools/workflow_management.py
steditt: Done — extracted a validate_workflow_type utility function in `validation_utils.py` following the same pattern as `validate_provider_type`. Both `list_workflows` and `get_workflow` now call the shared validator instead of inline `try/except` blocks. All 163 tests pass. See commit c3a4da2.

markjschreiber (approved):
LGTM. Thanks for the contribution.

src/aws-healthomics-mcp-server/awslabs/aws_healthomics_mcp_server/utils/validation_utils.py
```diff
@@ -589,6 +589,39 @@ async def validate_provider_type(
...
+        return None
+
+    valid_types = [wt.value for wt in GetWorkflowType]
+    if workflow_type not in valid_types:
+        error_message = (
+            f"Invalid workflow type '{workflow_type}'. Must be one of: {', '.join(valid_types)}"
+        )
+        logger.error(error_message)
+        await ctx.error(error_message)
+        raise ValueError(error_message)
```
markjschreiber: I have found this pattern to be problematic because not all agents communicate the `ctx` to the agents LLM. Therefore a tool failure can be silent and the model will not have access to the `error_message`. I recommend using this utility method https://github.com/awslabs/mcp/blob/a024ce485007cb6d9503455bf689c1407d87c3a0/src/aws-healthomics-mcp-server/awslabs/aws_healthomics_mcp_server/utils/error_utils.py#L22 in the appropriate place so that the agents LLM get's useable information about what happened.

markjschreiber (changes requested):

src/aws-healthomics-mcp-server/awslabs/aws_healthomics_mcp_server/utils/validation_utils.py
steditt: Hi Mark,

I changed how errors are surfaced to the LLM. I'm now using the `handle_tool_error()` function.

markjschreiber (approved):
Thanks for the fix

alxawan (approved):
LGTM - thank you for adding this!

PR #4121, opened 2026-07-13
fix(eks-mcp-server): support EKS CAs without AKI

### LINKED ISSUE #3791: eks-mcp-server: TLS verification fails ("Missing Authority Key Identifier") against EKS cluster CA on OpenSSL 3.x + urllib3>=2
### Describe the bug

Every Kubernetes operation in `eks-mcp-server` (`list_k8s_resources`,
`manage_k8s_resource`, `get_pod_logs`, `get_k8s_events`, …) fails while
establishing the TLS connection to the EKS cluster API server:

```
Failed to get cluster credentials: HTTPSConnectionPool(
  host='<cluster-id>.<sub>.<region>.eks.amazonaws.com', port=443):
  Max retries exceeded with url: /version
  (Caused by SSLError(SSLCertVerificationError(1,
   '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed:
    Missing Authority Key Identifier (_ssl.c:1032)')))
```

`eks:DescribeCluster` and STS token retrieval succeed — the failure is purely
the TLS handshake to the cluster API endpoint on the `/version` probe.

**Root cause:** EKS auto-generates the cluster CA as a minimal self-signed
root (`subject=CN=kubernetes`, `issuer=CN=kubernetes`) whose only X509v3
extension is `Basic Constraints: critical, CA:TRUE` — it has **no
Subject/Authority Key Identifier**. RFC 5280 §4.2.1.1 explicitly allows
omitting AKI for a self-signed CA, so the certificate is valid.

On OpenSSL 3.x (3.5.3 here), `ssl.create_default_context()` enables
`VERIFY_X509_STRICT`, which *requires* AKI. **urllib3 >= 2** builds its SSL
context from `ssl.create_default_context()`, so the `kubernetes` Python client
that eks-mcp-server uses inherits the strict flag and rejects the cluster CA.

`kubectl`, the AWS CLI, and Go-based clients accept the same endpoint without
issue — only this Python / urllib3-v2 / OpenSSL-3.x path is affected.

### Expected Behavior

`eks-mcp-server` connects to the EKS cluster API and verifies the API server
certificate against the cluster CA — regardless of the strict-TLS defaults of
the host Python / OpenSSL / urllib3 stack. The cluster CA it must trust is
generated by EKS itself.

### Current Behavior

Every k8s call returns `Failed to get cluster credentials: ... /version ...
CERTIFICATE_VERIFY_FAILED: Missing Authority Key Identifier`. The server is
effectively unusable against a standard EKS cluster on this stack.

Setting `PYTHONHTTPSVERIFY=0`, `SSL_CERT_FILE=/dev/null`, or empty
`REQUESTS_CA_BUNDLE` / `CURL_CA_BUNDLE` does **not** help — the kubernetes
client builds its own verifying SSL context and ignores those, and the failure
is a strictness check (missing AKI), not a missing-CA error.

### Reproduction Steps

```python
import ssl, socket
# Cluster CA, e.g.:
#   aws eks describe-cluster --name <c> --query cluster.certificateAuthority.data \
#     --output text | base64 -d > eks-ca.pem
ca   = "eks-ca.pem"
host = "<cluster-id>.<sub>.<region>.eks.amazonaws.com"   # cluster API endpoint

ctx = ssl.create_default_context(cafile=ca)
print("strict:", bool(ctx.verify_flags & ssl.VERIFY_X509_STRICT))   # True on OpenSSL 3.5
with socket.create_connection((host, 443)) as s:
    ctx.wrap_socket(s, server_hostname=host)        # -> Missing Authority Key Identifier

# Clearing the strict flag (still verifies against the cluster CA) -> succeeds:
ctx2 = ssl.create_default_context(cafile=ca)
ctx2.verify_flags &= ~ssl.VERIFY_X509_STRICT
with socket.create_connection((host, 443)) as s:
    ctx2.wrap_socket(s, server_hostname=host)        # -> OK (TLSv1.3)
```

Confirming the CA lacks AKI:

```
$ openssl x509 -in eks-ca.pem -noout -text | grep -A1 "Key Identifier\|Basic Constraints"
        X509v3 Basic Constraints: critical
            CA:TRUE
# (no Subject Key Identifier / Authority Key Identifier present)
```

Reproduced inside the server's own resolved environment
(`uvx --from awslabs.eks-mcp-server==0.1.31`): `urllib3 2.7.0`,
`create_urllib3_context().verify_flags & VERIFY_X509_STRICT == True`,
handshake fails with `Missing Authority Key Identifier`.

### Possible Solution

When building the kubernetes client's SSL context, clear the strict flag:

```python
ctx.verify_flags &= ~ssl.VERIFY_X509_STRICT
```

The connection still verifies the API server certificate against the cluster
CA, so there is **no security regression** — `X509_STRICT` / AKI adds nothing
for a self-signed cluster root. Alternatively, constrain `urllib3 < 2` in the
server's dependency set (its context is not strict).

**Workaround** (confirmed end-to-end on 0.1.31 and 0.1.22) — pin urllib3 < 2 at launch:

```jsonc
"awslabs.eks-mcp-server": {
  "command": "uvx",
  "args": ["--with", "urllib3<2", "awslabs.eks-mcp-server==0.1.31", "--allow-sensitive-data-access"]
}
```

Resolves to `urllib3 1.26.20` / `boto3 1.43.25` / `kubernetes 36.0.2`; with it,
`list_k8s_resources` / `get_pod_logs` etc. succeed against the cluster
(verified via the running MCP server, not just an isolated handshake).
Note the pin is fragile long-term (botocore will eventually drop urllib3<2
support), so a fix in the server is preferable.

### OS

macOS (Apple Silicon), Python 3.13

### Server

eks-mcp-server

### Server Version

0.1.31 (latest) and 0.1.22 — both affected

eu-central-1

### Other information

- urllib3 2.7.0, kubernetes 36.0.2, boto3 1.43.25, OpenSSL 3.5.3, Python 3.13
- EKS cluster API endpoint is public; standard EKS-generated cluster CA
- `kubectl` / `aws eks` against the same cluster work fine (lenient TLS stacks)
- Likely affects any EKS cluster reached from a modern OpenSSL 3.x + urllib3>=2
  environment — not specific to one cluster.

### Service quota

jstar0:
Fixes #3791

### Changes

- Install an explicit urllib3 SSL context on the IAM-backed EKS Kubernetes client.
- Keep certificate verification required while disabling only `VERIFY_X509_STRICT`, which rejects valid self-signed EKS cluster CAs that omit an Authority Key Identifier.
- Add regression coverage that verifies the client still uses the EKS CA and `CERT_REQUIRED` without strict AKI enforcement.

### User experience

Before this change, EKS MCP Kubernetes operations using IAM authentication can fail during the `/version` TLS probe on OpenSSL 3.x with urllib3 2.x with `certificate verify failed: Missing Authority Key Identifier`.

After this change, the IAM-backed client continues to verify the API server certificate against the EKS-provided CA, but accepts the self-signed CA shape generated by EKS. Kubeconfig-created clients are unchanged.

## Checklist

* [ ] Changes are documented

N

**RFC issue number**: N/A

Checklist:

## Verification

```bash
uv run --frozen pytest tests/test_k8s_client_cache.py tests/test_k8s_apis.py -q
uv run --frozen pytest --ignore=tests/test_path_validation.py -q
uv run --frozen ruff check awslabs/eks_mcp_server/k8s_apis.py tests/test_k8s_apis.py
uv run --frozen ruff format --check awslabs/eks_mcp_server/k8s_apis.py tests/test_k8s_apis.py
uv run --frozen pyright awslabs/eks_mcp_server/k8s_apis.py
```

Results: `83 passed`, `288 passed`, and all static checks passed.

The unfiltered local macOS suite passed 309 tests and retained four existing `tests/test_path_validation.py` failures because `/etc` resolves to `/private/etc` on macOS.

## Acknowledgment


src/eks-mcp-server/awslabs/eks_mcp_server/k8s_apis.py
```diff
@@ -134,6 +137,14 @@ def _configure_proxy_settings(self, config):
         logger.debug(f'Configuring proxy: {proxy_url}')
         config.proxy = proxy_url
 
+    def _configure_eks_ca_ssl_context(self, api_client):
+        """Configure Kubernetes TLS verification for EKS-generated cluster CAs."""
+        context = create_urllib3_context(cert_reqs=ssl.CERT_REQUIRED)
+        if hasattr(context, 'verify_flags'):
```
srhsrhsrhsrh, line 143: Thanks for the fix. For codecov to pass, could you patch `create_urllib3_context` to mock the other branch?

srhsrhsrhsrh (approved):

patrick-yu-amzn (approved):
LGTM

PR #4124, opened 2026-07-13
feat(openapi): migrate openapi-mcp-server to native FastMCP.from_openapi (RFC #4122)

### LINKED ISSUE #4122: RFC: Deprecate OpenAPI MCP Server
We propose deprecating `awslabs.openapi-mcp-server`: all of its features are believed to be covered by the upstream Apache-2.0 licensed **FastMCP** (`PrefectHQ/fastmcp`), which it already wraps. In-process users move to `FastMCP.from_openapi`; standalone users run a small entrypoint via `fastmcp run`. Please comment if you rely on a capability you believe FastMCP does not cover.

scottschreckengaust:
## Summary

**Proof-of-concept** that `awslabs.openapi-mcp-server` can be migrated onto the upstream Apache-2.0 **FastMCP** (`PrefectHQ/fastmcp`) it already wraps — validating the deprecation proposed in **#4122**.

This converts the package's internals from the **low-level** construction (`OpenAPIProvider(...)` + `FastMCP(providers=[...])`) to the **native high-level** `FastMCP.from_openapi(...)`, and replaces the one bespoke reimplementation (description enrichment) with FastMCP's own shipped utility. The CLI, config, and entrypoint are unchanged, so existing users keep working.

> Draft: not for merge as-is. See **Known follow-ups** before promoting.

## What changed

| Before (bespoke) | After (native FastMCP) |
|---|---|
| `OpenAPIProvider(**kwargs)` + `FastMCP(providers=[...])` | `FastMCP.from_openapi(spec, client=..., route_maps=..., mcp_component_fn=..., validate_output=...)` |
| Hand-rolled `enrich_component` building `desc \| Returns: ... \| Example: ...` | `fastmcp.utilities.openapi.format_description_with_responses(...)` (structured `**Responses:**` / `**Parameters:**`) |
| Additional specs appended to a `providers=[]` list | `server.add_provider(...)` on the `from_openapi` server |
| Server `instructions` string | preserved — forwarded through `from_openapi`'s `**settings` |

Files:
- `src/openapi-mcp-server/awslabs/openapi_mcp_server/server.py` — the migration.
- `src/openapi-mcp-server/tests/test_new_features.py` — enrichment tests updated to the native format.
- `src/openapi-mcp-server/MIGRATION.md` — customer-facing migration guide.
- `src/openapi-mcp-server/docs/migration/` — full deprecation evaluation + three upstream contribution drafts.

## Enrichment: before vs. after

For `listPets` (a GET with a `status` query param, 200 response):

**Before** (bespoke):
```
List all pets | Returns: 200 | Example: status=available
```
**After** (native `format_description_with_responses`):
```
List all pets

**Query Parameters:**

- **status**: No description.

**Responses:**

- **200** (Success): OK
```
The native formatter emits richer structure but does **not** inline `name=value` enum/example samples the old builder synthesized. This is the intended behavior change (see #4122 discussion on making enrichment first-class upstream).

## Test status

`uv run pytest` → **495 passed, 0 failed** (was 500 pre-change).

- **+5** enrichment tests updated to assert the native structured output.
- **−6** redundant tests removed: they asserted low-level `OpenAPIProvider`/`FastMCP` construction mechanics (`assert_called_once`, `call_args[1]` kwargs) that `from_openapi` now owns internally — behavior is covered by `test_new_features.py`. (Includes duplicate copies in `test_server_part1.py`.)
- **3** tests that assert real wrapper logic (auth-provider registration, httpx-version fallback, prompt generation) repointed to patch `FastMCP.from_openapi`.

## Known follow-ups (before this is merge-ready)

1. **Coverage check.** Re-run `pytest --cov` to confirm no real coverage regression once construction-mechanic tests are removed vs. relocated.
2. **Upstream give-backs (optional, non-blocking):** the three drafts in `docs/migration/` (enrichment DX, SSRF-safe spec fetch, Prometheus middleware) make the remaining glue first-class for the community.

## How to reproduce locally

```
git worktree add ../mcp-poc-fastmcp feat/poc-wrap-openapi-mcp-server-with-fastmcp
cd ../mcp-poc-fastmcp/src/openapi-mcp-server
uv sync --all-extras --dev
uv run pytest -q            # 495 passed, 0 failed
```

Related: **#4122** (RFC: Deprecate OpenAPI MCP Server).

---


src/openapi-mcp-server/MIGRATION.md
```diff
@@ -0,0 +1,128 @@
...
+**FastMCP** (`PrefectHQ/fastmcp`) — the same engine it already imports and
+builds on. Every capability the wrapper adds is reachable today with FastMCP's
+own primitives, so the wrapper can be deprecated without loss of function.
+
+## What maps to what
+
+| Wrapper feature | Native FastMCP equivalent |
+|---|---|
+| Server construction | `FastMCP.from_openapi(spec_dict, client=httpx_client)` |
+| Cognito auth | `fastmcp.server.auth.providers.aws.AWSCognitoProvider` |
```
theagenticguy: This row maps in the wrong direction and is the one defect here that can actively break a migrating user.

`awslabs/openapi_mcp_server/auth/cognito_auth.py` is **outbound** auth. `CognitoAuthProvider(BearerAuthProvider)` calls `initiate_auth` with `USER_PASSWORD_AUTH` (falling back to `admin_initiate_auth`), pulls the token out of `AuthenticationResult`, and `get_auth_headers()` returns an `Authorization: Bearer ...` header that `HttpClientFactory.create_client` puts on the httpx client used to call the **upstream API**.

`fastmcp.server.auth.providers.aws.AWSCognitoProvider` is **inbound** auth. Verified against 3.4.4:

```
MRO: AWSCognitoProvider -> OIDCProxy -> OAuthProxy -> OAuthProvider -> AuthProvider -> TokenVerifier
docstring: "add AWS Cognito OAuth protection to any FastMCP server ... mcp = FastMCP("My App", auth=auth)"
```

It validates JWTs presented by MCP **clients** connecting to the server. A user who follows this row ends up with their MCP server gated behind Cognito while their API requests go out unauthenticated, which is a silent failure in the direction that matters.

The correct native equivalent is the same one this table already gives for Basic/Bearer/API-key: set the header on the caller-supplied `httpx.AsyncClient`. There is no upstream equivalent for the token-acquisition and auto-refresh half (`_check_and_refresh_token_if_needed`), so that stays user glue. Suggest:

| Cognito auth (outbound, to the API) | acquire the token yourself (boto3 `initiate_auth`), then set `Authorization` on the `httpx.AsyncClient`. `AWSCognitoProvider` is **not** the equivalent: it protects the MCP server's own inbound connections. |

Grepped `server/auth/` in 3.4.4 for an outbound helper and found nothing applicable.

src/openapi-mcp-server/docs/migration/README.md
```diff
@@ -0,0 +1,34 @@
...
+| [`upstream-issue-3-prometheus-metrics.md`](./upstream-issue-3-prometheus-metrics.md) | Draft upstream `contrib/` enhancement — a Prometheus metrics middleware on the existing `Middleware` seam. |
+
+## Migration summary
+
+Every feature of this wrapper is reachable today with FastMCP primitives:
+
+| Wrapper feature | FastMCP equivalent |
+|---|---|
+| Server construction | `FastMCP.from_openapi(spec, client=...)` |
+| Cognito auth | `fastmcp.server.auth.providers.aws.AWSCognitoProvider` |
```
theagenticguy: Same inverted Cognito mapping as `MIGRATION.md:20`. See that comment for the MRO evidence and the suggested replacement row. Worth fixing in all three places at once so the docs do not disagree with each other.

src/openapi-mcp-server/docs/migration/deprecation-evaluation.md
```diff
@@ -0,0 +1,244 @@
...
+---
+
+## Details
+
+### Where the capabilities live now
+Source inspection of FastMCP 3.4.4 shows **every wrapper feature is reachable today** — natively, via a hook, or via a shipped primitive. (Full evidence + FOSS survey: Appendix G.)
+
+| Wrapper feature | Upstream today |
+|---|---|
+| Cognito auth | `AWSCognitoProvider` — native |
```
theagenticguy: Third instance of the inverted Cognito mapping, and this is the one that carries the most weight, since "native" here is doing real work in the recommendation.

`AWSCognitoProvider` is native for **inbound** MCP-client authentication. This wrapper's Cognito support is **outbound** API authentication. Nothing upstream covers the outbound path, so this row should read as glue rather than native.

Same correction applies to line 105 in Appendix E, which routes `cognito_auth.py` to `server/auth/providers/*` (`aws.py` for Cognito).

src/openapi-mcp-server/docs/migration/deprecation-evaluation.md
```diff
@@ -0,0 +1,244 @@
...
+## Appendix A — Verification & Confidence
+- Research method: 6 search angles → 16 sources fetched → 75 claims → 25 verified under 3-vote adversarial verification (24 confirmed, 1 refuted). Licenses confirmed against raw `LICENSE` files, not just GitHub badges.
+- **High confidence:** all license classifications; FastMCP maturity/backing; `ivo-toby` and `harsha-iiiv` feature sets.
+- **Medium confidence:** the gap analysis (rests on absence-of-documentation).
+- **Refuted (1):** a claim that higress lacks bearer/OAuth2/filtering — its exact auth surface is genuinely uncertain.
+
+---
+
+## Appendix B — Open Questions
+1. How many users depend on SSRF-fetch or Prometheus metrics? Cognito, tag filtering, and multi-spec are already upstream; only these two need sizing for the migration guide (not as a gate).
```
theagenticguy: This count depends on the Cognito row above being correct. Once Cognito is reclassified as outbound-only glue, it is three features needing sizing (Cognito, SSRF-fetch, Prometheus), not two.

That does not change the recommendation, since Cognito is still reachable with a few lines of boto3 plus a header. It does change the migration-support budget this appendix is trying to estimate, and Cognito is plausibly the most-used of the three.

laithalsaadoon (commented):
## Review: Comment

Verified every upstream claim against FastMCP 3.4.4 source, then mutation-tested the removed assertions to check whether they were load-bearing. Baseline: **517 passed** on this branch, **522** on `main`.

The migration itself is sound. `from_openapi` accepts every kwarg passed here and forwards `instructions` through `**settings` exactly as the comment claims; `add_provider(provider, namespace='')` is public and correct for the multi-spec path; `ssrf_safe_fetch`, `AWSCognitoProvider`, `Middleware`/`TimingMiddleware`, and `RouteMap(tags=...)` all exist as documented, and the 5 KB `max_size` plus HTTPS-only behavior match the notes. The Swagger 2.0 analysis is right: `prance` is what gives this wrapper 2.0 support today. `ruff` clean, CI green.

Two items block promoting this out of draft. Details inline.

**1. The Cognito row in the migration table is inverted** (`MIGRATION.md:20`, `docs/migration/README.md:25`, `docs/migration/deprecation-evaluation.md:20`)

The wrapper's `cognito_auth.py` is outbound: it calls `initiate_auth`/`admin_initiate_auth`, extracts the token, and sets `Authorization: Bearer ...` on the httpx client that calls the upstream API. FastMCP's `AWSCognitoProvider` is inbound: its MRO is `OIDCProxy -> OAuthProxy -> OAuthProvider -> AuthProvider -> TokenVerifier`, and its docstring describes adding "Cognito OAuth protection to any FastMCP server." It validates JWTs arriving from MCP clients.

A user who follows that row gets their MCP server gated behind Cognito while their API calls go out unauthenticated. This also moves the evaluation's headline count: Cognito belongs in the glue column alongside SSRF-fetch and Prometheus, so it is three glue cases, not two.

**2. A removed test was the only guard on SSRF flag forwarding** (`tests/test_server.py:42`)

`test_create_mcp_server_basic` asserted more than construction mechanics. It also pinned `allow_http` and `allow_private_networks` onto `load_openapi_spec`. Mutation proof below: an edit that silently opts every user into plaintext and private-network spec fetching now ships green. Codecov's "all modified lines covered" is consistent with this, since the loss is assertion coverage rather than line coverage, which sharpens follow-up #1 in the description: `pytest --cov` will not surface it.

**3. One justification comment claims coverage that does not exist** (`tests/test_server_extended.py:46`). Not a regression, since the forwarding is untested on both branches, but the note should say so.

---

### What would make this an Approve

1. Fix the Cognito row in all three docs to point at header-on-`httpx.AsyncClient` (matching the Basic/Bearer/API-key rows), note that the token-acquisition half has no upstream equivalent, and move Cognito into the glue column so the "only these two" count in `deprecation-evaluation.md:56` reads three.
2. Restore one `mock_load_spec.assert_called_once_with(...)` asserting `allow_http` and `allow_private_networks` are forwarded from config. This reintroduces no `from_openapi` internals. Verify by re-running the mutation below and confirming it fails.
3. Correct or drop the route-map justification comment in `tests/test_server_extended.py`.

Nice-to-have, not gating: hoist the `format_description_with_responses` import to module level, and consider a parity assertion for the auth headers / httpx auth / cookies reaching `HttpClientFactory.create_client`. That last gap predates this PR on `main`, so it is a separate change.

With those three landed, this is an Approve. The `from_openapi` conversion, the `add_provider` composition, and the enrichment swap all check out against 3.4.4, and the enrichment behavior change is documented honestly in both the description and the updated tests.

---

<details>
<summary>Mutation test used for finding 2</summary>

Hardcode the flags open at the primary spec load (`server.py:103-104`):

```python
openapi_spec = load_openapi_spec(
    url=config.api_spec_url,
    path=config.api_spec_path,
    allow_http=True,                 # was config.allow_insecure_http
    allow_private_networks=True,     # was config.allow_private_networks
)
```

| Branch | Result |
|---|---|
| `main` | 2 failed, 520 passed (`test_create_mcp_server_basic` in both `test_server.py` and `test_server_part1.py`) |
| this branch | **517 passed** |

</details>

src/openapi-mcp-server/tests/test_server.py
```diff
@@ -39,46 +39,11 @@ def mock_config():
...
-    mock_load_spec.assert_called_once_with(
-        url=mock_config.api_spec_url,
-        path=mock_config.api_spec_path,
-        allow_http=mock_config.allow_insecure_http,
-        allow_private_networks=mock_config.allow_private_networks,
-    )
-    mock_validate.assert_called_once()
-    mock_create_client.assert_called_once()
-    mock_openapi_provider.assert_called_once()
+# NOTE: ``test_create_mcp_server_basic`` was removed in the FastMCP-native
```
theagenticguy, line 42: This note is accurate about what most of the removed test did, but the test also carried one assertion that was not a construction mechanic:

```python
mock_load_spec.assert_called_once_with(
    url=mock_config.api_spec_url,
    path=mock_config.api_spec_path,
    allow_http=mock_config.allow_insecure_http,
    allow_private_networks=mock_config.allow_private_networks,
)
```

That pinned the two SSRF opt-in flags to config values, and it is now gone. Mutation test: hardcode `allow_http=True, allow_private_networks=True` at `server.py:103-104`.

- `main`: 2 failed, 520 passed (this test plus its `test_server_part1.py` twin)
- this branch: **517 passed**

So a change that silently opts every user into plaintext and private-network spec fetching ships green. `tests/test_ssrf_protection.py` covers `validate_url_for_spec` thoroughly but never asserts that `create_mcp_server` forwards the operator's flags into it, and nothing in `tests/` references `allow_insecure_http` outside that file.

This is also why follow-up #1 in the PR description will come back clean: the loss is assertion coverage, not line coverage, which matches Codecov reporting all modified lines covered.

Suggest keeping a single focused test that asserts only the forwarding, with no reference to `from_openapi` or `OpenAPIProvider`, so it survives future upstream refactors.

src/openapi-mcp-server/tests/test_server_extended.py
```diff
@@ -39,74 +39,14 @@ def mock_config():
...
-    # Verify that OpenAPIProvider was called with custom route mappings
-    call_args = mock_openapi_provider.call_args[1]
-
-    # Check that route_maps was included in the kwargs
-    assert 'route_maps' in call_args
+# NOTE: ``test_create_mcp_server_with_query_params_routes`` was removed in the
+# FastMCP-native migration. Its only assertion inspected
+# ``OpenAPIProvider.call_args[1]`` for a ``route_maps`` kwarg — a low-level
+# construction detail now passed through to ``FastMCP.from_openapi``. The
+# route-map builder itself is exercised behaviorally by the GET-with-query-params
```
theagenticguy: The claim in this sentence does not hold. FastMCP 3.4.4's `DEFAULT_ROUTE_MAPPINGS` is:

```python
DEFAULT_ROUTE_MAPPINGS = [RouteMap(mcp_type=MCPType.TOOL)]
```

Every operation becomes a tool with no `route_maps` passed at all, so a GET-with-query-params test asserting the operation shows up in `list_tools()` passes identically whether `_build_route_maps` output is forwarded or not. Confirmed by mutation: setting `route_maps=None` at `server.py:260` keeps all 517 tests green.

Not a regression, since `main` does not catch it either (same mutation against the `'route_maps': custom_mappings` kwarg leaves 522 passing). The `_build_route_maps` unit tests at `test_new_features.py:418-443` do cover the builder itself. What is untested on both branches is that its output reaches the provider.

Either drop the second sentence, or narrow it to "`_build_route_maps` is unit-tested directly in `test_new_features.py`; forwarding into the provider is not asserted on either branch."

src/openapi-mcp-server/awslabs/openapi_mcp_server/server.py
```diff
@@ -226,37 +226,43 @@ def handle_auth_error(auth_type, error_message):
...
+            """Enrich MCP tool/resource descriptions with OpenAPI spec details.
+
+            POC migration: this delegates to FastMCP's own shipped formatter
+            (`fastmcp.utilities.openapi.format_description_with_responses`)
+            instead of the previous bespoke string-building. The upstream
+            formatter emits richer, structured sections (Path/Query Parameters,
+            Request Body, Responses with examples) than the old
+            ``desc | Returns: ... | Example: ...`` format.
+            """
+            from fastmcp.utilities.openapi import format_description_with_responses
```
theagenticguy: Nit: this import runs on every route the provider builds. The rest of the file imports `fastmcp` at module level (`from fastmcp import FastMCP`, `from fastmcp.server.providers.openapi import ...`), so hoisting this one matches the surrounding style and drops the per-component work. The deferred imports further down for `json` and `url_validator` are inside a conditional branch, so those reads differently.

src/openapi-mcp-server/MIGRATION.md
scottschreckengaust: Fixed in bc3d79f8ce3eb14a427e797eec212d05921f485f. `MIGRATION.md:20` now reads the Cognito row as outbound-only glue — acquire the token via boto3 `initiate_auth`, set `Authorization` on the `httpx.AsyncClient` (matching the Basic/Bearer/API-key row), with an explicit note that `AWSCognitoProvider` is inbound and has no upstream equivalent for the token-acquisition/auto-refresh half.

src/openapi-mcp-server/docs/migration/README.md
scottschreckengaust: Fixed in bc3d79f8ce3eb14a427e797eec212d05921f485f. `docs/migration/README.md:25` reclassified to outbound glue with the inbound-vs-outbound note. All three docs now agree.

src/openapi-mcp-server/docs/migration/deprecation-evaluation.md
scottschreckengaust: Fixed in bc3d79f8ce3eb14a427e797eec212d05921f485f. Both spots corrected: the capability table (line ~20) now describes Cognito as outbound glue (`AWSCognitoProvider` MRO ends in `TokenVerifier` → inbound), and the Appendix E `cognito_auth.py` row no longer routes to `server/auth/providers/aws.py` — it points at boto3-token + header-on-httpx, with a note that `AWSCognitoProvider` is a different (inbound) direction.

src/openapi-mcp-server/docs/migration/deprecation-evaluation.md
scottschreckengaust: Fixed in bc3d79f8ce3eb14a427e797eec212d05921f485f. The count now reads three glue cases (Cognito, SSRF-fetch, Prometheus), not two — updated in Appendix B (the open question), the Executive Summary recommendation, and the Risks bullet, with a note that Cognito is plausibly the most-used of the three. The recommendation is unchanged.

src/openapi-mcp-server/tests/test_server.py
scottschreckengaust: Fixed in bc3d79f8ce3eb14a427e797eec212d05921f485f. Restored the forwarding assertion as a focused test — `test_create_mcp_server_forwards_ssrf_flags` in `tests/test_server.py`. It asserts only that `create_mcp_server` forwards `allow_insecure_http`/`allow_private_networks` into `load_openapi_spec`, with no reference to `from_openapi` or `OpenAPIProvider`, so it survives upstream construction refactors.

Verified against your mutation (hardcode `allow_http=True, allow_private_networks=True` at the primary spec load): the new test **fails** under the mutation and passes when reverted. Kept the single copy rather than duplicating into `test_server_part1.py`.

src/openapi-mcp-server/tests/test_server_extended.py
scottschreckengaust: Fixed in bc3d79f8ce3eb14a427e797eec212d05921f485f. Narrowed the comment to exactly your suggested wording: `_build_route_maps` is unit-tested directly in `test_new_features.py` (`test_build_route_maps_*`); forwarding into the provider is not asserted on either branch, since `DEFAULT_ROUTE_MAPPINGS` already maps every operation to TOOL. Dropped the incorrect 'exercised behaviorally by the GET-with-query-params → TOOL tests' claim.

src/openapi-mcp-server/awslabs/openapi_mcp_server/server.py
scottschreckengaust: Fixed in bc3d79f8ce3eb14a427e797eec212d05921f485f. Hoisted `format_description_with_responses` to a module-level import alongside the other `fastmcp` imports; removed the per-component import inside `enrich_component`. Left the `json`/`url_validator` imports deferred since they're inside the conditional additional-specs branch.

scottschreckengaust:
All review items addressed in bc3d79f8ce3eb14a427e797eec212d05921f485f (pushed to this branch). Mapped to your **What would make this an Approve** list:

1. **Cognito row (all three docs).** Reclassified to outbound-only glue — boto3 `initiate_auth` + `Authorization` on the `httpx.AsyncClient` (matching the Basic/Bearer/API-key rows), with an explicit note that `AWSCognitoProvider` is inbound (MRO ends in `TokenVerifier`) and that the token-acquisition/auto-refresh half has no upstream equivalent. Fixed in `MIGRATION.md`, `docs/migration/README.md`, the capability table in `deprecation-evaluation.md`, **and** the Appendix E `cognito_auth.py` row. Moved Cognito into the glue column so the count reads **three** (Executive Summary, Risks, Appendix B).

2. **SSRF flag forwarding.** Restored as `test_create_mcp_server_forwards_ssrf_flags` — asserts only that config's `allow_insecure_http`/`allow_private_networks` reach `load_openapi_spec`, no `from_openapi`/`OpenAPIProvider` internals. Confirmed against your mutation: fails hardcoded-open, passes reverted.

3. **Route-map comment.** Narrowed to your exact suggested wording in `test_server_extended.py`.

Nice-to-haves: hoisted the `format_description_with_responses` import to module level. The auth-headers/httpx-auth/cookies parity assertion predates this PR on `main`, so I've left it as the separate change you noted.

`ruff` clean. Test run: **516 passed** (was 515 — the +1 is the restored forwarding test). Two `tests/test_ssrf_protection.py` YAML-fallback tests fail in my local env due to a `ruamel-yaml` version quirk (they fail identically on the untouched branch, so unrelated to these changes); CI is the source of truth there.

src/openapi-mcp-server/tests/test_server.py
```diff
@@ -39,46 +39,55 @@ def mock_config():
...
+    """The operator's SSRF opt-in flags must reach ``load_openapi_spec``.
+
+    ``create_mcp_server`` is the only place that threads
+    ``allow_insecure_http`` / ``allow_private_networks`` from config into the
+    spec loader. Hardcoding either open at ``server.py`` would silently disable
+    the operator's SSRF protections; this assertion is the guard against that.
+    """
+    # Explicit, non-default values so the assertion pins forwarding, not defaults.
+    mock_config.allow_insecure_http = False
+    mock_config.allow_private_networks = False
```
theagenticguy: Non-blocking nit. The docstring says these values are "explicit, non-default" so the assertion pins forwarding rather than defaults, but `Config` declares both as `False` (`api/config.py:79-80`), so `False` **is** the default for each. More importantly, setting both to the same value means the test cannot tell the two flags apart.

Mutation demonstrating the residual gap, cross-wiring the two:

```python
allow_http=config.allow_private_networks,
allow_private_networks=config.allow_insecure_http,
```

`tests/test_server.py` → 4 passed. Both sides are `False`, so the swap is invisible.

The straight mutations are caught (both-`True` and `allow_http`-alone each fail this test), so the guard does its main job. Making the two values differ closes the remaining hole:

```python
mock_config.allow_insecure_http = True
mock_config.allow_private_networks = False
```

That also makes the docstring's "non-default" claim literally true for one of them, and it keeps the assertion below unchanged.

laithalsaadoon (dismissed):
## Review: Approve

All three blockers from [the previous review](https://github.com/awslabs/mcp/pull/4124#pullrequestreview-4803693186) are fixed. Re-verified `bc3d79f` independently rather than taking the commit message's word for it.

**1. Cognito mapping — corrected in all four places.** `MIGRATION.md:20-21`, `docs/migration/README.md:25`, and both the capability table and Appendix E row in `deprecation-evaluation.md` now describe Cognito as outbound glue (boto3 token, then `Authorization` on the `httpx.AsyncClient`) and state explicitly that `AWSCognitoProvider` is inbound and not the equivalent. The knock-on edits landed too: the executive summary, the risk bullet, and Appendix B all read three glue cases instead of two, and Appendix B notes Cognito is plausibly the most-used. Splitting the Appendix E auth row so Basic/Bearer/API-key point at header-on-client is a better fix than what was asked for, since those three were also pointing at inbound `server/auth/providers/*`.

**2. SSRF-flag guard — restored and confirmed effective.** `test_create_mcp_server_forwards_ssrf_flags` references neither `from_openapi` nor `OpenAPIProvider`, so it survives upstream construction refactors. Mutation-tested three ways:

| Mutation at `server.py:103-104` | Result |
|---|---|
| both flags hardcoded `True` | **1 failed**, 517 passed |
| `allow_http` alone hardcoded `True` | **1 failed** |
| flags cross-wired to each other | 4 passed (see nit below) |

The regression that shipped green before now fails. Suite is 518 passed, `ruff` clean.

**3. Route-map comment — corrected.** `tests/test_server_extended.py:42-50` now states plainly that forwarding into the provider is unasserted on either branch and explains why (`DEFAULT_ROUTE_MAPPINGS` already maps everything to TOOL). Accurate, and more useful than the original claim.

The import hoist to module level landed as well.

Approving. The `from_openapi` conversion, `add_provider` composition, enrichment swap, and `instructions` forwarding all check out against FastMCP 3.4.4 source, the behavior change in description formatting is documented honestly in both the description and the tests, and CI is green across the board.

One non-blocking nit inline: the new test cannot distinguish the two flags from each other while both default to `False`. Worth a one-line change if you touch the file again, not worth a new commit on its own.

Note that auto-merge is enabled with `reviewDecision: REVIEW_REQUIRED` and `mergeStateStatus: BLOCKED`, so this approval likely unblocks the merge. Flagging that deliberately since the description still carries the "Draft: not for merge as-is" banner and lists coverage re-verification as follow-up #1 — that item is now answered (the gap was assertion coverage, not line coverage, and it is closed), but you may want to drop the banner before this lands.

src/openapi-mcp-server/tests/test_server.py
```diff
@@ -39,46 +39,55 @@ def mock_config():
...
-    mock_fastmcp.return_value = mock_server
+    """The operator's SSRF opt-in flags must reach ``load_openapi_spec``.
+
+    ``create_mcp_server`` is the only place that threads
+    ``allow_insecure_http`` / ``allow_private_networks`` from config into the
+    spec loader. Hardcoding either open at ``server.py`` would silently disable
+    the operator's SSRF protections; this assertion is the guard against that.
+    """
+    # Explicit, non-default values so the assertion pins forwarding, not defaults.
+    mock_config.allow_insecure_http = False
```
prajwalendra: **Finding 1, recommended before merge.** Same point as @theagenticguy's thread at :76.

Both flags are set to the same value, so this test cannot tell them apart. Swapping the two kwargs at the spec load, or hardcoding both to `False`, still passes. Making the two values differ closes that.

```suggestion
    mock_config.allow_insecure_http = True
```

One small thing: the comment above calls these "non-default values", but `False` is the default for both (`api/config.py:79-80`).

src/openapi-mcp-server/awslabs/openapi_mcp_server/server.py
```diff
@@ -226,37 +227,41 @@ def handle_auth_error(auth_type, error_message):
...
+            instead of the previous bespoke string-building. The upstream
+            formatter emits richer, structured sections (Path/Query Parameters,
+            Request Body, Responses with examples) than the old
+            ``desc | Returns: ... | Example: ...`` format.
+            """
+            component.description = format_description_with_responses(
+                component.description or '',
+                route.responses if getattr(route, 'responses', None) else {},
+                getattr(route, 'parameters', None),
+                getattr(route, 'request_body', None),
```
prajwalendra, line 243: **Finding 2, recommended before merge.** This new argument has no test coverage. No spec that reaches this enrichment path has a `requestBody`, so the request-body branch of `format_description_with_responses` never runs. Adding a POST with a `requestBody` to the spec in `tests/test_new_features.py` would cover it.

**Finding 3, worth a line in `MIGRATION.md`.** `generate_example_from_schema` has no `$ref` branch and falls through to `"unknown_type"`. Specs loaded in-tree are fine, since prance resolves refs first. The raw-dict recipe in the guide leaves `$ref`s intact, so anyone following it verbatim gets `["unknown_type"]` as the response example, which contradicts the tool's own `outputSchema`.

prajwalendra (commented):
@scottschreckengaust reviewing this for merge readiness. The migration approach is correct: `from_openapi` + `add_provider` is equivalent to the previous `providers=[...]` construction, and credential isolation, tag filtering and operationId collision handling are all preserved.

**Findings**

1. `tests/test_server.py:75` - the restored SSRF forwarding test sets both flags to the same value, so it cannot tell them apart. Comment inline with a one-line suggestion.
2. `server.py:243` - the new `request_body` argument has no test coverage. No spec reaching that path has a `requestBody`, so the branch never runs.
3. `MIGRATION.md` - the raw-dict recipe leaves `$ref`s unresolved, which renders response examples as `["unknown_type"]`. In-tree loading is unaffected. Details inline on finding 2.

**Recommended before merge**

* Fix findings 1 and 2. Both are small and both are new in this PR.
* Follow-up 1 (coverage check): no line-coverage regression, `server.py` moves 90% to 91% with the same 25 uncovered lines. Please note `--cov` cannot settle this item on its own. What was lost when the construction tests came out is assertion coverage, not line coverage. Line 243 reports as covered but still passes when replaced with `None`. This is the same blind spot @theagenticguy raised at `test_server.py:42`, where Codecov showed all modified lines covered while the SSRF assertion was gone. Suggest closing this item on findings 1 and 2 rather than on the coverage number.
* Follow-up 2 (upstream give-backs): agreed these should not hold the PR. If you file only one, the SSRF-safe spec fetch draft is the strongest, since FastMCP accepts an already-parsed dict and never fetches a spec, so there is no upstream equivalent today.

**Not counted against this PR**

`validate_output` and `route_maps` forwarding are untested here, but equally untested on `main`. Missing non-2xx responses in descriptions and the cosmetic formatter artifacts are upstream behaviour, newly visible because this PR opts in.

Happy to approve once findings 1 and 2 are addressed.

src/openapi-mcp-server/tests/test_server.py
scottschreckengaust: Fixed in 5ef34223. Set `allow_insecure_http = True` / `allow_private_networks = False` so the two differ.

Your diagnosis was exactly right, and I confirmed it with your mutation. With both `False`, `assert_called_once_with` was only constraining the *multiset* of values, not the mapping of value→parameter — so any permutation satisfied it. Now:

| mutation at `server.py:104-105` | before | after |
|---|---|---|
| cross-wire the two kwargs | **passed** | **fails** |
| hardcode both `True` | fails | fails |

Also corrected the comment. You're right that `False` is the declared default for both (`api/config.py:79-80`), so "explicit, non-default" was simply wrong; it now says what the values are actually for, and `allow_insecure_http = True` makes the non-default part true for one of them.

src/openapi-mcp-server/tests/test_server.py
scottschreckengaust: **Finding 1 — fixed in 5ef34223.** Took your suggestion: `mock_config.allow_insecure_http = True`, leaving `allow_private_networks = False`. Confirmed by mutation that cross-wiring the two kwargs at `server.py:104-105` now fails where it previously passed (details in @theagenticguy's thread at :76). Also fixed the "non-default values" comment — you're right that `False` is the default for both.

src/openapi-mcp-server/awslabs/openapi_mcp_server/server.py
scottschreckengaust: Both fixed in 5ef34223.

**Finding 2 — covered.** Added a `requestBody` to `createPet` in `PETSTORE_SPEC` and a `test_enriched_descriptions_include_request_body` asserting the `**Request Body:**` and nested `**Request Properties:**` sections. Confirmed it's a real guard by mutation: deleting the `getattr(route, 'request_body', None)` argument from `enrich_component` fails **exactly** the new test and nothing else (`1 failed, 30 passed`), so nothing else in the suite was covering it.

One wrinkle worth recording, since it makes the difference between a real test and a vacuous one: the body needs a `description`. The formatter gates its *entire* Request Body section on that field being truthy —

```python
if request_body and request_body.description:   # formatters.py:241
```

— so a `requestBody` without one enriches byte-identically to no body at all, and the mutation above would still pass. The added spec carries `description` plus per-property descriptions to reach the nested sub-branch too.

**Finding 3 — documented.** Confirmed and reproduced. `generate_example_from_schema` (`formatters.py:100`) has no `$ref` branch, so a reference falls through to the `"unknown_type"` fallback at line 180. Running your scenario through `from_openapi` + `mcp_component_fn` with an unresolved spec:

```
- **200** (Success): OK
  - **Example:**
[
  "unknown_type"
]
```

while the same tool's `output_schema` is `{'type': 'array', 'items': {'type': 'object', 'properties': {'id': ..., 'name': ...}}}` — the reference resolved correctly. So the contradiction you predicted is literally visible side by side in one tool definition.

`MIGRATION.md` now carries the caveat plus a `prance`/`RESOLVE_INTERNAL` dereference recipe (verified: the placeholder becomes `[{"id": 1, "name": "string"}]`), scoped to `mcp_component_fn` users since `from_openapi` resolves refs itself for input/output schemas. `RESOLVE_INTERNAL` rather than the default, deliberately — external `$ref` resolution is the SSRF/LFI sink #4141 closed, and it'd be poor form for the migration guide to hand people back the hole. `docs/migration/deprecation-evaluation.md` Appendix E carries the same enrichment recipe, so it got a short cross-referencing note.

scottschreckengaust:
## Review round 2 addressed — 5ef34223

All three outstanding threads are remediated and replied to individually. Branch also synced to current `main` (`2fbc9b2b`).

| Finding | Reviewer | Change | Verified by |
|---|---|---|---|
| SSRF flags indistinguishable | @theagenticguy, @prajwalendra | `allow_insecure_http = True` / `allow_private_networks = False` | reviewer's cross-wire mutation now **fails** (previously passed) |
| `request_body` argument uncovered | @prajwalendra | `requestBody` on `createPet` + `test_enriched_descriptions_include_request_body` | dropping the argument fails **exactly** the new test |
| `$ref` → `"unknown_type"` | @prajwalendra | caveat + dereference recipe in `MIGRATION.md`; cross-ref in `deprecation-evaluation.md` | reproduced the bad example, then verified the recipe clears it |

Every fix was checked by mutation rather than asserted — both reviewers used that method to find these, and a test that doesn't fail when the thing it guards is broken isn't coverage.

### Two things worth flagging

**The request-body test needed a `description` to be non-vacuous.** The upstream formatter gates its whole Request Body section on `if request_body and request_body.description:` (`formatters.py:241`). A `requestBody` without one enriches byte-identically to no body at all, so the obvious version of this test would have passed while still covering nothing. Called out in the spec comment so a future trim doesn't silently re-hollow it.

**Finding 3 is an upstream gap, not just a docs gap.** `generate_example_from_schema` having no `$ref` branch means *any* `mcp_component_fn` user who skips dereferencing gets a description that contradicts the tool's own `output_schema`. The guide now covers our migrating users, which is what this PR owes them. But the same class of paper-cut is what `docs/migration/upstream-issue-1-enrichment-dx.md` already proposes fixing — if that enrichment-DX contribution goes upstream, a `$ref` branch (or a resolve-first guard) belongs in the same change, since making enrichment the *default* would turn this from an opt-in footgun into one everybody hits. Recorded here rather than expanding this PR's scope.

### Recommendations for reviewers

- **The two local `tests/test_ssrf_protection.py` failures are not from this PR.** `test_basic_parse_falls_back_to_ruamel_when_pyyaml_rejects` and `test_parse_spec_bytes_yaml_ref_blocked_when_pyyaml_rejects` fail identically on **unmodified `origin/main`** (checked in a clean worktree, same `ruamel-yaml` 0.18.10): the tab-separated YAML those tests use as a "PyYAML rejects, ruamel accepts" fixture is now rejected by ruamel too, so the fallback assertion no longer holds. CI is green because it resolves a different ruamel. It predates this branch (the file last changed in #4141) and is worth its own issue — flagging so it isn't misread as a regression here.
- Otherwise: **517 passed**, `pyright` clean (0 errors), `pre-commit` clean on all changed files.

src/openapi-mcp-server/tests/test_server.py
theagenticguy: Verified independently at head `4e45486`: re-ran the cross-wire mutation (`allow_http=config.allow_private_networks` / `allow_private_networks=config.allow_insecure_http` at `server.py:104-105`) and `test_create_mcp_server_forwards_ssrf_flags` now fails — 1 failed, 3 passed — where it previously passed. The comment correction is accurate too. Resolving.

src/openapi-mcp-server/tests/test_server.py
theagenticguy: Independently re-verified at head `4e45486` (same finding as my thread above): the suggested `allow_insecure_http = True` landed, the cross-wire mutation now fails the test, and the "non-default values" comment was corrected. Resolving — @prajwalendra flag if you disagree.

src/openapi-mcp-server/awslabs/openapi_mcp_server/server.py
theagenticguy: Re-verified both at head `4e45486`.

**Finding 2:** mutating `getattr(route, 'request_body', None)` → `None` at `server.py:243` fails exactly `test_enriched_descriptions_include_request_body` (1 failed, 30 passed in `test_new_features.py`), so the new test is the sole and sufficient guard. The `description`-gating caveat in the spec comment checks out against the installed formatter source (`if request_body and request_body.description:`).

**Finding 3:** reproduced both halves of the claim against the installed FastMCP: `generate_example_from_schema({'type': 'array', 'items': {'$ref': ...}})` → `['unknown_type']`, and the `MIGRATION.md` `ResolvingParser`/`RESOLVE_INTERNAL` recipe clears it (`[{'id': 1, 'name': 'string'}]`). Restricting to internal refs is the right call given #4141.

Resolving — @prajwalendra flag if you disagree.

laithalsaadoon (approved):
## Review: Approve

Re-review of round 2 (`5ef3422`, merged up to `main` in `4e45486`). My earlier approval was dismissed by the new push, so this re-establishes it. As before, I verified rather than took the remediation comment's word for it — fresh clone of head, full suite, and every claimed mutation re-run locally.

**Finding 1 (SSRF flag distinguishability) — closed.** `allow_insecure_http = True` / `allow_private_networks = False` landed with a corrected comment. Re-ran my cross-wire mutation at `server.py:104-105`: the test now fails (1 failed, 3 passed) where it previously passed. Hardcoding both open also still fails. The guard now pins each flag to its own parameter.

**Finding 2 (`request_body` coverage) — closed.** Mutating `getattr(route, 'request_body', None)` → `None` at `server.py:243` fails exactly `test_enriched_descriptions_include_request_body` and nothing else (1 failed, 30 passed in `test_new_features.py`), confirming it is the sole guard on that argument. The `description`-gating caveat recorded in the spec comment is accurate against the formatter source (`if request_body and request_body.description:`) — good catch to make the test non-vacuous, and worth having in-tree so a future spec trim doesn't hollow it out.

**Finding 3 (`$ref` → `"unknown_type"`) — closed.** Reproduced both halves independently: an unresolved array-of-`$ref` schema yields `["unknown_type"]` from `generate_example_from_schema`, and the `MIGRATION.md` `ResolvingParser`/`RESOLVE_INTERNAL` recipe clears it to a correct example. Scoping resolution to internal refs is the right call given the #4141 SSRF/LFI fix; the guide would otherwise reopen the hole it sits next to.

**Overall state at head:** 519 passed locally (fresh clone, `uv sync --all-extras --dev`, Python 3.10) including the two `test_ssrf_protection.py` YAML-fallback tests that fail in the author's local env — consistent with the ruamel-version explanation given, and CI is green across the board. `ruff` clean on the changed files. All review threads from both rounds are now addressed and resolved.

Two non-blocking notes on the PR description before this lands: the "Draft: not for merge as-is" banner and the "495 passed" test count are stale (follow-up #1, the coverage question, was answered by @prajwalendra — assertion coverage was the gap, and it is closed), and since auto-merge is enabled this approval may unblock the merge, so update the description first if you want it accurate at merge time.

prajwalendra (approved):
LGTM

PR #4127, opened 2026-07-14
fix(redshift-mcp-server): correct serverless db name, map SHOW output by column name, drop execution_time_ms

saeedma8:
## Summary

Fixes three issues in the redshift-mcp-server discovery/query tools and clarifies datashare behavior in the docs.

### Changes

- Serverless `list_clusters` reported a bogus database_name (e.g. `true`) because it read an arbitrary config parameter; report the built-in `dev` database instead. Every serverless workgroup exposes `dev` and can always connect to it. Callers can still pass an explicit database_name to the other tools.
- SHOW output is now parsed by column name. Introduces a small `RedshiftDataModel` base whose `from_redshift_response` builds models by matching result columns to model fields by name (via the result-set column metadata), so the discovery tools (`list_databases`, `list_schemas`, `list_tables`, `list_columns`) are resilient to column reordering or additional columns instead of relying on fixed positions. The discovery models inherit it; `execute_query` shares the same value decoder. 
- Removed `execution_time_ms` from query results. It reported server-side wall time including session/queue overhead, so it could be wildly larger than the actual query time and was misleading. Clients can time tool calls directly. `QueryResult` now returns `columns`, `rows`, `row_count`, and `query_id`.
- Docs: clarified that discovery reflects the privileges of the principal the server connects as — datashare (shared) databases and their schemas appear only when that principal has been granted access to the consumer database.

### User experience

<details>

```
% kiro-cli chat \
    --model claude-opus-4.8 \
    --no-interactive \
    --trust-tools "@{awslabsredshift-mcp-server}/*,execute_bash,fs_read" \
'Use Redshift MCP Server to run end-to-end (integration) tests covering the scenarios for the changes introduced in the current branch only. Get the test scenario ideas from the unit tests under the project directory. Provide a short testing summary, one line per scenario.'
```
...

> All six scenarios pass against live clusters. Here's the testing summary.

## Integration Test Summary — branch redshift-mcp-show-fixes (commit 08ddc5a1)

Tested live via the Redshift MCP Server against us-east-1 (serverless workgroup redshift-cluster-serverless + provisioned test-drive-consumer-personal). Scenarios derived from the branch's changed unit tests.

| # | Scenario (unit test) | Result |
|---|---|---|
| 1 | Serverless db name fix — list_clusters reports database_name="dev" for serverless, not a bogus config value (test_discover_clusters_serverless) | ✅ Both serverless workgroups returned "dev"; provisioned +
serverless all returned as typed models |
| 2 | parameters field rename + column-name mapping — list_databases exposes parameters (was database_options) (test_discover_databases) | ✅ parameters field present; connecting to serverless dev succeeded (
validates the 'dev' connect fix) |
| 3 | Schema discovery mapping — list_schemas maps all 7 fields by column name (test_discover_schemas) | ✅ tickit/public/information_schema returned with correct owner (long), type, acl |
| 4 | Table type/acl not swapped — list_tables maps table_type vs table_acl by name, not position (test_discover_tables) | ✅ 7 tables; table_type="TABLE" and table_acl=ACL string correctly assigned (the exact
regression fixed) |
| 5 | Column discovery mapping — list_columns maps 12 fields incl. mixed null/long/string (test_discover_columns) | ✅ All 10 sales columns returned with correct ordinal, data_type, precision/scale, nullable |
| 6 | Cell decoder + drop execution_time_ms — execute_query decodes string/long/double/bool/null and omits execution_time_ms (test_execute_query_success) | ✅ Row [1,"lbl",728.0,true,null] decoded correctly; result
had only columns/rows/row_count/query_id — no execution_time_ms |
| + | Datashare visibility (doc change) — shared DBs appear for granted principals | ✅ Consumer cluster showed sample_data_dev as shared with populated datashare parameters JSON |

All branch-introduced behaviors verified end-to-end with no failures.

```
% kiro-cli chat \
    --model claude-opus-4.8 \
    --no-interactive \
    --trust-tools "@{awslabsredshift-mcp-server}/*,execute_bash,fs_read" \
'Use Redshift MCP Server to run a complete set of end-to-end (integration) tests covering all its tools. Check both provisioned and serverless clusters, including the database schema exploration in both. Check the SQL read-only protection, transaction breaker protection, failed user SQL behavior. Get the test scenario ideas from the unit tests under the project directory. Provide a short testing summary, one line per tool.'
```

...

> All tests passed. Here's the end-to-end testing summary.

## Redshift MCP Server — E2E Test Summary

Environment: us-east-1 · provisioned = redshift-cluster-1 · serverless = redshift-cluster-serverless · sample dataset sample_data_dev.tickit

Per-tool results (one line each):

- **list_clusters** — ✅ Returned all 6 resources (4 provisioned, 2 serverless) with correct type/status/config.
- **list_databases** — ✅ Worked on both cluster types; listed local + auto-mounted catalog DBs (sample_data_dev present on both).
- **list_schemas** — ✅ Explored sample_data_dev on both; returned tickit/public/information_schema (+tpcds/tpch on provisioned)^[[O with ACLs.
- **list_tables** — ✅ Listed all 7 tickit tables on both cluster types identically.
- **list_columns** — ✅ Returned all 10 tickit.sales columns with types/precision/nullability, identical across both.
- **execute_query** — ✅ Valid SELECTs succeeded on both (provisioned aggregate: 344,912 rows / $221.5M; serverless top-3 recent sales) and all protection paths held (see below).
- **review_cluster** — ✅ Provisioned: 11 signals → 11 findings + recommendations; Serverless: 8 signals (provisioned-only diagnostics correctly skipped, ServerlessScaling added) → 0 findings (healthy).

Safety / negative-path checks (all behaved correctly):

| Scenario | Input | Result |
|---|---|---|
| Read-only guard (deny-list) | TRUNCATE tickit.sales | Rejected: Statement type not allowed in read-only mode: TRUNCATE |
| Read-only guard (DCL) | GRANT SELECT … TO PUBLIC | Rejected: …not allowed…: GRANT |
| Multi-statement guard | SELECT 1; SELECT 2 | Rejected: Only a single SQL statement is allowed |
| Transaction breaker (provisioned) | CREATE TABLE … | Passed guard, blocked by engine: transaction is read-only |
| Transaction breaker (serverless) | INSERT INTO tickit.sales … | Blocked by engine: transaction is read-only |
| No side effects | SELECT … FROM mcp_e2e_should_not_exist | relation … does not exist — confirms the CREATE was rolled back |
| Failed user SQL (unparseable) | SELECT FROM WHERE | Fails closed: SQL could not be parsed |
| Failed user SQL (bad relation) | SELECT * FROM tickit.no_such_table_xyz | Engine error surfaced cleanly: relation … does not exist |

All 7 tools function correctly across both provisioned and serverless clusters. The two-layer write protection is verified end-to-end: the AST deny-list stops transaction-control/DDL-egress statements before
execution, and the BEGIN READ ONLY transaction backstop rejects ordinary writes (CREATE/INSERT) that intentionally pass the guard — with no residual side effects. Failed-query behavior (both parse-time and engine-
time) returns clear errors without leaking internals.

</details>

## Checklist

* [x] Changes are documented

Is this a breaking change? (Y/N) - Yes, minor. `execution_time_ms` is removed from `execute_query` results; consumers that read that field should drop it (all other fields are unchanged).

**RFC issue number**:

Checklist:

* [N/A] Migration process documented
* [N/A] Implement warnings (if it can live side by side)

## Acknowledgment

### DIFF
--- a/src/redshift-mcp-server/README.md
+++ b/src/redshift-mcp-server/README.md
@@ -398,7 +398,7 @@ execute_query(cluster_identifier: str, database_name: str, sql: str) -> QueryRes
 
 - Column names and data types
 - Result rows with proper type conversion
-- Row count and execution time
+- Row count
 - Query ID for reference
 
 ### review_cluster
--- a/src/redshift-mcp-server/awslabs/redshift_mcp_server/models.py
+++ b/src/redshift-mcp-server/awslabs/redshift_mcp_server/models.py
@@ -16,7 +16,40 @@
 
 from datetime import datetime
 from pydantic import BaseModel, Field
-from typing import Dict, Optional
+from typing import Any, Dict, Optional, TypeVar
+
+
+RedshiftDataModelT = TypeVar('RedshiftDataModelT', bound='RedshiftDataModel')
+
+
+class RedshiftDataModel(BaseModel):
+    """Base for models built from a Redshift Data API result set.
+
+    Subclasses declare their fields named to match the SHOW result columns.
+    `from_redshift_response` maps result columns to those fields by name, so
+    parsing is independent of column order; unknown columns are ignored.
+    """
+
+    @staticmethod
+    def cell_value(cell: dict) -> Any:
+        """Unwrap a single Redshift Data API result cell to a Python scalar."""
+        if cell.get('isNull'):
+            return None
+        for key in ('stringValue', 'longValue', 'doubleValue', 'booleanValue'):
+            if key in cell:
+                return cell[key]
+        return str(cell)
+
+    @classmethod
+    def from_redshift_response(
+        cls: type[RedshiftDataModelT], results_response: dict
+    ) -> list[RedshiftDataModelT]:
+        """Build a list of model instances from a Data API result set."""
+        names = [col.get('name') for col in results_response.get('ColumnMetadata', [])]
+        return [
+            cls.model_validate({name: cls.cell_value(cell) for name, cell in zip(names, record)})
+            for record in results_response.get('Records', [])
+        ]
 
 
 class RedshiftCluster(BaseModel):
@@ -40,7 +73,7 @@ class RedshiftCluster(BaseModel):
     )
 
 
-class RedshiftDatabase(BaseModel):
+class RedshiftDatabase(RedshiftDataModel):
     """Information about a database in a Redshift cluster."""
 
     database_name: str = Field(..., description='The name of the database')
@@ -51,14 +84,14 @@ class RedshiftDatabase(BaseModel):
     database_acl: Optional[str] = Field(
         None, description='Access control information (for internal use)'
     )
-    database_options: Optional[str] = Field(None, description='The properties of the database')
+    parameters: Optional[str] = Field(None, description='The properties of the database')
     database_isolation_level: Optional[str] = Field(
         None,
         description='The isolation level of the database (Snapshot Isolation or Serializable)',
     )
 
 
-class RedshiftSchema(BaseModel):
+class RedshiftSchema(RedshiftDataModel):
     """Information about a schema in a Redshift database."""
 
     database_name: str = Field(..., description='The name of the database where the schema exists')
@@ -78,7 +111,7 @@ class RedshiftSchema(BaseModel):
     )
 
 
-class RedshiftTable(BaseModel):
+class RedshiftTable(RedshiftDataModel):
     """Information about a table in a Redshift database."""
 
     database_name: str = Field(..., description='The name of the database where the table exists')
@@ -94,7 +127,7 @@ class RedshiftTable(BaseModel):
     remarks: Optional[str] = Field(None, description='Remarks about the table')
 
 
-class RedshiftColumn(BaseModel):
+class RedshiftColumn(RedshiftDataModel):
     """Information about a column in a Redshift table."""
 
     database_name: str = Field(..., description='The name of the database')
@@ -123,7 +156,4 @@ class QueryResult(BaseModel):
     columns: list[str] = Field(..., description='List of column names in the result set')
     rows: list[list] = Field(..., description='List of rows, where each row is a list of values')
     row_count: int = Field(..., description='Number of rows returned')
-    execution_time_ms: Optional[int] = Field(
-        None, description='Query execution time in milliseconds'
-    )
     query_id: str = Field(..., description='Unique identifier for the query execution')
--- a/src/redshift-mcp-server/awslabs/redshift_mcp_server/redshift.py
+++ b/src/redshift-mcp-server/awslabs/redshift_mcp_server/redshift.py
@@ -32,6 +32,14 @@ from awslabs.redshift_mcp_server.consts import (
     SESSION_KEEPALIVE,
     TABLES_SQL,
 )
+from awslabs.redshift_mcp_server.models import (
+    RedshiftCluster,
+    RedshiftColumn,
+    RedshiftDatabase,
+    RedshiftDataModel,
+    RedshiftSchema,
+    RedshiftTable,
+)
 from awslabs.redshift_mcp_server.sql_guard import assert_executable
 from botocore.config import Config
 from botocore.exceptions import ClientError
@@ -147,14 +155,14 @@ class RedshiftSessionManager:
         return existing
 
     async def session(
-        self, cluster_identifier: str, database_name: str, cluster_info: dict
+        self, cluster_identifier: str, database_name: str, cluster_info: RedshiftCluster
     ) -> str:
         """Get or create a session for the given cluster and database.
 
         Args:
             cluster_identifier: The cluster identifier to get session for.
             database_name: The database name to get session for.
-            cluster_info: Cluster information dictionary from discover_clusters.
+            cluster_info: Cluster information model from discover_clusters.
 
         Returns:
             Session ID for use in ExecuteStatement calls.
@@ -182,14 +190,14 @@ class RedshiftSessionManager:
         return session_id
 
     async def _create_session_with_app_name(
-        self, cluster_identifier: str, database_name: str, cluster_info: dict
+        self, cluster_identifier: str, database_name: str, cluster_info: RedshiftCluster
     ) -> str:
         """Create a new session by executing SET application_name.
 
         Args:
             cluster_identifier: The cluster identifier.
             database_name: The database name.
-            cluster_info: Cluster information dictionary.
+            cluster_info: Cluster information model.
 
         Returns:
             Session ID from the ExecuteStatement response.
@@ -272,7 +280,7 @@ async def _execute_protected_statement(
     clusters = await discover_clusters()
     cluster_info = None
     for cluster in clusters:
-        if cluster['identifier'] == cluster_identifier:
+        if cluster.identifier == cluster_identifier:
             cluster_info = cluster
             break
 
@@ -364,7 +372,7 @@ async def _execute_protected_statement(
 
 
 async def _execute_statement(
-    cluster_info: dict,
+    cluster_info: RedshiftCluster,
     cluster_identifier: str,
     database_name: str,
     sql: str,
@@ -377,7 +385,7 @@ async def _execute_statement(
     """Execute a single statement with optional session support and parameters.
 
     Args:
-        cluster_info: Cluster information dictionary.
+        cluster_info: Cluster information model.
         cluster_identifier: The cluster identifier.
         database_name: The database name.
         sql: The SQL statement to execute.
@@ -398,12 +406,12 @@ async def _execute_statement(
     # Add database and cluster/workgroup identifier only if not using session
     if not session_id:
         request_params['Database'] = database_name
-        if cluster_info['type'] == 'provisioned':
+        if cluster_info.type == 'provisioned':
             request_params['ClusterIdentifier'] = cluster_identifier
-        elif cluster_info['type'] == 'serverless':
+        elif cluster_info.type == 'serverless':
             request_params['WorkgroupName'] = cluster_identifier
         else:
-            raise Exception(f'Unknown cluster type: {cluster_info["type"]}')
+            raise Exception(f'Unknown cluster type: {cluster_info.type}')
 
     # Add parameters if provided
     if parameters:
@@ -446,7 +454,7 @@ async def _execute_statement(
     return statement_id
 
 
-async def discover_clusters() -> list[dict]:
+async def discover_clusters() -> list[RedshiftCluster]:
     """Discover all Redshift clusters and serverless workgroups.
 
     Discovery is best-effort for each type: if either provisioned or serverless
@@ -454,7 +462,7 @@ async def discover_clusters() -> list[dict]:
     if both fail (i.e., no clusters could be discovered at all).
 
     Returns:
-        List of cluster information dictionaries.
+        List of RedshiftCluster models.
 
     Raises:
         Exception: If both provisioned and serverless discovery fail.
@@ -488,7 +496,7 @@ async def discover_clusters() -> list[dict]:
                     'encrypted': cluster.get('Encrypted'),
                     'tags': {tag['Key']: tag['Value'] for tag in cluster.get('Tags', [])},
                 }
-                clusters.append(cluster_info)
+                clusters.append(RedshiftCluster(**cluster_info))
 
         logger.info(f'Found {len(clusters)} provisioned clusters')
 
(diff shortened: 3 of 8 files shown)


src/redshift-mcp-server/awslabs/redshift_mcp_server/models.py
```diff
@@ -51,14 +87,17 @@ class RedshiftDatabase(BaseModel):
     database_acl: Optional[str] = Field(
         None, description='Access control information (for internal use)'
     )
-    database_options: Optional[str] = Field(None, description='The properties of the database')
+    # SHOW DATABASES names this column 'parameters'.
```
grayhemp: Shall we rename it in the model and all downstream to comply with SHOW?

src/redshift-mcp-server/awslabs/redshift_mcp_server/models.py
```diff
@@ -15,8 +15,44 @@
 """Redshift MCP Server Pydantic models."""
 
 from datetime import datetime
-from pydantic import BaseModel, Field
-from typing import Dict, Optional
+from pydantic import BaseModel, ConfigDict, Field
+from typing import Any, Dict, Optional, TypeVar
+
+
+T = TypeVar('T', bound='RedshiftDataModel')
```
grayhemp: T is too examplish. Let's use RedshiftDataModelT.

src/redshift-mcp-server/awslabs/redshift_mcp_server/models.py
```diff
@@ -15,8 +15,44 @@
...
+class RedshiftDataModel(BaseModel):
+    """Base for models built from a Redshift Data API result set.
+
+    Subclasses declare their fields (and, where the SHOW column name differs from
+    the field, a validation alias). `from_redshift_response` maps result columns
+    to those fields by name, so parsing is independent of column order; unknown
+    columns are ignored.
+    """
+
+    model_config = ConfigDict(populate_by_name=True)
```
grayhemp: Just curious, what is this for?

src/redshift-mcp-server/awslabs/redshift_mcp_server/redshift.py
```diff
@@ -494,13 +501,27 @@ async def discover_clusters() -> list[dict]:
...
 
+                # The database name comes from the workgroup's namespace
+                database_name = 'dev'
+                namespace_name = workgroup_detail.get('namespaceName')
+                if namespace_name:
+                    try:
+                        namespace = serverless_client.get_namespace(namespaceName=namespace_name)[
+                            'namespace'
+                        ]
+                        database_name = namespace.get('dbName') or 'dev'
```
grayhemp: .get() supports default value to not do 'or'.

src/redshift-mcp-server/awslabs/redshift_mcp_server/redshift.py
```diff
@@ -494,13 +501,27 @@ async def discover_clusters() -> list[dict]:
...
+                    except Exception as e:
+                        logger.warning(
+                            f'Could not resolve dbName for namespace {namespace_name} '
```
grayhemp: 'Could not get namespace to resolve dbName'

src/redshift-mcp-server/awslabs/redshift_mcp_server/redshift.py
grayhemp: Also we have database_name defined as 'dev' which should be used as the default, not the constant.

src/redshift-mcp-server/awslabs/redshift_mcp_server/redshift.py
```diff
@@ -494,13 +501,27 @@ async def discover_clusters() -> list[dict]:
...
+                            f'(check the redshift-serverless:GetNamespace permission); '
+                            f"falling back to '{database_name}', which may be inaccurate: {e}"
```
grayhemp: to the default '{database_name}'

grayhemp: ", which may be inaccurate" - remove

src/redshift-mcp-server/awslabs/redshift_mcp_server/models.py
```diff
@@ -15,8 +15,44 @@
...
+        names = [col.get('name') for col in results_response.get('ColumnMetadata', [])]
+        return [
+            cls.model_validate(
+                {name: redshift_cell_value(cell) for name, cell in zip(names, record)}
+            )
+            for record in results_response.get('Records', [])
+        ]
 
 
 class RedshiftCluster(BaseModel):
```
grayhemp, line 55: Just curious, why did you decide to skip clusters?

src/redshift-mcp-server/awslabs/redshift_mcp_server/models.py
```diff
@@ -15,8 +15,44 @@
...
-from pydantic import BaseModel, Field
-from typing import Dict, Optional
+from pydantic import BaseModel, ConfigDict, Field
+from typing import Any, Dict, Optional, TypeVar
+
+
+T = TypeVar('T', bound='RedshiftDataModel')
+
+
+def redshift_cell_value(cell: dict) -> Any:
```
grayhemp: Let's make cell_value() a public class method of RedshiftDataModel.

src/redshift-mcp-server/README.md
```diff
@@ -417,6 +417,7 @@ Your AWS credentials need the following IAM permissions:
         "redshift:DescribeClusters",
         "redshift-serverless:ListWorkgroups",
         "redshift-serverless:GetWorkgroup",
+        "redshift-serverless:GetNamespace",
```
grayhemp: Wonder if there is a feasible way of getting the dbName without the GetNamespace requirement. If there is no, then ok, but if we can avoid introducing another permission we should try.

grayhemp (changes requested):
Left several comments.

src/redshift-mcp-server/awslabs/redshift_mcp_server/models.py
saeedma8: Renamed.

src/redshift-mcp-server/awslabs/redshift_mcp_server/models.py
saeedma8: Added.

src/redshift-mcp-server/awslabs/redshift_mcp_server/models.py
saeedma8: Using ConfigDict(populate_by_name=True) to support both name and alias,  because I was using alias for 'parameters' and when using validation_alias by default pydantic use alias only.

src/redshift-mcp-server/awslabs/redshift_mcp_server/models.py
saeedma8: RedshiftCluster isn't built from a Data API SHOW result and it comes from boto3 describe_clusters/list_workgroups. There is no ColumnMetadata/Records. The column-name mapping doesn't apply.

src/redshift-mcp-server/awslabs/redshift_mcp_server/models.py
saeedma8: Renamed to parameters to match the SHOW column, dropped the alias, and updated all downstream.

src/redshift-mcp-server/awslabs/redshift_mcp_server/redshift.py
saeedma8: Modified.

src/redshift-mcp-server/awslabs/redshift_mcp_server/redshift.py
saeedma8: Corrected.

src/redshift-mcp-server/awslabs/redshift_mcp_server/redshift.py
saeedma8: Modified.

src/redshift-mcp-server/README.md
saeedma8: Looked for an alternative and there isn't a clean one. get_workgroup doesn't expose the namespace's dbName.

src/redshift-mcp-server/awslabs/redshift_mcp_server/redshift.py
grayhemp: Let's check if it would be beneficial to return list[RedshiftClusters] here.

src/redshift-mcp-server/awslabs/redshift_mcp_server/redshift.py
```diff
@@ -494,13 +501,27 @@ async def discover_clusters() -> list[dict]:
...
 
+                # The database name comes from the workgroup's namespace
+                database_name = 'dev'
+                namespace_name = workgroup_detail.get('namespaceName')
+                if namespace_name:
+                    try:
+                        namespace = serverless_client.get_namespace(namespaceName=namespace_name)[
+                            'namespace'
+                        ]
+                        database_name = namespace.get('dbName') or database_name
```
grayhemp: `.get('dbName', database_name)`?

src/redshift-mcp-server/awslabs/redshift_mcp_server/server.py
```diff
@@ -234,7 +234,7 @@ async def list_databases_tool(
     - database_owner: The database owner user ID.
     - database_type: The type of database (local or shared).
     - database_acl: Access control information (for internal use).
-    - database_options: The properties of the database.
+    - database_parameters: The properties of the database.
```
grayhemp: parameters

src/redshift-mcp-server/README.md
grayhemp: This is a breaking change. Can we double-check that the dev DB is always present unconditionally? In this case we can avoid the get_namespace call.

grayhemp (changes requested):

src/redshift-mcp-server/awslabs/redshift_mcp_server/redshift.py
saeedma8: n/a anymore. Namespace removed.

src/redshift-mcp-server/awslabs/redshift_mcp_server/server.py
saeedma8: fixed.

grayhemp:
## Testing Summary — branch pr-4127, Redshift MCP Server (E2E)

Ran against live us-east-1: provisioned redshift-cluster-analytics + serverless redshift-workgroup-etl. Scenarios drawn from 
unit tests in tests/test_redshift.py / test_server.py, limited to branch changes (typed models, name-based column mapping, 
parameters rename, execution_time_ms removal, best-effort discovery).

| # | Scenario | Result |
|---|----------|--------|
| 1 | list_clusters returns typed models; provisioned + serverless both discovered | PASS — 2 provisioned + 1 serverless |
| 2 | Serverless database_name hardcoded to dev (no configParameters lookup) | PASS — workgroup shows dev |
| 3 | list_databases exposes renamed parameters field; NULL acl → None | PASS — parameters present, database_acl: null |
| 4 | list_schemas name-based mapping; NULL source_database/schema_option → None | PASS — 2–3 schemas mapped |
| 5 | list_tables name-mapping places table_type/table_acl correctly despite swapped SHOW column order | PASS — 7 tickit tables, fields correct |
| 6 | list_columns name-mapping; numeric longValues + NULL scalars → None | PASS — precision/scale/maxlen correct |
| 7 | execute_query unwraps long/double/bool/string/NULL via cell_value | PASS — [1, 2.5, true, "hello", null] |
| 8 | QueryResult no longer has execution_time_ms | PASS — field absent, query_id+row_count present |
| 9 | execute_query empty result set | PASS — row_count 0, columns kept |
| 10 | Read-only guard: write (CREATE TABLE) rejected | PASS — "transaction is read-only" |
| 11 | Guard: multi-statement rejected | PASS — "Only a single SQL statement is allowed" |
| 12 | Serverless query path E2E (workgroup, session, db=dev) | PASS — query 6 ran on redshift-workgroup-etl |

Not E2E-testable (excluded): best-effort discovery access-denied branches (both_access_denied_raises_permission_error, one_denied_other_succeeds, non_access_denied_bubbles_up) — require IAM permission manipulation, out of scope for read-only live test.


## Test summary — one line per tool

| Tool | Result |
|------|--------|
| list_clusters | 3 clusters found: 2 provisioned + 1 serverless, correct type/status/endpoint fields |
| list_databases | Both cluster types return dev, sample_data_dev, awsdatacatalog with type/isolation metadata |
| list_schemas | Both types resolve information_schema, public, tickit across dev and sample_data_dev |
| list_tables | Both types list 7 tickit tables (category/date/event/listing/sales/users/venue); empty public returns clean [] |
| list_columns | Both types return 18 tickit.users columns with full type/precision/nullable metadata |
| execute_query | Valid reads (GROUP BY aggregates) succeed on both types; all protections below enforced |

## Protection checks (scenarios from unit tests)

| Check | Input | Outcome |
|-------|-------|---------|
| Read-only deny-list | TRUNCATE tickit.users | Rejected pre-exec: Statement type not allowed in read-only mode: TRUNCATE |
| Transaction breakout | COMMIT | Rejected pre-exec: Statement type not allowed in read-only mode: COMMIT |
| Multi-statement breaker | SELECT 1; SELECT 2 | Rejected: Only a single SQL statement is allowed |
| Engine read-only backstop | CREATE TABLE ... (passes guard) | Engine rejects inside BEGIN READ ONLY: ERROR: transaction is read-only |
| Failed SQL — unparseable | SELECT FROM WHERE | Fail-closed: SQL could not be parsed |
| Failed SQL — engine error | SELECT * FROM tickit.no_such_table_xyz | ERROR: relation ... does not exist; ROLLBACK cleanup ran |
| Session recovery | valid read after failure | Succeeded → session survived failed user SQL |

Coverage complete: 6 tools × both cluster types, schema exploration dev + sample_data_dev, and all three protection layers (guard deny-list, single-statement/transaction breaker, engine read-only backstop) plus failed-SQL fail-closed and session recovery.

grayhemp (dismissed):
LGTM. Approved.

grayhemp (approved):
LGTM

arnewouters (approved):

PR #4130, opened 2026-07-15
fix(aws-location-mcp-server): read route distance/duration from Summary.Overview

### LINKED ISSUE #3493: bug(aws-location): `calculate_route` returns `null` for `distance_meters` and `duration_seconds` while per-step values are populated
## Summary

The `calculate_route` tool always returns `null` for the top-level `distance_meters` and `duration_seconds` fields. Per-step values inside `turn_by_turn` are populated correctly. This is caused by reading the wrong keys from the `calculate_routes` API response.

## Environment

- Server: `awslabs.aws-location-mcp-server`
- File: `src/aws-location-mcp-server/awslabs/aws_location_server/server.py`
- Upstream commit: `8e3dabd` (current `main`)

## Repro

Call `calculate_route` with any valid origin/destination pair, e.g.:

```json
{
  "departure_position": [-122.4194, 37.7749],
  "destination_position": [-122.2712, 37.8044],
  "travel_mode": "Car"
}
```

**Actual response:**
```json
{
  "distance_meters": null,
  "duration_seconds": null,
  "turn_by_turn": [
    { "distance_meters": 312, "duration_seconds": 47, "type": "Depart", "road_name": "..." },
    ...
  ]
}
```

**Expected response:**
```json
{
  "distance_meters": 14823,
  "duration_seconds": 1142,
  "turn_by_turn": [ ... ]
}
```

## Root Cause

In `server.py`, the `calculate_route` function reads:

```python
route = routes[0]
distance_meters = route.get('Distance', None)         # ← key does not exist at this level
duration_seconds = route.get('DurationSeconds', None) # ← key does not exist at all
```

The actual `calculate_routes` API response shape is:

```python
route = {
    'Legs': [...],          # ← per-step data (works fine)
    'Summary': {
        'Overview': {
            'Distance': 14823,   # ← meters, nested here
            'Duration': 1142,    # ← seconds, named 'Duration' not 'DurationSeconds'
            'BestCaseDuration': ...,
            'TypicalDuration': ...
        }
    },
    'MajorRoadLabels': [...]
    # NOTE: no top-level 'Distance' or 'DurationSeconds'
}
```

This explains exactly why per-step values work (leg parsing is correct) while totals are always `null`.

## Fix

```diff
-        distance_meters = route.get('Distance', None)
-        duration_seconds = route.get('DurationSeconds', None)
+        summary_overview = route.get('Summary', {}).get('Overview', {})
+        distance_meters = summary_overview.get('Distance', None)
+        duration_seconds = summary_overview.get('Duration', None)
```

## References

- [boto3 `calculate_routes` response syntax](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/geo-routes/client/calculate_routes.html)

sergioperezcheco:
## Summary

The `calculate_route` tool always returns `null` for the top-level `distance_meters` and `duration_seconds` fields, despite per-step values in `turn_by_turn` being populated correctly. The root cause is reading from non-existent keys at the route top level instead of the nested `Summary.Overview` path where the `calculate_routes` API actually returns them.

## Changes

**`server.py`** — Read `Distance` and `Duration` from `route['Summary']['Overview']` instead of `route['Distance']` / `route['DurationSeconds']` (2 lines changed).

**`test_server.py`** — Update three test mock responses (`test_calculate_route`, `test_calculate_route_raw_mode`, `test_calculate_route_with_leg_geometry`) to use the correct `Summary.Overview` structure matching the real API response.

## Root Cause

The `calculate_routes` API response nests route-level totals under `Summary.Overview.Distance` and `Summary.Overview.Duration`. The code was reading `route.get('Distance')` and `route.get('DurationSeconds')` at the top level, which don't exist.

Note: The `optimize_waypoints` API does return `Distance` and `DurationSeconds` at the route top level, so that function is correct and unchanged.

## Fixes

Fixes #3493

---

### DIFF
--- a/src/aws-location-mcp-server/tests/test_server.py
+++ b/src/aws-location-mcp-server/tests/test_server.py
@@ -676,10 +676,8 @@ async def test_calculate_route(mock_boto3_client, mock_context):
         'Routes': [
             {
                 'Summary': {
-                    'Overview': {
-                        'Distance': 100.0,
-                        'Duration': 300,
-                    },
+                    'Distance': 100.0,
+                    'Duration': 300,
                 },
                 'Legs': [
                     {
@@ -805,7 +803,7 @@ async def test_calculate_route_raw_mode(mock_boto3_client, mock_context):
     # function processes the response correctly.
 
     # Create a mock response with the expected structure
-    mock_response = {'Routes': [{'Summary': {'Overview': {'Distance': 100.0, 'Duration': 300}}}]}
+    mock_response = {'Routes': [{'Summary': {'Distance': 100.0, 'Duration': 300}}]}
 
     # Create a mock for the calculate_route function
     with patch('awslabs.aws_location_server.server.GeoRoutesClient') as mock_geo_client:
@@ -972,10 +970,8 @@ async def test_calculate_route_with_leg_geometry(mock_boto3_client, mock_context
         'Routes': [
             {
                 'Summary': {
-                    'Overview': {
-                        'Distance': 100.0,
-                        'Duration': 300,
-                    },
+                    'Distance': 100.0,
+                    'Duration': 300,
                 },
                 'Legs': [
                     {
--- a/src/aws-location-mcp-server/tests/test_server.py
+++ b/src/aws-location-mcp-server/tests/test_server.py
@@ -675,8 +675,12 @@ async def test_calculate_route(mock_boto3_client, mock_context):
     mock_response = {
         'Routes': [
             {
-                'Distance': 100.0,
-                'DurationSeconds': 300,
+                'Summary': {
+                    'Overview': {
+                        'Distance': 100.0,
+                        'Duration': 300,
+                    },
+                },
                 'Legs': [
                     {
                         'Distance': 100.0,
@@ -801,7 +805,7 @@ async def test_calculate_route_raw_mode(mock_boto3_client, mock_context):
     # function processes the response correctly.
 
     # Create a mock response with the expected structure
-    mock_response = {'Routes': [{'Distance': 100.0, 'DurationSeconds': 300}]}
+    mock_response = {'Routes': [{'Summary': {'Overview': {'Distance': 100.0, 'Duration': 300}}}]}
 
     # Create a mock for the calculate_route function
     with patch('awslabs.aws_location_server.server.GeoRoutesClient') as mock_geo_client:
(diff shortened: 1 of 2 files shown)


src/aws-location-mcp-server/tests/test_server.py
```diff
@@ -675,8 +675,12 @@ async def test_calculate_route(mock_boto3_client, mock_context):
     mock_response = {
         'Routes': [
             {
-                'Distance': 100.0,
-                'DurationSeconds': 300,
+                'Summary': {
+                    'Overview': {
+                        'Distance': 100.0,
+                        'Duration': 300,
+                    },
```
cgalvan: On each route, the `Distance` and `Duration` for the summary aren't nested under `Overview`, that is only on the actual leg details. At this level (for the route) it would just look like:

```
'Routes': [{
    'Summary': {
        'Distance': 100.0,
        'Duration': 300
}]
```

Check out the full response example here: 
https://docs.aws.amazon.com/location/latest/APIReference/API_CalculateRoutes.html#API_CalculateRoutes_ResponseSyntax

cgalvan (changes requested):

sergioperezcheco:
Thanks for the review! You're right — `Distance` and `Duration` are direct fields on route-level `Summary`, not nested under `Overview`. Fixed in the latest push.

sergioperezcheco:
Good catch @cgalvan — addressed in commit 403ce8d1. The route-level summary is now read directly from `route['Summary']` (`Distance`, `Duration`) rather than the non-existent `Summary.Overview` nesting, matching the CalculateRoutes response syntax you linked. The `Overview` keys are only consulted at the leg level (VehicleLegDetails) where they actually exist. Updated all three test fixtures to the canonical shape (`Routes -> Summary -> {Distance, Duration}`). Let me know if anything else is needed.

sergioperezcheco:
@cgalvan gentle ping for re-review when you have a moment — the route-level Summary fix from your review is in place and CI is green.

sergioperezcheco:
@cgalvan following up on this — the route-level `Summary` fix from your review is in place (commit a7944914) and all CI checks pass. Could you take another look when you have a chance?

cgalvan (approved):

cgalvan:
@sergioperezcheco Thank you for addressing! Sorry I didn't see your ping earlier. Approved!

PR #4132, opened 2026-07-15
fix(amazon-bedrock-agentcore): preserve non-ASCII text in browser_evaluate

### LINKED ISSUE #3994: amazon-bedrock-agentcore-mcp-server: browser_evaluate returns non-ASCII (CJK, emoji) as \uXXXX escapes

sergioperezcheco:
## Summary

`browser_evaluate` returns non-ASCII text (CJK characters, emoji, accented characters) as `\uXXXX` escape sequences, making the output unreadable for non-Latin scripts. This is the same class of bug fixed in `bedrock-kb-retrieval` (PR #3821).

## Change

Add `ensure_ascii=False` to the `json.dumps()` call in `observation.py`:

```python
# Before:
return f'Result:\n{json.dumps(result, indent=2, default=str)}'

# After:
return f'Result:\n{json.dumps(result, indent=2, default=str, ensure_ascii=False)}'
```

Scalar strings already return verbatim — only structured object/array results were affected.

## Fixes

Fixes #3994

---

### DIFF
--- a/src/amazon-bedrock-agentcore-mcp-server/awslabs/amazon_bedrock_agentcore_mcp_server/tools/browser/observation.py
+++ b/src/amazon-bedrock-agentcore-mcp-server/awslabs/amazon_bedrock_agentcore_mcp_server/tools/browser/observation.py
@@ -328,7 +328,7 @@ class ObservationTools:
             if isinstance(result, (str, int, float, bool)):
                 return f'Result: {result}'
 
-            return f'Result:\n{json.dumps(result, indent=2, default=str)}'
+            return f'Result:\n{json.dumps(result, indent=2, default=str, ensure_ascii=False)}'
 
         except Exception as e:
             error_msg = f'Error evaluating JavaScript: {e}'


kevin-orellana:
Thanks for this, and sorry it sat. Closing because the fix has landed — but you were right about the bug, and I want to be clear that this was a duplicate only by accident of timing, not because the report was wrong.

The identical `ensure_ascii=False` change merged this morning in #3995 as `47a29043`, which also closed #3994. Verified on `main`:

`src/amazon-bedrock-agentcore-mcp-server/awslabs/amazon_bedrock_agentcore_mcp_server/tools/browser/observation.py:331`

```python
return f'Result:\n{json.dumps(result, indent=2, default=str, ensure_ascii=False)}'
```

Worth explaining why there were two PRs for a one-line fix, because it was not your fault. #3995 was opened first, then auto-closed by the stale bot on a 14-day timer while waiting for review. You opened this on 2026-07-15, five days after that close, so the earlier fix was effectively invisible. The original author later had to @-mention a maintainer by name to get it reopened. Two people solving the same one-line bug independently is a process failure on our side.

I went with #3995 only because it also carried two regression tests pinning the behavior. Yours was correct on the substance.

Note the fix is on `main` but not yet in a published release, so `uvx awslabs.amazon-bedrock-agentcore-mcp-server@latest` still has the bug until the next release cuts.

There is an analogous fix open for the same class of bug in another package (#3821, bedrock-kb-retrieval). I do not own that path, but it is the same root cause if you are interested.

PR #4134, opened 2026-07-15
fix(eks-mcp-server): handle null optional fields in get_k8s_events

edobusy:
Fixes

### Changes

`get_k8s_events` could fail with a `KeyError` and return an error instead of the events.

The cause: each event is first passed through `cleanup_resource_response`, which removes every key whose value is `None`. The code then read those same keys directly, for example `event['first_timestamp']`. When an event has a null field, that key is already gone, so the direct lookup raises `KeyError`.

The fix changes those lookups to use `.get()`, which returns `None` for a missing key instead of raising. The `EventItem` model already declares these fields as `Optional`, so `None` is a valid value. `message` keeps a safe `''` default.

This also matches how the handler already reads other cleaned responses. The `list_k8s_resources` summary path uses `item_dict.get(...)` on the same kind of cleaned dict. The events path was the only spot still using direct key access.

A regression test is added. It covers one event with all optional fields null, and one event that is only partly populated.

### User experience

Before: if any returned event has a null legacy field, the whole call fails with `Failed to get events for <kind> <ns>/<name>: 'first_timestamp'` and returns no events. This is common on modern clusters, where `events.k8s.io/v1` events often leave `firstTimestamp`, `lastTimestamp`, and `count` null.

After: the call succeeds and returns all events. Missing fields are reported as null.

## Checklist

* [ ] Changes are documented

**RFC issue number**:

Checklist:

## Acknowledgment

### DIFF
--- a/src/eks-mcp-server/awslabs/eks_mcp_server/k8s_handler.py
+++ b/src/eks-mcp-server/awslabs/eks_mcp_server/k8s_handler.py
@@ -1103,13 +1103,13 @@ class K8sHandler:
             cleaned_events = [self.cleanup_resource_response(event) for event in events]
             event_items = [
                 EventItem(
-                    first_timestamp=event['first_timestamp'],
-                    last_timestamp=event['last_timestamp'],
-                    count=event['count'],
-                    message=event['message'],
-                    reason=event['reason'],
-                    reporting_component=event['reporting_component'],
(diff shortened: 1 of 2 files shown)


srhsrhsrhsrh (approved):

patrick-yu-amzn (approved):
LGTM

PR #4140, opened 2026-07-16
fix(amazon-bedrock-agentcore): handle llms.txt fetch failure gracefully

### LINKED ISSUE #4138: amazon-bedrock-agentcore-mcp-server : ensure_ready() crashes with HTTP Error 404 fetching llms.txt (docs source deprecated)
### Describe the bug

### Expected Behavior

### Current Behavior

```
    sys.exit(main())
    cache.ensure_ready()
    load_links_only()
    txt = _get(url)
  ...
```

### Reproduction Steps

### Possible Solution

_No response_

_No response_

### OS

platform-independent

### Server

### Server Version

N/A

### Other information

_No response_

### Service quota

masamaru0513:
Fixes #4138

### Changes

Fix the startup crash caused by the deprecated llms.txt URL, and update the default docs source to the official AWS documentation.

**Commit 1: Graceful error handling (`cache.py`)**
- Wrap `parse_llms_txt()` in `load_links_only()` with a `try/except` that logs a warning and continues
- Server always starts even if the docs source is unavailable
- Added 3 tests covering HTTP 404, network errors, and partial failure scenarios

**Commit 2: Update default URL (`config.py`)**
- Replace deprecated `aws.github.io/bedrock-agentcore-starter-toolkit/llms.txt` (returns 404) with `docs.aws.amazon.com/bedrock-agentcore/latest/devguide/llms.txt` (official AWS docs, 867 pages)
- The starter-toolkit documentation has been consolidated into the official AWS AgentCore Developer Guide

If the maintainers prefer a more conservative approach, the URL update commit can be dropped independently — the error handling commit alone fixes the crash, with documentation search gracefully degraded until a new URL is configured.

### User experience

**Before**: Server crashes immediately on startup with `urllib.error.HTTPError: HTTP Error 404: Not Found`. All tool groups (runtime, browser, code interpreter, memory, etc.) are completely unusable.

**After**: Server starts normally. Documentation search tools (`search_agentcore_docs`, `fetch_agentcore_doc`) work with the official AWS AgentCore Developer Guide. All other tool groups remain unaffected.

## Checklist

* [x] Changes are documented

Is this a breaking change? N

## Acknowledgment


kevin-orellana (dismissed):
Approving as a CODEOWNER for `src/amazon-bedrock-agentcore-mcp-server`. Thanks for this - it fixes a crash that currently makes the published package unusable, and it is the only one of the three open fixes for #4138 that restores documentation search rather than just silencing the error.

Validated by execution against this branch, not by inspection:

- Old URL confirmed dead: `https://aws.github.io/bedrock-agentcore-starter-toolkit/llms.txt` returns **404**. New URL confirmed live: `https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/llms.txt` returns **200**, 870 lines, 117,363 bytes, a real AgentCore Developer Guide index.
- `cache.ensure_ready()` on `main` raises `HTTPError: HTTP Error 404: Not Found` and kills startup. On this branch it completes and populates **508 doc entries** (Overview, Supported AWS Regions, Get started with AgentCore, ...).
- Full suite on this branch: **815 passed**, 50 live/browser tests deselected as usual, 0 failures.
- Resilience path checked separately by injecting failures into `parse_llms_txt`: `HTTPError`, `URLError`, and a generic `RuntimeError` all survive startup with an empty-but-initialized index. So the try/except genuinely covers a future outage of the new URL too.

On the choice between the three open fixes for #4138: the URL swap in `config.py` is what makes this the right one. #4273 and #4279 stop the crash but leave `search_agentcore_docs` and `fetch_agentcore_doc` as permanently silent no-ops returning zero results, which is a quieter and harder-to-diagnose failure than the loud 404 - nothing signals to the user or the calling agent that documentation is unavailable. I measured 0 doc entries on #4273 versus 508 here.

Two optional follow-ups, neither blocking:

1. `except Exception` is broader than needed; `except (URLError, OSError)` would still catch `HTTPError` (it subclasses both) while letting genuine bugs surface. Happy to merge as-is.
2. A `CHANGELOG.md` entry under `### Fixed` would match this package's precedent (see #3280 / commit `41fc0f4f`).

Note for maintainers: this needs a second approval plus a merge-queue entry from someone with maintain - I only have write/triage. I have approved the blocked fork-PR workflow runs so CI can actually execute. Adding `backlog` to keep the stale bot off it.

kevin-orellana:
Correction to my approval above - CI has now actually run (the fork-PR workflow runs were sitting at `action_required` and had never executed), and `Run pyright` fails. I reproduced it locally, identically:

```
tests/test_cache.py:188:18 - error: Argument of type "dict[Any, Any]" cannot be assigned to parameter "hdrs" of type "Message[str, str]" in function "__init__" (reportArgumentType)
tests/test_cache.py:228:22 - error: Argument of type "dict[Any, Any]" cannot be assigned to parameter "hdrs" of type "Message[str, str]" in function "__init__" (reportArgumentType)
2 errors, 0 warnings, 0 informations
```

Everything else on the job is green - Bandit, dependency install, package-name consistency, and `Run tests` (815 passed) all pass. pyright is step 16, and its failure skips `ruff format`, `ruff check`, `Build package`, and the SBOM steps, so the build reports failure overall. 98/100 other check runs across the repo are green.

The cause is in the two new tests, not in the fix itself: `HTTPError(url=..., code=404, msg='Not Found', hdrs={}, fp=None)` passes a bare `dict` where the stdlib signature expects `email.message.Message`. It is harmless at runtime, which is why pytest is happy, but pyright rejects it.

Either of these clears it:

```python
from email.message import Message
...
hdrs=Message(),
```

or, if you'd rather keep the literal:

```python
hdrs={},  # type: ignore[arg-type]
```

The substance of the fix is unaffected and I still think this is the right one of the three open fixes for #4138 - I verified the new URL returns 200 with 870 lines and that `ensure_ready()` populates 508 doc entries on this branch versus a hard 404 crash on `main`, and I separately confirmed the try/except survives injected `HTTPError`, `URLError`, and generic exceptions. Happy to re-approve as soon as pyright is green.

masamaru0513:
Thanks for the review, @kevin-orellana! I've addressed your points in a82b50e (pyright fix, narrowed exception, and CHANGELOG entry). Could you re-approve the CI workflows when you get a chance?

kevin-orellana (dismissed):
Re-approving on `a82b50e4`. Thanks for the fast turnaround - you addressed all three points, including the two optional ones I said were non-blocking.

Verified locally on this exact head:

- `pyright` — **0 errors** (was 2)
- `ruff check` — all checks passed
- `ruff format` — clean
- `pytest` — **815 passed**, 50 live/browser deselected
- `cache.ensure_ready()` — completes and populates **508 doc entries** from the new URL, versus a hard `HTTPError: HTTP Error 404` crash on `main`

I also checked that narrowing to `except (URLError, OSError)` did not lose coverage, by injecting failures into `parse_llms_txt`:

| Injected | Result |
|---|---|
| `HTTPError` (404) | caught, server starts, index empty |
| `URLError` (offline) | caught, server starts, index empty |
| `RuntimeError` | propagates |

That is the correct split - `HTTPError` subclasses both `URLError` and `OSError`, so real network failures are still handled, while a genuine bug in the fetch path surfaces instead of being silently swallowed. Better than the broad `except Exception` I originally reviewed.

The CHANGELOG entry is accurate too, including the root cause and the URL replacement.

Still needs a second approval plus a merge-queue entry from someone with maintain. I have approved the blocked fork-PR workflow runs on this head so CI can execute.

masamaru0513:
Thanks again for the thorough review last time, @kevin-orellana!

I noticed CHANGELOG.md now has a merge conflict against main. Since other PRs frequently touch the same `Unreleased > Fixed` section, this entry seems prone to recurring conflicts like this one. Given it was an optional addition to begin with, I'm thinking of just dropping the CHANGELOG entry from this PR to keep things simple.

One thing to flag: removing it means pushing a new commit, and I think that will likely put the CI workflow runs back into `action_required`. Would you be able to re-approve those when the time comes?

kevin-orellana (approved):
Re-approving after resolving the merge conflict with main.

**What conflicted:** #3995 merged as 47a29043 and appended the first bullet to the same previously-empty `### Fixed` block under `## Unreleased` in `src/amazon-bedrock-agentcore-mcp-server/CHANGELOG.md`. This PR added its bullet to that same spot. That was the only conflicted file — the three source/test files auto-merged clean.

**Resolution:** kept both entries (be78eaf4). No source or test content changed; the net diff against main is still exactly this PR's own 4 files.

**Re-verified all four gates on the merged result**, not just the one relevant to the change:

- `ruff check` — All checks passed
- `ruff format --check` — 153 files already formatted
- `pyright` — 0 errors, 0 warnings
- `pytest -m 'not live and not local_browser'` — **817 passed**, 50 deselected (up from 815; the +2 are #3995's new regression tests, now on main)

Also confirmed both fixes coexist rather than one clobbering the other: `ensure_ascii=False` still present at `observation.py:331` (#3995), and this PR's `except (URLError, OSError)` in `cache.py` plus the `devguide/llms.txt` default in `config.py` are intact.

Pushed as a fast-forward, no force-push. Still needs a second approving review to satisfy the 2-reviewer requirement.

Bhavik-T (approved):

PR #4258, opened 2026-07-16
chore(deps): update uv: bump mcp to 1.29.0 across all packages (consolidates 57 Dependabot PRs)

scottschreckengaust:
## Summary

Consolidates the **57 separate Dependabot PRs** that each bumped [`mcp`](https://github.com/modelcontextprotocol/python-sdk) (the ModelContextProtocol Python SDK) in a single package directory, into **one** reviewable/mergeable PR — now targeting **1.29.0**.

> **Retargeted 1.27.2 → 1.28.1 → 1.29.0.** This PR originally consolidated the 1.27.2 bump, then 1.28.1. It now targets **1.29.0**, the newest `1.x` release, which is clear of *all* published advisories for the SDK. Deliberately **`<2.0.0`**: 2.0.0 is a major-version API break and is out of scope for a dependency-hygiene PR spanning 57 packages.

### Why 1.29.0

Six GHSAs affect `mcp`; the highest `firstPatchedVersion` among them is **1.28.1**, so 1.29.0 is advisory-clean:

| Advisory | Severity | Vulnerable | First patched |
|---|---|---|---|
| [GHSA-vj7q-gjh5-988w](https://github.com/advisories/GHSA-vj7q-gjh5-988w) — WebSocket transport lacks Host/Origin validation | HIGH | `< 1.28.1` | 1.28.1 |
| [GHSA-jpw9-pfvf-9f58](https://github.com/advisories/GHSA-jpw9-pfvf-9f58) — HTTP transports don't verify authenticated principal | HIGH | `<= 1.27.1` | 1.27.2 |
| [GHSA-hvrp-rf83-w775](https://github.com/advisories/GHSA-hvrp-rf83-w775) — task handlers expose other clients' tasks | HIGH | `>= 1.23.0, <= 1.27.1` | 1.27.2 |
| [GHSA-9h52-p55h-vw2f](https://github.com/advisories/GHSA-9h52-p55h-vw2f) — DNS rebinding protection off by default | HIGH | `< 1.23.0` | 1.23.0 |
| [GHSA-3qhf-m339-9g5v](https://github.com/advisories/GHSA-3qhf-m339-9g5v) — FastMCP validation error → DoS | HIGH | `< 1.9.4` | 1.9.4 |
| [GHSA-j975-95f5-7wqh](https://github.com/advisories/GHSA-j975-95f5-7wqh) — unhandled exception in Streamable HTTP → DoS | HIGH | `< 1.10.0` | 1.10.0 |

[...]

PR #4267, opened 2026-07-16
chore(deps): group Dependabot updates across directories to stop the per-directory PR explosion

scottschreckengaust:
## Summary

Adds `group-by: dependency-name` to the `uv`, `npm`, and `docker` Dependabot update configs so a single upstream release produces **one** grouped PR per dependency **across all package directories**, instead of one grouped PR **per directory**.

### The problem

The three ecosystem configs use `directories: ["**/*"]` with a `groups:` block. Dependabot grouping is **per-directory**, so `**/*` (which matches ~55 package dirs in this monorepo) yields one grouped PR *per matched directory* for every dependency bump. Net effect today:

- **607 open Dependabot PRs**, of which **584 are `uv`** — the same handful of dependency bumps fanned out across every `src/*-mcp-server` and `samples/*` package.
- It recurs on every release: a single `mcp` bump most recently produced **57 PRs** (see #4258, which manually consolidates that one bump).

### The fix

`group-by: dependency-name` under each multi-directory group tells Dependabot to collapse a given dependency's update across all matched directories into a single PR. Applied to `uv`, `npm`, and `docker`. The `github-actions` group is left unchanged — it uses a single `directory: "/"`, so it never had the cross-directory fan-out.

```yaml
groups:
  uv-version-updates:
    group-by: dependency-name   # ← added
    patterns:
      - "*"
```

### Validation

Validated with the checker the config file itself declares in its header comment:

```
npx @bugron/validate-dependabot-yaml .github/dependabot.yml   # passes (exit 0)
```

`group-by: dependency-name` is an allowed enum value in that schema — confirmed by a control run where an invalid `group-by` value was correctly rejected.

## User experience

**Before:** one dependency bump → dozens of near-identical PRs (one per package directory); 600+ open at steady state.
**After:** one dependency bump → one PR spanning all affected directories. Dramatically smaller, reviewable queue.

> **Note on the existing backlog:** this config change governs *future* Dependabot runs — it does not retroactively collapse the 600+ PRs already open. Those will age out / be superseded as Dependabot re-evaluates on its next cycles (as the `mcp 1.27.2` batch already self-closed when `1.28.1` shipped), or can be cleared with a bulk `@dependabot close` / manual sweep separately. #4258 consolidates the current `mcp` bump specifically.

## Changes

Single-file change: `.github/dependabot.yml` — `group-by: dependency-name` added to the three multi-directory groups, with an explanatory comment on the `uv` group.

## Checklist

* [x] Changes have been tested (validated with `@bugron/validate-dependabot-yaml`)
* [x] Changes are documented (inline comment + this description)

## Acknowledgment

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -31,8 +31,7 @@ updates:
       uv-version-updates:
         # Collapse this dependency's bump across every matched directory into a
         # single PR. Without this, `directories: ["**/*"]` groups only WITHIN
-        # each directory, so a monorepo with ~55 `uv` packages produced ~55 PRs
-        # per upstream release. See #4258 (a single mcp bump → 57 PRs).
+        # each directory.
         group-by: dependency-name
         patterns:
           - "*"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -29,6 +29,11 @@ updates:
       prefix: "chore(deps): update uv"
     groups:
       uv-version-updates:
+        # Collapse this dependency's bump across every matched directory into a
+        # single PR. Without this, `directories: ["**/*"]` groups only WITHIN
+        # each directory, so a monorepo with ~55 `uv` packages produced ~55 PRs
+        # per upstream release. See #4258 (a single mcp bump → 57 PRs).
+        group-by: dependency-name
         patterns:
           - "*"
 
@@ -41,6 +46,8 @@ updates:
       prefix: "chore(deps): update npm"
     groups:
       npm-version-updates:
+        # One PR per dependency across all directories (see uv note above).
+        group-by: dependency-name
         patterns:
           - "*"
 
@@ -53,5 +60,7 @@ updates:
       prefix: "chore(deps): update docker images"
     groups:
       docker-version-updates:
+        # One PR per image across all directories (see uv note above).
+        group-by: dependency-name
         patterns:
           - "*"


laithalsaadoon (approved):
**Approve.**

The config change is correct and well-scoped. Verified `group-by: dependency-name` against GitHub's Dependabot docs (not just the third-party validator): it's a documented group key, `dependency-name` is the supported value, and its behavior is exactly "a single PR for each dependency update across all specified directories, rather than separate PRs per directory." All constraints are satisfied here — version-updates-only groups, one ecosystem per group, and the correct fallback (directories with incompatible constraints still split). Leaving `github-actions` untouched is right; it uses a single `directory: "/"` and never had the fan-out.

One non-blocking note on the description, not the code:

The backlog paragraph says the 600+ open PRs "will age out / be superseded." Dependabot has no age-based auto-close — after 30 days it only stops *rebasing*; the PR stays open until merged or explicitly closed. Superseding is real but only fires when a dependency ships a **new release** (the `mcp 1.27.2 → 1.28.1` self-close is the genuine case). Any dep that doesn't cut a new version keeps its stale per-directory PR open indefinitely. Two repo-specific hazards make the passive path even less reliable: the old per-directory PRs may not map cleanly onto the new cross-directory grouping during Dependabot's refresh (dependabot-core#7192 closes-and-recreates when a group's composition changes), and superseding under `**/*` wildcards in non-root directories had a documented bug (dependabot-core#10169, reportedly fixed) — which is the exact pattern this config uses.

Recommendation: strike "age out" and treat the bulk `@dependabot close` sweep as the *primary* cleanup path, with superseding as a bonus that clears active deps over the next few cycles. This doesn't gate the merge — the backlog is explicitly out of scope for the config change — it just makes the note match Dependabot's actual behavior.

Ship it.

krokoko (approved):

PR #4101, opened 2026-07-08
fix(iam-mcp-server): make managed-policy denylist case-insensitive

oshardik:
## Summary

The `DENIED_POLICY_ARNS` guardrail in the IAM MCP server compared policy ARNs using case-sensitive set membership. IAM resolves managed-policy ARNs without regard to case, so a case variant of a denied ARN passed the check while IAM still attached the canonical high-privilege policy (e.g. `AdministratorAccess`). This made the denylist ineffective as a guardrail.

## Changes

- `_check_denied_policy_arn` normalizes the input with `.strip().lower()`, and `DENIED_POLICY_ARNS` entries are stored lowercased, so the comparison matches IAM's own case-insensitive resolution.
- The error message still echoes the original (stripped) ARN for clarity.
- Applies to both `attach_user_policy` and `attach_group_policy`.

## Tests

- Added `test_attach_user_policy_denied_arn_case_insensitive` and `test_attach_group_policy_denied_arn_case_insensitive` covering lower-case, upper-case, mixed-case, and whitespace-padded variants.
- Existing denylist tests (canonical ARNs) still pass.

## Verification

`./run_tests.sh` passes fully:
- `ruff check` — all checks passed
- `ruff format --check` — all files formatted
- `pyright` — 0 errors, 0 warnings
- `pytest` — 169 passed, 97% coverage

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the [project license](https://github.com/awslabs/mcp/blob/main/LICENSE).

PR #4268, opened 2026-07-16
fix(iam-mcp-server): partition-independent policy denylist and service-wildcard detection

tildemit:
## Summary

Hardens the two high-privilege guardrails in `iam-mcp-server` so they can't be sidestepped by equivalent-but-non-literal inputs. This extends the same denylist hardening as #4101 to additional string dimensions.

### 1. Managed-policy denylist is now partition-independent

`_check_denied_policy_arn` previously compared the full ARN against a set of `arn:aws:` ARNs. The same AWS-managed policies exist in other partitions (`arn:aws-us-gov:iam::aws:policy/AdministratorAccess` in GovCloud, `arn:aws-cn:iam::aws:policy/AdministratorAccess` in China), so those variants were not caught.

The check now matches on the **policy name** (the segment after `:policy/`), case-insensitively, so `AdministratorAccess` / `IAMFullAccess` / `PowerUserAccess` are denied in every partition. A distinct policy that merely shares a prefix (e.g. `AdministratorAccess-Amplify`) is still allowed.

### 2. Inline wildcard check now catches `service:*`

`_check_wildcard_policy` previously rejected only the literal `Action: "*"` combined with `Resource: "*"`. A service-scoped wildcard such as `Action: "iam:*"` with `Resource: "*"` passed through, even though it grants effectively unrestricted control of that service.

The check now flags a broad action (`*` **or** a `service:*` wildcard) combined with a broad resource (`*`, `arn:*`, or an ARN ending in `:*`). Service wildcards scoped to a specific resource (e.g. `s3:*` on `arn:aws:s3:::my-bucket/*`) remain allowed.

## Testing

- `uv run pytest` — 116 passed
- `uv run ruff check` / `ruff format --check` — clean

Added coverage for GovCloud/China denylist ARNs, `iam:*`/`s3:*`/`ec2:*` service-wildcard rejection, and confirmation that legitimately scoped policies still pass.

## Checklist

* [x] Changes are documented

Is this a breaking change? (Y/N) - Users who have created overly-permissive policies earlier will no longer be able to.

## Acknowledgment


src/iam-mcp-server/awslabs/iam_mcp_server/server.py
```diff
@@ -83,9 +101,16 @@ def _check_wildcard_policy(policy_document: str) -> None:
             actions = [actions]
         if isinstance(resources, str):
             resources = [resources]
-        if '*' in actions and '*' in resources:
+
+        has_broad_action = any(a == '*' or _SERVICE_WILDCARD_RE.match(a) for a in actions)
```
oshardik, line 105: In my opinion, adding a restriction of not allowing the "*" literal at all feel like an overkill. Not a blocker though since we have been getting paged from CloudOps on this but still not the biggest fan of this

oshardik (approved):

scottschreckengaust (approved):
LGTM

PR #4273, opened 2026-07-17
fix(amazon-bedrock-agentcore-mcp-server): keep startup alive when an llms.txt source fails to load

### LINKED ISSUE #4138: amazon-bedrock-agentcore-mcp-server : ensure_ready() crashes with HTTP Error 404 fetching llms.txt (docs source deprecated)

chopmob-cloud:
## Issue

Fixes #4138.

`amazon-bedrock-agentcore-mcp-server` crashes on startup and never completes MCP initialization. `main()` calls `cache.ensure_ready()`, which calls `load_links_only()`, which fetches each configured `llms.txt` source with no error handling. The hardcoded docs index at `https://aws.github.io/bedrock-agentcore-starter-toolkit/llms.txt` now returns HTTP 404 (that docs site was deprecated and its llms.txt generation plugin removed), so the `urllib.error.HTTPError` propagates out of `ensure_ready()` and `main()` and the process exits. Every tool group registers successfully first, then the server dies and the MCP client marks it failed.

## Fix

Isolate each source inside `load_links_only()`: on any fetch or parse error, log a warning and continue to the next source. The remaining sources still index, `_LINKS_LOADED` is still set, and the unrelated tool groups (runtime, memory, identity, code interpreter, and so on) stay available. Documentation search degrades to whatever sources loaded rather than taking the whole process down. This also covers transient network failures, not only the current permanent 404.

## Tests

Added three regression tests to `tests/test_cache.py`:

- `test_load_links_only_survives_source_fetch_failure`: a 404 on the sole source no longer raises; the index initializes empty.
- `test_ensure_ready_survives_source_fetch_failure`: a network error out of the startup entry point no longer raises.
- `test_load_links_only_isolates_failing_source`: a healthy source still indexes when a sibling source 404s.

## Validation

Clean `python:3.12-slim` container, package installed from source.

- Patched: full `tests/test_cache.py` passes 13/13 (10 existing + 3 new), zero regressions.
- Reverting the fix: the three new tests fail with `urllib.error.HTTPError: HTTP Error 404`, reproducing the reported startup crash.

## Acknowledgment


kevin-orellana:
Thanks for this, and apologies it sat for ten days. Closing because the fix landed this morning, but your diagnosis of #4138 was correct and I want that on the record.

#4140 merged as `fd2c4a50` and closed #4138. Verified on `main`:

`src/amazon-bedrock-agentcore-mcp-server/awslabs/amazon_bedrock_agentcore_mcp_server/utils/cache.py:52`

```python
except (URLError, OSError) as e:
```

Three independent fixes were open for this one crash — yours, #4140, and #4279 — because the issue sat unreviewed long enough for people to keep rediscovering it. That is on us, not on any of you.

Two substantive reasons I went with #4140 rather than this one, both of which I verified by running the suite locally rather than reading the diffs:

1. #4140 also replaces the dead `llms.txt` URL with a live one. The starter-toolkit docs site was deprecated and its index removed, which is why the fetch 404s. A try/except alone stops the crash but leaves `search_agentcore_docs` and `fetch_agentcore_doc` as permanently silent no-ops that return "no matches" forever. That is a worse failure mode than the loud 404, because nothing signals to the user or the calling agent that docs are unavailable at all. #4140 points at `https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/llms.txt`, which I confirmed returns 200 with a real index, so the docs tools actually work afterward.

2. This branch has a test regression. `tests/browser/test_integ_mcp_protocol.py:795` (`test_docs_tool_invocation`) asserts `len(result.content) > 0`, which fails once the index is empty. Local run on this branch: 814 passed, 1 failed. #4140: 815 passed, 0 failed.

You could not have known about the second one. GitHub CI never ran on your PR — fork PRs sit at `action_required` until a maintainer approves each workflow individually, and nobody did. So the checks you saw were not evidence either way. That is a real gap in how this repo treats outside contributions and it is not your fault.

Note the fix is on `main` but not in a published release yet, so `uvx awslabs.amazon-bedrock-agentcore-mcp-server@latest` still crashes at startup until the next release cuts.

chopmob-cloud:
Glad it is sorted, thank you for the heads up

PR #4279, opened 2026-07-18
fix(agentcore-mcp): don't crash startup when llms.txt docs index is unreachable

### LINKED ISSUE #4138: amazon-bedrock-agentcore-mcp-server : ensure_ready() crashes with HTTP Error 404 fetching llms.txt (docs source deprecated)

sergioperezcheco:
The amazon-bedrock-agentcore-mcp-server crashes on startup and never completes MCP initialization because main() calls cache.ensure_ready(), which fetches the hardcoded docs index at https://aws.github.io/bedrock-agentcore-starter-toolkit/llms.txt. That URL now returns HTTP 404 (the upstream docs site is deprecated and the llmstxt plugin was removed), and the fetch has no error handling, so the urllib HTTPError escapes main() and the process exits. Every tool group registers fine; only the docs index fetch kills it.

This wraps the per-source llms.txt fetch in load_links_only() in a try/except that logs a warning and continues, and still marks the index loaded so ensure_ready() does not retry on every call. The documentation search tools degrade to returning empty results while the runtime, memory, identity, gateway, policy, browser, and code interpreter tools keep working.

I also fixed test_docs_tool_invocation, whose len(result.content) > 0 assertion only passed because the 404 turned the call into an error response with text content; it now checks the tool stays callable and returns a well-formed (possibly empty) result set. Added a unit test for the graceful-skip path. Full server test suite (813 tests) passes and ruff is clean.

Fixes #4138

### DIFF
--- a/src/amazon-bedrock-agentcore-mcp-server/awslabs/amazon_bedrock_agentcore_mcp_server/utils/cache.py
+++ b/src/amazon-bedrock-agentcore-mcp-server/awslabs/amazon_bedrock_agentcore_mcp_server/utils/cache.py
@@ -14,6 +14,7 @@
 
 from ..config import doc_config
 from . import doc_fetcher, indexer, text_processor
+from loguru import logger
 from typing import Dict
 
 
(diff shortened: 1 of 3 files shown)


kevin-orellana:
Thanks for this, and sorry it sat. Closing because the fix landed this morning — your diagnosis of #4138 was correct.

#4140 merged as `fd2c4a50` and closed #4138. Verified on `main`:

`src/amazon-bedrock-agentcore-mcp-server/awslabs/amazon_bedrock_agentcore_mcp_server/utils/cache.py:52`

```python
except (URLError, OSError) as e:
```

Three independent fixes were open for this one crash — yours, #4140, and #4273 — because the issue sat unreviewed long enough for people to keep rediscovering it independently. That is our process failure, not yours.

Your PR was close to the one I picked. It fixed the crash and repaired the test that breaks once the docs index is empty, and it ran clean locally for me (813 passed, 0 failed). The deciding difference is that #4140 also swaps the dead `llms.txt` URL for a live one. The starter-toolkit docs site was deprecated and its index removed, which is the root cause of the 404. Catching the exception stops the crash but leaves `search_agentcore_docs` and `fetch_agentcore_doc` as permanent silent no-ops returning "no matches" forever, which is arguably worse than the loud failure because nothing tells the user or the calling agent that docs are unavailable. #4140 points at `https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/llms.txt`, which I verified returns 200 with a real index, so the docs tools work again afterward rather than just failing quietly.

Also flagging, since it affected all three PRs here: GitHub CI never actually ran on any of them. Fork PRs sit at `action_required` until a maintainer approves each workflow run individually, and nobody did. So the check status you saw was not meaningful in either direction. I verified all three locally instead.

Note the fix is on `main` but not in a published release yet, so `uvx awslabs.amazon-bedrock-agentcore-mcp-server@latest` still crashes at startup until the next release cuts.

Separately, your #4132 was a duplicate of another already-merged fix and I have closed that one too, with the reasoning there.

PR #4282, opened 2026-07-20
fix(stepfunctions-tool-mcp-server): paginate list_state_machines

### LINKED ISSUE #2050: mcp-server-aws-stepfunctions: list_state_machines returns only the first 100 results (missing pagination)
### Describe the bug

list_state_machines in the AWS Step Functions MCP server does not handle AWS pagination (nextToken).
In accounts with more than 100 state machines, the tool only returns the first page (max 100) and there is no way to retrieve the remaining state machines.

This leads to incomplete inventories and incorrect results for any workflow relying on a full list of state machines.
This silently truncates results, which is easy to miss

### Expected Behavior

When listing state machines, the server should return all state machines (or provide a supported way to page through results, e.g. exposing nextToken / maxResults).
At minimum, it should not silently truncate results at 100.

### Current Behavior

The tool returns only the first page of results (≤ 100 state machines).
Additional pages are not fetched / nextToken is ignored, so state machines beyond the first 100 are never returned.

### Reproduction Steps

	1.	Use an AWS account/region with > 100 Step Functions state machines.
	2.	Run the MCP Step Functions server.
	3.	Call the tool that lists state machines (e.g. list_state_machines / equivalent tool in the server).
	4.	Observe the tool returns only 100 items (or fewer), even though the account contains more.

Verification with AWS CLI (shows the full set; AWS CLI paginates automatically):
`aws stepfunctions list-state-machines --region <REGION> --output json | jq '.stateMachines | length'`

The MCP tool returns 100, while AWS CLI returns a larger number (e.g. 142).

### Possible Solution

Implement pagination by looping on nextToken until it is absent, concatenating results (optionally bounded by a user-provided limit).

I opened a PR with a fix: https://github.com/awslabs/mcp/pull/1923

	•	AWS ListStateMachines API supports pagination via nextToken and maxResults.
	•	This is not region- or OS-specific; it reproduces anywhere there are >100 state machines.
	•	Happy to adjust behavior if maintainers prefer exposing paging tokens instead of returning the full list by default.

### OS

macOS 15.x (Apple Silicon)

### Server

other

### Server Version

0.1.16

all

### Other information

_No response_

### Service quota

arimu1:
Fixes #2050

### Changes

`register_state_machines()` in `src/stepfunctions-tool-mcp-server/awslabs/stepfunctions_tool_mcp_server/server.py` called `sfn_client.list_state_machines()` exactly once and never followed `nextToken`. In an AWS account with more than 100 state machines, everything past the first page was silently dropped: those state machines never got registered as MCP tools, with no error surfaced anywhere.

This adds a `get_all_state_machines()` helper that uses the boto3 paginator (`sfn_client.get_paginator('list_state_machines')`) to walk every page and concatenate `stateMachines` across all of them. `register_state_machines()` now calls this helper instead of the single unpaginated call. This mirrors the pattern `lambda-tool-mcp-server` already uses for `list_functions` (`get_all_lambda_functions()`), so the fix follows the existing in-repo convention for AWS API pagination rather than introducing a new one.

Existing unit tests in `test_register_state_machines.py` were updated to mock the paginator instead of the direct `list_state_machines` call. Added `test_get_all_state_machines.py` covering single-page, multi-page, and empty results, plus a new `test_register_machines_across_multiple_pages` regression test in `test_register_state_machines.py` that mocks a two-page paginated response and asserts state machines from both pages get registered as tools.

### User experience

Before: in an account with >100 state machines, only the first 100 (by API return order) were ever exposed as MCP tools. State machines past that boundary were invisible to the agent with no warning.

After: all state machines across all pages are registered as tools, regardless of account size.

### Prior art

This exact fix was previously submitted in #1923 by @JulienJBO but received no maintainer review and was auto-closed by the stale bot in January 2026. The bug was independently reconfirmed on `main` and the same paginator approach was recommended in a comment on this issue by @edobusy (who also pointed at the `lambda-tool-mcp-server` precedent) while asking to be assigned; as of this PR that request had gone unanswered for several days and no PR had been opened against it. I'm resubmitting the fix along these same lines, crediting both of them, rather than presenting it as a new discovery.

## Checklist

- [x] Changes are documented

N

**RFC issue number**:

N/A

Checklist:

- [ ] Migration process documented
- [ ] Implement warnings (if it can live side by side)

## Acknowledgment

### DIFF
--- a/src/stepfunctions-tool-mcp-server/awslabs/stepfunctions_tool_mcp_server/server.py
+++ b/src/stepfunctions-tool-mcp-server/awslabs/stepfunctions_tool_mcp_server/server.py
@@ -332,14 +332,24 @@ def filter_state_machines_by_tag(state_machines, tag_key, tag_value):
     return tagged_state_machines
 
 
+def get_all_state_machines():
+    """Retrieve all available Step Functions state machines using pagination."""
+    paginator = sfn_client.get_paginator('list_state_machines')
+    all_state_machines = []
+
+    for page in paginator.paginate():
+        all_state_machines.extend(page.get('stateMachines', []))
+
+    return all_state_machines
+
+
 def register_state_machines():
     """Register Step Functions state machines as individual tools."""
     try:
         logger.info('Registering Step Functions state machines as individual tools...')
-        state_machines = sfn_client.list_state_machines()
 
         # Get all state machines
-        all_state_machines = state_machines['stateMachines']
+        all_state_machines = get_all_state_machines()
         logger.info(f'Total Step Functions state machines found: {len(all_state_machines)}')
 
         # First filter by state machine name if prefix or list is set
--- a/src/stepfunctions-tool-mcp-server/tests/test_get_all_state_machines.py
+++ b/src/stepfunctions-tool-mcp-server/tests/test_get_all_state_machines.py
@@ -1,0 +1,67 @@
+"""Tests for the get_all_state_machines function."""
+
+import pytest
+from unittest.mock import MagicMock, patch
+
+
+with pytest.MonkeyPatch().context() as CTX:
+    CTX.setattr('boto3.Session', MagicMock)
+    from awslabs.stepfunctions_tool_mcp_server.server import get_all_state_machines
+
+
+class TestGetAllStateMachines:
+    """Tests for the get_all_state_machines function."""
+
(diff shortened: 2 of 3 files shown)


arimu1:
Still relevant; waiting on review.

PR #4291, opened 2026-07-21
fix(aws-network-mcp-server): detect_tgw_inspection KeyError on VpcId

### LINKED ISSUE #4286: detect_tgw_inspection throws KeyError: 'VpcId' on valid Transit Gateways
## Description

`detect_tgw_inspection` fails with an unhandled `KeyError: 'VpcId'` when called against Transit Gateways that have at least one attachment type other than a plain VPC attachment (e.g. Connect, peering, or Direct Connect Gateway attachments).

## Environment

- `awslabs.aws-network-mcp-server` version: 0.0.13 (via `uvx awslabs.aws-network-mcp-server@latest`)
- MCP client: Claude Code
- Region: eu-west-1

## Steps to reproduce

1. Call `list_transit_gateways(region="eu-west-1", profile_name="<profile>")` — returns 2+ TGWs, at least one with mixed attachment types (VPC + Connect/peering).
2. Call `detect_tgw_inspection(transit_gateway_id="<tgw-id>", region="eu-west-1", profile_name="<profile>")` on that TGW.

## Actual behavior

```
Error detecting firewall attachments: 'VpcId'. REQUIRED TO REMEDIATE BEFORE CONTINUING
```

This happened consistently on 2 out of 2 TGWs tested in the account, one of which is a long-standing production TGW and one a newer TGW used for third-party VPC attachment.

## Expected behavior

The tool should either:
- Skip/ignore attachments that don't have a `VpcId` (e.g. Connect, peering) when checking for firewall-appliance-mode VPC attachments, or
- Return a clear error identifying which attachment(s) lack the expected field, rather than an unhandled `KeyError`.

## Additional context

This appears to be an unguarded dict access (e.g. `attachment['VpcId']`) somewhere in the TGW attachment-inspection logic that doesn't account for non-VPC attachment types in the TGW's attachment list.

arimu1:
## Summary

`detect_tgw_inspection` raised an unhandled `KeyError: 'VpcId'` any time the account/region had at least one AWS Network Firewall — regardless of the Transit Gateway's attachment mix.

Root cause: `nfw_client.list_firewalls()['Firewalls']` items only ever contain `FirewallName`/`FirewallArn`/`TransitGatewayAttachmentId` (see the [ListFirewalls API reference](https://docs.aws.amazon.com/network-firewall/latest/APIReference/API_ListFirewalls.html)) — `VpcId` is never present there. `VpcId` only shows up in the [DescribeFirewall](https://docs.aws.amazon.com/network-firewall/latest/APIReference/API_DescribeFirewall.html) response. The unguarded `fw['VpcId']` access at line 68 of `detect_transit_gateway_inspection.py` therefore failed on every call that returned any firewalls, independent of the TGW's attachment types — the TGW-attachment filtering code (`ResourceType`/`ResourceId`) was already correct and unaffected.

## Fix

Look up each listed firewall's `VpcId` via `describe_firewall()` (the same call already used a few lines down for TGW network-function attachments), guarding with `.get('VpcId')` so non-VPC firewalls or describe failures degrade gracefully instead of crashing the whole tool call.

Closes #4286

## Test plan

- [x] Added a regression test (`test_vpc_firewall_detection_no_vpc_id_on_list_firewalls`) that mocks `list_firewalls()` with the real API shape (no `VpcId`) and asserts the tool still correctly finds the VPC-attached firewall via `describe_firewall()`.
- [x] Updated existing fixtures/tests that previously (incorrectly) assumed `list_firewalls()` returns `VpcId`.
- [x] Verified the regression test fails with the exact reported error (`ToolError: Error detecting firewall attachments: 'VpcId'.`) against the pre-fix code, and passes after the fix.
- [x] `uv run pytest tests/` — 245 passed
- [x] `uv run ruff check` / `uv run ruff format --check` on changed files — clean

### DIFF
--- a/src/aws-network-mcp-server/CHANGELOG.md
+++ b/src/aws-network-mcp-server/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+
+- `detect_tgw_inspection` raising `KeyError: 'VpcId'` because `list_firewalls()` doesn't return `VpcId` (#4286)
+
 ### Added
 
 - Initial project setup
--- a/src/aws-network-mcp-server/awslabs/aws_network_mcp_server/tools/transit_gateway/detect_transit_gateway_inspection.py
+++ b/src/aws-network-mcp-server/awslabs/aws_network_mcp_server/tools/transit_gateway/detect_transit_gateway_inspection.py
@@ -64,8 +64,20 @@ async def detect_tgw_inspection(
         nfw_client = get_aws_client('network-firewall', region, profile_name)
 
         # Get AWS Network Firewalls (VPC-attached)
+        # list_firewalls() only returns FirewallName/FirewallArn, not VpcId, so we
+        # describe each firewall to find which ones are VPC-attached. TGW-attached
+        # firewalls (and any other non-VPC firewall) simply won't have a VpcId.
         aws_firewalls = nfw_client.list_firewalls()['Firewalls']
-        aws_firewall_vpcs = {fw['VpcId'] for fw in aws_firewalls}
+        aws_firewall_vpcs = set()
+        for fw in aws_firewalls:
+            try:
+                firewall_details = nfw_client.describe_firewall(FirewallName=fw['FirewallName'])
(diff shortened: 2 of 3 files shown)


arimu1:
Still relevant; waiting on review.

PR #4299, opened 2026-07-21
feat(postgres-mcp-server): add least-privilege startup/connect guardrail

kennthhz:
Reject over-privileged database connections by default. After a connection is established (startup probe and the connect_to_database tool), the server runs a privilege check and refuses roles that are a superuser or a member of rds_superuser.

Server (server.py):
- Add --privilege_check {enforce,warn,off} (default: enforce) and a matching privilege_check_policy global.
    - enforce: reject over-privileged connections; fail closed when the privilege cannot be verified (query error / missing columns).
    - warn: log a warning but allow the connection.
    - off: connectivity check only (SELECT 1).
- Add validate_connection(), ConnectionValidationError, and the POSTGRES_PRIVILEGE_QUERY (rolsuper + rds_superuser membership, tolerant of environments without the role). The check is connection-agnostic, so it works for both the psycopg pool and the RDS Data API paths.
- Wire it into connect_to_database (remove the connection from the map and fail on rejection) and into the startup path (exit 1 on rejection).

Tests:
- tests/test_privilege_guardrail.py: unit coverage for enforce/warn/off across superuser, rds_superuser, clean, and unverifiable cases for both connection types.
- tests/test_server_error_handling.py: cover the new startup/connect wiring.

E2E (tests/e2e):
- Provision a dedicated non-superuser role + Secrets Manager secret on the fly for the express and serverless clusters, pin its ARN, and run the functional and security suites as that role under 'enforce' (production-like), while keeping the master (rds_superuser) path to assert enforce rejects it.
- Add a privilege-enforcement suite asserting enforce rejects the superuser, warn/off allow it, and enforce allows the least-privilege role.
- Decouple the provisioning master-connection method from the lp IAM setup so serverless can connect-as-master over RDS_API yet still authorize the lp role for IAM DB auth only when --test-non-express-cluster exercises it; track iam_policy_created so deprovision cleans up exactly what was created.

Docs:
- README: document the guardrail, the --privilege_check modes, startup-vs-tool behavior, fail-closed semantics, and rds_superuser tolerance.
- tests/e2e/README: document the provisioned least-privilege prerequisites.

Verified: 704 unit tests pass; ruff clean. Live e2e in us-west-2 passed for both express (--test-serverless-cluster off) and serverless via RDS_API (engine 16.9): 164 passed, 0 failed, 0 skipped, including the serverless enforce:least_priv_allowed case. Clusters and provisioned artifacts torn down with no leaks.

Fixes

### Changes

### User experience

## Checklist

* [x] Changes are documented

**RFC issue number**:

Checklist:

## Acknowledgment


thephantomthief (dismissed):

src/postgres-mcp-server/awslabs/postgres_mcp_server/server.py
```diff
@@ -359,6 +473,18 @@ async def connect_to_database(
...
                 raise
 
+        # Post-connect validation: connectivity + least-privilege guardrail.
+        # A superuser / rds_superuser connection is rejected under the default
+        # 'enforce' policy. Remove the connection from the map on failure so a
+        # later attempt starts fresh.
+        try:
+            await validate_connection(db_connection, privilege_check_policy)
+        except Exception:
+            db_connection_map.remove(
```
scottschreckengaust: **Guardrail bypass: rejected over-privileged connection is not evicted from the map.**

This cleanup `remove()` uses the *original tool-supplied* `db_endpoint` and `port`, but `internal_create_connection` stored the connection under a **different key**:

- `internal_create_connection` calls `db_connection_map.set(connection_method, cluster_identifier, db_endpoint, database, db_connection)` (server.py:1013) — 5 positional args, so `port` **defaults to 5432**, and `db_endpoint` is the **AWS-resolved** host (the local variable is overwritten from cluster/instance metadata at lines 860/892/909 and never returned to the caller).
- Here `remove(connection_method, cluster_identifier, db_endpoint, database, port)` passes the caller’s original values.

When they differ, `remove()` hits `KeyError` and **silently no-ops** (db_connection_map.py:101), leaving the connection cached.

Repro: connect via an APG cluster with `db_endpoint=''` (valid for RDS_API — the docstring says endpoint is ignored) as a superuser role under the default `enforce`. `set()` stores the key under the resolved writer endpoint + port 5432; `validate_connection` raises; this line calls `remove(..., '', ..., port)` → key miss → the **superuser connection stays in the map**. `get_database_connection_info` then exposes the resolved key to the LLM, and `run_query` (which looks up with port default 5432, server.py:289) will happily execute against the very superuser connection the guardrail rejected.

A custom `port != 5432` triggers the same miss even when the endpoint matches.

Fix: evict by the key that was actually stored (resolved endpoint, port 5432) — e.g. have `internal_create_connection` return the stored key, or add a remove-by-connection-object path — rather than reconstructing the key from caller input.

src/postgres-mcp-server/awslabs/postgres_mcp_server/server.py
```diff
@@ -127,6 +138,109 @@ def parse_execute_response(response: dict) -> list[dict]:
...
+        response = await db_connection.execute_query(POSTGRES_PRIVILEGE_QUERY)
+        rows = parse_execute_response(response)
+    except Exception as e:
+        # The check (which also proves connectivity) failed. Fail closed
+        # under enforce; allow with a warning under warn.
+        message = f'Could not verify connection role privileges: {type(e).__name__}: {e}'
+        if policy == PRIVILEGE_CHECK_ENFORCE:
+            raise ConnectionValidationError(f'{message}. Rejecting connection (fail-closed).')
+        logger.warning(f'{message}. Allowing connection (privilege_check=warn).')
+        return
```
scottschreckengaust: **`warn` mode silently drops the startup connectivity guarantee.**

The old startup path ran `run_query('SELECT 1', ...)` and `sys.exit(1)` on *any* error result, so an unreachable host / bad credentials aborted startup. The new `validate_connection` under `policy == 'warn'` catches every `execute_query` failure here — including genuine connectivity/auth failures, not just privilege-check ambiguity — logs a warning, and returns normally. `main()` then logs "Successfully validated database connection" and proceeds to `mcp.run()`.

Result: starting with `--privilege_check=warn` against an unreachable DB or wrong credentials no longer fails fast; every subsequent query fails at runtime instead. (`off` still propagates `SELECT 1` failures and exits; only `warn` loses the guarantee.)

Consider distinguishing "privilege query failed but connectivity is fine" from "connection is dead" — e.g. still prove connectivity (`SELECT 1`) under `warn` before allowing, or only downgrade privilege-*ambiguity* to a warning while still failing on a hard connection error.

src/postgres-mcp-server/awslabs/postgres_mcp_server/server.py
scottschreckengaust: (Related, same root cause — the pre-existing `remove()` at line 471, in the pool-init-failure `except` just above this block.) It has the identical key mismatch: it reconstructs the key from the caller’s original `db_endpoint`/`port` while `set()` stored the AWS-resolved endpoint with port 5432. For an APG cluster connected with `db_endpoint=''` or a non-5432 port, that `remove()` also misses and leaves the broken pool cached, so the "next connect attempt creates a fresh one" comment there no longer holds. Both cleanup paths should evict by the actually-stored key.

scottschreckengaust (changes requested):
## Review Summary — least-privilege startup/connect guardrail

Solid, well-documented security feature with genuinely good test coverage (the `off`/`warn`/`enforce` matrix, fail-closed-on-missing-columns, and the startup-vs-tool wiring are all exercised). Two correctness issues stand out, one of which undermines the guardrail's own purpose. Inline comments posted.

### Must fix (blocking)

1. **Guardrail bypass — rejected over-privileged connection is not evicted** (`server.py:483`). The rejection-cleanup `remove()` rebuilds the map key from the *caller-supplied* `db_endpoint`/`port`, but `internal_create_connection` stored the connection under the *AWS-resolved* endpoint with `port` defaulted to 5432 (`set()` omits `port`). When they differ (e.g. RDS_API with `db_endpoint=''`, or any non-5432 port), `remove()` hits `KeyError` and silently no-ops, so the superuser connection the guardrail just rejected **stays cached** and remains reachable via `run_query`. This defeats the feature under `enforce`. Fix by evicting the actually-stored key (return it from `internal_create_connection`, or remove-by-object).

2. **Same key mismatch on the pre-existing pool-init-failure cleanup** (`server.py:471`) — a broken pool is left cached under the same conditions. Fix both paths together.

### Should fix

3. **`warn` mode silently drops the startup connectivity guarantee** (`server.py:~207`). The old startup path failed fast on any `SELECT 1` error; `warn` now swallows genuine connectivity/auth failures and proceeds to `mcp.run()`. Prove connectivity under `warn` before allowing.

### Suggestions (tests — not blocking)

- `test_rds_superuser_rejected` asserts `'rds_superuser' in str(exc.value)`, but the rejection message statically contains that phrase for *any* over-privileged rejection, so it can't actually distinguish rds_superuser detection. The `enforce`/`least_privilege` tests that assert `conn.queries == [POSTGRES_PRIVILEGE_QUERY]` are the strong ones; the `warn` tests don't assert the privilege query was issued.
- The `remove()` key-mismatch bug above is not caught by the unit test `test_connect_to_database_rejects_superuser_and_removes_connection` because it mocks `db_connection_map` and asserts `remove` was called with the tool-supplied args — i.e. it locks in the buggy key rather than verifying eviction actually happened. Consider a test with a real map that asserts the connection is gone after rejection.
- E2E `provision_least_privilege_access` (`e2e_integration_test.py:368`) does `get(..., port=port)` after a `connect_to_database` that stored under port 5432 — this masks/degrades on any non-default `--port`. Since e2e defaults to 5432 it currently works, but it shares the same latent assumption as bug #1.

**Recommendation: Request changes** — the guardrail bypass (#1) means the default `enforce` policy does not reliably prevent operating as superuser, which is the entire point of the PR.

src/postgres-mcp-server/awslabs/postgres_mcp_server/server.py
kennthhz: fixed

src/postgres-mcp-server/awslabs/postgres_mcp_server/server.py
kennthhz: fixed

src/postgres-mcp-server/awslabs/postgres_mcp_server/server.py
kennthhz: fixed

thephantomthief (dismissed):

src/postgres-mcp-server/awslabs/postgres_mcp_server/server.py
```diff
@@ -353,12 +477,31 @@ async def connect_to_database(
...
+        # 'enforce' policy. Remove the connection from the map on failure so it
+        # cannot be reached later via run_query. Evict by object identity:
+        # internal_create_connection stored the connection under the
+        # AWS-resolved endpoint/port, which may differ from the caller-supplied
+        # db_endpoint/port here (empty db_endpoint, non-5432 port, host casing).
+        # A key-based remove() rebuilt from the caller args could miss the
+        # actual key, leave the rejected connection cached, and defeat the
+        # guardrail. See the connection-map key-normalization follow-up.
+        try:
+            await validate_connection(db_connection, privilege_check_policy)
```
scottschreckengaust, line 610: **Re-validating a cached connection can evict a healthy one (transient-failure hazard).**

`internal_create_connection` returns a cached connection unchanged via the `existing_conn` early-return (server.py:827-842), but `connect_to_database` still runs `validate_connection` here on *every* call, and the `except` below evicts on *any* exception.

`validate_connection` under `enforce` re-raises **any** probe failure — including transient ones (RDS Data API `ThrottlingException`, Aurora Serverless resume, a momentarily saturated psycopg pool → `PoolTimeout`). There's no retry. So a hiccup during a redundant re-connect tears a still-good shared connection out of the map, and subsequent `run_query` returns `No database connection available`.

Consider validating **only newly-created** connections (skip the probe when `internal_create_connection` returned the cached `existing_conn`), or not evicting a cached connection on a transient (non-`ConnectionValidationError`) error.

src/postgres-mcp-server/awslabs/postgres_mcp_server/server.py
```diff
@@ -353,12 +477,31 @@ async def connect_to_database(
...
+        except Exception:
+            db_connection_map.remove_connection(db_connection)
```
scottschreckengaust, line 612: **Connection leak: eviction never closes the connection.**

By the time we reach here for a `PsycopgPoolConnection`, `initialize_pool()` (line 477) has already opened a live pool with `min_size` backend sessions. `remove_connection` only does `del self.map[key]` — it never awaits `conn.close()` — so the pool and its open Postgres sessions leak. Under the default `enforce`, every rejected over-privileged connect leaks one pool.

`close_all()` closes connections explicitly, confirming `close()` is the required teardown. Suggest `await db_connection.close()` alongside the `remove_connection` call here (and in the pool-init `except` at line 486, though psycopg self-closes there). RDS_API is negligible (boto3 client, GC-reclaimed).

src/postgres-mcp-server/awslabs/postgres_mcp_server/connection/db_connection_map.py
```diff
@@ -103,6 +103,32 @@ def remove(
...
+        by identity sidesteps that entirely. See the tracked follow-up on
+        connection-map key normalization for the broader fix.
+
+        Returns:
+            True if a matching connection was found and removed, else False.
+        """
+        with self._lock:
+            for key, existing in list(self.map.items()):
+                if existing is conn:
+                    del self.map[key]
```
scottschreckengaust, line 128: `remove_connection` deletes the map entry but doesn't close the connection, unlike `close_all()` which awaits `conn.close()`. Callers in `connect_to_database` use this to evict an already-opened pool on validation failure, leaking the pool's backend sessions (see comment on server.py:502). Consider having callers `await conn.close()`, or documenting that this method intentionally does not close and the caller must.

src/postgres-mcp-server/awslabs/postgres_mcp_server/server.py
```diff
@@ -127,6 +138,119 @@ def parse_execute_response(response: dict) -> list[dict]:
...
+        message = 'Could not determine connection role privileges (unexpected result shape).'
+        if policy == PRIVILEGE_CHECK_ENFORCE:
+            raise ConnectionValidationError(f'{message} Rejecting connection (fail-closed).')
+        logger.warning(f'{message} Allowing connection (privilege_check=warn).')
+        return
+
+    is_superuser = bool(rows[0].get('is_superuser'))
+    is_rds_superuser = bool(rows[0].get('is_rds_superuser'))
+
+    if is_superuser or is_rds_superuser:
```
scottschreckengaust: **Fail-open on an unrecognized policy value.**

If `privilege_check_policy` is ever anything other than exactly `'off'`/`'enforce'` (a mis-cased or typo'd string set directly on the global in a test or the e2e harness, bypassing argparse's `choices`), execution skips the `OFF` branch, runs the probe, and then — because `policy != PRIVILEGE_CHECK_ENFORCE` — takes the `warn` branch here and **allows** a superuser connection with only a warning.

For a security guardrail, an unknown policy should fail closed. Consider an explicit `else: raise ConnectionValidationError(...)` (or validating the policy string once at entry).

src/postgres-mcp-server/awslabs/postgres_mcp_server/server.py
```diff
@@ -1203,31 +1366,25 @@ def main():
...
-                        ConnectionMethod[args.connection_method],
-                        cluster_identifier,
-                        args.db_endpoint,
-                        args.database,
+                try:
+                    asyncio.run(validate_connection(db_connection, privilege_check_policy))
+                    logger.success('Successfully validated database connection to Postgres')
+                except ConnectionValidationError as e:
+                    logger.error(
+                        f'Refusing to start: {e} '
```
scottschreckengaust: **Misleading remediation for connectivity failures.**

`validate_connection` wraps a pure connectivity/auth failure (DB unreachable, bad password) as `ConnectionValidationError` under `enforce`, so it lands in this branch and tells the operator to "set --privilege_check to warn/off to override." But `warn`/`off` also require a working connection (`off` runs `SELECT 1`; `warn` re-raises connectivity errors), so that advice can't fix an unreachable DB — it only applies to an actual over-privileged role.

Consider only appending the warn/off hint when the error is a genuine privilege violation, not a wrapped connectivity error (e.g. distinguish the two at the raise site in `validate_connection`).

src/postgres-mcp-server/awslabs/postgres_mcp_server/server.py
```diff
@@ -127,6 +138,119 @@ def parse_execute_response(response: dict) -> list[dict]:
...
+
+    Args:
+        db_connection: the freshly established data-plane connection.
+        policy: one of 'enforce', 'warn', 'off'.
+
+    Raises:
+        ConnectionValidationError: under 'enforce', on a privilege violation
+            or when privileges cannot be verified.
+    """
+    if policy == PRIVILEGE_CHECK_OFF:
```
scottschreckengaust, line 210: Minor (reuse): this `off`-path connectivity probe re-implements `AbstractDBConnection.check_connection_health()`, which is defined in both connection classes as exactly `execute_query('SELECT 1')`. Calling the existing method keeps the connectivity contract in one place.

src/postgres-mcp-server/awslabs/postgres_mcp_server/server.py
```diff
@@ -353,12 +477,31 @@ async def connect_to_database(
                 await db_connection.initialize_pool()
```
scottschreckengaust: Minor (simplification): these two back-to-back `try/except: db_connection_map.remove_connection(db_connection); raise` blocks — for `initialize_pool()` and for `validate_connection()` — do the same cleanup and carry near-identical multi-line rationale comments. They can be collapsed into a single `try` covering both calls with the rationale stated once.

src/postgres-mcp-server/tests/e2e/e2e_integration_test.py
```diff
@@ -281,6 +281,208 @@ def configure_server_secret_for_cluster(cluster_identifier: str, region: str) ->
...
+        # Drop the master provisioning connection so suites reconnect as lp.
+        server.db_connection_map.remove(
+            connection_method, cluster_identifier, valid_endpoint, database, port
+        )
+        server.readonly_query = saved_readonly
+
+    # 4. Store the role's credentials in a tagged secret.
+    sm = boto3.client('secretsmanager', region_name=region)
+    secret_name = f'mcp-e2e-lp-{cluster_identifier}-{datetime.now().strftime("%Y%m%d%H%M%S")}'
+    created = sm.create_secret(
```
scottschreckengaust: **IAM policy leak on the `create_secret` failure window.**

The IAM policy is created/attached at line 393 (inside the `try`), but `sm.create_secret` runs here *after* the `try/finally`. If `create_secret` raises (missing `secretsmanager:CreateSecret`, throttling, tag policy), `provision_least_privilege_access` raises before returning `lp_info`, so it never lands in `lp_info_by_kind` and `deprovision_least_privilege_access` never deletes the `AuroraIAMAuth-<lp_role>` policy it just created. Each failed run then accumulates an orphaned IAM policy attached to the test principal.

Consider moving `create_secret` inside a try that cleans up the IAM policy on failure, or tracking the created policy for teardown before attempting `create_secret`.

scottschreckengaust (commented):
## Code review — least-privilege guardrail

Thanks for this — the guardrail is a real security improvement, the fail-closed-under-`enforce` semantics are correct, the `EXISTS(...)` tolerance for clusters without `rds_superuser` is a nice touch, and the unit/e2e coverage is genuinely thorough (I verified the enforce-rejection tests actually raise `ConnectionValidationError` rather than passing vacuously).

I'm marking this **Comment** rather than Approve because of two correctness issues that bite the long-lived server path, plus a resource-leak in the e2e provisioner. None are blockers if you'd rather fast-follow, but they're worth a look before merge.

### Verified findings (details inline)

**Correctness**
1. **Re-validating a cached connection can evict a healthy, in-use one.** `connect_to_database` runs `validate_connection` unconditionally, even when `internal_create_connection` returned a *cached* connection. A transient probe failure (RDS Data API throttle, Aurora Serverless resume, momentarily saturated psycopg pool) then evicts a shared connection that was fine, and later `run_query` calls fail with "No database connection available." (server.py:500)
2. **`remove_connection` leaks the connection — it never `close()`s.** On an `enforce` rejection the psycopg pool was already opened by `initialize_pool()`, so evicting by map-delete alone orphans a live `AsyncConnectionPool` with open backend sessions. Repeated rejections leak one pool each. `close_all()` shows `close()` is the required teardown. (db_connection_map.py:106, server.py:502)
3. **e2e: IAM policy leaks if `create_secret` fails.** The IAM policy is created inside the `try`, but `sm.create_secret` runs *after* the `try/finally`; if it raises, the function never returns `lp_info`, so `deprovision_least_privilege_access` never deletes the just-created `AuroraIAMAuth-<lp_role>` policy. (e2e_integration_test.py:408)

**Robustness / UX**
4. **Unknown policy value fails OPEN.** Any `privilege_check_policy` that isn't exactly `'off'`/`'enforce'` (e.g. a mis-cased string assigned directly in a test/e2e, bypassing argparse) runs the probe but takes the `warn` branch on a violation → allows a superuser silently. A defensive `else: raise` would fail closed. (server.py:233)
5. **Misleading operator guidance on connectivity failure.** Under `enforce`, a pure connectivity/auth failure is wrapped as `ConnectionValidationError`, so startup prints "set --privilege_check to warn/off to override" — but `warn`/`off` also require a working connection, so that advice can't help. (server.py:1378)

**Cleanup (non-blocking)**
6. Redundant privilege re-probe on every `connect_to_database` (incl. cached hits) — extra round-trip / billable Data API call per connect. (server.py:500)
7. The two back-to-back `try/except: remove_connection; raise` blocks + near-identical rationale comments can collapse into one. (server.py:473)
8. The `off` path hand-rolls `execute_query('SELECT 1')` — `AbstractDBConnection.check_connection_health()` already is exactly that probe. (server.py:196)

### Notes / not-bugs (for the record)
- The boolean columns come back as real Python `bool` on **both** paths (RDS `booleanValue`; psycopg checks `isinstance(bool)` before `int`), so `bool('f')`-style false-negatives can't occur — the guardrail correctly detects superusers. ✅
- Column-name casing (`is_superuser`/`is_rds_superuser`) matches on both connection types. ✅
- `remove_connection`'s first-match-return is fine: no code stores one connection object under two keys. ✅
- The connection-map key divergence (`get()` uses caller endpoint, `set()` uses AWS-resolved endpoint) that the PR documents is **pre-existing on `main`**, not introduced here — reasonable to defer to the tracked follow-up, though note it also affects `run_query`'s lookup, not just eviction.
- `current_user` vs `session_user`: the guardrail keys on `current_user`, which a role-lowering connect option could dodge; low severity given it's explicitly a guardrail and read-only mode blocks `RESET ROLE`. Worth considering `session_user` for hardening.

src/postgres-mcp-server/awslabs/postgres_mcp_server/server.py
kennthhz: fixed

src/postgres-mcp-server/awslabs/postgres_mcp_server/connection/db_connection_map.py
kennthhz: fixed

src/postgres-mcp-server/awslabs/postgres_mcp_server/server.py
kennthhz: fixed

src/postgres-mcp-server/awslabs/postgres_mcp_server/server.py
kennthhz: fixed

src/postgres-mcp-server/awslabs/postgres_mcp_server/server.py
kennthhz: fixed

src/postgres-mcp-server/awslabs/postgres_mcp_server/server.py
kennthhz: fixed

src/postgres-mcp-server/awslabs/postgres_mcp_server/server.py
kennthhz: fixed

src/postgres-mcp-server/tests/e2e/e2e_integration_test.py
kennthhz: fixed

src/postgres-mcp-server/awslabs/postgres_mcp_server/server.py
```diff
@@ -336,6 +475,21 @@ async def connect_to_database(
             db_endpoint must be set
     """
     try:
+        # Detect a cache hit using the SAME key internal_create_connection uses
+        # for its dedup early-return. internal_create_connection is synchronous,
+        # so nothing interleaves between this check and the call. A cached
+        # connection was already validated when it was first created; re-running
```
scottschreckengaust: **Finding #2 (correctness of the assumption).** This comment states a cached connection "was already validated when it was first created," and the `was_cached` early-return at line 503 skips `validate_connection` on that basis. But connections seeded by `create_cluster` / `create_cluster_worker` (and the express path) are cached via `internal_create_connection` **without** ever being validated. So for those keys the assumption is false: a later `connect_to_database` on the same key sees `was_cached=True` and returns without validating, meaning an explicit reconnect can never surface or evict the over-privileged connection — it stays cached and usable. This is the same root cause as the blocking finding #1; fixing the cluster paths to validate-on-create makes this comment true again.

src/postgres-mcp-server/awslabs/postgres_mcp_server/server.py
```diff
@@ -127,6 +138,134 @@ def parse_execute_response(response: dict) -> list[dict]:
...
+
+    if is_superuser or is_rds_superuser:
+        flags = []
+        if is_superuser:
+            flags.append('a superuser')
+        if is_rds_superuser:
+            flags.append('a member of rds_superuser')
+        message = (
+            f'The MCP server is connecting as an over-privileged Postgres role '
+            f'({" and ".join(flags)}). A superuser / rds_superuser role bypasses '
```
scottschreckengaust: **Finding #3 (message vs. what's actually checked).** This message justifies rejection with "bypasses row-level security," but `POSTGRES_PRIVILEGE_QUERY` (line 155) only tests `rolsuper` and `rds_superuser` membership. A non-superuser role carrying `rolbypassrls` directly — or `CREATEROLE` / `rds_replication` — passes the guardrail while still bypassing RLS, exactly the outcome the message claims to prevent. Consider either (a) adding `rolbypassrls` (and possibly `rolcreaterole`) to the query and flags so the check matches the stated rationale, or (b) softening the message to describe only what's enforced (superuser / rds_superuser).

src/postgres-mcp-server/tests/e2e/e2e_integration_test.py
```diff
@@ -1830,6 +2051,139 @@ async def _run(sql):
...
+    finally:
+        server.privilege_check_policy = saved_policy
+        server.configured_secret_arns.clear()
+        server.configured_secret_arns.update(saved_secret_arns)
+        try:
+            server.db_connection_map.remove(
+                connection_method, cluster_identifier, valid_endpoint, test_database, port
+            )
+        except Exception as e:
+            logger.warning('Non-fatal cleanup failure removing test DB connection: %s', e)
```
scottschreckengaust: **Finding #4 (loguru formatting).** loguru formats with `str.format()` and brace placeholders, not printf `%s`. With no `{}` in the message string, the positional `e` argument is dropped and the literal `%s` is logged — so a cleanup failure prints `...removing test DB connection: %s` with the real exception silently discarded. Use an f-string: `logger.warning(f'Non-fatal cleanup failure removing test DB connection: {e}')`. Note this exact line is duplicated at line 2049 — both need the fix.

scottschreckengaust (changes requested):
## Automated code review (xhigh)

The least-privilege guardrail is well-designed: `validate_connection`'s fail-closed logic (enforce + any unrecognized policy reject; only explicit `warn` softens), the connectivity-vs-privilege distinction, and the evict-by-object-identity fix in `connect_to_database` are all well-reasoned and well-tested. The findings below are verified against the code at `acd8e17`.

### 🟥 Blocking — finding #1: `create_cluster` / `create_cluster_worker` bypass the guardrail

`create_cluster` (express path, `server.py:637`) and `create_cluster_worker` (`server.py:762`) both call `internal_create_connection(...)` to cache a connection to the freshly-created cluster as the **master / `rds_superuser`** role, but neither calls `validate_connection`. `internal_create_connection` itself only caches (it returns the connection at `server.py:855-870` with no privilege probe).

Under the default `enforce` policy, the resulting over-privileged connection lands in `db_connection_map` and is immediately reachable by `run_query` (which only does `db_connection_map.get`). So the PR's headline guarantee — *"reject over-privileged connections by default"* — has a real hole reachable through a shipped MCP tool. `connect_to_database` is the only entry point that runs the guardrail; the cluster-creation paths and the express path do not.

**Suggested direction:** run `validate_connection` (with the same evict-and-close-on-failure cleanup used in `connect_to_database:519-531`) after the `internal_create_connection` calls in both cluster paths — or centralize validation inside `internal_create_connection` for the not-cached branch so every caller inherits it.

---

### Also worth addressing (see inline comments)

- **#2** — the `was_cached` early-return comment claims cached connections were "already validated when first created," which is false for the cluster-seeded path above.
- **#3** — the privilege query checks only `rolsuper` / `rds_superuser` membership, but the rejection message justifies itself with "bypasses row-level security"; a role with `BYPASSRLS` directly (not superuser) passes.
- **#4** — loguru `%s` printf-style formatting drops the exception (loguru uses `str.format`/braces).

### Lower-severity (non-blocking, not inlined)

- **#5** — startup `asyncio.run(validate_connection(...))` (`server.py:1402`) opens the psycopg pool in a throwaway loop later reused by `mcp.run()`; latent "Future attached to a different loop" risk on the first `run_query`. Not covered by unit tests (they mock pool init).
- **#6** — two eviction tests mutate the real module-global `db_connection_map` and rely on `finally` cleanup; a fixture-scoped map would be safer against future parallelization / early-raise state leakage.
- **#7** — `connect_to_database` does a `db_connection_map.get` for `was_cached` and `internal_create_connection` immediately does an identical locked lookup; minor redundant work per connect.

---

src/postgres-mcp-server/awslabs/postgres_mcp_server/server.py
kennthhz: fixed the comment

src/postgres-mcp-server/awslabs/postgres_mcp_server/server.py
kennthhz: Fixed with suggestion (a)

src/postgres-mcp-server/tests/e2e/e2e_integration_test.py
kennthhz: Fixed

src/postgres-mcp-server/awslabs/postgres_mcp_server/server.py
```diff
@@ -346,18 +535,35 @@ async def connect_to_database(
             database=database,
         )
 
-        # Eagerly initialize the connection pool so it's ready for queries
-        # and created_time is set at connect time, not at first query time
-        if isinstance(db_connection, PsycopgPoolConnection):
-            try:
+        if was_cached:
```
scottschreckengaust, line 589: **Enforce-mode bypass (blocking): `create_cluster` cached connection is reachable via the `was_cached` fast-return without validation.**

`create_cluster` / `create_cluster_worker` seed the map with the freshly-created cluster's `rds_superuser` master connection but never call `validate_connection`. Under `--privilege_check=enforce` (the recommended production mode), the natural next step after `create_cluster` is a `connect_to_database` to the exact endpoint/database the tool result just handed the LLM. That call hits `was_cached=True` here and returns success immediately — skipping `validate_connection` — so every subsequent `run_query` executes as `rds_superuser`, silently defeating the guardrail in precisely the create-then-use flow it's meant to protect.

The comment above (lines 511-520) documents this as the "known create_cluster bootstrap gap tracked as a follow-up," but under `enforce` it's a real bypass of the feature's primary purpose, not just a moot `warn`-mode detail. Consider validating-on-create with an explicit bootstrap exemption (e.g. mark bootstrap-seeded entries and reject/re-validate them on first `connect_to_database`) before merge, or at minimum gate the create-then-use path so `enforce` can't be silently defeated.

src/postgres-mcp-server/awslabs/postgres_mcp_server/server.py
```diff
@@ -346,18 +535,35 @@ async def connect_to_database(
...
+        #   - evict so a rejected/broken connection can't be reached later via
+        #     run_query (fail-closed: a fresh connection we could not validate
+        #     must not remain cached);
+        #   - close so we don't orphan an already-opened AsyncConnectionPool
+        #     with live backend sessions.
+        # Evict by object identity because the map key uses the AWS-resolved
+        # endpoint/port, which may differ from the caller-supplied args here
+        # (empty db_endpoint, non-5432 port, host casing); a key-based remove()
+        # rebuilt from caller args could miss and leave the connection cached.
+        try:
```
scottschreckengaust, line 607: **Concurrency TOCTOU (blocking, lower likelihood): the `was_cached` check races with the not-yet-awaited validation below.**

`internal_create_connection` stores the connection in the map synchronously, but validation happens at the `await validate_connection(...)` a few lines down. If two `connect_to_database` coroutines run for the same `(method, cluster, endpoint, database, port)`: call #1 takes the fresh path and stores the entry, then awaits `initialize_pool()` / `validate_connection()` here. At that await the event loop can switch to call #2, whose `was_cached` check now finds #1's entry and returns success at line 538 — before #1's validation has run or rejected.

Under `enforce`, if #1 is about to reject an over-privileged role, #2 has already reported success to its caller; #1 then evicts the connection, so #2's later `run_query` fails confusingly. The safety comment above reasons only about the synchronous gap (`internal_create_connection` is sync) and never about this cross-coroutine window opened by the await. A per-key async lock around the create+validate sequence would close it.

src/postgres-mcp-server/awslabs/postgres_mcp_server/connection/db_connection_map.py
```diff
@@ -103,17 +103,53 @@ def remove(
                     f'Try to remove a non-existing connection. {method} {cluster_identifier} {db_endpoint} {database} {port}'
                 )
 
+    def remove_connection(self, conn: AbstractDBConnection) -> bool:
```
scottschreckengaust, line 106: **Cleanup (non-blocking): identity-based `remove_connection()` is a bandaid over a real key divergence that leaves `get()`/`remove()` broken elsewhere.**

As this docstring itself documents: `set()` never receives `port`, so every connection is stored under the 5432 default, while `get()`/`remove()` are called with the caller-supplied endpoint/port. This new identity-based method sidesteps the mismatch for the `connect_to_database` eviction path, but the underlying divergence remains — any *key-based* `remove()` caller passing a non-5432 port or an endpoint that differs from the AWS-resolved key will silently no-op (info-logged only) and leak a live cached backend session. The e2e provisioning teardown is one such key-based caller.

Preferred fix is to normalize the key: pass `port` into `set()` so the stored key matches what `get()`/`remove()` reconstruct, rather than keeping a known-broken key-based `remove()` callable alongside the identity workaround. Not blocking, but worth resolving so the map has one correct keying scheme.

src/postgres-mcp-server/awslabs/postgres_mcp_server/server.py
```diff
@@ -1085,6 +1299,18 @@ def main():
...
     parser.add_argument('--db_endpoint', help='Instance endpoint address')
     parser.add_argument('--region', help='AWS region')
+    parser.add_argument(
+        '--privilege_check',
+        choices=[PRIVILEGE_CHECK_ENFORCE, PRIVILEGE_CHECK_WARN, PRIVILEGE_CHECK_OFF],
+        default=PRIVILEGE_CHECK_WARN,
+        help=(
+            'Least-privilege guardrail applied when a database connection is established. '
+            "'warn' (default) logs a warning but allows a connection whose Postgres role is a "
+            "superuser or a member of rds_superuser; 'enforce' rejects it (recommended for "
```
scottschreckengaust, line 1362: **Docs (non-blocking): help text omits `BYPASSRLS`, which `enforce` also rejects.**

`validate_connection` treats a role carrying the `BYPASSRLS` attribute as over-privileged and rejects it under `enforce` (see the `is_bypassrls` handling above). But this `--privilege_check` help text — and the `privilege_check_policy` module docstring — list only "superuser or a member of rds_superuser." An operator who reads the CLI help and grants their dedicated role `BYPASSRLS` (a plausible way to let the MCP see all rows without being a superuser) will hit an unexpected hard rejection under `enforce` with no forewarning. The README and the `ConnectionValidationError` docstring already mention `BYPASSRLS`, so this is just an internal inconsistency — please add `BYPASSRLS` to the help text and module docstring for a consistent documented contract.

scottschreckengaust (dismissed):
Thanks for this — the validation core is well-built: `validate_connection` is fail-closed (any policy that isn't literally `warn` takes the reject branch, and connectivity errors are never swallowed under `warn`), it's connection-agnostic across the psycopg and Data API paths, and it's thoroughly unit-tested (746 tests pass). My concerns are not in the validation logic but in the **caching lifecycle around it**: under `enforce` — the mode this feature exists to support — there are paths where an over-privileged connection becomes usable without the guardrail firing. Details inline.

Blocking (enforce-mode bypasses):
1. `create_cluster` bootstrap fast-return (server.py ~538) — most impactful, realistically reachable in a create-then-use flow.
2. Concurrent-connect await-window race (server.py ~554) — lower likelihood.

Non-blocking (cleanup + docs), noted inline:
3. `remove_connection()` by identity papers over the underlying map-key divergence (db_connection_map.py).
4. `--privilege_check` help text / module docstring omit `BYPASSRLS`, which `enforce` also rejects.

One more, not a diff line: the PR **description** says the default privilege mode is `enforce`, but the code and README both default to `warn` (`PRIVILEGE_CHECK_WARN`, server.py:83/1305). Worth correcting the description for accuracy.

src/postgres-mcp-server/awslabs/postgres_mcp_server/server.py
kennthhz: Thanks — agreed the mechanics are exactly as you describe, and good catch that it reaches beyond the was_cached branch: a bootstrap-seeded connection is validated by nothing, so run_query on that key hits it directly too. I want to argue this is a real, tracked limitation rather than a merge blocker, on four points.

1. It cannot affect anyone under the shipped default. The default is warn (PRIVILEGE_CHECK_WARN, server.py:83/1305), and this PR intentionally does not turn enforce on for anyone. Every existing user and every fresh install gets warn, where this path is explicitly a no-op. The gap is reachable only for an operator who has opted into enforce and then runs a create-then-use flow. So merging this changes the security posture of exactly zero deployments in the direction of this gap.

2. The bootstrap connection is over-privileged by construction, not by guardrail failure. A freshly created cluster has only the rds_superuser master and no least-privilege role yet. There is nothing else to connect as. If we validated at create time under enforce, create_cluster would reject its own bootstrap connection and the tool would be non-functional — you could never create-then-use at all. The "rds_superuser connection right after create" isn't the guardrail being defeated; it's the only credential that exists at that moment. That's precisely why the exemption is there.

3. The blast radius is bounded to a resource the caller just provisioned with credentials it was already granted. This isn't arbitrary lateral access to existing clusters. To reach the gap, the caller must be able to invoke create_cluster, which already requires rds:CreateDBCluster and access to the master secret. An actor who can provision a cluster and read its master credentials has, by configuration, already been trusted with master-level privilege on brand-new resources. Reusing that same bootstrap connection for immediate queries on that same just-created cluster does not grant access the caller didn't already have. The guardrail's actual purpose — refusing to operate as an over-privileged role against existing clusters in the steady-state query path — is fully intact.

4. Doing it right is a lifecycle change that shouldn't be rushed into this PR. The correct fix (mark bootstrap-seeded entries as exempt-but-unvalidated, then validate-and-reject on first real use — covering run_query, not just reconnect) lives in the connection-map identity/keying redesign we're already tracking. Bundling that into this PR means touching the caching lifecycle right next to the validation core that's currently green across 746 tests, trading a well-tested, warn-default improvement for regression risk in the hot path. I'd rather ship the validated core now and land the bootstrap fix as its own reviewed unit.

Concretely, I propose: keep the in-code comment documenting the gap, add a one-line note to the README enforce section stating that create-then-use bypasses the check until the least-privilege role exists, and gate the follow-up so the bootstrap validate-on-first-use fix lands before enforce is promoted from opt-in to recommended-for-this-flow. That keeps enforce honest for its primary target (existing-cluster queries) today and closes the provisioning gap before we tell anyone to rely on it there.

src/postgres-mcp-server/awslabs/postgres_mcp_server/server.py
kennthhz: see above

src/postgres-mcp-server/awslabs/postgres_mcp_server/connection/db_connection_map.py
kennthhz: not in scope for this change

src/postgres-mcp-server/awslabs/postgres_mcp_server/server.py
kennthhz: See above, disagree this is blocking merge

laithalsaadoon (dismissed):
LGTM

src/postgres-mcp-server/awslabs/postgres_mcp_server/server.py
```diff
@@ -336,6 +499,32 @@ async def connect_to_database(
...
+        # startup path) are validated at creation, so skipping re-validation for
+        # those is safe. Connections seeded by create_cluster /
+        # create_cluster_worker are cached WITHOUT validation (an intentional
+        # bootstrap: the freshly created cluster only has the rds_superuser
+        # master and no least-privilege role yet). For those keys this skip means
+        # an over-privileged connection is not surfaced/evicted on reconnect —
+        # under the default 'warn' this is moot, and under 'enforce' it is the
+        # known create_cluster bootstrap gap tracked as a follow-up
+        # (validate-on-create with a bootstrap exemption).
+        was_cached = (
```
thephantomthief: I think was_cached will always be `False` when db_endpoint is empty, because we key off the resolved cluster writer endpoint on `set()` but look up with the caller's value.

kennthhz: fixed

thephantomthief (approved):

laithalsaadoon (approved):
lgtm

PR #4319, opened 2026-07-22
feat(security-agent-mcp-server): Add custom User-Agent

jigar-lab:
## Summary

Adds a custom User-Agent string to all boto3 API calls from the security-agent-mcp-server, enabling service-side tracking of MCP usage by IDE/tool.

### User-Agent format

```
md/awslabs#mcp#security-agent-mcp-server#0.1.2 md/client#kiro/1.5.0
```

Follows the repo-wide `md/awslabs#mcp#<server-name>#<version>` convention, with an additional `md/client#<name>/<version>` suffix for IDE tracking.

### How it works

1. **`aws_client.py`**: `SecurityAgentClient` builds a `botocore.config.Config(user_agent_extra=...)` and passes it to every `.client()` call (securityagent, sts, s3, iam). Uses the package `__version__` constant directly.
2. **`server.py`**: `@ensure_client_ua` decorator extracts `clientInfo` from the MCP session context (sent by the IDE during `initialize`) and calls `set_mcp_client_info()` on the shared AWS client singleton.

### Design decisions

- **`__version__` constant**: Uses the package-level `__version__` instead of `importlib.metadata.version()` to avoid silent degradation to "unknown" in editable/source installs.
- **Sanitized client name**: `clientInfo.name` is normalized with `.lower().replace(" ", "-")` before injection into the UA string, preventing malformed tokens (e.g. "Claude Code" → "claude-code").
- **Decorator pattern**: `@ensure_client_ua` applied to all tool functions as a single choke point — no per-tool boilerplate, and new tools cannot forget it.
- **Lazy injection**: The `SecurityAgentClient` singleton is created at module level before any MCP session connects, so `clientInfo` is injected on first tool invocation.
- **No-op optimization**: `set_mcp_client_info()` short-circuits if the info has not changed.
- **Graceful fallback**: Falls back to `unknown` when `clientInfo` is unavailable.
- **Concurrency note**: Current stdio transport guarantees one client per process. Under future HTTP/SSE transport with concurrent sessions, `_config` would need per-session isolation.

### Tracking values by IDE

| IDE | md/client# value |
|-----|-----------------|
| Kiro | kiro/\<version\> |
| Claude Code | claude-code/\<version\> |
| Cursor | cursor/\<version\> |
| Unknown | unknown |

### Tested

- All 41 unit tests pass (197 total including existing)
- E2E verified: API call to SecurityAgent service confirmed the custom UA on the wire
- Works with Kiro, Claude Code, and any MCP-compliant client that sends clientInfo

---

### DIFF
--- a/src/security-agent-mcp-server/awslabs/security_agent_mcp_server/aws_client.py
+++ b/src/security-agent-mcp-server/awslabs/security_agent_mcp_server/aws_client.py
@@ -15,25 +15,70 @@
 """AWS SecurityAgent API client using boto3 SDK."""
 
 import boto3
+import botocore.config
 import json
 import re
 from typing import Any, Optional
 
 
+DEFAULT_MCP_CLIENT_NAME = 'unknown'
+
+
 class SecurityAgentClient:
     """Client for AWS SecurityAgent APIs using boto3."""
 
-    def __init__(self, region: str = 'us-east-1'):
+    def __init__(
+        self,
+        region: str = 'us-east-1',
+        mcp_client_name: str = DEFAULT_MCP_CLIENT_NAME,
+        mcp_client_version: str = '',
+    ):
         """Initialize SecurityAgent client."""
         self.region = region
+        self._mcp_client_name = mcp_client_name
+        self._mcp_client_version = mcp_client_version
+        self._config = self._build_config(mcp_client_name, mcp_client_version)
+
+    def _build_config(
+        self, mcp_client_name: str, mcp_client_version: str
+    ) -> botocore.config.Config:
+        """Build a botocore Config with a custom user_agent_extra string."""
+        try:
+            from importlib.metadata import version as pkg_version
+
+            mcp_server_version = pkg_version('awslabs.security-agent-mcp-server')
+        except Exception:
+            mcp_server_version = 'unknown'
+
+        ua_extra = (
+            f'awslabs-security-agent-mcp-server/{mcp_server_version} md/client#{mcp_client_name}'
+        )
+        if mcp_client_version:
+            ua_extra += f'/{mcp_client_version}'
+
+        return botocore.config.Config(user_agent_extra=ua_extra)
+
+    def set_mcp_client_info(self, mcp_client_name: str, mcp_client_version: str = '') -> None:
+        """Update the MCP client identity and rebuild the botocore config.
+
+        Called after MCP session initialization when clientInfo becomes available.
+        """
+        if (
+            mcp_client_name == self._mcp_client_name
+            and mcp_client_version == self._mcp_client_version
+        ):
+            return  # No change needed
+        self._mcp_client_name = mcp_client_name
+        self._mcp_client_version = mcp_client_version
+        self._config = self._build_config(mcp_client_name, mcp_client_version)
 
     def _get_session(self):
         """Fresh session each call to pick up rotated credentials."""
         return boto3.Session(region_name=self.region)
 
     def _client(self):
-        """Get a fresh securityagent boto3 client."""
-        return self._get_session().client('securityagent')
+        """Get a fresh securityagent boto3 client with custom user-agent."""
+        return self._get_session().client('securityagent', config=self._config)
 
     def call(self, operation: str, params: dict) -> dict:
         """Call any SecurityAgent API operation generically."""
@@ -50,7 +95,7 @@ class SecurityAgentClient:
 
     def get_caller_identity(self) -> dict:
         """Get the current AWS caller identity."""
-        return self._get_session().client('sts').get_caller_identity()
+        return self._get_session().client('sts', config=self._config).get_caller_identity()
 
     def list_agent_spaces(self) -> list[dict]:
         """List all SecurityAgent agent spaces."""
@@ -224,7 +269,7 @@ class SecurityAgentClient:
 
         Public-access block, SSE-S3, TLS-only policy, and 30-day lifecycle.
         """
-        s3 = self._get_session().client('s3')
+        s3 = self._get_session().client('s3', config=self._config)
         create_args: dict[str, Any] = {'Bucket': bucket_name}
         if self.region != 'us-east-1':
             create_args['CreateBucketConfiguration'] = {'LocationConstraint': self.region}
@@ -287,7 +332,7 @@ class SecurityAgentClient:
 
     def create_service_role(self, role_name: str, account_id: str, bucket_name: str) -> str:
         """Create IAM service role for Security Agent with S3 + CloudWatch Logs access."""
-        iam = self._get_session().client('iam')
+        iam = self._get_session().client('iam', config=self._config)
 
         trust_policy = json.dumps(
             {
@@ -347,5 +392,5 @@ class SecurityAgentClient:
 
     def upload_to_s3(self, bucket: str, key: str, file_path: str) -> str:
         """Upload a file to S3."""
-        self._get_session().client('s3').upload_file(file_path, bucket, key)
+        self._get_session().client('s3', config=self._config).upload_file(file_path, bucket, key)
         return f's3://{bucket}/{key}'
--- a/src/security-agent-mcp-server/awslabs/security_agent_mcp_server/aws_client.py
+++ b/src/security-agent-mcp-server/awslabs/security_agent_mcp_server/aws_client.py
@@ -18,6 +18,7 @@ import boto3
 import botocore.config
 import json
 import re
+from awslabs.security_agent_mcp_server import __version__
 from typing import Any, Optional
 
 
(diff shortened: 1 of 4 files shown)


src/security-agent-mcp-server/awslabs/security_agent_mcp_server/aws_client.py
```diff
@@ -15,25 +15,70 @@
...
+        """Build a botocore Config with a custom user_agent_extra string."""
+        try:
+            from importlib.metadata import version as pkg_version
+
+            mcp_server_version = pkg_version('awslabs.security-agent-mcp-server')
+        except Exception:
+            mcp_server_version = 'unknown'
+
+        ua_extra = (
+            f'awslabs-security-agent-mcp-server/{mcp_server_version} md/client#{mcp_client_name}'
```
scottschreckengaust: This isn't following the repo's convention

scottschreckengaust (commented):
Thanks for adding usage telemetry — the lazy-injection approach is reasonable and every failure path degrades gracefully (no crashes). I reviewed at high effort and verified the technical claims below empirically. Two items keep the PR from delivering its own stated contract and are worth fixing before merge; the rest are cleanup/altitude.

**Should fix (contract gaps):**
1. **`get_api_guide` sends an untagged User-Agent.** It builds its own `session.client('securityagent')` without `config=`, so it's the one `.client()` site the PR didn't thread the custom UA through — contradicting "passes it to every `.client()` call." The `_ensure_client_ua(ctx)` at the top of that tool is also inert, since it mutates the `_client` singleton this tool bypasses.
2. **Version reports `unknown` from any uninstalled/editable checkout.** `importlib.metadata.version('awslabs.security-agent-mcp-server')` raises `PackageNotFoundError` when run from source (verified in this repo), silently degrading to `unknown`. `__version__ = '0.1.2'` is already exported by `__init__.py`, and ~30 other servers in this repo build the UA from that constant. Reusing it also removes the per-rebuild metadata lookup.

**Consider (cleanup / consistency):**
3. `clientInfo.name` is injected into the UA unsanitized, unlike `_client_prefix` which normalizes it — a name like `Claude Code` produces a broken UA token (botocore does not sanitize `user_agent_extra`; confirmed).
4. `_ensure_client_ua(ctx)` is duplicated as the first line of all 11 tools, and duplicates the clientInfo-extraction logic already in `_client_prefix` (with a divergent fallback: `unknown` vs `ide`).
5. The shared mutable `_client._config` is safe under today's stdio transport but would cross-contaminate the UA across concurrent sessions under HTTP/SSE.

Inline comments below. Nothing here is a crash; posting as a Comment, not a blocking Request Changes.

src/security-agent-mcp-server/awslabs/security_agent_mcp_server/server.py
```diff
@@ -619,6 +660,7 @@ async def get_api_guide(ctx: Context) -> str:
     Returns operation names dynamically from the service model,
     plus a link to full API documentation with parameter details.
     """
+    _ensure_client_ua(ctx)
```
scottschreckengaust: **`get_api_guide` sends an untagged User-Agent — the one `.client()` site the PR missed.**

A few lines below (`session.client('securityagent')`, ~line 670), this tool builds its own boto3 client **without** `config=`, so its call goes out with the default UA rather than `md/awslabs#mcp#security-agent-mcp-server#… md/client#…`. Every site in `aws_client.py` got `config=self._config`; this one in `server.py` did not — directly contradicting the PR summary's "passes it to every `.client()` call."

Also note the `_ensure_client_ua(ctx)` on this line is **inert here**: it mutates the shared `_client` singleton, but this tool never uses `_client` — it constructs a separate client. So the enrichment call gives a false impression the path is covered.

Suggest routing this through `_client` (e.g. expose `operation_names` on the client) or at minimum passing `config=_client._config` to this `session.client(...)` call.

src/security-agent-mcp-server/awslabs/security_agent_mcp_server/aws_client.py
```diff
@@ -15,25 +15,68 @@
...
+        self._mcp_client_name = mcp_client_name
+        self._mcp_client_version = mcp_client_version
+        self._config = self._build_config(mcp_client_name, mcp_client_version)
+
+    def _build_config(
+        self, mcp_client_name: str, mcp_client_version: str
+    ) -> botocore.config.Config:
+        """Build a botocore Config with a custom user_agent_extra string."""
+        try:
+            from importlib.metadata import version as pkg_version
```
scottschreckengaust: **Version silently degrades to `unknown` from a source/editable checkout; reuse the existing `__version__` constant.**

`importlib.metadata.version(...)` reads *distribution* metadata (the `.dist-info` written by `pip install`), which is absent in an uninstalled/editable source tree or a frozen bundle — it raises `PackageNotFoundError` (a subclass of `Exception`, caught by the bare `except`), so the UA reports `security-agent-mcp-server#unknown`. Verified in this repo:

```
>>> from importlib.metadata import version
>>> version('awslabs.security-agent-mcp-server')
PackageNotFoundError: No package metadata was found for awslabs.security-agent-mcp-server
```

`__init__.py` already exports `__version__ = '0.1.2'`, and the repo-wide convention (translate, cloudwatch, eks, ccapi, iam, aws-pricing, … ~30 servers) builds the UA from that constant:

```python
from awslabs.security_agent_mcp_server import __version__
...
ua_extra = f'md/awslabs#mcp#security-agent-mcp-server#{__version__} md/client#{mcp_client_name}'
```

This is always available, keeps a single source of truth, and as a bonus removes the metadata lookup that currently reruns on every `set_mcp_client_info` config rebuild (the server version is a process-lifetime constant).

src/security-agent-mcp-server/awslabs/security_agent_mcp_server/aws_client.py
```diff
@@ -15,25 +15,68 @@
...
+
+            mcp_server_version = pkg_version('awslabs.security-agent-mcp-server')
+        except Exception:
+            mcp_server_version = 'unknown'
+
+        ua_extra = f'md/awslabs#mcp#security-agent-mcp-server#{mcp_server_version} md/client#{mcp_client_name}'
```
scottschreckengaust: **`mcp_client_name` is injected into the UA unsanitized — inconsistent with `_client_prefix`.**

`_client_prefix` (server.py) normalizes the same untrusted `clientInfo.name` with `.lower().replace(' ', '-')`, but here it flows in verbatim. botocore does **not** sanitize `user_agent_extra` (verified on botocore 1.43.32 — spaces and CR/LF pass through to the emitted UA string). So a client reporting `Claude Code` yields the malformed token `md/client#Claude Code/1.0` (the space breaks the UA token), while the scan-title path renders the same client cleanly as `claude-code`. Same input, two serializations — UA-based telemetry parsing becomes unreliable.

Suggest normalizing `mcp_client_name` (and `mcp_client_version`) the same way `_client_prefix` does before interpolating, ideally via a shared helper so the two paths can't drift.

src/security-agent-mcp-server/awslabs/security_agent_mcp_server/server.py
```diff
@@ -140,6 +143,34 @@ def _client_prefix(ctx: Context) -> str:
         return 'ide'
 
 
+def _ensure_client_ua(ctx: Context) -> None:
```
scottschreckengaust, line 147: **Altitude: `_ensure_client_ua` is a per-tool bandaid that duplicates existing extraction logic.**

Two observations on depth:

1. `_ensure_client_ua(ctx)` is pasted as the literal first line of all 11 `@mcp.tool()` functions. `clientInfo` is fixed for a session's lifetime, so re-running extract-compare on every call is the wrong altitude, and any new tool must remember to re-add the line or its calls silently ship as `unknown`. A single choke point — a shared decorator wrapping the tools, or a session-init hook — would set it once.
2. This function duplicates the `ctx.session → client_params → clientInfo.name` traversal (and its `except (AttributeError, TypeError)` guard) already in `_client_prefix`. The two copies even disagree on the fallback for the identical missing field: `unknown` here vs `ide` there. Extracting one helper that returns the parsed name would remove the drift risk.

src/security-agent-mcp-server/awslabs/security_agent_mcp_server/aws_client.py
```diff
@@ -15,25 +15,68 @@
...
+        Called after MCP session initialization when clientInfo becomes available.
+        """
+        if (
+            mcp_client_name == self._mcp_client_name
+            and mcp_client_version == self._mcp_client_version
+        ):
+            return  # No change needed
+        self._mcp_client_name = mcp_client_name
+        self._mcp_client_version = mcp_client_version
+        self._config = self._build_config(mcp_client_name, mcp_client_version)
```
scottschreckengaust, line 70: **Shared mutable `_config` on a module-level singleton — safe today, a cross-session hazard under HTTP/SSE.**

`_client` is a process-wide singleton and `set_mcp_client_info` rebuilds `self._config` in place. Under the current stdio transport this is fine (one client per process, boto3 calls block the loop). But if this server ever runs over streamable-http/SSE with concurrent sessions from different clients, each session's `_ensure_client_ua` overwrites the shared `_config`, so `md/client#…` becomes nondeterministic across clients — and even single-session, two concurrent async tool calls can race the read-modify-write with no lock.

Not blocking given today's transport, but worth a comment recording the single-session assumption so a future HTTP transport doesn't silently inherit the bug.

src/security-agent-mcp-server/tests/test_aws_client.py
```diff
@@ -457,3 +457,86 @@ def test_batch_get_threats(self, mock_boto3):
...
+    @patch('awslabs.security_agent_mcp_server.aws_client.boto3')
+    def test_upload_to_s3_passes_config(self, mock_boto3):
+        """upload_to_s3 passes the botocore config to s3 client."""
+        mock_session = MagicMock()
+        mock_boto3.Session.return_value = mock_session
+        client = SecurityAgentClient(region='us-east-1')
+        client.upload_to_s3('bucket', 'key', '/path/to/file')
+        mock_session.client.assert_called_once_with('s3', config=client._config)
+
+    def test_build_config_handles_missing_package_metadata(self):
```
scottschreckengaust: **This test entrenches the fragile fallback rather than flagging it.**

`test_build_config_handles_missing_package_metadata` patches `importlib.metadata.version` to raise and asserts the UA contains `…#unknown`. It passes (and locks in `unknown` as intended behavior) precisely because of the fragility called out on `aws_client.py:47` — so it would keep passing even though real installs from source lose the version. If you switch to the `__version__` constant, this test becomes unnecessary; otherwise it's giving false confidence that version resolution is robust.

PR #4344, opened 2026-07-24
fix(aws-transform-mcp-server): discover instruction docs inside artifact-store folders

jabouayash:
`load_instructions` discovers instruction documents by calling `ListArtifacts` with no `pathPrefix`. That listing returns only the artifacts at the top level of the job's artifact store and the names of its folders; it does not descend into them. Customer uploads (both the `upload_artifact` tool and the AWS Transform web console) are stored inside the store's `User Uploads/` folder, so an uploaded instruction document can never be discovered, regardless of its file name. The same one-sided check affects `guidance_nudge`: it runs once per job at creation time, before any upload has happened, and never re-runs.

This change makes discovery walk the store:

- The root listing runs as before; if no instruction document is found there, each folder the root listing reports is scanned with `pathPrefix`. The canonical `AWSTransform/Workspaces/{workspaceId}/Jobs/{jobId}/User Uploads/` prefix is appended as a fallback in case the root listing omits it. Per-folder scans are individually guarded so a folder that fails path validation is skipped rather than aborting discovery.
- Label matching compares against the path basename, so `User Uploads/JOB_INSTRUCTIONS` matches the `JOB_INSTRUCTIONS` label the same way a root-level upload does.
- `upload_artifact` re-arms the one-shot guidance check when the uploaded file matches an instruction label, so a document uploaded mid-session is picked up on the next `load_instructions` call.

Verified end to end against a live AWS Transform job: with an instruction document uploaded through `upload_artifact`, the released v0.1.6 server returns `instructionsFound=false`, and this patched server finds the document and returns its content. Ten new unit tests cover the folder walk (including the `pathPrefix` values passed), the canonical-prefix fallback, basename label matching, and the re-arm on upload.

### DIFF
--- a/src/aws-transform-mcp-server/awslabs/aws_transform_mcp_server/tools/load_instructions.py
+++ b/src/aws-transform-mcp-server/awslabs/aws_transform_mcp_server/tools/load_instructions.py
@@ -16,7 +16,10 @@
 
 from awslabs.aws_transform_mcp_server.audit import audited_tool
 from awslabs.aws_transform_mcp_server.config_store import is_fes_available
-from awslabs.aws_transform_mcp_server.guidance_nudge import mark_job_checked
+from awslabs.aws_transform_mcp_server.guidance_nudge import (
+    mark_job_checked,
+    matches_instruction_label,
+)
 from awslabs.aws_transform_mcp_server.tool_utils import (
     READ_ONLY,
     download_s3_content,
@@ -30,16 +33,43 @@ from awslabs.aws_transform_mcp_server.transform_api_models import (
     JobFilter,
     ListArtifactsRequest,
 )
+from loguru import logger
 from mcp.server.fastmcp import Context
 from pydantic import Field
-from typing import Annotated, Any, Dict
+from typing import Annotated, Any, Dict, List, Optional, Tuple
 
 
 _NOT_CONFIGURED_CODE = 'NOT_CONFIGURED'
 _NOT_CONFIGURED_MSG = 'Not connected to AWS Transform.'
 _NOT_CONFIGURED_ACTION = 'Call configure with authMode "cookie" or "sso".'
 
-INSTRUCTION_ARTIFACT_LABELS = ['JOB_INSTRUCTIONS']
+
+async def _scan_for_instruction_artifact(
+    workspace_id: str, job_id: str, path_prefix: Optional[str] = None
+) -> Tuple[Optional[Dict[str, Any]], List[str]]:
+    """Paginated ListArtifacts scan for an instruction artifact.
+
+    Returns (matching artifact or None, folders seen across all pages).
+    """
+    folders: List[str] = []
+    next_token: Optional[str] = None
+    while True:
+        list_req = ListArtifactsRequest(
+            workspaceId=workspace_id,
+            jobFilter=JobFilter(jobId=job_id),
+            nextToken=next_token,
+            pathPrefix=path_prefix,
+        )
+        result = await call_transform_api('ListArtifacts', list_req)
+        for artifact in result.get('artifacts', []):
+            if matches_instruction_label((artifact.get('fileMetadata') or {}).get('path') or ''):
+                return artifact, folders
+        for folder in result.get('folders') or []:
+            if folder not in folders:
+                folders.append(folder)
+        next_token = result.get('nextToken')
+        if not next_token:
+            return None, folders
 
 
 class LoadInstructionsHandler:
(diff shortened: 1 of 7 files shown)


jabouayash:
Adding testing details, including regression coverage for jobs and agent types that have no instruction documents.

**How the fix was verified (instruction document present).** A/B against the same live AWS Transform job in an internal test stage, with a `JOB_INSTRUCTIONS` document uploaded through this server's `upload_artifact` (stored by the service under `User Uploads/`): the released v0.1.6 server returns `instructionsFound=false`; the patched server discovers the document, downloads it, and returns its content with the expected artifactId. This was the customer-reported .NET scenario.

**No-regression check (no instruction document — the normal case for every job).** Same A/B on a job with no instruction documents anywhere in its store:

| | released v0.1.6 | patched |
|---|---|---|
| `load_instructions` result | success, `instructionsFound=false` | success, `instructionsFound=false` |
| gate opens afterwards (no `INSTRUCTIONS_REQUIRED` from job-scoped tools) | yes | yes |
| `load_instructions` latency | 0.8s | 2.5s |

The tool does not block or error when nothing is found: the not-found path still returns `success` with "Proceed normally" and marks the job checked, exactly as before. The latency delta is the folder walk itself (one extra `ListArtifacts` call per folder the root listing reports, plus the canonical uploads prefix) and is only paid when the root listing finds no instruction document; a per-folder failure is caught and skipped, so a folder that fails path validation cannot fail the call.

**Why this covers VMware/mainframe agents.** This server has no per-agent-type code path — `load_instructions` and the guidance check run identically for every job type. The only behavior change is additional read-only `ListArtifacts` calls on the not-found path, and the no-document outcome above is byte-identical to the released server. Unit tests pin both sides: the folder walk finding a document (including the exact `pathPrefix` values sent), the walk finding nothing and still returning the normal "proceed" result, and the gate opening after either outcome.

src/aws-transform-mcp-server/awslabs/aws_transform_mcp_server/tools/load_instructions.py
```diff
@@ -70,29 +99,29 @@ async def load_instructions(
...
                 )
-                if instruction_artifact:
-                    break
-                next_token = artifacts_result.get('nextToken')
-                if not next_token:
-                    break
+                if canonical_uploads_prefix not in folders:
+                    folders.append(canonical_uploads_prefix)
+                for folder in folders:
+                    instruction_artifact, _ = await _scan_for_instruction_artifact(
```
youtuyy: it will fail the entire when a single folder scan errors? Does not align with the PR  drscription `Per-folder scans are individually guarded so a folder that fails path validation is skipped rather than aborting discovery.`

jabouayash: Good catch — that guard was in the tested internal variant of this change but got dropped in the port to this repo. Restored in the latest push: each folder scan is wrapped in try/except-continue so a prefix the store rejects skips to the next candidate instead of failing the call, with a unit test pinning it (first folder scan raises, walk continues, document found in the second).

youtuyy (dismissed):

src/aws-transform-mcp-server/awslabs/aws_transform_mcp_server/tools/load_instructions.py
```diff
@@ -30,16 +33,43 @@
...
+async def _scan_for_instruction_artifact(
+    workspace_id: str, job_id: str, path_prefix: Optional[str] = None
+) -> Tuple[Optional[Dict[str, Any]], List[str]]:
+    """Paginated ListArtifacts scan for an instruction artifact.
+
+    Returns (matching artifact or None, folders seen across all pages).
+    """
+    folders: List[str] = []
+    next_token: Optional[str] = None
+    while True:
```
occhangg, line 56: Is this draining all pages? What if there are 1000 pages and the instruction is in the very last element? This means MCP client would need to page all 1000, save all 1000 item into in memory. We would need some sort of filter to get that instruction artifact from the api.

jabouayash: Memory is fine - each page gets checked and thrown away as it comes in, nothing accumulates. And the full page drain on the not-found path isn't new, the released code already pages through the whole unprefixed listing the same way.

Agreed a server-side filter would be the right fix though. There isn't one today - no filename/label filter on ListArtifacts, and pathPrefix 400s on a file-level prefix (tried it against a live job). I can file that as a follow-up.

In the meantime I pushed a small tweak: after a root miss we now scan User Uploads/ first, since that's where customer docs land. So the common case is one extra single-page call, and we only page everything when there's genuinely nothing to find.

occhangg (approved):

youtuyy (approved):

PR #4345, opened 2026-07-24
feat(billing-cost-management-mcp-server): Add multi-region support to the AWS Compute Optimizer Automation MCP tool

mattkelsey:
# Add multi-region support to the AWS Compute Optimizer Automation MCP tool

Compute Optimizer Automation exposes regional APIs behind a global Console. This change lets a single MCP call span several regions and return one merged result, so users and agents no longer have to make repeat per-region calls and stitch the responses together by hand.

Region selection is explicit: the tool takes an optional `regions` list. Omitting it performs a single-region request, preserving the existing behavior.

A similar change will be needed for the AWS Compute Optimizer sub-server, but it needs to be onboarded to SQL offloading before we fan out, as the response size will balloon. This tool routes merged responses through the existing SQL offload path.

## Changes

- Replaces the single `region` string with an optional `regions` list. With no list, one call targets the `AWS_REGION` env var or us-east-1; with a list, the tool queries those regions concurrently (bounded at 8) and merges the results.
- Preserves resource region context: each returned item is stamped with the region that produced it, and the response reports `regions_queried`.
- Reports regional failures without discarding successful results — per-region errors are classified into `region_errors` while the regions that succeeded still return their data.
- Adds cross-region pagination via an opaque `next_token` that encodes per-region state and resumes only the regions with results outstanding. `max_pages` applies per region.
- Adds cross-region lookup for `get_automation_event`, returning the first hit with `found_in_region`, and distinguishes "absent from every region searched" from "could not be determined" when some regions fail.
- Keeps the five account-global operations (`get_automation_rule`, `get_enrollment_configuration`, `list_accounts`, `list_automation_rules`, `list_tags_for_resource`) single-region. Rules are global resources and enrollment/account data is account-scoped, so fanning them out would repeat identical data; they accept at most one region and reject more.
- Validates structured inputs, filter names, and the region list before any client is created, so malformed input fails once with an actionable message rather than once per region.
- Extracts the fan-out mechanics into a reusable `regional_fanout` utility — bounded concurrency, AWS error classification, and pagination-token encode/decode — for other regional services to adopt.
- Extends regression coverage for region routing, pagination, partial and total regional failure, not-found semantics, and the rule-preview operations.

## Behavior

| `regions` | Result |
|---|---|
| Omitted or `[]` | One call to the default region, native response shape |
| `["us-east-1"]` | Merged shape: `region` per item, `regions_queried` |
| `["us-east-1", "eu-west-1", …]` | Concurrent fan-out, merged items, per-region `region_errors` |
| Account-global operation, one region | That region as the endpoint, native shape |
| Account-global operation, 2+ regions | Validation error |

Duplicate entries collapse so no region is queried twice; blank entries are rejected. A multi-region `next_token` must be passed back with the same `regions` list it was produced with. Region names are not validated locally, so an unavailable or misspelled region surfaces as that region's entry in `region_errors`.

Because the region set now comes from the caller, the hard-coded list of Automation regions is removed.

## User experience

Before this change, covering multiple regions meant repeat calls and manually combining the results.

After this change, a caller passes the regions it cares about and gets one merged response with region provenance, pagination, and actionable regional errors — the same cross-region aggregation the ACO Console performs. Omitting `regions` continues to perform a targeted single-region request.

## Testing

- 1109 tests pass in the package.
- 100% line and branch coverage on all three changed modules — `compute_optimizer_automation_tools.py`, `compute_optimizer_automation_operations.py`, and `regional_fanout.py` (0 missed branches).
- pyright: 0 errors, 0 warnings. `ruff check` and `ruff format`: clean. bandit at the CI gate (medium severity, medium confidence): clean.

## Checklist

* [X] Changes are documented

Is this a breaking change? (Y/N) Y

**RFC issue number**: N/a, update existing server

Checklist:

* [ ] Migration process documented N/a
* [ ] Implement warnings (if it can live side by side) N/a

## Acknowledgment

## Testing
**Unit tests** — 116 tests in `tests/tools/test_compute_optimizer_automation_tools.py` (full package suite: 1059). New coverage: multi-region merge, per-item region stamping, opaque global `next_token` round-trip/resume/rejection (invalid base64, empty map, unknown regions, non-string tokens, cross-mode misuse in both directions), partial and total region failure, typed per-region error classification, not-found vs indeterminate semantics for `get_automation_event` and `list_automation_event_steps`, single-region passthrough, account-global routing, and SQL offload of merged responses with metadata survival. `ruff check`/`format` and `pyright` clean.

Spun up mini-eval system, validated against real cross-region account data:

**Live end-to-end eval** against accounts with real data:

1. **Ground truth**: plain boto3 (no MCP) paginates each Automation API across all 17 regions and records totals, per-region counts, rule names, enrollment status, and a real event ID.
2. **Agent runs**: one headless Sonnet-5 agent per route (8 routes covering every dispatch path: global fan-out lists, cross-region summary sums, single-region passthrough, account-global ops, fan-out event lookup by ID).
3. **Eval agent**: a tool-less evaluator agent compared ground truth vs agent answers per route (ignoring formatting).

**Result: 16/16 routes MATCH, zero drift, both accounts PASS.** Large merged responses (1,768 summary rows) exercised the SQL-offload path end-to-end: agents queried the offloaded table via `session-sql` and still produced correct totals.

**Performance** (live, per global call): 17 regions queried concurrently in ~2.6s wall vs ~12.7s sum of individual call times (semaphore cap 8); single-region calls unchanged (~0.5s).


src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/compute_optimizer_automation_tools.py
```diff
@@ -46,11 +65,35 @@
...
 
 
 _SERVICE_NAME = 'Compute Optimizer Automation'
 _BOTO_SERVICE_NAME = 'compute-optimizer-automation'
+_MAX_CONCURRENT_REGIONS = 8
+
+# The AWS regions where Compute Optimizer Automation is available. The service is
+# absent from botocore's endpoints.json (it ships only an endpoint rule set), so
+# there is no local API to enumerate its regions; this list is maintained by hand.
+COMPUTE_OPTIMIZER_AUTOMATION_REGIONS = [
```
mattkelsey: Open to ideas on this. Some other sub-servers also hardcode regions. Would be better to pull from SSM but that would require an extra permission. IMO doesn't make sense to cordon off the entire MCP behind an SSM permission.

endpoints.json is deprecated to my understanding

callnmm (approved):

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/compute_optimizer_automation_tools.py
```diff
@@ -306,25 +548,199 @@ async def compute_optimizer_automation(
...
+        offload_metadata['region_errors'] = region_errors
+
+    response_data = await convert_response_if_needed(
+        ctx,
+        response_data,
+        f'compute_optimizer_automation_{operation}',
+        pagination_token_key='next_token',
+        **offload_metadata,
+    )
+    return format_response('success', response_data)
```
Comusus, line 694: Because we are now accepting a list of regions to tool, how do we expect to handle partial and full failures when any subset of regional calls fail in tool?


#### Partial Error
When `region_errors` is not empty, it seems like we return top level `'success'` which is expected behavioral of batch APIs like https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_BatchGetItem.html. 

#### Full Error
When all region calls fails in single tool call, it seems like the tool top-level response would be `error`. An AWS Batch API would still return a top level success in such case with region_errors.

---

Are these behavioral expected? Believe this is the first tool that accepts >1 region.

Comusus: If LLM is able to recover from nonatomic error to rescope new tool call with the subset of regions from nested region_errors, change looks good to me.

Comusus (approved):

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/compute_optimizer_automation_tools.py
mattkelsey: The errors are returned to the agent, and the LLM can decide what to do with them (surface to user, retry tool call, etc) I think that's the right pattern, but open to alternatives

PR #4351, opened 2026-07-27
fix(redshift-mcp-server): remove review_cluster false positives

vsiniavin:
### Changes

Three fixes to `review_cluster`, all from one report against a 2-node `rg.xlarge` cluster that received two findings it should not have.

**1. Node type is read from the API instead of inferred from storage capacity**

  `NodeDetails` derived the node type from `stv_partitions.capacity` via a hardcoded CASE covering only dc2, ds2 and ra3. RG node types have no entry, so RG nodes fell into an `ra3.*` bucket and matched every `node_type` list, including `should consider migrating to RG instances`.

**2. The Spectrum QMR signal is gated on Spectrum activity**

The branch reporting a missing QMR rule for `spectrum_scan_size_mb` or `spectrum_scan_row_count` read WLM configuration only, never whether the cluster queries the data lake. QMR rules are opt-in, so it fired on nearly every cluster, and it reads as an active risk that may not exist.

That branch now also requires S3 external query activity in the last 7 days from `SYS_EXTERNAL_QUERY_DETAIL`. `source_type` is `S3` for Spectrum, and `PG` for federated queries, which these metrics do not govern. Seven days matches the documented system view log retention and the windows already used by `UsagePattern` and `ServerlessScaling`. The other three REC_019 branches are unchanged, so REC_019 still surfaces on clusters that need QMR rules.

**3. Access documentation points at `sys:monitor`**

The permission error, tool docstring, server instructions and README all recommended `ALTER USER <user> CREATEUSER`. That cannot work for the identity this server connects as: no `DbUser` or `SecretArn` is sent to the Data API, so the database user is the caller's IAM identity (`IAM:<user>` or `IAMR:<role>`), whose password is
disabled, and a superuser cannot have a disabled password. The attempt fails with `ERROR: Superusers cannot have disabled passwords`.

All four now recommend `GRANT ROLE sys:monitor TO "<database_user>"`, state that the name comes from `SELECT current_user` and must be quoted for IAM identities, and note that the grant must be issued by a superuser such as the cluster admin user.

  ## Testing

  * 242 unit tests pass; `ruff check` and `ruff format --check` clean.
  * New tests assert an RG cluster's node type appears exactly once in `NodeDetails` (in the projection, in no predicate), that pre-RG node types still match every branch, that only the Spectrum QMR branch is gated, and that REC_019 still has four branches. Each was mutation-checked: reverting the fix fails the test.
  * Verified live against a 2-node `rg.xlarge`, a 1-node `ra3.large` and a serverless workgroup: 11 signals on provisioned, 8 on serverless, no unsubstituted placeholder reached the database, and `WLMConfig` executes with the new `SYS` subquery alongside the `stv_wlm_*` tables.
  * Permission handling verified by revoking `sys:monitor` from the connecting IAM database user and re-running: the new message is returned and the calling agent proposed the correct grant.

  ## Checklist

  * [x] I have reviewed the [contributing
  guidelines](https://github.com/awslabs/mcp/blob/main/CONTRIBUTING.md)
  * [x] Changes are documented

  `review_cluster` output changes for RG clusters and for clusters without recent Spectrum activity, but the tool signature and response schema are unchanged.

  ## Acknowledgment


src/redshift-mcp-server/awslabs/redshift_mcp_server/review/executor.py
```diff
@@ -28,6 +31,35 @@
...
+        node_type: The node type reported by the Redshift DescribeClusters API,
+            for example 'ra3.xlplus' or 'rg.4xlarge'. May be None for serverless
+            workgroups or when the API omits it.
+
+    Returns:
+        The node type when it matches NODE_TYPE_PATTERN, otherwise NODE_TYPE_UNKNOWN.
+        NODE_TYPE_UNKNOWN appears in no signal's node_type list, so an unresolvable
+        node type yields no node-type findings instead of misattributed ones.
+    """
+    if node_type and NODE_TYPE_PATTERN.match(node_type):
```
grayhemp: Why are we checking if it matches the pattern? What is the reason we don't trust the API?

src/redshift-mcp-server/awslabs/redshift_mcp_server/review/executor.py
```diff
@@ -62,9 +94,13 @@ async def review_cluster(
...
 
-    # Stage 1: Select queries, filtering by cluster type scope
+    # Stage 1: Select queries, filtering by cluster type scope, and inline the cluster's
+    # node type from the Redshift API so node-type signals evaluate it directly.
+    # Serverless workgroups have no node type, and every query using the placeholder is
+    # provisioned-only, so nothing is substituted for them.
+    node_type = NODE_TYPE_UNKNOWN if is_serverless else resolve_node_type(cluster_info.node_type)
     queries = [
-        (name, sql)
+        (name, sql.replace(NODE_TYPE_PLACEHOLDER, node_type))
```
grayhemp: I wonder if .format() with kw args would work better here https://docs.python.org/3/tutorial/inputoutput.html#the-string-format-method. We control the input (from our own API), so it should be safe and native. IMHO, NODE_TYPE_PLACEHOLDER - a bit of overengineering smell.

grayhemp (changes requested):

src/redshift-mcp-server/awslabs/redshift_mcp_server/review/definitions.py
```diff
@@ -351,6 +351,12 @@
 }
 
 
+# Every query is rendered with str.format(node_type=...) before it runs, so a
+# {node_type} field is substituted with the cluster's node type from the Redshift
+# DescribeClusters API. Only NodeDetails uses it. Any other brace in a query would be
```
grayhemp: "Only NodeDetails uses it. Any other brace in a query would be read as a format field, which test_no_unrecognized_braces guards against." - no need in maintaining that as we will need to remember to update it when the situation changes.

src/redshift-mcp-server/awslabs/redshift_mcp_server/server.py
```diff
@@ -657,11 +657,20 @@ async def review_cluster_tool(
     - Ensure your AWS credentials are properly configured (via AWS_PROFILE or default credentials).
     - The cluster must be available and accessible.
     - Required IAM permissions: redshift-data:ExecuteStatement, redshift-data:DescribeStatement, redshift-data:GetStatementResult.
-    - The connected user must have superuser (CREATEUSER) privileges to access the required system views.
-      If it does not, the review fails fast with "Review requires superuser (CREATEUSER) privileges"
-      (for example "permission denied for relation sys_auto_table_optimization"). This is by design -
-      an expected signal, not a tool defect - so the review never returns partial or misleading
-      results. Run the review as a superuser to get a complete assessment.
+    - The connected database user must be able to read Redshift system views, some of which
```
grayhemp: Please make it more concise.

src/redshift-mcp-server/awslabs/redshift_mcp_server/review/executor.py
```diff
@@ -62,9 +62,14 @@ async def review_cluster(
 
     is_serverless = cluster_info.type == 'serverless'
 
-    # Stage 1: Select queries, filtering by cluster type scope
+    # Stage 1: Select queries, filtering by cluster type scope, and render each with the
```
grayhemp: Please make it concise.

src/redshift-mcp-server/awslabs/redshift_mcp_server/review/definitions.py
```diff
@@ -967,9 +963,15 @@
 WHERE service_class_id <> 5 and service_class_id <> 14 and service_class_id <> 15 AND ((select count(1) from data where coalesce(qmr_rule,'') not like '%query_temp_blocks_to_disk%') = (select count(1) from data))
 UNION ALL
 -- Signal: no QMR defined for spectrum_scan_size_mb or spectrum_scan_row_count metric
+-- Gated on recent Spectrum activity: those metrics only bound queries that scan S3
```
grayhemp: Please make it concise.

src/redshift-mcp-server/tests/test_review_executor.py
```diff
@@ -250,18 +305,43 @@ async def _side_effect(*a, **kw):
...
+            Exception, match='Review requires read access to Redshift system views'
+        ):
             await review_cluster(
                 cluster_identifier='test-cluster',
                 execute_query_func=execute_query_func,
                 discover_clusters_func=_make_discover_clusters(),
             )
 
+    @pytest.mark.asyncio
+    async def test_permission_denied_message_names_the_working_grant(self):
```
grayhemp: No need in testing the content of the message. Let's remove this test.

src/redshift-mcp-server/awslabs/redshift_mcp_server/review/definitions.py
```diff
@@ -967,9 +963,15 @@
...
+-- external data, so on a cluster that runs none the missing rules are not actionable.
+-- source_type is 'S3' for Spectrum (including Glue and S3 Tables external schemas) and
+-- 'PG' for federated queries, which these metrics do not govern. If a lake source type
+-- other than 'S3' is ever introduced this signal goes quiet rather than firing on every
+-- cluster, which is the safer direction.
 SELECT count(*), 'REC_019', 'no QMR defined for spectrum_scan_size_mb or spectrum_scan_row_count metric'
 FROM data
-WHERE service_class_id <> 5 and service_class_id <> 14 and service_class_id <> 15 AND ((select count(1) from data where coalesce(qmr_rule,'') not like '%spectrum_scan%') = (select count(1) from data))
+WHERE service_class_id <> 5 and service_class_id <> 14 and service_class_id <> 15 AND ((select count(1) from data where coalesce(qmr_rule,'') not like '%spectrum_scan%') = (select count(1) from data)) AND ((SELECT count(1) FROM SYS_EXTERNAL_QUERY_DETAIL WHERE trim(source_type) = 'S3' AND start_time >= dateadd(day, -7, getdate())) > 0)
```
grayhemp: I wonder if `EXISTS ()` would be more efficient than `SELECT count()` here?

src/redshift-mcp-server/tests/test_review_executor.py
```diff
@@ -376,6 +456,232 @@ async def test_missing_recommendation_id_skipped(self):
...
+        await review_cluster(
+            cluster_identifier='test-cluster',
+            execute_query_func=execute_query_func,
+            discover_clusters_func=_make_discover_clusters(node_type='rg.xlarge'),
+        )
+
+        assert "'rg.xlarge'::varchar(32) AS node_type" in recorded['NodeDetails']
+
+    @pytest.mark.asyncio
+    async def test_no_executed_query_retains_the_placeholder(self):
```
grayhemp: That is not needed because it basically tests str.format().

src/redshift-mcp-server/tests/test_review_executor.py
```diff
@@ -376,6 +456,232 @@ async def test_missing_recommendation_id_skipped(self):
...
+        await review_cluster(
+            cluster_identifier='test-cluster',
+            execute_query_func=execute_query_func,
+            discover_clusters_func=_make_discover_clusters(node_type=None),
+        )
+
+        assert "'unknown'::varchar(32) AS node_type" in recorded['NodeDetails']
+
+    @pytest.mark.asyncio
+    async def test_legacy_node_types_still_get_the_rg_migration_signal(self):
```
grayhemp: Not needed - the code is indifferent to what the API returns whether it's legacy or not.

src/redshift-mcp-server/awslabs/redshift_mcp_server/review/definitions.py
```diff
@@ -549,24 +555,14 @@
...
-       WHEN capacity = 952455 THEN 'ds2.xlarge'
-       WHEN capacity = 945026 THEN 'ds2.8xlarge'
-       WHEN capacity < 2002943 THEN 'ra3.large'
-       WHEN capacity = 2002943 AND part_count = 1 THEN 'ra3.xlplus'
-       WHEN capacity > 2002943 AND part_count = 1 THEN 'ra3.4xlarge'
-       WHEN capacity > 2002943 AND part_count = 4 THEN 'ra3.16xlarge'
-       ELSE 'unknown'
-       END AS node_type,
+-- node_type is substituted from the cluster's actual node type as reported by the Redshift DescribeClusters API.
+SELECT '{node_type}'::varchar(32) AS node_type,
```
grayhemp: Just `::text`.

src/redshift-mcp-server/tests/test_review_executor.py
```diff
@@ -376,6 +456,232 @@ async def test_missing_recommendation_id_skipped(self):
...
+            execute_query_func=execute_query_func,
+            discover_clusters_func=_make_discover_clusters(node_type='ra3.xlplus'),
+        )
+
+        predicate = _branch_where(recorded['NodeDetails'], 'REC_001', 'migrating to RG')
+        assert "'ra3.xlplus'" in predicate
+
+    @pytest.mark.asyncio
+    @pytest.mark.parametrize('node_type', ['rg.xlarge', 'rg.4xlarge'])
+    async def test_rg_cluster_matches_no_node_type_signal(self, node_type):
```
grayhemp: No need in testing SQL constants. We rely on SQL to be correct.

src/redshift-mcp-server/tests/test_review_executor.py
```diff
@@ -376,6 +456,232 @@ async def test_missing_recommendation_id_skipped(self):
...
+    @pytest.mark.parametrize(
+        ('rec_id', 'label_fragment'),
+        [
+            ('REC_001', 'migrating to RG'),
+            ('REC_002', 'recommended storage threshold of 70%'),
+            ('REC_006', 'storage threshold of 50%'),
+            ('REC_011', 'under-utilized storage'),
+        ],
+    )
+    async def test_rg_excluded_from_each_node_type_branch(self, rec_id, label_fragment):
```
grayhemp: Same here, no testing SQL constants.

src/redshift-mcp-server/tests/test_review_executor.py
```diff
@@ -376,6 +456,232 @@ async def test_missing_recommendation_id_skipped(self):
...
+    @pytest.mark.parametrize(
+        ('rec_id', 'label_fragment'),
+        [
+            ('REC_001', 'migrating to RG'),
+            ('REC_002', 'recommended storage threshold of 70%'),
+            ('REC_006', 'storage threshold of 50%'),
+            ('REC_011', 'under-utilized storage'),
+        ],
+    )
+    async def test_pre_rg_node_types_keep_storage_signals(self, rec_id, label_fragment):
```
grayhemp: Same here, no testing SQL constants.

src/redshift-mcp-server/tests/test_review_executor.py
```diff
@@ -376,6 +456,232 @@ async def test_missing_recommendation_id_skipped(self):
...
+            cluster_identifier='test-cluster',
+            execute_query_func=execute_query_func,
+            discover_clusters_func=_make_discover_clusters(node_type='dc2.large'),
+        )
+
+        predicate = _branch_where(recorded['NodeDetails'], rec_id, label_fragment)
+        assert "'dc2.large'" in predicate
+
+    @pytest.mark.asyncio
+    async def test_data_skew_signal_still_applies_to_rg(self):
```
grayhemp: Same here, no testing SQL constants.

src/redshift-mcp-server/tests/test_review_executor.py
```diff
@@ -376,6 +456,232 @@ async def test_missing_recommendation_id_skipped(self):
...
+
+
+def _query_sql(name: str) -> str:
+    """Return the SQL of a named diagnostic query."""
+    matches = [sql for query_name, _ct, sql in SIGNAL_EVALUATION_SQL if query_name == name]
+    assert len(matches) == 1, f'expected 1 query named {name}, got {len(matches)}'
+    return matches[0]
+
+
+class TestSpectrumActivityGate:
```
grayhemp: Same here, no testing SQL constants.

grayhemp: No need in all this class and the _query_sql() helper.

src/redshift-mcp-server/tests/test_review_executor.py
grayhemp: No testing constants. Please remove all that.

src/redshift-mcp-server/README.md
```diff
@@ -459,6 +459,12 @@ In addition to AWS IAM permissions, you need appropriate database-level permissi
...
 - **Schema Access**: `USAGE` permissions on schemas you want to explore
 - **Database Access**: Connection permissions to databases you want to access
-- **Review Access**: The `review_cluster` tool requires superuser (CREATEUSER) privileges to read system views such as `SYS_AUTO_TABLE_OPTIMIZATION`, `SYS_QUERY_HISTORY`, and `SVV_TABLE_INFO`. Grant via `ALTER USER <username> CREATEUSER;` or `GRANT ROLE sys:monitor TO <username>;` for a narrower scope.
+- **Review Access**: The `review_cluster` tool reads system views such as `SYS_AUTO_TABLE_OPTIMIZATION`, `STV_NODE_STORAGE_CAPACITY`, and `SVV_TABLE_INFO`, some of which Redshift makes visible only to superusers. Grant the connecting database user the `sys:monitor` role, which is the narrowest grant that covers them:
+
+  ```sql
+  GRANT ROLE sys:monitor TO "<database_user>";
+  ```
+
+  `<database_user>` is the output of `SELECT current_user`. When the server authenticates with IAM credentials it is `IAM:<user>` or `IAMR:<role>`, and the double quotes are required for those names. The grant must be issued by a superuser, such as the cluster's admin (master) user.
```
grayhemp: "(master)" remove.

grayhemp (changes requested):
Few cosmetics. Mostly tests overkill, including some previously missed.

src/redshift-mcp-server/tests/test_review_executor.py
grayhemp: I meant we don't need the TestReviewQueriesConstants class at all.

src/redshift-mcp-server/tests/test_review_executor.py
```diff
@@ -376,6 +401,41 @@ async def test_missing_recommendation_id_skipped(self):
         assert len(result.recommendations) == 0
 
 
+# ---------------------------------------------------------------------------
+# Node type substitution
+# ---------------------------------------------------------------------------
+
+
+class TestNodeTypeSubstitution:
```
grayhemp, line 402: Missing one more test method for not-'unknown' node type (ie. 'ra3.xlplus').

src/redshift-mcp-server/README.md
```diff
@@ -459,6 +459,12 @@ In addition to AWS IAM permissions, you need appropriate database-level permissi
 - **Read Access**: `SELECT` permissions on tables/views you want to query
 - **Schema Access**: `USAGE` permissions on schemas you want to explore
 - **Database Access**: Connection permissions to databases you want to access
-- **Review Access**: The `review_cluster` tool requires superuser (CREATEUSER) privileges to read system views such as `SYS_AUTO_TABLE_OPTIMIZATION`, `SYS_QUERY_HISTORY`, and `SVV_TABLE_INFO`. Grant via `ALTER USER <username> CREATEUSER;` or `GRANT ROLE sys:monitor TO <username>;` for a narrower scope.
+- **Review Access**: The `review_cluster` tool reads system views such as `SYS_AUTO_TABLE_OPTIMIZATION`, `STV_NODE_STORAGE_CAPACITY`, and `SVV_TABLE_INFO`, some of which Redshift makes visible only to superusers. Grant the connecting database user the `sys:monitor` role, which is the narrowest grant that covers them:
```
grayhemp: "some of which Redshift makes visible only to superusers" -> "which require superuser or monitor role"

src/redshift-mcp-server/awslabs/redshift_mcp_server/server.py
```diff
@@ -657,11 +657,12 @@ async def review_cluster_tool(
     - Ensure your AWS credentials are properly configured (via AWS_PROFILE or default credentials).
     - The cluster must be available and accessible.
     - Required IAM permissions: redshift-data:ExecuteStatement, redshift-data:DescribeStatement, redshift-data:GetStatementResult.
-    - The connected user must have superuser (CREATEUSER) privileges to access the required system views.
-      If it does not, the review fails fast with "Review requires superuser (CREATEUSER) privileges"
-      (for example "permission denied for relation sys_auto_table_optimization"). This is by design -
-      an expected signal, not a tool defect - so the review never returns partial or misleading
-      results. Run the review as a superuser to get a complete assessment.
+    - The connected database user must be able to read Redshift system views, some of
```
grayhemp: "some of which are superuser-visible" -> which require superuser or monitor role.

grayhemp: "A superuser must grant" -> "if current user is not superuser, it must be granted"

src/redshift-mcp-server/awslabs/redshift_mcp_server/review/executor.py
```diff
@@ -94,7 +96,11 @@ async def review_cluster(
             logger.error('Review query {} failed: {}', query_name, str(e))
             if 'permission denied' in str(e).lower():
                 raise Exception(
-                    f'Review requires superuser (CREATEUSER) privileges. '
+                    f'Review requires read access to Redshift system views. Ask a superuser '
```
grayhemp: "read access to Redshift system views." -> "superuser or monitor role."

grayhemp: "Ask a superuser (for example the cluster admin user)" -> "Request administrator"

grayhemp (changes requested):
Few more wording fixes and one removal needed.

grayhemp:
## Branch E2E

E2E run via Redshift MCP Server on real clusters (us-east-1). Branch changes under test: node type from DescribeClusters inlined into diagnostic SQL, Spectrum QMR signal 
gated on recent S3 activity, sys:monitor permission message.

| # | Scenario | Result |
|---|---|---|
| 1 | RG (Graviton) cluster redshift-cluster-appsec (rg.large) reviewed → no NodeDetails findings, so no bogus "migrate to RG" | PASS |
| 2 | Old capacity heuristic cross-check: stv_partitions.capacity=195897, part_count=1 → old CASE yield ra3.large → pre-fix code would flag RG cluster for RG migration | PASS (regression confirmed fixed) |
| 3 | Legacy node type redshift-cluster-marketing (ra3.large) → REC_001 "should consider migrating to RG instances", 2 nodes | PASS |
| 4 | Rendered NodeDetails CTE run direct → node_type='rg.large' inlined, exactly 1 row per node (simplified stv_partitions subquery no longer duplicates by capacity/part_count) | PASS |
| 5 | Serverless redshift-workgroup-etl → 8 signals, NodeDetails/WLMConfig/WorkloadEvaluation skipped, no render attempt with null node type | PASS |
| 6 | No sys:monitor on redshift-cluster-analytics → fail-fast: "Review requires superuser or sys:monitor access… GRANT ROLE sys:monitor TO \"<database_user>\"" | PASS |
| 7 | Spectrum gate positive: appsec 17 S3 queries / marketing 79 in last 7d → REC_019 spectrum signal fires on both | PASS |
| 8 | Spectrum gate predicate negative: same SYS_EXTERNAL_QUERY_DETAIL filter over 1-min window → 0 rows → EXISTS false → branch suppressed | PASS (predicate level) |
| 9 | str.format render safety: all 11 provisioned + 8 serverless queries rendered and executed, no brace/format errors | PASS |

Not reproducible live:
- Review-level Spectrum suppression (scenario 8 end-to-end) — both permitted provisioned clusters have Spectrum activity in the 7-day window; predicate verified instead.
- node_type=None → 'unknown' fallback for provisioned clusters — API reports node type for all clusters here; unit-test-only coverage.
- dc2.large cluster (redshift-cluster-analytics) node signals — blocked by missing sys:monitor grant (that block is scenario 6).


## Full E2E

| Tool | Result | Evidence |
|---|---|---|
| list_clusters | PASS | 4 targets: 3 provisioned + 1 serverless. Serverless correctly nulls node_type/number_of_nodes/master_username. |
| list_databases | PASS both types | dev, sample_data_dev (local, Snapshot Isolation), awsdatacatalog (auto mounted catalog). Identical shape provisioned vs serverless. |
| list_schemas | PASS both types | tickit/public/information_schema. ACL differs by auth: awsuser=UCDA (provisioned) vs "IAMR:Admin"=UCDA (serverless). Bad db → engine FATAL: database ... does not exist surfaced. |
| list_tables | PASS both types | 7 TABLE objects in tickit, identical on both targets, ACL per-principal. |
| list_columns | PASS both types | venue 5 cols, matching types/precision/scale/nullability on both. |
| execute_query | PASS | Aggregate+GROUP BY on real data, all data types (int/decimal/varchar/bool/NULL/date/timestamp), CTE, SHOW, parenthesized SELECT, SET (empty result set, no HasResultSet fetch). |
| review_cluster | PASS both types | Provisioned marketing: 11 signals, 15 findings, 10 recs, provisioned-only sections run (NodeDetails, WLMConfig, WorkloadEvaluation, CopyPerformance). Serverless: 8 signals, 0 findings, ServerlessScaling substituted, provisioned-only skipped. analytics: fail-fast with sys:monitor remediation text — no 
partial results. |

## Protections

Read-only deny-list — 15 statements rejected pre-flight: COMMIT, ROLLBACK TRANSACTION, ;COMMIT, /* block */ end work, ABORT TRANSACTION, BEGIN TRANSACTION, START TRANSACTION, 
truncate"...", UNLOAD, GRANT, REVOKE, ANALYZE, VACUUM, CALL, COMMENT ON. Case, leading semicolon, and comment trivia did not hide keywords.

Transaction breaker — non-deny-listed writes pass guard and get blocked by engine: UPDATE, CREATE TABLE, WITH cte ... UPDATE all returned ERROR: transaction is read-only. Post
-check confirmed probe table absent (probe_table_exists = 0), so BEGIN READ ONLY ... ROLLBACK persisted nothing.

Failed user SQL — planner error (relation does not exist), runtime error (Divide by zero), and multi-statement/unparseable rejections all surfaced verbatim. After each failure
the next query succeeded on the same backend pid (1073988036), proving ROLLBACK closed the transaction and left no aborted-transaction state. Distinct pid per target (
1073955099 serverless) confirms per-cluster:database session isolation.

No false positives — '; COMMIT;' literal, $$ ; COMMIT ; $$ dollar-quoted body, and AS grant alias all executed. Nested-comment payload /* a /* b */ SELECT 1 FROM */ SELECT 99 
returned 99 — guard classification matches what the engine runs, no comment desync.

grayhemp (approved):
LGTM. Approved.

saeedma8 (approved):
Reviewed. No further issue.

PR #4368, opened 2026-07-28
fix(amazon-bedrock-agentcore-mcp-server): expire cached browser clients to pick up refreshed credentials

kevin-orellana:
## Problem
`get_browser_client()` caches `BrowserClient` instances per region at module level and never invalidates them. A BrowserClient wraps boto3 clients whose credential providers resolve and cache credentials at construction time — so after credential rotation (e.g. `ada credentials update`, IAM role rotation), the cached client keeps using stale tokens until the process restarts.

## Fix
TTL-based invalidation (default 300s, configurable via `AGENTCORE_BROWSER_CLIENT_TTL_SECONDS`):
- `_get_ttl_seconds()` resolves the TTL from the environment, warns and falls back to the default on invalid input; a non-positive TTL disables caching entirely
- `get_browser_client()` discards and recreates entries older than the TTL (monotonic clock)

## Testing
- `tests/browser/test_unit_browser_client.py`: 11 passed, including the regression guard `test_recreates_client_after_ttl_expiry` (monotonic mocked past TTL asserts a second construction) and invalid-TTL fallback cases

Draft for review.

### DIFF
--- a/src/amazon-bedrock-agentcore-mcp-server/CHANGELOG.md
+++ b/src/amazon-bedrock-agentcore-mcp-server/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+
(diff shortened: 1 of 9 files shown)


kevin-orellana:
Superseded by a clean rebase onto current main — the original branch was accidentally based on unmerged local work (invoke_agent_runtime_command), pulling unrelated runtime files into the diff. Replacement PR has only the browser-client cache fix.

Issue #4370, opened 2026-07-28
Watch: Dependabot group-by:dependency-name not collapsing cross-directory PRs (upstream dependabot-core#14286)


scottschreckengaust:
### Summary

Track the upstream Dependabot regression that prevents `group-by: dependency-name` from collapsing version-update PRs across directories, so we can drop our interim throttles once it's fixed.

### Background

PR #4267 added `group-by: dependency-name` to the `uv`, `npm`, and `docker` groups in `.github/dependabot.yml` to collapse a single dependency's bump across all package directories into **one** PR (instead of one PR per directory). The config is valid and correct per GitHub's docs, but **it is not working** — Dependabot still opens one PR per directory.

Evidence (all opened well after #4267 merged 2026-07-16):

- `gitpython 3.1.54` → #4357, #4358, #4359 (three PRs; #4358/#4359 are byte-identical `3.1.44 → 3.1.54` bumps that a working group-by would have merged into one)
- `setuptools 80.10.2 → 83.0.0` → #4303, #4305, #4306, #4309, #4310 (five PRs)
- `pillow 12.1.1 → 12.3.0` → #4292, #4293, #4294, #4295 (four PRs, identical from→to versions)

The tell is the PR **title**: ours read `bump <dep> ... in /<dir>` — Dependabot's *individual/ungrouped* format. A working group-by yields `bump the <group>/<dep> group across N directories` (which we do see for ecosystems that don't use `group-by`, e.g. #4339 "the pip group across 2 directories", #4327 "the github-actions-version-updates group").

### Root cause (upstream, not our config)

- `group-by: dependency-name` is a valid schema key and went **default-on** when its feature flag was removed in dependabot/dependabot-core#14292 (merged 2026-02-27).
- Open upstream bug **dependabot/dependabot-core#14286** — *"group-by: individual PRs are now created for each dependency within a group, not for the whole group"* — reproduces our exact symptom. Notably it repros with **explicitly enumerated** directories, not just a `**/*` glob, so rewriting our `directories: ["**/*"]` to an explicit list is **not** guaranteed to fix it.

### Interim mitigation (this repo)

Since the feature we rely on is broken upstream, PR #4371 adds throttles that don't depend on it, to the `uv`/`npm`/`docker` (and `github-actions`) entries:

- `open-pull-requests-limit: 10` — caps the open queue (default is only 5; a large release like the earlier `mcp` bump produced ~57 PRs — see #4258)
[...]

PR #4371, opened 2026-07-28
chore(deps): throttle Dependabot PR volume while cross-directory grouping is broken (upstream #14286)

scottschreckengaust:
Relates to #4370 · follow-up to #4267

## Summary

`group-by: dependency-name` (added in #4267) is **not** collapsing version-update PRs across directories — Dependabot still opens one PR per directory per dependency. The config is valid and correct per GitHub's docs; the cause is an upstream regression, [dependabot/dependabot-core#14286](https://github.com/dependabot/dependabot-core/issues/14286) (the feature went default-on when its flag was removed in dependabot/dependabot-core#14292). It reproduces even with explicitly enumerated directories, so this is not something we can fix from our config.

The open Dependabot queue is dominated by **security** updates (~1,928 open alerts, ~1,872 pip). That matters because the throttles below (`open-pull-requests-limit`, `cooldown`) apply to **version updates only** — [security updates are exempt from both](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file). So the security fan-out is bounded separately, by grouping.

Recent version-update fan-out, all opened after #4267 merged:

- `gitpython 3.1.54` → #4357, #4358, #4359 (#4358/#4359 are identical `3.1.44 → 3.1.54` bumps)
- `setuptools 80.10.2 → 83.0.0` → #4303, #4305, #4306, #4309, #4310
- `pillow 12.1.1 → 12.3.0` → #4292, #4293, #4294, #4295

### Changes

`.github/dependabot.yml` — for the four update entries (`github-actions`, `uv`, `npm`, `docker`):

- **`applies-to: security-updates` group per ecosystem** — collapses each directory's security bumps into a single PR. `groups` defaults to `applies-to: version-updates`, so the existing grouping never touched security updates; this addresses the majority of the queue and is independent of #14286.
- **`cooldown: { default-days: 7 }`** — delays a version-update PR until a release has been out 7 days. Dependabot already applies a 3-day cooldown by default, so this is a **+4-day** increment (not new behavior), smoothing back-to-back releases. Does not apply to security updates.
- **Removed `open-pull-requests-limit: 10`.** The default is **5**, so setting `10` *raised* the ceiling rather than throttling. Reverted to the default; it can be set **below 5** later if version-update volume needs a hard cap. (The cross-directory duplication this PR targets is one-PR-per-directory, which this limit — per resolved directory — does not bound anyway.)

Deliberately **not** changed:

- `schedule.interval` stays `weekly` (per request — no move to `monthly`).
- `group-by: dependency-name` on the version-update groups is left in place so cross-directory grouping resumes automatically once the upstream fix ships. The security groups intentionally omit `group-by` (that cross-directory path is the feature broken by #14286).
- The existing backlog is not bulk-closed here (handled manually).

### User experience

**Before:** every dependency release can fan out to one PR per directory, unbounded; the security backlog fans out per-advisory-per-directory.
**After:** each directory's security bumps collapse into one PR, and a 7-day cooldown smooths version-update churn — a more reviewable queue — without waiting on the upstream grouping fix.

### Validation

```
npx @bugron/validate-dependabot-yaml .github/dependabot.yml   # exit 0
```
Also checked against the SchemaStore `dependabot-2.0.json` schema (`cooldown.default-days`: integer 1–90; `groups.*.applies-to`: `version-updates` | `security-updates`).

## Follow-up

#4370 tracks the upstream bug; when dependabot/dependabot-core#14286 is fixed we can confirm cross-directory grouping works (title becomes `bump the <group>/<dep> group across N directories`) and relax these settings if desired.

## Checklist

* [x] Changes have been tested (validated with `@bugron/validate-dependabot-yaml`, exit 0)
* [x] Changes are documented (inline comments + this description + #4370)

**RFC issue number**:

## Acknowledgment


laithalsaadoon (changes requested):
Thanks for the thorough writeup, and for filing #4370 to track the upstream cause. The YAML is valid, the diagnosis of dependabot/dependabot-core#14286 holds, and leaving `group-by: dependency-name` in place so grouping resumes automatically is the right call.

Requesting changes on the two throttles themselves. Verified against the Dependabot options reference, `dependabot-core`, and this repo's own Dependabot history (lookback: all 1,392 Dependabot PRs and 1,926 open alerts, checked 2026-07-28).

## 1. `open-pull-requests-limit: 10` raises the ceiling, because the default is 5

The limit is scoped per resolved directory, not per ecosystem. Two observations from this repo settle it:

- `npm` resolves to exactly one directory (`/docusaurus` holds the only `package.json`), and it sits at exactly **5** open version-update PRs (#4322-#4326) while `typescript`, `@docusaurus/types`, and `@docusaurus/tsconfig` all have newer releases with no PR raised. The cap is live and binding at the documented default of 5.
- On 2026-07-16, under `directories: ["**/*"]` with no limit configured, the `mcp` bump to 1.28.1 held **58 PRs open simultaneously**. A per-ecosystem cap would have stopped at 5.

Upstream dependabot/dependabot-core#10395 is the same experiment run deliberately: 26 directories, one shared base-image bump, `open-pull-requests-limit: 10` set, 26 PRs delivered. It is still open as a feature request, so nothing has changed since.

Net effect of this diff: for `uv`, `npm`, and `docker`, the per-directory ceiling goes from 5 to 10. For `github-actions` the value is inert, since `patterns: ["*"]` already collapses that entry to a single PR.

If bounding version updates is the goal, the value has to go **below** 5.

## 2. The fan-out cited as motivation is security updates, which both options exempt

Every example in the description targets an advisory's `first_patched_version` exactly: pillow 12.3.0, setuptools 83.0.0, gitpython 3.1.54, pyasn1 0.6.4. 20 of the 28 currently open Dependabot PRs match an open alert in that same directory.

From the options reference:

> *Security update* pull requests are not subject to this limit and do not count toward it. There is no limit on the number of open pull requests for security updates.

Cooldown is exempted the same way in core, in `update_all_versions.rb`:

```ruby
update_cooldown: job.security_updates_only? ? nil : job.cooldown
```

The 57-PR `mcp` event in #4258 was half of a 114-PR day: 57 PRs to 1.27.2 (the patched version for the 177 `mcp` advisories opened that morning) and 57 to 1.28.1 (version updates). Neither new option would have prevented the security half.

## 3. `cooldown: default-days: 7` adds 4 days, not 7

A 3-day default cooldown already applies to version updates when `cooldown` is unset. Worth stating in the description, since "delays a PR until a release has been out 7 days" reads as new behavior rather than an increment. This part is harmless and mildly useful; it delays rather than skips, so churny releases still land once the window passes.

## Suggested change

The repo carries 1,926 open alerts (1,872 of them pip). `groups` defaults to `applies-to: version-updates`, so none of the current grouping touches any of them. Adding a second group per ecosystem scoped to security updates addresses the majority of the queue:

```yaml
    groups:
      uv-version-updates:
        group-by: dependency-name
        patterns:
          - "*"
      uv-security-updates:
        applies-to: security-updates
        patterns:
          - "*"
```

That is independent of #14286 and does not depend on the broken cross-directory grouping.

Happy to re-review once the limit is either dropped or lowered below 5.

scottschreckengaust:
Verified all three points against the [Dependabot options reference](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file) and this repo's live data (1,928 open alerts, 1,872 pip). They hold. Pushed 4c0aa5c:

**1. `open-pull-requests-limit: 10` raised the ceiling.** Confirmed — default is 5, so the value throttled nothing. Removed the line entirely (reverts to the default 5) rather than hard-coding a number, so it can be dropped below 5 later if version-update volume needs a real cap. Also confirmed the limit is per resolved directory, and the #14286 fan-out is one-PR-per-directory, so this lever never bound the actual problem regardless of value.

**2. The fan-out is security updates, exempt from both throttles.** Confirmed — limit and cooldown are version-updates-only. Spot-checked the cited bumps: they land on or above each advisory's `first_patched_version` (gitpython → 3.1.54 vs floor 3.1.52/3.1.53; pillow → 12.3.0 vs 12.2.0/12.3.0). Added an `applies-to: security-updates` group per ecosystem, as suggested — no `group-by`, since cross-directory grouping is exactly what #14286 breaks; these collapse each directory's security bumps into one PR, which is independent of the upstream regression.

**3. cooldown adds 4 days, not 7.** Confirmed — a 3-day default cooldown already applies to version updates when unset, so `default-days: 7` is a +4 increment. Kept it (delays rather than skips) and corrected the description to state it as an increment.

Net diff now: security-update groups added (uv/npm/docker + github-actions), `open-pull-requests-limit` removed, cooldown unchanged, `group-by: dependency-name` left on the version-update groups so cross-directory grouping resumes when #14286 ships. `@bugron/validate-dependabot-yaml` exits 0. Description updated to match. Ready for re-review.

laithalsaadoon (approved):
All three points from the previous review are addressed as of 4c0aa5cd. Verified against the head branch and the Dependabot options reference (checked 2026-08-03):

1. **The limit bump is gone.** The head file has no `open-pull-requests-limit` line, and the net diff against main no longer touches it, so all four ecosystems sit at the documented default of 5 for version updates.
2. **Security-update groups added per ecosystem**, each with `applies-to: security-updates`, `patterns: ["*"]`, and no `group-by`, matching the suggested shape. This collapses each directory's security bumps into one PR, addresses the majority of the open queue (1.9k alerts), and is independent of dependabot-core#14286. Defining these groups also enables grouped security updates for the repo automatically, so no settings change is needed.
3. **Cooldown framed as an increment.** The inline comment and description now state the built-in 3-day default, so `default-days: 7` reads as the +4 days it is.

The inline comments match documented behavior (cooldown and the open-PR limit apply to version updates only; the security fan-out is bounded by the new groups), GitHub's own dependabot.yml check passes on the head commit, and the PR touches only `.github/dependabot.yml`. Leaving `group-by: dependency-name` in place so cross-directory grouping resumes when the upstream fix ships is still the right call.

Approving.

arnewouters (approved):

PR #4373, opened 2026-07-29
fix(aurora-dsql-mcp-server): raise dollar-quote timing threshold to fix flaky CI

### LINKED ISSUE #4366: aurora-dsql-mcp-server: flaky timing test test_dollar_quote_scan_is_linear (2.04s vs 2.0s threshold)
### Describe the bug

`tests/test_readonly_enforcement.py::TestReadonlyEnforcement::test_dollar_quote_scan_is_linear` is a **flaky wall-clock timing test**. It feeds a 500k-`$` payload through `detect_mutating_keywords` / `check_sql_injection_risk` and asserts the run finishes in under 2 seconds:

```python
# src/aurora-dsql-mcp-server/tests/test_readonly_enforcement.py:399
assert elapsed < 2.0, f'dollar-heavy input too slow ({elapsed:.2f}s) — possible O(n^2)'
```

On a loaded CI runner the absolute wall-clock time occasionally creeps just over the threshold and fails the whole `Build aurora-dsql-mcp-server` job, even though the code is correct.

### Observed failure

From the `main` post-merge run of #4360 ([job log](https://github.com/awslabs/mcp/actions/runs/30384985162/job/90361979166)):

```
AssertionError: dollar-heavy input too slow (2.04s) — possible O(n^2)
assert 2.042716128000002 < 2.0
tests/test_readonly_enforcement.py:399: AssertionError
1 failed, 237 passed in 7.03s
```

The failure (2.04s vs. a 2.0s threshold — ~2% over) is unrelated to that PR's dependency change; it is pure timing sensitivity. This is at least the second sighting of this same test tripping right at ~2.04s.

### Why it's a false positive

The regression this test guards against — the old O(n²) `sql[i:]` slicing in `_end_of_dollar_quote` — takes *tens of seconds* on a 500k-`$` string. The linear implementation runs in well under a second, so a 2.0s bar leaves almost no headroom against runner jitter while providing little extra protection.

Measured locally on the linear implementation:

- `test_dollar_quote_scan_is_linear` elapsed: **~0.59–0.61s** (≈3.3× margin under the 2.0s bar)
- Full suite (238 tests), 5 consecutive runs: **238 passed** every time

### Expected behavior

The test should reliably pass on shared/loaded CI runners while still catching a true O(n²) regression.

### Possible solutions

1. **Raise the absolute threshold** to something like `< 5.0s` (or higher). The O(n²) version takes tens of seconds at 500k `$`, so detection still works and jitter no longer trips it. Smallest, lowest-risk change.
2. **Assert relative slowdown** instead of absolute wall-clock — e.g. compare the `$`-heavy payload's time against a baseline linear-size run and require the ratio stay within a bound. More robust to hardware variance, slightly more code.
3. **Bound the algorithm directly** (e.g. `pytest-timeout` or an operation/step counter) rather than measuring seconds, decoupling correctness from machine speed.

Recommend option 1 as the immediate fix; option 2/3 as a more durable follow-up if this class of timing test recurs elsewhere.

### Additional context

- Affected file: `src/aurora-dsql-mcp-server/tests/test_readonly_enforcement.py:383-399`
- Sibling ReDoS guards in the same file (`elapsed < 1.0`) have far larger margins (measured ~2.7–27ms) and are not currently flaky, but share the same absolute-wall-clock design.
- Discovered while investigating the post-merge CI failure on #4360; the merge itself was not the cause.

### Server

aurora-dsql-mcp-server

sergioperezcheco:
Raise the timing threshold from 2.0s to 5.0s in test_dollar_quote_scan_is_linear to stop flaky CI failures on loaded runners. The linear implementation runs in about 0.6s locally but occasionally hits 2.04s on shared CI runners, failing the whole build. The O(n-squared) regression this test catches takes 30+ seconds at this payload size, so 5.0s still detects it reliably.

Closes #4366 .

### DIFF
--- a/src/aurora-dsql-mcp-server/tests/test_readonly_enforcement.py
+++ b/src/aurora-dsql-mcp-server/tests/test_readonly_enforcement.py
@@ -391,12 +391,15 @@ class TestReadonlyEnforcement:
 
         # 500k `$`: linear cost is well under 1s; the reverted O(n^2) slice version
         # takes tens of seconds at this size, so this reliably catches a regression.
+        # The threshold is set generously (5s) so that a loaded CI runner does not
+        # trip a false positive while still detecting a genuine O(n^2) regression
+        # (which takes 30+ seconds at this payload size).
         payload = '$' * 500_000
         start = time.perf_counter()
         detect_mutating_keywords(payload)
         check_sql_injection_risk(payload)
         elapsed = time.perf_counter() - start
-        assert elapsed < 2.0, f'dollar-heavy input too slow ({elapsed:.2f}s) — possible O(n^2)'
+        assert elapsed < 5.0, f'dollar-heavy input too slow ({elapsed:.2f}s) — possible O(n^2)'
 
     def test_double_quoted_identifier_apostrophe_does_not_bypass(self):
         """An apostrophe inside a double-quoted identifier must not desync scanning.


scottschreckengaust:
Thank you @sergioperezcheco, would you please update with a contribution statement to your PR please? Thank you!

sergioperezcheco:
Done — the contribution statement was already in the PR description, but I've confirmed it here as well: by submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the project license.

spencercorwin (approved):

anwesham-lab (approved):

PR #4384, opened 2026-07-31
fix(dynamodb-mcp-server): validate and escape names in CDK generator

LeeroyHannigan:
## Description

Names from `dynamodb_data_model.json` were interpolated into the generated CDK stack without validation or escaping, so a name containing a quote could alter the generated TypeScript.

Two changes, split by what the DynamoDB API itself permits:

1. **`TableName` and `IndexName`** are validated in `models.py` at parse time against the character set and length the DynamoDB API enforces (`[a-zA-Z0-9_.-]`, 3 to 255 characters, per the `TableName` and `IndexName` service model shapes). These two values reach TypeScript *identifier* positions in the template (method and variable names), where escaping cannot be applied, so they have to be validated rather than escaped.
2. **Attribute names** (partition key, sort key, GSI keys, TTL, `NonKeyAttributes`) carry no character restriction in the DynamoDB API, so they cannot be rejected. Every string-literal interpolation in `stack.ts.j2` now uses the `tojson` filter, which the template already used for `nonKeyAttributes`. Attribute names are length-checked (1 to 255) to match the service model.

Also in `_to_camel_case`: `.` is now treated as a word separator, and a name that starts with a digit or consists only of separators is prefixed, so any name DynamoDB accepts produces a legal TypeScript identifier. `my.table` previously generated `const my.tableTable = ...`, which does not compile.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Testing

Added to `tests/cdk_generator/`:

* `TestTableNameValidation` / `TestIndexNameValidation`: quote-bearing and out-of-range names rejected with a clear field path; every name DynamoDB accepts still parses.
* `TestAttributeNameValidation`: quote-bearing attribute and TTL names are accepted (they are valid in DynamoDB) and length bounds enforced.
* `TestTemplateEscaping`: renders the real template and asserts each interpolated value round-trips through `json.loads`, proving it stayed inside its string literal. Covers partition key, sort key, TTL, GSI keys and `nonKeyAttributes`.
* `TestIdentifierSafety`: dot separators, digit-leading names, separator-only names, and an end-to-end check that identifiers rendered for `2024.order-events_v2` are legal.

159 tests pass in `tests/cdk_generator/`. `ruff format --check` and `ruff check` are clean.

## Note on behaviour change

The `TableName` **and `IndexName`** minimum length of 3 comes from the DynamoDB service model (the `TableName` and `IndexName` shapes both specify `min: 3`). In this generator `table_name` is a logical name (CDK generates the physical table name), so a 1 or 2 character name was previously accepted. It is now rejected, matching the API contract the input document is modelled on.

`IndexName` is passed straight through to the `TableV2` construct and therefore reaches DynamoDB, so a shorter name such as `g1` would be rejected at deploy time regardless. Validating it at parse time surfaces that as a clear error instead of a CloudFormation failure, but it does mean a model file using a 1 or 2 character logical index name now fails earlier than before.


willykev (dismissed):

scottschreckengaust (commented):
## Automated code review

The core escaping approach here is **sound** — Jinja's `tojson` escapes control chars, quotes, `<`, `>`, `&`, `'`, and the U+2028/U+2029 line separators, and identifier positions are sanitized through `_to_camel_case`/`_to_pascal_case` (output restricted to `[A-Za-z0-9]`). No residual TypeScript-injection vector was found. The findings below are validation/behavior/cleanup items, offered as a non-blocking comment.

---

### 1. `re.match` vs `re.fullmatch` — trailing-newline bypass (🐞 most defensible)
`src/dynamodb-mcp-server/awslabs/dynamodb_mcp_server/cdk_generator/models.py:178`

`_validate_ddb_name` uses `re.match`, so the `$` anchor matches *just before* a trailing newline. A name like `'Users\n'` passes `DDB_NAME_PATTERN.match()` even though it contains a character DynamoDB forbids, defeating the validator's purpose.

The PR's own test only covers `\n` in the *middle* of a name (`'Users\nTable'`), so the trailing-newline gap is untested.

**Suggested fix:** use `DDB_NAME_PATTERN.fullmatch(value)`.

### 2. Undocumented `IndexName` minimum-length regression (⚠️)
`models.py:553`

`MIN_NAME_LENGTH = 3` is now enforced on `IndexName` (a *logical* name) as well as `TableName`. A previously-valid short GSI name like `'g1'` or `'id'` now raises `ValueError: IndexName must be between 3 and 255 characters`. The PR description's "Note on behaviour change" only calls out the `TableName` change, not `IndexName` — a silent regression for existing inputs. Worth documenting and/or reconsidering whether logical index names should share the DynamoDB minimum.

### 3. Widened collision surface from prefix fallbacks (⚠️)
`src/dynamodb-mcp-server/awslabs/dynamodb_mcp_server/cdk_generator/generator.py:199`

The digit-leading and separator-only fallbacks funnel distinct DynamoDB-valid names onto the same identifier. E.g. `'123'` and `'_123'` both map through `_to_camel_case` to `'table123'`, so `_check_table_name_collisions` raises `CdkGeneratorError` for two legal, distinct DynamoDB table names. No test covers cross-name collisions introduced by prefixing.

### 4. Redundant `_validate_attribute_name` (🧹 cleanup)
`models.py:174`

The min-length-1 branch can never fire for values that already passed `_validate_string_field` (which rejects empty/whitespace strings), so the method effectively reduces to an upper-bound (255) length check. An inline bound check or a shared length helper reused by `_validate_ddb_name` would avoid the parallel `MIN_NAME_LENGTH`/`MIN_ATTRIBUTE_NAME_LENGTH` constant pairs.

---

LeeroyHannigan:
Thanks for the review, and for confirming the escaping approach independently. Pushed `6459e856` in response.

**1. `re.match` vs `fullmatch`: agreed and fixed.** You are right, and it reproduces: `TableName='Users\n'` was accepted before, rejected now. Switched to `DDB_NAME_PATTERN.fullmatch(value)` and added `\n`, `\r` and `\t` at the end of both `TableName` and `IndexName` to the rejection cases, since the existing coverage only had `\n` mid-name. Also left a comment on the pattern so the next reader does not reintroduce `match`. Best catch of the four.

**2. `IndexName` minimum length: documented, kept at 3.** Fair that the PR description only called out `TableName`. I have added the note. Keeping the value at 3 though, because `IndexName` is not purely logical: it passes straight through to the `TableV2` construct's `indexName` and therefore reaches DynamoDB, whose `IndexName` shape specifies `min: 3`. A name like `g1` would fail at deploy time regardless, so validating at parse time converts a CloudFormation failure into a clear message. The behaviour change is real and now stated.

**3. Collision surface: pinned with a test, with one correction.** The specific pair you cite was already colliding before this PR: `_to_camel_case` treats `_` as a separator, so `'123'` and `'_123'` both mapped to `'123'` on `main` too. What the prefix genuinely introduces is a collision between `'123'` and a literal `'table123'`, which is a real new case. I do not think there is a clean fix, since the derived identifier is restricted to `[A-Za-z0-9]` and any alphabetic prefix can collide with some legal name. It is also fail-safe: `_check_table_name_collisions` raises `CdkGeneratorError` rather than emitting two methods with the same name. So I have added `test_prefixed_name_collision_is_reported` to pin that, rather than hide it. Happy to revisit if you would rather reject digit-leading names outright, though DynamoDB permits them.

**4. `_validate_attribute_name`: leaving as is.** The premise does not hold. `_validate_string_field` only checks key presence and `isinstance(value, str)`; it does not reject empty or whitespace strings, so an empty attribute name reaches `_validate_attribute_name` and the lower bound does fire. Worth double checking that one, since it is the kind of thing that reads as dead code but is not.

One ask: `willykev`'s approval predates these commits, so it may need re-confirming before this can merge. Also flagging that the branch carries a `Merge branch 'main'` commit, which some queues eject; happy to squash-linearise if that is a problem here.

scottschreckengaust (commented):
## Automated re-review (follow-up to [my 2026-08-07 review](https://github.com/awslabs/mcp/pull/4384#pullrequestreview-4884584983))

Re-reviewed at `6459e856`. **The shipped fix is correct** — I re-verified empirically that every hostile `TableName` is rejected (`Users'`, `Users"x`, `a;b`, `ev(il)`, `` q'`${x} ``, `Users\n`, `Users\n `, `ab`, 256 chars), and that all 11 string-literal interpolations in `stack.ts.j2` go through `tojson`. I found **no residual injection vector**. 164 tests pass.

### On my previous four items — you were right on both pushbacks

**#4 (`_validate_attribute_name`): your rebuttal is correct, I withdraw it.** I independently re-checked: `_validate_string_field` (`models.py:136-140`) only tests key presence and `isinstance(str)`, and `''` passes it. The `MIN_ATTRIBUTE_NAME_LENGTH` lower bound is live and is the *only* guard against a zero-length attribute name. Leaving it as-is is right. (See finding 3 below — that branch turns out to be unpinned by tests, which makes the "reads as dead code" risk you identified worse, not better.)

**#3 (collisions): your correction is also right** — `'123'`/`'_123'` collided on `main` too, since `_` was already a separator. The genuinely new case is the one you tested. Agreed there's no clean fix, and agreed that fail-safe detection plus a test is the right resolution rather than rejecting digit-leading names DynamoDB permits.

**#1** is properly fixed (verified `Users\n` and `Users\n ` both rejected; reverting `fullmatch`→`match` is caught by a test). **#2** — accepting your reasoning: `IndexName` reaching the `TableV2` construct means `min: 3` is the real API contract, so parse-time rejection converts a CloudFormation failure into a clear message. Documented is sufficient.

The new findings below are all *regression-detection* gaps, not live defects.

---

### 1. `indexName` is the one interpolation site with no escaping test (Important)
`templates/stack.ts.j2:31`

**Empirically verified:** I deleted ` | tojson` from this line in a scratch copy and **all 164 tests still passed.** It is the only one of the 11 escaped sites with no test coverage.

Cause: `test_gsi_key_breakout_is_escaped` (`test_generator.py:587`) builds a GSI but with a clean `IndexName: 'by-breakout'`, so nothing observes that literal; and `test_no_single_quoted_literals_remain` (`test_generator.py:622`) uses `sample_data_model`, which has **no GSI at all** — so the `indexName:`, `nonKeyAttributes:` and GSI `name:` lines are never seen by the one test asserting double-quoted output.

Defense-in-depth only (validation blocks quotes in `IndexName` today), but worth one line:
```python
assert 'indexName: "by-breakout"' in rendered
```

### 2. Widening `DDB_NAME_PATTERN` is invisible to the suite — and validation is the *sole* defense at identifier positions (Important)
`models.py:42`

**Empirically verified:** adding `'` to the charset leaves **all 164 tests passing**, and the generator then emits non-compiling TypeScript:
```
private createUsers'Table(): void {
const users'Table = ...
```
The tests assert *"a quote is rejected"* via the error **message**, never *"the charset is exactly this set"* — so the pattern's contract is unpinned.

This matters more than it first appears. `_to_camel_case` (`generator.py:180-201`) only replaces `-`/`_`/`.` with spaces and title-cases; it has **no allowlist filter**, so it passes quotes, `;`, and `${}` straight through (verified: `_to_camel_case("Users'") == "users'"`, `_to_camel_case('a;b') == 'a;b'`). So `DDB_NAME_PATTERN` is not one layer of two at identifier positions — **it is the only layer**. That argues for pinning it directly:
```python
assert not DDB_NAME_PATTERN.fullmatch("a'b")
assert generator._to_camel_case('Users') == 'users'   # output stays identifier-legal
```

### 3. Attribute min-length check is unpinned (Minor)
`models.py:210`, tests at `test_models.py:2139/2162/2175`

**Empirically verified:** removing `MIN_ATTRIBUTE_NAME_LENGTH <=` leaves the suite green, and `AttributeName: ''` and an empty TTL attribute are then both **accepted**. The three attribute-length tests only exercise the 256-char upper bound and match the shared message `must be between 1 and 255`. Worth adding `AttributeName: ''` and TTL `''` rejection cases. This is exactly the branch you correctly defended above — pinning it with a test is what stops the next reviewer (or a future me) from deleting it as apparent dead code.

### 4. A rejected name leaves an orphaned `cdk/` directory (Minor, pre-existing)
`generator.py:86-92`

`cdk_dir.mkdir()` and `_run_cdk_init()` run *before* `_parse_data_model()`, so a validation failure leaves a populated `cdk/` behind; the retry then hits `CDK directory already exists` and needs a manual `rm -rf`.

To be clear about attribution: **this ordering is unchanged from `main`** — not introduced here. But `main` had zero name validation, so this PR is what makes the trap reachable in normal use (including via the new min-length-3 `IndexName` rule). Moving `_parse_data_model` + `_check_table_name_collisions` above `cdk_dir.mkdir()` fixes it and also skips a pointless ~2min `npx cdk init` on invalid input. Reasonable to defer to a follow-up.

### 5. `Projection` of a non-dict raises `TypeError`, not `ValueError` (Minor, pre-existing)
`models.py:479-505` — `Projection: "ALL"` → `TypeError: string indices must be integers`. Inconsistent with the `_validate_is_object` pattern used elsewhere. Untouched by this PR; noting only because it is adjacent to the validation being hardened.

---

### What's good

The threat model is split deliberately rather than accidentally: validate where escaping is impossible (identifiers), escape where rejection is impossible (attribute names, which DynamoDB does not restrict). That's the right analysis, and it's the part most such fixes get wrong. `projectionType` and `AttributeType` are the only other unescaped interpolations and both are allowlist-constrained enums.

`_decoded_literals` (`test_generator.py:505`) is a genuinely strong helper — `json.loads` on each literal means raw single-quoted output raises `JSONDecodeError` rather than passing, and each escaping test pairs a negative (`BREAKOUT not in rendered`) with a positive round-trip. Together those pin exact behavior; either alone would be weak. **9 of 10 `tojson` deletions I attempted were caught**, and no vacuous `pytest.raises(Exception)` assertions — every one uses a specific `match=` with a field path. Boundary coverage is exact on both sides (2/3, 255/256, 0/1), not just mid-range. `test_prefixed_name_collision_is_reported` catches a subtle interaction, and documenting the `fullmatch` rationale at `models.py:42` *and* pinning it with a test is exactly the pairing that stops a comment from rotting.

### Verdict

**Comment, not Request changes.** The security fix is sound and I could not break it. Findings 1–3 are one-to-three-line test additions that close regression-detection gaps in code that is currently correct; 4 and 5 are pre-existing and fine to defer. Items 1 and 2 are the ones I'd genuinely like before merge — not because the code is wrong, but because a future edit could silently make it wrong.

willykev (approved):

willykev (approved):

src/dynamodb-mcp-server/awslabs/dynamodb_mcp_server/cdk_generator/templates/stack.ts.j2
```diff
@@ -13,26 +13,26 @@ export class {{ stack_class_name }} extends cdk.Stack {
...
+        name: {{ table.sort_key.name | tojson }},
         type: dynamodb.AttributeType.{{ table.sort_key.to_cdk_type() }},
       },
 {% endif %}
 {% if table.global_secondary_indexes %}
       globalSecondaryIndexes: [
 {% for gsi in table.global_secondary_indexes %}
         {
-          indexName: '{{ gsi.index_name }}',
+          indexName: {{ gsi.index_name | tojson }},
```
scottschreckengaust, line 31: **Important (raised in round 2, still open) — the one interpolation site with no escaping test.**

Re-verified by mutation at `8946c4b5`: I removed ` | tojson` from this line, and separately reverted it to the exact pre-fix form `indexName: '{{ gsi.index_name }}',`. **Both times all 164 tests passed.** `grep -rn indexName tests/cdk_generator/` returns zero hits — no test anywhere asserts on this line's output. It is the only one of the ~11 escaped sites in that position; every other site was caught by both mutations.

Why the existing tests miss it: `test_gsi_key_breakout_is_escaped` builds a GSI but with a clean `IndexName: 'by-breakout'`, so nothing observes that literal; and `test_no_single_quoted_literals_remain` uses `sample_data_model`, which has **no GSI at all** — so this line, the GSI `name:` lines, and `nonKeyAttributes:` are never seen by the one test asserting double-quoted output.

This is defense-in-depth today, since `_validate_ddb_name` rejects quotes in `IndexName`. But it stops being defense-in-depth the moment the pattern is widened — see my comment on `models.py:42`, where that widening is *also* undetected. Two individually-silent regressions compose into a live injection.

One line closes it — add a GSI to the model in `test_no_single_quoted_literals_remain` and assert:

```python
assert 'indexName: "by-status"' in rendered
```

src/dynamodb-mcp-server/awslabs/dynamodb_mcp_server/cdk_generator/models.py
```diff
@@ -23,6 +24,27 @@
...
+# interpolates those two values into TypeScript identifier positions (method names and
+# variable names), where escaping cannot be applied, so they must be validated here.
+#
+# Attribute names carry no character restriction in the DynamoDB API, so they cannot be
+# rejected. They reach only TypeScript string-literal positions, which the template escapes
+# with the tojson filter.
+#
+# DDB_NAME_PATTERN must be applied with fullmatch, not match: `$` also matches immediately
+# before a trailing newline, so `re.match` would accept a name like 'Users\n'.
+DDB_NAME_PATTERN = re.compile(r'^[a-zA-Z0-9_.-]+$')
```
scottschreckengaust, line 42: **Important (raised in round 2, still open) — this pattern's charset is unpinned, and it is the *sole* defense at identifier positions.**

Re-verified by mutation: widening this to `r"^[a-zA-Z0-9_.\-']+$"` leaves **all 164 tests passing**, and the generator then emits non-compiling, injectable TypeScript for `TableName: "Users'"`:

```
  private createUsers'Table(): void {
      value: users'Table.tableName,
```

The `tojson` sites correctly emit `"Users\u0027"` — but identifier positions are raw by design, and `_to_camel_case` has **no allowlist filter**: it only replaces `-`/`_`/`.` with spaces and title-cases, so it passes `'`, `;` and `${}` straight through. So this pattern is not one layer of two here; **it is the only layer.**

The current tests assert *"a quote is rejected"* via the error **message**, which passes just as happily when the charset changes underneath. Pinning the constant directly is what makes a widening edit fail immediately, rather than only when a table name happens to contain the newly-allowed character:

```python
@pytest.mark.parametrize('bad', ["'", '"', '`', ';', ' ', '$', '{', '}', '/', '\n'])
def test_pattern_rejects_charset_outliers(bad):
    assert not DDB_NAME_PATTERN.fullmatch(f'Users{bad}')
```

Credit where due: the `fullmatch` fix from round 1 *is* properly pinned — reverting it to `.match` is caught by `test_index_name_trailing_newline_rejected` and the `'Users\n'` case. This is the same treatment applied to the charset instead of the anchoring.

src/dynamodb-mcp-server/awslabs/dynamodb_mcp_server/cdk_generator/models.py
```diff
@@ -138,6 +160,60 @@ def _validate_array_field(data: dict, field_name: str, context: str) -> list:
...
+            value: The attribute name to validate
+            field_path: Full field path for error messages
+
+        Returns:
+            The validated attribute name
+
+        Raises:
+            ValueError: If the attribute name is the wrong length
+        """
+        if not MIN_ATTRIBUTE_NAME_LENGTH <= len(value) <= MAX_ATTRIBUTE_NAME_LENGTH:
```
scottschreckengaust, line 210: **Minor (raised in round 2, still open) — this lower bound is live but untested.**

Re-verified: deleting `MIN_ATTRIBUTE_NAME_LENGTH <=` from this line leaves all 164 tests passing, and `AttributeName: ''` is then accepted, rendering `name: "",` — valid TypeScript, invalid DynamoDB. The three attribute-length tests only exercise the 256-char upper bound; there is no `AttributeName: ''` case anywhere in `test_models.py`.

This matters specifically because of our round-1 exchange: you correctly pushed back that this bound is **not** redundant — `_validate_string_field` only checks presence and `isinstance(str)`, so `''` sails through it, making this the only guard against a zero-length attribute name. I withdrew that finding, and I still agree. But an unpinned branch that a reviewer already mistook for dead code is exactly the one that gets deleted next time. A test is what makes your reasoning durable:

```python
@pytest.mark.parametrize('path', ['AttributeDefinitions', 'ttl', 'non_key'])
def test_empty_attribute_name_rejected(path):
    ...  # assert ValueError, match=r'must be between 1 and 255'
```

Related, and cheap to fold in: the upper bound is only tested from the reject side, so narrowing `MAX_ATTRIBUTE_NAME_LENGTH` to 254 would also go undetected. An accepting case at `'a' * 255` covers it.

src/dynamodb-mcp-server/awslabs/dynamodb_mcp_server/cdk_generator/models.py
```diff
@@ -443,6 +522,7 @@ def _parse_gsi_projection(cls, projection: dict, context: str) -> tuple:
                     raise ValueError(
                         f'{context}.Projection.NonKeyAttributes[{i}] must not be empty'
                     )
+                cls._validate_attribute_name(attr, f'{context}.Projection.NonKeyAttributes[{i}]')
```
scottschreckengaust, line 525: Not about this line — about the function it sits in, and **pre-existing on `main`, so out of scope for this PR.** Commenting here only because it is the highest-impact issue anywhere in this file and you are already in `_parse_gsi_projection`.

`_parse_gsi_projection` is annotated `projection: dict` but only truthiness-checked — there is no `_validate_is_object`, unlike every other container in this module. A non-dict `Projection` therefore either crashes or is **silently accepted with the wrong value**:

```
Projection='KEYS_ONLY'      -> projection_type='ALL'   <-- silently wrong
Projection='INCLUDE'        -> projection_type='ALL'   <-- silently wrong
Projection='ProjectionType' -> TypeError: string indices must be integers
Projection=42               -> TypeError: argument of type 'int' is not iterable
```

The silent rows are the real problem: `Projection: "KEYS_ONLY"` is a plausible hand-edit of the JSON, and it deploys `ALL`-projected indexes — a real cost and capacity difference — with no warning. The `TypeError` rows are merely unhelpful: through the MCP tool, `handle_exceptions` flattens them to `{'error': 'string indices must be integers'}`, with no field path or file name to locate the offending table.

No injection impact — `projection_type` stays inside its `ALL`/`KEYS_ONLY`/`INCLUDE` whitelist. And I confirmed every *other* container access in this module is properly guarded (`KeySchema[i]`, `AttributeDefinitions[i]`, `TimeToLiveSpecification`, `GlobalSecondaryIndexes[i]`, `tables[i]` all give a clean `ValueError ... must be an object`), so this is the only instance of the shape.

One line, after the `if not projection` short-circuit:

```python
cls._validate_is_object(projection, f'{context}.Projection')
```

Entirely reasonable to leave for a separate PR — it is not yours and it is not a security issue.

scottschreckengaust (approved):
## Automated review, round 3 (follow-up to [2026-08-07](https://github.com/awslabs/mcp/pull/4384#pullrequestreview-4884584983) and [2026-08-10](https://github.com/awslabs/mcp/pull/4384#pullrequestreview-4901731040))

Re-reviewed at `8946c4b5`. The only change since my last round is the merge of `main` (version bumps and lockfiles); the `cdk_generator` source and tests are byte-identical to `6459e856`, and 164 tests still pass at the new head. This time I ran three independent reviewers — security/injection, test-coverage/mutation, and error-handling — over the diff without showing them my earlier conclusions, plus my own checks. **Approving.** Rationale for the change from `Comment` is at the bottom.

### The security fix holds up under independent attack

An independent reviewer enumerated all 19 interpolations in `stack.ts.j2` and classified each: `stack_class_name` is derived from the constant `CDK_DIRECTORY_NAME = 'cdk'` and never from JSON; `to_cdk_type()` and `projectionType` are whitelist-constrained; the two `~` concatenation sites at lines 64/66 apply `tojson` **after** concatenation, which is the correct order. Everything else is either `DDB_NAME_PATTERN`-validated or `tojson`-escaped. **No bypass was found.**

Two gaps I had left open are now closed by evidence rather than argument:

* **Unicode:** brute-forcing all 1,114,112 codepoints against `DDB_NAME_PATTERN.fullmatch` accepts exactly `-.0-9A-Z_a-z` — zero non-ASCII. Python's default `re` flags do not widen `a-z`/`0-9` (only `\w`/`\d` are Unicode-aware, and neither is used). So the `str.isdigit()` call in `_to_camel_case` cannot disagree with `[0-9]`: `'١٢٣'` is rejected at validation and, even if reached, yields `table١٢٣`.
* **`KeySchema` indirection is airtight.** `_parse_key_schema:314` / `_parse_gsi_key_schema:452` require `attr_name in attr_types`, and `attr_types` is written **only** at `_parse_attribute_definitions:278`, one line after `_validate_attribute_name`. A key name that skipped length validation cannot exist in the map.
* Fuzzing ~40k valid names of length 3–5 over `aA0_.-9zZ` produced **zero** illegal camelCase or PascalCase identifiers, and zero PascalCase collisions among distinct camelCase names — confirming the camelCase-only collision check is sufficient, since uppercasing index 0 is injective on this domain.

Also: the new `.` separator does introduce collisions that `main` did not have (`my.table` and `myTable` both → `myTable`; on `main` they did not collide), but I verified it **fails safe** — `_check_table_name_collisions` raises `CdkGeneratorError` rather than emitting two identical methods. That is the right outcome.

`ruff format --check`, `ruff check`, and `pyright` are all clean; 164 tests pass.

### Test quality: 31 of 34 mutations caught

I had each `tojson` site mutated two ways — dropped entirely, and reverted to the exact pre-fix `'{{ x }}'` form — plus mutations to the pattern, the bounds, every validation call site, and each `_to_camel_case` branch. **31 of 34 were caught.** That is a strong suite, and `_decoded_literals` is the reason: I confirmed it fails hard (`JSONDecodeError`) on raw single-quoted output, and that its zero-match case returns `[]` so `assert X in _decoded_literals(...)` fails rather than passing vacuously.

The 3 survivors are the three items inline below — all previously raised, all still regression-detection gaps in code that is **correct today**.

### One new finding, pre-existing and out of scope

`_parse_gsi_projection` never calls `_validate_is_object`, so a non-dict `Projection` either crashes as `TypeError` or is **silently accepted with the wrong value**. Verified:

```
Projection='KEYS_ONLY'  -> projection_type='ALL'   <-- silently wrong
Projection='INCLUDE'    -> projection_type='ALL'   <-- silently wrong
Projection='ProjectionType' -> TypeError: string indices must be integers
```

The silent cases are worse than the crash: the user asks for `KEYS_ONLY` and gets `ALL` projected infrastructure with no warning. End-to-end through the MCP tool, `handle_exceptions` flattens the `TypeError` to `{'error': 'string indices must be integers'}` — no field path, no file. **Identical on `origin/main`; not introduced or worsened here**, and no injection impact since `projection_type` stays inside its whitelist. Every other container access is guarded. A one-line `_validate_is_object(projection, f'{context}.Projection')` fixes it, but it belongs in its own change. Flagging it because it is the highest-impact thing found anywhere in this file.

Similarly out of scope, for whoever touches this file next: `generator.py:134` reports a missing/unreadable `cwd` as `npx command not found` (verified with npx installed), and `_render_template`'s `write_text` truncates before writing, so an ENOSPC mid-write destroys a previously-good stack file and leaves invalid TypeScript on disk. Both pre-existing.

### On CHANGELOG

I checked, and **no entry is needed** — `[tool.commitizen]` in `pyproject.toml` sets `update_changelog_on_bump = true`, so the file is generated at release. (Noting this because I have asked for CHANGELOG entries on other PRs in this repo; that was the wrong call for this package.)

### Verdict: Approve

My two previous rounds said `Comment` on the strength of findings 1–3 below. I am approving now because: the shipped behavior is correct and has survived three independent attempts to break it; the remaining items are test additions that guard against *future* edits, not defects; and the highest-impact issue found in this file is pre-existing and outside the PR's scope. Withholding approval a third time over unchanged test-coverage nits would be obstructive.

I would still like items 1 and 2 — they are one-liners, and notably they **compose**: widening `DDB_NAME_PATTERN` is undetected, and `indexName` losing `tojson` is undetected, so two individually-silent regressions together produce a live injection. Worth closing whenever this file is next touched.

### What's good

The threat model is split deliberately rather than accidentally — validate where escaping is impossible (identifiers), escape where rejection is impossible (attribute names, which DynamoDB genuinely does not restrict). That analysis is the part most fixes of this shape get wrong, and it is right here. Boundary coverage is exact on both sides (2/3, 255/256), not mid-range. Zero vacuous assertions: all 11 new `pytest.raises` carry `match=` with anchored field paths, which is why removing a validation call site produces 18 failures instead of coincidentally passing. And documenting the `fullmatch` rationale at `models.py:40-41` *and* pinning it with a test is exactly the pairing that keeps a comment from rotting.

PR #4450, opened 2026-08-05
fix(aws-documentation-mcp-server): re-validate redirect targets agaist domain allowlist

zjerath:
Fixes

### Changes

`read_documentation`, `read_sections`, and `search_table` validate the initial URL against the AWS-docs allowlist, then fetch with `follow_redirects=True` without re-validating redirect targets. A 3xx from an allow-listed page was followed to any host, including link-local metadata, with the body returned to the caller.

This change re-validates **every redirect hop** against the same allowlist:

* **`util.py`**: add `url_matches_allowlist()` (domain-only check) and `enforce_redirect_allowlist()` — an httpx response event hook that resolves each redirect's `Location` (including relative redirects) against the request URL and refuses any hop whose target is not allow-listed.
* **`server_utils.py`**: add `_docs_client()` that attaches the hook, plus `DEFAULT_ALLOWED_DOMAIN_REGEXES` (commercial partition). All three read impls now fetch through this client and accept an `allowed_domain_regexes` argument (defaulted, so callers are unchanged).
* **`server_aws_cn.py`**: apply the same hook to both China-partition fetch sites (`read_documentation` and `get_available_services`) with `CN_ALLOWED_DOMAIN_REGEXES`.
* **Tests**: unit tests for the allowlist check and the redirect hook (off-allowlist/IMDS blocked, on-allowlist and relative redirects allowed, protocol-relative escape blocked, non-redirect and no-`Location` no-ops), plus end-to-end `httpx.MockTransport` tests confirming a read that 302s to IMDS returns a fetch error and never leaks the redirect body, while a same-domain redirect is followed.

`follow_redirects=True` is retained — legitimate AWS-docs pages rely on same-domain canonicalization redirects (verified: ~4.3% of doc reads are 3xx, all staying on the docs host).

### User experience

No change for legitimate usage. Direct 200 responses and same-domain redirects behave exactly as before. Only a redirect whose target leaves the allowlist changes: previously followed silently (with the off-host body returned), now refused and surfaced as a normal "Failed to fetch" error.

## Checklist

* [X] Changes are documented

Is this a breaking change? N

**RFC issue number**: N/A

## Acknowledgment

### DIFF
--- a/src/aws-documentation-mcp-server/awslabs/aws_documentation_mcp_server/server_aws_cn.py
+++ b/src/aws-documentation-mcp-server/awslabs/aws_documentation_mcp_server/server_aws_cn.py
@@ -14,7 +14,6 @@
 """awslabs AWS China Documentation MCP Server implementation."""
 
 import httpx
-import re
 import uuid
 from awslabs.aws_documentation_mcp_server.server_utils import (
     DEFAULT_USER_AGENT,
@@ -23,9 +22,11 @@ from awslabs.aws_documentation_mcp_server.server_utils import (
 
 # Import utility functions
 from awslabs.aws_documentation_mcp_server.util import (
+    enforce_redirect_allowlist,
     extract_content_from_html,
     format_documentation_result,
     is_html_content,
+    url_matches_allowlist,
 )
 from loguru import logger
 from mcp.server.fastmcp import Context, FastMCP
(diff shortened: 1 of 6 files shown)


alexisareyn (dismissed):

artb30 (dismissed):

JonLim (approved):

artb30 (approved):

SyedAnas01:
I reported this one to AWS Security on August 3 through the VDP. Good to see the redirect targets re-validated rather than only the initial URL.

Same shape as CVE-2026-14540, an SSRF I reported in Google's MCP Toolbox earlier this year, where the initial URL was checked but the destination after a redirect was not. It is a recurring pattern in MCP fetch tools: validation happens once, at the wrong point, and the redirect carries the request past it.

Nice to see the test coverage added alongside the fix. Happy to help verify similar patterns.

PR #4449, opened 2026-08-05
fix(lambda-tool-mcp-server): migrate to mcp SDK v2, remove <2.0.0 cap, add rollout codemod

scottschreckengaust:
## Description

First step of #4448: migrates `lambda-tool-mcp-server` to MCP Python SDK v2 and removes its
`mcp[cli]<2.0.0` cap, plus a codemod to automate the remaining 47 servers.

SDK 2.0.0 **removed** `mcp.server.fastmcp` (gone, not deprecated) and renamed `FastMCP` ->
`MCPServer`, so #4360 capped 44 servers at `<2.0.0` as a stopgap for #4354. That cap is
load-bearing — it can't be dropped without the rename.

**One server here, deliberately.** The remaining 47 follow in separate PRs, each gated on
its own test suite so one failure doesn't block the rest.

### Commits

1. **`fix(lambda-tool-mcp-server)`** — imports moved to `mcp.server.mcpserver`, cap raised
   to `>=2.0.0,<3.0.0`, `uv.lock` relocked onto `mcp 2.0.0`. Constructor kwargs, the tool
   decorator and `run()` are unchanged in v2, so stdio behavior is identical.
2. **`feat(scripts)`** — `scripts/migrate-mcp-v2.py`, dry-run by default.

### Why a codemod

The rename spans **246 files across 48 servers**. The script handles all seven v1 import
forms and three cases that a naive find-and-replace gets wrong:

- **Standalone `fastmcp` package** — a different project that also exports `FastMCP`.
  `FastMCP` is renamed only where it came from `mcp.server.fastmcp`; those files are skipped.
- **String monkeypatch targets** — `patch('mcp.server.fastmcp.FastMCP')` needs both halves
  renamed. Fixing only the module yields `mcp.server.mcpserver.FastMCP`, which doesn't exist
  and fails at patch time rather than import time.
- **Declared extras** — preserved; one server declares bare `mcp`, not `mcp[cli]`.

It does **not** touch `uv.lock` (generated — run `uv lock`) and does not move
`stateless_http=` from the constructor to `run()`; those are reported for review instead.

## Testing

- `lambda-tool-mcp-server`: **56 tests pass**, 99% coverage on `server.py`.
- **Live stdio handshake** — a real `ClientSession` completes `initialize()` and
  `list_tools()` returns the registered tool, so the v2 protocol path works end-to-end, not
  just import.
- **Codemod equivalence** — replaying the script over pristine v1 `lambda-tool-mcp-server`
[...]

PR #4360, opened 2026-07-28
fix: pin ceiling mcp SDK <2.0.0 for servers using the removed mcp.server.fastmcp module

scottschreckengaust:
Fixes #4354
Fixes #4356

## Summary

MCP Python SDK `2.0.0` (a [pre-release](https://pydantic.dev/articles/mcp-python-sdk-v2-beta)) **removed the vendored `mcp.server.fastmcp` module**. 45 servers import `from mcp.server.fastmcp import ...` but declared the SDK with **no upper bound** (`mcp[cli]>=1.23.0`), so `uvx awslabs.<server>@latest` resolves `mcp==2.0.0` and the server dies at import:

```
ModuleNotFoundError: No module named 'mcp.server.fastmcp'
```

The process exits before the `initialize` handshake (clients report a connection error, e.g. `-32000`).

### Changes

Add `,<2.0.0` to the `mcp` / `mcp[cli]` dependency in **45 `pyproject.toml` files** (one line each, no code touched), with each `uv.lock` updated in lockstep — specifier line only, so no resolved versions move and `uv lock --check` passes. This matches upstream guidance for downstream packages: *"add an upper bound constraint like `mcp>=1.27,<2`."*

<details>
<summary>All 45 affected servers</summary>

[...]

Issue #4354, opened 2026-07-28
(aws-pricing-mcp-server, aws-serverless-mcp-server): fail to start with mcp SDK 2.0.0 — ModuleNotFoundError: No module named 'mcp.server.fastmcp'


rnoap:
### Describe the bug

`awslabs.aws-pricing-mcp-server` and `awslabs.aws-serverless-mcp-server` crash at import time on any fresh dependency resolution, because the MCP Python SDK `2.0.0` removed the vendored `mcp.server.fastmcp` module.

Both servers do:

```python
from mcp.server.fastmcp import Context, FastMCP
```

and both declare the dependency without an upper bound:

```toml
# src/aws-pricing-mcp-server/pyproject.toml:8
# src/aws-serverless-mcp-server/pyproject.toml:9
"mcp[cli]>=1.23.0",
```

So `uvx ...@latest` resolves `mcp==2.0.0`, the import fails, the process exits with code 1, and any MCP client reports a connection error (Claude Code shows `-32000`, i.e. the server died before the `initialize` handshake).

For reference, `mcp 2.0.0` no longer ships that module:

```
mcp.server submodules: ['__main__', '_otel', '_streamable_http_modern', 'apps', 'auth', 'caching',
 'connection', 'context', 'elicitation', 'extension', 'lowlevel', 'mcpserver', 'models',
 'request_state', 'runner', 'session', 'sse', 'stdio', 'streamable_http',
 'streamable_http_manager', 'subscriptions', 'transport_security', 'validation']
mcp.server exports: [..., 'MCPServer', 'Server', ...]   # no 'fastmcp'
```

This is the same class of failure as #2533 (`openapi-mcp-server` broken by a missing version constraint on `fastmcp`).

### Expected Behavior

`uvx awslabs.aws-pricing-mcp-server@latest` and `uvx awslabs.aws-serverless-mcp-server@latest` start and answer the `initialize` request.

### Current Behavior

`aws-pricing-mcp-server`:

```
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "...\Scripts\awslabs.aws-pricing-mcp-server.exe\__main__.py", line 4, in <module>
    from awslabs.aws_pricing_mcp_server.server import main
  File "...\site-packages\awslabs\aws_pricing_mcp_server\server.py", line 51, in <module>
    from mcp.server.fastmcp import Context, FastMCP
[...]

PR #4452, opened 2026-08-05
feat: migrate the server fleet to mcp SDK v2 and remove the <2.0.0 cap (43 servers)

### LINKED ISSUE #4448: Migrate servers to MCP Python SDK v2 and remove the mcp[cli]<2.0.0 cap (48 servers)
### Summary

PR #4360 added `mcp[cli]>=...,<2.0.0` to 44 servers as a stopgap for #4354, pinning the
ecosystem to SDK 1.x. This issue tracks removing that cap by porting the servers to v2.

SDK 2.0.0 **removed** `mcp.server.fastmcp` (it is gone, not deprecated) and renamed
`FastMCP` -> `MCPServer`. Every capped server would crash at import under 2.x, so the cap
is load-bearing and cannot simply be dropped.

### Actual scope: 48 servers / 246 files

Larger than PR #4360's pyproject count suggests. Four servers (`amazon-keyspaces`,
`amazon-translate`, `aws-iot-sitewise`, `ecs`) declare the standalone `fastmcp` package
*and* import `mcp.server.fastmcp`, so they are affected too and can't be set aside as
"different dependency."

### API surface

Verified by importing from an installed `mcp 2.0.0` rather than reading the migration
guide. All seven v1 import forms present in `src/` move under `mcp.server.mcpserver`:

| v1 import | files | v2 import |
|----|----|----|
| `from mcp.server.fastmcp import Context` | 143 | `from mcp.server.mcpserver import Context` |
| `from mcp.server.fastmcp import Context, FastMCP` | 45 | `from mcp.server.mcpserver import Context, MCPServer` |
| `from mcp.server.fastmcp import FastMCP` | 41 | `from mcp.server.mcpserver import MCPServer` |
| `from mcp.server.fastmcp.tools import Tool` | 10 | `from mcp.server.mcpserver.tools import Tool` |
| `from mcp.server.fastmcp.prompts import Prompt` | 5 | `from mcp.server.mcpserver.prompts import Prompt` |
| `from mcp.server.fastmcp.exceptions import ToolError` | 3 | `from mcp.server.mcpserver.exceptions import ToolError` |
| `from mcp.server.fastmcp.server import Context` | 1 | `from mcp.server.mcpserver.server import Context` |

Constructor kwargs (`instructions`, `dependencies`), the `mcp.tool()` decorator, and
`mcp.run()` are unchanged. For a stdio server the migration is effectively an import
rename. No file in `src/` uses the 3-parameter `Context[ServerSessionT, ...]` form, so
that v2 delta doesn't apply here.

### Approach

Per-server: rename imports -> bump cap to `>=2.0.0,<3.0.0` -> `uv lock` -> run that
server's own test suite. Rename and cap bump must land **atomically** per server, since
v2-only imports break under a 1.x resolution.

Stateless / streamable-HTTP / cache-hint adoption is **out of scope** — it matters for
remote-deployed servers and is a separate per-server decision.

Rejected: a `try/except ImportError` shim supporting both 1.x and 2.x. It would add
dual-path complexity to 48 servers for a transitional benefit, and the repo pins via
lockfiles anyway.

### Needs individual review

- **Heaviest surface** — `amazon-bedrock-agentcore` (33 files), `aws-dataprocessing` (32),
  `aws-serverless` (26), `aws-healthomics` (25), `eks` (16), `aws-iot-sitewise` (15),
  `aws-transform` (15).
- **`stateless_http=`** — `aws-api-mcp-server` passes it to the constructor; in v2 it moves
  to `run()`. The one confirmed non-mechanical change. (Not among the 48 — it uses the
  standalone `fastmcp` package.)
- **String monkeypatch targets** — 3 files patch `'mcp.server.fastmcp.FastMCP'` as a
  string. Both halves must be renamed together; rewriting only the module yields
  `mcp.server.mcpserver.FastMCP`, which doesn't exist and fails at patch time rather than
  import time.
- **Standalone `fastmcp` servers** — the 4 above need a dependency decision alongside the
  rename.

### Separate latent bug

`ecs-mcp-server` imports `mcp.server.fastmcp` but declares **no `mcp` dependency at all**
(only `fastmcp>=3.0.0`) — it receives `mcp` transitively. Worth fixing independently of
this migration.

### Plan

First server + tooling, then the rest in follow-up PRs (individually or batched), each
gated on its own test suite so one failure doesn't block the others.

- [ ] `lambda-tool-mcp-server` + `scripts/migrate-mcp-v2.py` codemod
- [ ] Remaining 47 servers
- [ ] `ecs-mcp-server` missing `mcp` dependency (separate)

Related: #4354, #4360, #2533.

scottschreckengaust:
Fixes #4448

### Changes

Migrates the server fleet to MCP Python SDK v2, reversing the `mcp[cli]<2.0.0` cap that
#4360 added as a stopgap for #4354. Continues the `lambda-tool-mcp-server` port from #4449 —
**this branch contains #4449's two commits verbatim at its base**, so merging that first is a
clean fast-forward.

| | Count |
|---|---|
| Servers now on `mcp>=2.0.0,<3.0.0` | **43** |
| ↳ `pyright` errors across all 43 | **0** |
| ↳ failures reproducing identically on unmodified `main` | 6 |
| Blocked upstream, deliberately left on v1 | 6 |
| No `mcp` SDK dependency to bump | 4 |
| **Unexplained regressions** | **0** |

289 Python files. Most of the diff is mechanical, applied with `scripts/migrate-mcp-v2.py`
from #4449; `uv.lock` churn (43 locks) accounts for most of the rest.

**Two v2 changes not in #4448's original API table.** Both found by running suites rather
than reading the migration guide, and both share one trap: *construction still works, reads
and assignment break.* That asymmetry is why this surfaced as ~964 test failures across 9
servers instead of import errors.

1. **`MCPError` constructor flattened** — takes fields directly, not a prebuilt `ErrorData`:

   ```python
   raise McpError(ErrorData(code=INTERNAL_ERROR, message=msg))   # v1
   raise MCPError(code=INTERNAL_ERROR, message=msg)              # v2
   ```

   Reads of `.error`/`.code`/`.message` are unchanged, so only `raise` sites break — with a
   `TypeError`, at raise time. 11 files across `mssql`, `oracle`, `postgres`,
   `aws-for-sap-management`. Automated via AST, not regex: the inner call spans lines and
   `message=` routinely holds an f-string with its own parens.

2. **Transport settings moved off `Settings` onto `run()`** — v2's `Settings` has 8 fields,
   down from 17:

   ```python
   mcp.settings.port = args.port; mcp.run(transport='sse')   # v1
   mcp.run(transport='sse', port=args.port)                  # v2
   ```

   Pydantic intercepts the removed-field assignment, so the symptom is
   `ValueError: "Settings" object has no field "port"`, **not** `AttributeError`. Same delta
   #4448 flagged for `aws-api`'s `stateless_http=`; here it hit
   `well-architected-security-mcp-server`.

Also: `MCPServer.get_context()` and `create_connected_server_and_client_session` are gone.

**Four changes needed hands** (everything else was mechanical):

- `well-architected-security` — the `Settings`→`run()` move above, plus two test assertions.
- `amazon-qindex` — tests build `Context(mcp_server=mcp)` directly, which is what the removed
  `get_context()` did outside a live request.
- `aws-transform` — a test asserting the `PROFILE_SELECTION_REQUIRED` fallback had been
  **passing by accident**: v1's `elicit_with_validation` ended in `raise ValueError(...)`,
  which a bare `AsyncMock` triggered and an `except Exception` swallowed into the fallback
  path. v2 correctly returns `CancelledElicitation`, exposing that the test never declared
  the non-elicitation client its own assertion requires. I fixed the **setup**, not the
  assertion, matching `TestSelectProfileFallback`. *Flagging for a second opinion — this is a
  behavior question, not a mechanical one.*
- The 6 blocked servers below, and the type work in its own section further down.

**Adds `scripts/verify-mcp-v2-locks.py`.** `uv run` resolves from `uv.lock`, not
`pyproject.toml`, so a lock predating a cap change keeps installing the **old** SDK — the
suite passes *against v1* while the manifest claims v2. A green run that proves nothing. This
bit during the rollout and is invisible without a cross-check. Worth wiring into CI
regardless of how this PR lands.

### User experience

No behavioral change for end users. All 43 servers keep identical stdio behavior — constructor
kwargs, the `mcp.tool()` decorator and `run()` are unchanged in v2. What changes is that these
servers are no longer pinned to a 1.x SDK and can take v2 fixes and features.

Before: `mcp[cli]>=1.23.0,<2.0.0` — cap load-bearing, since v2-only imports break under a 1.x
resolution. After: `mcp[cli]>=2.0.0,<3.0.0`.

## Blocked: 6 servers stay on v1

`fastmcp` 3.x depends on `fastmcp-slim`, which pins `mcp>=1.24.0,<2.0`. **Not a code problem**
— and it bites in two ways, the second of which cost real time to find:

**1. Loud (3 servers).** The `fastmcp` floor is already `>=3`, so `mcp>=2.0.0` is
*unsatisfiable* and `uv lock` fails outright: `aws-api-mcp-server`,
`billing-cost-management-mcp-server`, and `dynamodb-mcp-server` (which inherits it through
`awslabs-aws-api-mcp-server`). Verified with `uv pip compile`: `fastmcp` 3.4.6 fails against
`mcp 2.0.0`; only the unreleased `fastmcp` 4.0.0b1 resolves.

**2. Silent (3 servers) — this is the dangerous one.** `amazon-keyspaces-mcp-server`,
`amazon-translate-mcp-server` and `aws-iot-sitewise-mcp-server` declare a `fastmcp` floor that
*predates* the pin (`>=2.14.0`, `>=2.13.1`). The resolver therefore does **not** fail — it
walks *backward* to a `fastmcp` old enough to accept mcp 2.x. In practice fastmcp
**3.4.3 → 2.14.1**, which carries **4 known advisories, one CRITICAL** (confirmed against
GitHub's advisory API).

I had all three migrated and green before catching this. A silent dependency downgrade
produces a passing test suite and a security regression *at the same time*, which is precisely
the combination a green CI run cannot distinguish from success. All three are reverted to v1
with a comment in each `pyproject.toml` explaining why, so the next person doesn't re-migrate
them. **A raised dependency floor is not a raised resolved version — diff the lock.**

`verify-mcp-v2-locks.py` exists to catch case 2; case 1 catches itself.

**Revisit when fastmcp 4.0 ships** — that release is the single blocker for all six. The
codemod hard-errors if you name one via `--server`, since a half-migrated blocked server is
broken against *both* SDKs; remove them from `BLOCKED_SERVERS` once 4.0 is out.

## The 371 pyright errors, and why none of them are this PR's fault

CI failed with 371 type errors. Root cause is a single line of the SDK's own typing:

```python
# v1
def tool(...) -> Callable[[AnyFunction], AnyFunction]:   # AnyFunction = Callable[..., Any]
# v2
def tool(...) -> Callable[[_CallableT], _CallableT]:     # TypeVar -- signature preserved
```

v1 **erased every decorated tool's signature to `(...) -> Any`**. v2 preserves it. **No call
site changed.** v2 simply stopped hiding errors that were always there — so the error count is
proportional to accumulated debt, not to API surface changed. Worth knowing before starting a
similar upgrade.

Every one of the 371 was a genuine pre-existing defect. The interesting ones:

- **21 mock context classes across 8 servers** were duck-typed and never structurally matched
  `Context`. They now subclass it. Two wrinkles: `Context` is a pydantic `BaseModel`, so stubs
  accumulating state via `self.errors = []` raise `ValueError: object has no field` — those
  became declared fields (which also gives each instance its own list, since pydantic
  deep-copies field defaults). And each subclass needs an explicit no-arg `__init__`, because
  pyright otherwise re-synthesizes a pydantic `__init__` from `Context`'s private attrs and
  demands `_request_context`, `_mcp_server`, `_input_params`, `_subscriptions`. That, in
  hindsight, is what the pre-existing `DummyContext(_request_context=None, _fastmcp=None)` in
  `aws-location` was working around.
- **`aws-location`'s integration `main()` passed the pytest fixture *function* as `ctx`** to
  nine tools. It type-checked under v1 and "worked" only because each tool hits an AWS error
  path before touching `ctx`.
- **4 `iam` tests asserted nothing.** `with pytest.raises(Exception): await get_user(user_name=…)`
  was catching `TypeError: missing 1 required positional argument: 'ctx'` — not the AWS error
  its docstring claimed. They now pass a context and assert the specific error type *and*
  message. `create_user` additionally needed `Context.initialize`, or the call short-circuits
  on the readonly guard and never reaches the mock.
- **`aurora-dsql`** passed tuples where `List[str]` was required, and `transact((sql), ctx)` —
  parens without a comma are not a 1-tuple, so that was passing a bare `str`.
- **`oracle`** subscripted a `str | dict` return with a string key; **`aws-documentation`** read
  an `Optional[Dict]` without narrowing and imported `Context` from the private
  `mcp.server.mcpserver.server` submodule.

The remainder are tests deliberately passing invalid input to exercise validation. Those get
targeted `# pyright: ignore[reportArgumentType]` at the call site — the convention `ccapi`'s
own tests already use — rather than weakening the assertion.

**Why this was invisible locally:** `pyright` runs at pre-commit **`stages: [pre-push]`**, so
`pre-commit run --files …` never exercises it. `scripts/README.md` now documents this, with the
per-server invocation (and the fact that `elasticache` declares dev tools as an optional extra,
needing `--extra dev`).

**One suppression left in place.** `aws-dataprocessing-mcp-server` sets
`reportCallIssue = false`, masking 354 alias-spelled keyword arguments (`isError=` vs
`is_error=`). Forcing the rule on reveals all 354. They are *not* part of CI's 371 and they
work at runtime — v2 keeps the camelCase alias, so construction resolves and serializes
identically. Renaming them would add 354 no-op changes to an already-large PR, so I left them
and am flagging it rather than burying it.

Relatedly, the codemod now rewrites keyword construction too, but **gated on the server
actually being on v2** — a distinction that turned out to matter. Under v1,
`Tool.model_fields` literally contains `inputSchema` as the *real field name*, not an alias
(verified by runtime introspection), so a blanket rename would have broken
`healthlake-mcp-server`: still on `mcp>=1.23.0`, using the low-level `mcp.server.Server`, and
constructing `Tool(inputSchema=…)` 11 times.

## Testing

**Every suite was run against unmodified `origin/main` before any failure was called a
regression.** The repo has a steady background of environment-dependent failures, and a suite
failing identically on both sides is not this migration's doing.

- **All 43 v2 servers: 0 `pyright` errors.**
- Full suites green at their baseline counts, including every server touched by the type work:
  `aws-documentation` 280+23s, `postgres` 690, `eks` 315, `mysql` 428, `mssql` 288, `ccapi` 274,
  `oracle` 246, `aurora-dsql` 238, `s3-tables` 219, `iam` 174, `aws-appsync` 173,
  `cloudwatch-applicationsignals` 1444, `aws-pricing` 251, `aws-location` 53, `amazon-qindex` 34.
- **6 servers**: failures reproduce identically on `origin/main` — `amazon-kendra-index` (5),
  `amazon-mq` (2), `amazon-qbusiness-anonymous` (1), `amazon-sns-sqs` (1),
  `aws-dataprocessing` (2), `timestream-for-influxdb` (1). Listed so a reviewer doesn't
  re-investigate them.
- **0 unexplained regressions.**

`aws-pricing`'s 3 "failures" and `ccapi`'s 2 were mine, not the repo's: a local `AWS_PROFILE`
leaking into assertions expecting `Session(profile_name=None)`. Both are green under
`env -u AWS_PROFILE`.

Also verified:

- 2366 files compile; codemod idempotent (a re-run on the migrated tree reports zero work).
- `verify-mcp-v2-locks.py`: every server declaring `mcp>=2.0.0` has a 2.x lock — no
  stale-lock false PASSes.
- `ruff-check` and `ruff-format` clean over all 289 changed files. These found real work:
  removing the `mcp.settings.port` assertion orphaned an import that `F401` caught, so the
  affected suites were re-run afterwards.
- Only residual v1 API references anywhere in `src/` are inside the 6 deliberately-blocked
  servers.
- Replaying the codemod against pristine `origin/main` reproduces #4449's branch
  **byte-for-byte on all 4 files**.

`main` moved during this work: the merge commit resolves two mechanical conflicts with #4450
(which touched the same `aws-documentation-mcp-server` import lines). Resolved keeping both
sides; that server is at **280 passed, 23 skipped** — #4450's 23 new tests all pass under
SDK v2.

## Out of scope

- **Stateless / streamable-HTTP / cache-hint adoption** — matters for remote-deployed servers,
  a separate per-server decision.
- **`ecs-mcp-server`'s missing `mcp` dependency** — imports `mcp.server.fastmcp` but declares
  only `fastmcp>=3.0.0`, receiving `mcp` transitively. Latent packaging bug, tracked in #4448.
- **The 6 blocked servers**, until fastmcp 4.0.
- **`aws-dataprocessing`'s 354 suppressed alias-spelled keyword arguments** — see above; they
  work at runtime and are hidden by that server's own `reportCallIssue = false`.

## Reviewer notes

**How to land this is an open question, and I'd rather you chose than assume.** 43 servers in
one review is a lot. The commit slices cleanly along `src/<server>/` if you'd prefer per-server
PRs off this branch — #4448's plan allows "individually or batched", and this is the batched
form. Happy to split it.

Suggested checks:

```bash
git fetch origin && git checkout feat/mcp-v2-fleet
python3 scripts/verify-mcp-v2-locks.py   # every server declaring mcp>=2.0.0 has a 2.x lock
python3 scripts/migrate-mcp-v2.py        # no work remaining (idempotent)
```

The four interesting files to read, if you read only four:
`src/well-architected-security-mcp-server/.../server.py` (the `run()` move),
`src/aws-transform-mcp-server/tests/test_configure.py` and
`src/iam-mcp-server/tests/test_server.py` (tests that were passing vacuously), and the
`BLOCKED_SERVERS` comments in the 6 blocked `pyproject.toml`s — the three *silent* ones are the
finding I'd most want a second pair of eyes on.

## Checklist

* [x] Changes are documented

Is this a breaking change? (Y/N) **N** for end users — no behavioral change, and the servers
are lock-pinned. **Y** in the packaging sense: these 43 distributions now require SDK 2.x, so
anything installing them alongside a `mcp<2.0` pin will fail to resolve.

**RFC issue number**: #4448

Checklist:

* [x] Migration process documented — `scripts/README.md` covers every transformation, the
  `Settings`→`run()` non-goal, and the blocked servers; #4448 has the full API surface.
* [ ] Implement warnings (if it can live side by side) — **N/A, and deliberately so.** #4448
  rejected a `try/except ImportError` shim: v1 and v2 cannot coexist in one install, so there
  is no side-by-side period to warn during. `mcp.server.fastmcp` was *removed*, not
  deprecated, so v2 itself raises `ImportError` on the old path.

## Acknowledgment


scottschreckengaust:
## CI is green — 205 pass, 0 fail

Both failures are fixed in `7205b2d7`. One was mine; the other was a real security regression.

### 1. `dependency-review` — a genuine `cryptography` downgrade

The job flagged `cryptography` 45.0.7 / 46.0.0 as **added** to `src/lambda-tool-mcp-server/uv.lock`, carrying 7 advisories (4 high). This was *not* incidental lock churn:

| | `cryptography` |
|---|---|
| `origin/main` | `50.0.0` (single entry) |
| merge-base `c1738dba` | `50.0.0` |
| this branch (before fix) | `45.0.7` + `46.0.0` (marker-forked) |

Nothing constrains the version on either side — `mcp` pulls `pyjwt[crypto]`, which requires `cryptography` unbounded. So this was resolution *vintage*, not a new bound. The migration commit was authored against a tree predating #4433 (*"bump the uv-security-updates group"*), which had raised it to 50.0.0. Re-locking `mcp` to 2.x recomputed the entire graph, and with nothing pinning `cryptography` upward the resolver had no reason to preserve the security bump — it re-derived exactly the marker-forked pair that #4433 had collapsed.

`uv lock --upgrade-package cryptography` restores a single `50.0.0`. The lock diff touches **no other package** and `mcp` stays at `2.0.0`. I audited all 43 changed locks against `origin/main` for any version that moved *backward*; this was the only one.

> **This is the second silent downgrade this migration has surfaced**, after the `fastmcp` 3.4.3 → 2.14.1 walk on the blocked servers. Worth generalising: a re-lock is a whole-graph recomputation, so **unpinned transitive dependencies can quietly lose security bumps landed by an earlier commit**. A green test suite says nothing about it — diff the lock. The structural tell here was that our entry was *marker-forked* into two versions while main's was a single unmarked entry; that fork was an artifact of the stale resolution, not a real platform constraint.

### 2. `Build amazon-qindex-mcp-server` — displaced pyright pragma (my error)

A `# pyright: ignore[reportArgumentType]` sat on the wrong argument: ruff-format reflowed the call to one-arg-per-line *after* I added the pragma, landing it on `qbuiness_region` instead of the `application_id=None` it was meant to suppress. Inline pragmas are position-sensitive — add them **after** formatting. Re-verified placements in all three files that use them.

### Not a failure: the 236 `invalid-license-changes` entries

These did not fail the job (`denied-changes` was empty; only `vulnerable-changes` failed it), and they aren't fixable here. They are four packages — `httpx2`, `httpcore2`, `mcp-types`, `opentelemetry-api` — repeated across the 43 locks. All four declare **PEP 639 `License-Expression`**, which GitHub's dependency graph doesn't yet read, so it reports `license: null`.

Provenance verified rather than assumed, since `httpx2`/`httpcore2` are exactly the shape a typosquat takes:

- `httpx2` / `httpcore2` — pydantic's fork of httpx (`github.com/pydantic/httpx2`, same author, `BSD-3-Clause`)
- `mcp-types` — the official SDK split-out (`modelcontextprotocol/python-sdk`, `MIT`)

All resolve from PyPI with hashes, ship a license file, and are declared by `mcp==2.0.0` itself.

### Verification

- `uv run --frozen pyright` → **0 errors** for all 43 v2 servers (CI's exact command, per server directory)
- `lambda-tool` 56 tests pass against the re-locked graph
- Full CI: **205 pass, 0 fail** (the 2 "skipping" are the uninstantiated `${{ matrix.package }}` placeholder)

Ready to move out of draft. #4449 remains untouched.

scottschreckengaust:
## Correction: I got the `cryptography` root cause wrong

My [earlier comment](https://github.com/awslabs/mcp/pull/4452#issuecomment-5198916698) and commit `7205b2d7` blamed `uv lock` for discarding the `cryptography` 50.0.0 bump when the `mcp` bound moved to v2. **That is wrong**, and it's the wrong lesson — it would have people distrusting re-locks, when re-locking is actually the fix. Corrected in `df38e11a`. The fix itself (50.0.0 restored) was right; only the explanation was wrong.

### What I actually tested

`uv lock` is **conservative** — it keeps what the input lock pins and moves only what the changed constraint forces:

| scenario | result |
|---|---|
| bump `mcp` to v2 against an **up-to-date** lock (`015ec459`) | `cryptography 50.0.0` ✅ |
| lock **from scratch**, no input lock | `cryptography 50.0.0` ✅ |
| bump `mcp` to v2 against the **stale** pre-#4433 lock | `45.0.7`/`46.0.0` ❌ reproduced |

So the resolver never discarded anything. Only a stale *input* lock loses the bump.

### The real cause

The lambda-tool lock was **generated 2026-07-20 in the earlier PoC worktree**, then committed (`9848a3cc`) onto a parent `015ec459` that already carried #4433's bump. The stale lock still pinned 45.0.7/46.0.0, so the commit silently reverted it.

The corroborating evidence is what I should have looked for the first time — *several* packages stale in the same direction, not one suspicious version:

| package | `origin/main` | branch lock |
|---|---|---|
| `cryptography` | 50.0.0 | 45.0.7 + 46.0.0 |
| `sse-starlette` | 3.4.8 | **3.4.6** |
| `certifi` | 2025.1.31 | **absent** |

One old generation date. I also checked and **ruled out** the branch merge `b5356e1f` — it discarded nothing from main's side (audited every lock main touched).

I'd also retract the "marker-fork is the tell" claim: a `resolution-markers` split is a weak hint at best, since forks can be legitimate.

**So the guidance inverts:** sync to `origin/main` *before* re-locking. `git fetch && git merge origin/main`, then `uv lock`.

### Meanwhile, the `fastmcp` case *is* real — and worse than I said

Worth separating, because this one is a genuine resolver walk, reproducible from a fully current lock. On `amazon-translate-mcp-server`, raising only the `mcp` cap walks `fastmcp` **3.4.3 → 2.14.1**, which is exposed to:

- **GHSA-vv7q-7jx5-f767 — CRITICAL**, SSRF + path traversal in the OpenAPI provider
- GHSA-rww4-4w9c-7733 (high), GHSA-m8x7-r2rg-vh5g (moderate) — missing OAuth consent verification, command injection
- GHSA-5h2m-4q8j-pqpj (high) — OAuth token reuse across servers

The first three are patched only in **3.2.0, which requires mcp 1.x**. There is no `fastmcp` version that is both mcp-v2-compatible and free of them.

**This is worth fixing on `main` independently of this PR.** Raising the floor to `fastmcp>=3.2.0` *while staying on mcp v1* still resolves 3.4.3 / mcp 1.27.1 today (verified), and it makes the silent walk **structurally impossible** — a future v2 attempt then fails loudly rather than quietly downgrading. That's a guardrail worth having before anyone retries these three servers.

`fastmcp` 4.0 is what genuinely lifts the ceiling — `fastmcp-slim` 4.0.0b1 requires `mcp>=2.0.0,<3.0.0`, and `fastmcp>=4.0.0b1` + `mcp>=2.0.0` resolves cleanly — but only `4.0.0a1`/`4.0.0a2`/`4.0.0b1` exist so far, needing `--prerelease=allow`. Not something to ship; the blocked six should wait for stable 4.0.

### Re-verified after this merge

Since the lesson is that merges bring newer locks, I re-ran the audit rather than trusting the earlier pass — and fixed a real flaw in it: my first version compared versions with a tuple-of-int parse wrapped in a silent `except: pass`, which could have swallowed non-numeric versions. Now using `packaging.version.Version`.

**All 61 `uv.lock` files, 0 packages moved backward vs `origin/main`, 0 unparseable.**

src/amazon-bedrock-agentcore-mcp-server/pyproject.toml
```diff
@@ -9,7 +9,7 @@ readme = "README.md"
 requires-python = ">=3.10"
 dependencies = [
     "loguru>=0.7.0",
-    "mcp[cli]>=1.23.0,<2.0.0",
+    "mcp[cli]>=2.0.0,<3.0.0",
```
theagenticguy, line 12: Representative of all 43 migrated servers: raising the `mcp` floor from `<2.0.0` to `>=2.0.0` changes this distribution's install-compatibility envelope (co-installation with any `mcp<2.0` pin stops resolving), while the server's own tool surface is unchanged. That is a **minor** bump under semver, but the release automation only ever does patch bumps unless signaled. Please set `version = "0.1.9223372036854775807"` here (and the equivalent `X.Y.9223372036854775807` in the other 42 migrated servers) so the next `release-initiate-branch` run cuts `0.2.0` instead of `0.1.5` — full per-package table in the review body.

src/s3-tables-mcp-server/pyproject.toml
```diff
@@ -9,7 +9,7 @@ readme = "README.md"
 requires-python = ">=3.11"
 dependencies = [
     "loguru==0.7.3",
-    "mcp[cli]==1.23.0",
+    "mcp[cli]>=2.0.0,<3.0.0",
```
theagenticguy: This one changes character as well as version: `mcp[cli]==1.23.0` (exact pin) becomes `>=2.0.0,<3.0.0` (range), in a file where every other runtime dependency is exact-pinned (`loguru==0.7.3`, `pydantic==2.12.5`, `boto3==1.42.41`, `pyiceberg==0.10.0`). If the exact-pin policy is deliberate for this server, this should be `mcp[cli]==2.<x>.<y>` matching what `uv.lock` resolved; if the policy is being abandoned, fine, but worth a conscious call rather than a side effect of the codemod.

laithalsaadoon (dismissed):
## Semver review: per-package version-bump analysis

**Verdict: Request changes** — the migration itself is thorough and well-evidenced, but the PR leaves every `version` field untouched, and this repo's release automation (`.github/workflows/release.py::bump_version`) only ever does a **patch** bump per changed directory. Your own checklist answers "Is this a breaking change?" with **Y in the packaging sense** — so as written, the next `release-initiate-branch` run will ship 43 packaging-breaking packages as patch releases, which is the one semver violation the automation cannot catch on its own.

### The "magic number" mechanics (verified against `release.py` on `main`)

`MAX_VERSION_COMPONENT = sys.maxsize = 9223372036854775807`. `bump_version` does `patch += 1`, then rolls over on overflow:

| You set in this PR | Next release produces | Signal |
|---|---|---|
| `X.Y.9223372036854775807` | `X.(Y+1).0` | **minor** |
| `X.9223372036854775807.9223372036854775807` | `(X+1).0.0` | **major** |

`validate_version_format` explicitly allows oversized minor/patch components ("Allow large components for bumping to zero") and rejects only an oversized major. The precedent is already in-repo: `src/aws-api-mcp-server/pyproject.toml` carries the `# NOTE: "Patch"=9223372036854775807 bumps next release to zero.` comment, and `amazon-mq` used the same trick historically (`2.0.9223372036854775807` ahead of its 2.0.0 release).

### Why minor, not patch or major

- **Patch understates it.** Repo precedent does treat dependency work as patch (#4360, #1929), but #4360 added a cap that matched what environments already resolved — no existing install broke. This PR flips the floor *above* every 1.x release: any consumer co-installing one of these 43 distributions alongside an `mcp<2.0` pin resolves today and fails after the next release. That is a changed compatibility contract, not a bug fix.
- **Major overstates it.** The servers' own contract — stdio transport, tool names, tool schemas, console entry points — is byte-for-byte unchanged (your own testing section demonstrates this). These distribute primarily as `uvx`-isolated applications, so the breakage surface is the co-installation case only. Reserving major for changes to the servers' own tool surface keeps that signal meaningful.
- **Minor is the honest middle:** same behavior, new requirements. (For the 0.x packages semver technically promises nothing, but the same roll gives downstream pins like `~=0.1` a fighting chance.)

### Requested change

Set `version = "X.Y.9223372036854775807"` in the `pyproject.toml` of each of the **43 migrated servers** (keep `X.Y` as-is). Expected result at next release:

| Package | Current | Next release |
|---|---|---|
| amazon-bedrock-agentcore-mcp-server | 0.1.4 | 0.2.0 |
| amazon-kendra-index-mcp-server | 1.0.21 | 1.1.0 |
| amazon-mq-mcp-server | 2.0.26 | 2.1.0 |
| amazon-neptune-mcp-server | 1.0.21 | 1.1.0 |
| amazon-qbusiness-anonymous-mcp-server | 0.0.20 | 0.1.0 |
| amazon-qindex-mcp-server | 0.0.14 | 0.1.0 |
| amazon-sns-sqs-mcp-server | 2.0.23 | 2.1.0 |
| aurora-dsql-mcp-server | 1.0.39 | 1.1.0 |
| aws-appsync-mcp-server | 0.1.17 | 0.2.0 |
| aws-dataprocessing-mcp-server | 0.1.35 | 0.2.0 |
| aws-documentation-mcp-server | 1.1.30 | 1.2.0 |
| aws-for-sap-management-mcp-server | 0.0.6 | 0.1.0 |
| aws-healthomics-mcp-server | 0.0.44 | 0.1.0 |
| aws-location-mcp-server | 2.0.21 | 2.1.0 |
| aws-pricing-mcp-server | 1.0.34 | 1.1.0 |
| aws-serverless-mcp-server | 0.1.22 | 0.2.0 |
| aws-transform-mcp-server | 0.1.11 | 0.2.0 |
| bedrock-kb-retrieval-mcp-server | 1.0.25 | 1.1.0 |
| ccapi-mcp-server | 1.0.25 | 1.1.0 |
| cloudtrail-mcp-server | 0.0.18 | 0.1.0 |
| cloudwatch-applicationsignals-mcp-server | 0.1.40 | 0.2.0 |
| cloudwatch-mcp-server | 0.1.8 | 0.2.0 |
| documentdb-mcp-server | 1.0.15 | 1.1.0 |
| eks-mcp-server | 0.1.37 | 0.2.0 |
| elasticache-mcp-server | 0.1.23 | 0.2.0 |
| finch-mcp-server | 0.1.18 | 0.2.0 |
| healthimaging-mcp-server | 0.0.10 | 0.1.0 |
| iam-mcp-server | 1.0.25 | 1.1.0 |
| lambda-tool-mcp-server | 2.0.22 | 2.1.0 |
| memcached-mcp-server | 1.0.21 | 1.1.0 |
| mssql-mcp-server | 0.1.5 | 0.2.0 |
| mysql-mcp-server | 1.0.26 | 1.1.0 |
| oracle-mcp-server | 0.1.5 | 0.2.0 |
| postgres-mcp-server | 1.1.11 | 1.2.0 |
| prometheus-mcp-server | 0.2.20 | 0.3.0 |
| redshift-mcp-server | 0.0.34 | 0.1.0 |
| s3-tables-mcp-server | 0.0.26 | 0.1.0 |
| sagemaker-ai-mcp-server | 1.0.12 | 1.1.0 |
| security-agent-mcp-server | 0.1.5 | 0.2.0 |
| stepfunctions-tool-mcp-server | 0.1.27 | 0.2.0 |
| timestream-for-influxdb-mcp-server | 0.0.21 | 0.1.0 |
| valkey-mcp-server | 1.0.22 | 1.1.0 |
| well-architected-security-mcp-server | 0.1.10 | 0.2.0 |

### The 6 blocked servers: no bump signal needed, but expect no-op patch releases

`amazon-keyspaces`, `amazon-translate`, `aws-iot-sitewise`, `aws-api`, `billing-cost-management`, and `dynamodb` receive comment-only `pyproject.toml` changes — no dependency, code, or behavior change. Semver-wise nothing is required. Be aware, though, that `release-initiate-branch.yml` keys on *any* changed file under `src/<dir>` since the last release tag, so all six will still be swept into the next release as functionally identical patch releases. Harmless, just expected noise.

### Minor observations (non-blocking)

- `stepfunctions-tool-mcp-server` has a pre-existing version drift the release script won't fix: `pyproject.toml` (0.1.27) and `__init__.py` (0.1.27) are synced, but `server.py` still carries `__version__ = '0.1.5'`, and `release.py` only rewrites `__init__.py`. Pre-dates this PR; noting it since the pyproject comment asks for manual sync.
- The `aws-api` NOTE comment says the magic patch "bumps next release to zero" — under current `release.py` it also increments minor (patch overflow does `patch=0; minor+=1`), which is exactly the minor-bump semantics recommended here, but the comment reads as if only patch resets.

Everything else in the PR reads sound for its stated scope: the blocked-server comments match the real fastmcp-slim constraint, the silent-downgrade catch on the three low-floor fastmcp servers is a genuinely good find, and `verify-mcp-v2-locks.py` addresses a real stale-lock blind spot.

src/amazon-bedrock-agentcore-mcp-server/pyproject.toml
scottschreckengaust: Done in 52c7d517. Set `version = "X.Y.9223372036854775807"` (keeping `X.Y`) across all 43 migrated servers per the review table — this one is now `0.1.9223372036854775807`, so the next `release-initiate-branch` run cuts `0.2.0`. The 6 blocked servers were left untouched (comment-only changes). Verified only the `[project]` version line changed in each file; `[tool.commitizen]` versions are untouched.

src/s3-tables-mcp-server/pyproject.toml
scottschreckengaust: Good catch — the exact-pin policy is deliberate for this server, so this was an unintended side effect of the codemod. Fixed in 52c7d517: `mcp[cli]==2.0.0`, matching what `uv.lock` resolved and consistent with the other exact-pinned runtime deps (`loguru==0.7.3`, `pydantic==2.12.5`, `boto3==1.42.41`, `pyiceberg==0.10.0`). `uv.lock` regenerated and confirmed recording `specifier = "==2.0.0"`.

laithalsaadoon (dismissed):
## Follow-up review: requested changes verified on 52c7d517

**Verdict: Approve.** Both items from the previous (dismissed) review are addressed exactly as requested, and I verified them against the head commit rather than the reply text.

### What was checked

1. **Version-bump signal (43 servers).** Every one of the 43 migrated servers now carries `version = "X.Y.9223372036854775807"` with `X.Y` unchanged from the table in the prior review — checked all 43 `pyproject.toml`s at `52c7d51`, zero mismatches. Commit touches exactly the 43 files plus `s3-tables`' `uv.lock`; the 6 blocked servers are untouched, `[tool.commitizen]` sections untouched. Next `release-initiate-branch` run cuts minor releases (`0.2.0` / `1.1.0` / `2.1.0` etc.) as intended.
2. **`s3-tables` exact-pin policy.** `mcp[cli]>=2.0.0,<3.0.0` → `mcp[cli]==2.0.0`, consistent with the server's other exact-pinned runtime deps, and `uv.lock` records `specifier = "==2.0.0"` resolving `mcp 2.0.0`. Matches the requested "pin to what the lock resolved."

Both inline threads are resolved.

### One non-blocking observation

`release.py bump-package` rewrites `pyproject.toml`/`__init__.py`, and the workflow's follow-up `uv sync` refreshes each lock — but the magic-version commit only regenerated `s3-tables`' lock, so the other 42 locks still record their old self-version (e.g. `amazon-mq`'s `uv.lock` says `2.0.26` while `pyproject.toml` says `2.0.9223372036854775807`). This is self-healing: the release workflow runs `uv sync --directory=…` right after bumping, which rewrites the lock's self-version along with the real `X.(Y+1).0`. Historical precedent agrees — `aws-api`'s magic-version commits never synced the lock either, and releases came out clean. No action needed; noting it so nobody burns time on the "lock drift" during a later audit.

CI is green on head (205 successes, 2 skips). The migration substance was reviewed in the previous round and stands unchanged.

src/security-agent-mcp-server/awslabs/security_agent_mcp_server/server.py
```diff
@@ -136,7 +136,7 @@ def _client_prefix(ctx: Context) -> str:
         client_params = session.client_params
         if client_params is None:
             return 'ide'
-        name = client_params.clientInfo.name  # type: ignore[union-attr]
+        name = client_params.client_info.name  # type: ignore[union-attr]
```
theagenticguy, line 139: This merge brought in main's new `_ensure_client_ua()` (further down this file, ~line 161), which reads `client_params.clientInfo` — removed in SDK v2, the same rename this line already handles. Under this server's own v2 lock that read raises `AttributeError`, which the function's `except (AttributeError, TypeError)` swallows into a debug log — so user-agent enrichment silently never runs, and the 9 matching test sites can't catch it because they set `clientInfo` on a `MagicMock`.

The repo's own codemod already detects and fixes all 10 sites:

```bash
python3 scripts/migrate-mcp-v2.py --server security-agent-mcp-server --apply
```

Verified on a scratch checkout: suite stays green afterward (207 passed).

laithalsaadoon (changes requested):
## Re-review of merge commit 7e0bef06: the declared conflict is resolved correctly, but the merge silently reintroduced a v1 API call in `security-agent-mcp-server`

**Verdict: Request changes** — one file needs a re-run of the PR's own codemod. Everything else about the merge checks out.

### The declared conflict (postgres): resolved correctly ✅

`git show --remerge-diff 7e0bef06` confirms the only textual conflict was `src/postgres-mcp-server/.../server.py`, and the resolution is right: `mcp = MCPServer(` kept from the branch side, main's new `ConnectionValidationError` / privilege-guardrail code (#4510-era work) kept intact. Verified beyond the diff:

- No v1 API residue in postgres (`mcp.server.fastmcp`, `FastMCP(` — zero hits outside tests).
- Full suite green under the v2 lock: **749 passed** (up from 690 — main added guardrail tests, all passing under SDK v2).
- `pyproject.toml` still `mcp[cli]>=2.0.0,<3.0.0`, `uv.lock` still resolves `mcp 2.0.0`.

### The undeclared problem (security-agent): auto-merged v1 code ⚠️

Main's #4489-era change added `_ensure_client_ua()` to `src/security-agent-mcp-server/awslabs/security_agent_mcp_server/server.py`, written against SDK v1. It merged without conflict, so nothing flagged it — but it reads `client_params.clientInfo`, an attribute **removed in v2** (the branch already migrated the older read at line 139 to `.client_info`; this new function at line ~161 arrived after that migration).

Confirmed as a real runtime defect, not a style nit — under the server's own v2 lock:

```python
p.client_info.name   # 'kiro'
p.clientInfo         # AttributeError: 'InitializeRequestParams' object has no attribute 'clientInfo'
```

Two aggravating factors:

1. **It fails silently.** `_ensure_client_ua` wraps everything in `except (AttributeError, TypeError)` with a debug log, so under v2 the user-agent enrichment simply never happens — `set_mcp_client_info` becomes unreachable. This is precisely the silent-degradation class this PR was built to eliminate.
2. **The tests can't catch it.** The 9 corresponding test sites drive `ctx.session.client_params.clientInfo` on a `MagicMock`, which fabricates any attribute — so all 207 tests pass against the broken path, and CI is green (the `# type: ignore[union-attr]` on the read also placates pyright).

This also breaks two invariants the PR body asserts: "codemod idempotent — a re-run reports zero work" and "only residual v1 API references anywhere in src/ are inside the 6 deliberately-blocked servers." A dry run of `scripts/migrate-mcp-v2.py` on head now reports **1 field rename in server.py + 9 in tests/test_server.py** for security-agent (it reported zero for this server at 52c7d517, the commit previously approved).

### The fix is one command

```bash
python3 scripts/migrate-mcp-v2.py --server security-agent-mcp-server --apply
```

We applied exactly that on a scratch checkout: it rewrites the 10 sites (`clientInfo` → `client_info`) and the suite stays green — **207 passed**. No lock changes needed.

### Everything else swept clean

- All 43 magic versions (`X.Y.9223372036854775807`) survived the merge — exactly 43 `version` fields carry it, none dropped, none gained.
- `verify-mcp-v2-locks.py` passes on head; `s3-tables` still exact-pinned `==2.0.0` with a matching lock.
- Main's other additions to migrated servers are v2-clean: `timestream-for-influxdb` (+132 lines of new tools) **154 passed**; the codemod reports nothing for it.
- The `aws-dataprocessing` dry-run findings are the same 354 alias-spelled kwargs already documented as out of scope in the PR body — present at 52c7d517 too, not introduced by this merge.
- The billing-cost-management v1 API hits in the merge are expected: it's one of the 6 deliberately-blocked v1 servers.

Once the security-agent codemod run lands, this is an approve — the merge is otherwise faithful and the prior approval's substance stands.

src/security-agent-mcp-server/awslabs/security_agent_mcp_server/server.py
scottschreckengaust: Good catch! fixing now

scottschreckengaust: Good catch — confirmed and fixed. You're right that this was a live bug: the `main` merge re-introduced the pre-v2 `client_params.clientInfo` read in `_ensure_client_ua()`, and under this server's v2 lock that raises `AttributeError`, which the `except (AttributeError, TypeError)` swallows into a debug log — so user-agent enrichment silently never ran. And as you noted, the 9 `MagicMock`-based test sites auto-vivify either attribute name, so they couldn't catch it.

Fixed via the repo codemod exactly as suggested:

```bash
python3 scripts/migrate-mcp-v2.py --server security-agent-mcp-server --apply
```

It migrated all 10 field reads (1 in `server.py`, 9 in `tests/test_server.py`). Suite green locally (207 passed) and CI is now all green on the branch. Pushed in 1d9b70aac.

laithalsaadoon (approved):
## Re-review of 1d9b70aa: the security-agent fix is verified — Approve

**Verdict: Approve.**

The single blocking finding from the previous review is fixed exactly as prescribed. Verified against the head commit, not the description:

- **The fix is precisely scoped.** `git diff 7e0bef06..1d9b70aa` touches only the two flagged files: `server.py` (1 line — the `client_params.clientInfo` read in `_ensure_client_ua` is now `.client_info`) and `tests/test_server.py` (the 9 matching mock sites). Byte-identical to what `migrate-mcp-v2.py --server security-agent-mcp-server --apply` produced in our scratch verification last round.
- **Codemod idempotency restored.** A fleet-wide dry run on head reports work remaining in exactly 1 server: `aws-dataprocessing`'s 354 alias-spelled kwargs, which the PR body already documents as deliberately out of scope. Security-agent reports zero — the invariant from the PR body ("re-run reports zero work") holds again, modulo the documented exception.
- **Suite green under the v2 lock:** 207 passed, `--frozen`, on `1d9b70aa`.
- **CI green on head:** 205 passing checks, 2 skipped.

All three inline threads are now resolved. Prior verification of the postgres conflict resolution, the 43 magic versions, the s3-tables exact pin, and `verify-mcp-v2-locks.py` carries over unchanged — nothing outside the two security-agent files moved since the last review.

arnewouters (approved):

PR #4454, opened 2026-08-06
fix(deps): raise fastmcp floor to >=3.2.0 to exclude a CRITICAL advisory

scottschreckengaust:
Fixes

## Summary

Raises the declared `fastmcp` lower bound to `>=3.2.0` on the 12 servers whose floor permitted versions carrying known advisories.

**Nothing is exposed today** — all 13 standalone-`fastmcp` servers already resolve `fastmcp 3.4.3`, and this PR changes **no installed version**. It is a *guardrail*: the floors permitted a resolver walk back into the vulnerable range, and that walk is not hypothetical — the mcp SDK v2 migration (#4452) hit it for real.

### The advisories

| GHSA | Severity | Issue | Patched in |
|---|---|---|---|
| GHSA-vv7q-7jx5-f767 | **CRITICAL** | SSRF + path traversal in the OpenAPI provider | 3.2.0 |
| GHSA-rww4-4w9c-7733 | high | Missing consent verification in OAuth proxy callback | 3.2.0 |
| GHSA-m8x7-r2rg-vh5g | moderate | Command injection (Gemini CLI) | 3.2.0 |
| GHSA-5h2m-4q8j-pqpj | high | OAuth proxy token reuse across MCP servers | 2.14.2 |

### Why a floor raise, when nothing is currently vulnerable

`fastmcp-slim` pins `mcp>=1.24.0,<2.0`, so **only older `fastmcp` accepts mcp 2.x**. That turns an unrelated dependency bump into a silent security regression. Reproduced on `amazon-translate-mcp-server` from a fully up-to-date lock:

```
fastmcp>=2.13.1 + mcp[cli]>=1.23.0,<2.0.0   ->  fastmcp 3.4.3, mcp 1.27.1
fastmcp>=2.13.1 + mcp[cli]>=2.0.0,<3.0.0    ->  fastmcp 2.14.1, mcp 2.0.0   # 4 advisories, 1 CRITICAL
```

The resolver doesn't fail — it walks *backward* to a `fastmcp` old enough to accept mcp 2.x. Tests stay green while the advisories come back. With the floor at 3.2.0 that resolution has no solution, so it **fails loudly instead of silently downgrading**.

Note the corollary: **no `fastmcp` release is both mcp-v2-compatible and free of these advisories.** `fastmcp` 4.0 is the real unblock (`fastmcp-slim` 4.0.0b1 requires `mcp>=2.0.0,<3.0.0`), but only `4.0.0a1`/`a2`/`b1` exist so far — so v2 for these servers still waits on a stable 4.0, and this PR does not attempt it.

`openapi-mcp-server` already declared `>=3.3.1,<4` and needed no change.

### User experience

No change. No installed version moves, no API or behaviour changes. The only user-visible effect is that a future dependency bump which would have quietly downgraded `fastmcp` into the vulnerable range now fails to resolve instead.

## Verification

- **No resolved version changed** in any of the 12 relocked files — compared as per-package version *sets* across every lock, not by reading the diff.
- The small lock diffs are the `requires-dist` line plus `marker = "..."` annotations uv adds when a narrower floor lets it prove a conditional edge is reachable only under a specific interpreter — strictly more precise metadata for an identical graph.
- An **unmodified** re-lock is a byte-for-byte no-op, which rules out ambient uv-version drift as the source of that churn; every line here is attributable to the floor raise.
- Floors re-audited afterwards: all 13 servers now exclude `2.13.1`, `2.14.0`, `2.14.1`, `3.0.0`, `3.0.1`, `3.1.x`.
- **5,125 tests pass** across all 12 servers.

<details><summary>Per-server test counts</summary>

| server | tests |
|---|---|
| amazon-keyspaces | 81 |
| amazon-translate | 994 |
| aws-api | 727 |
| aws-bedrock-custom-model-import | 126 |
| aws-iac | 149 |
| aws-iot-sitewise | 590 |
| aws-network | 244 |
| aws-support | 194 |
| billing-cost-management | 1154 |
| document-loader | 54 |
| ecs | 722 |
| roda | 90 |

</details>

`roda-mcp-server` declares its dev tools under `[project.optional-dependencies]` rather than `[dependency-groups]`, so it needs `uv run --frozen --extra dev pytest`. Pre-existing and unrelated to this change.

## Checklist

* [x] Changes are documented

**RFC issue number**:

Checklist:

## Acknowledgment


src/aws-bedrock-custom-model-import-mcp-server/pyproject.toml
```diff
@@ -8,7 +8,7 @@ dependencies = [
     "boto3>=1.28.0",
     "pydantic>=2.0.0",
     "loguru>=0.7.0",
-    "fastmcp>=2.14.0",
+    "fastmcp>=3.2.0",
```
theagenticguy, line 11: Non-blocking: consider `fastmcp>=3.2.0,<4` here and on the other servers that don't pin `mcp<2.0` (aws-network, billing-cost-management, aws-iac, this one). For servers pinning `mcp>=1.23,<2.0.0` the resolver can never pick fastmcp 4.x anyway (fastmcp-slim 4.x requires `mcp>=2`), but for the unpinned ones a stable fastmcp 4.0 release would be adopted silently on the next re-lock — a major bump with breaking changes, which is the same "silent resolver walk" failure mode this PR exists to prevent, just in the forward direction. `openapi-mcp-server` already caps at `<4`, so there's in-repo precedent. Fine as a follow-up given the v2 migration will revisit all these floors anyway.

laithalsaadoon (approved):
Verified the substance of this PR independently rather than taking the description on faith:

- **Advisories**: all four GHSA IDs confirmed against the GitHub Advisory Database. GHSA-vv7q-7jx5-f767 (critical), GHSA-rww4-4w9c-7733 (high), and GHSA-m8x7-r2rg-vh5g (medium) are patched in fastmcp 3.2.0; GHSA-5h2m-4q8j-pqpj (high) in 2.14.2. A `>=3.2.0` floor excludes the vulnerable range for all four.
- **No resolved version moves**: grepped the full diff for changed `version = ` lines in any `uv.lock` — zero. Every lock still resolves `fastmcp 3.4.3` on this branch (spot-checked keyspaces, translate, aws-api, aws-support, ecs, roda). The lock churn is exactly what the description says: `requires-dist` lines plus uv marker annotations (`caio`, `secretstorage`, `zipp`, `pycparser`) that tighten conditional-edge metadata without changing the graph.
- **Coverage**: 13 `pyproject.toml` files in the repo declare `fastmcp`; this PR touches 12 and the 13th (`openapi-mcp-server`) already declares `>=3.3.1,<4`. No server missed.
- **Mechanism**: the silent-downgrade path is real — `fastmcp-slim` pins `mcp<2`, so only pre-3.2 fastmcp satisfies an `mcp>=2` constraint, and the resolver will happily walk backward into the advisory range. Turning that into a resolution failure is the right guardrail.
- **CI**: all completed checks pass (codecov reports all tests green, project coverage unchanged at 93.08%); remaining checks were still queued at review time.

One non-blocking suggestion inline about an upper bound. Approving — this is a pure floor raise with no behavior change and a well-verified no-op lock diff.

arnewouters (approved):

src/aws-bedrock-custom-model-import-mcp-server/pyproject.toml
scottschreckengaust: Followup

PR #4468, opened 2026-08-07
feat: add icon.svg to license header check patterns

scottschreckengaust:
Fixes broken license header checks

### Changes

### User experience

## Checklist

* [x] Changes are documented

**RFC issue number**:

Checklist:

## Acknowledgment

### DIFF
--- a/.github/workflows/check-license-header.json
+++ b/.github/workflows/check-license-header.json
@@ -23,7 +23,8 @@
       "**/*.properties",
       "**/*.xml",
       "**/*.drawio",
-      "**/dist/**"
+      "**/dist/**",
+      "**/icon.svg"
     ],
     "include": [
       "**/*.py",
@@ -85,7 +86,8 @@
       "**/skills/*/references",
       "**/skills/*/mcp",
       "**/DO_NOT_RELEASE",
-      "**/*.pyi"
+      "**/*.pyi",
+      "**/icon.svg"
     ]
   }
 ]


.github/workflows/check-license-header.json
```diff
@@ -85,7 +86,8 @@
       "**/skills/*/references",
       "**/skills/*/mcp",
       "**/DO_NOT_RELEASE",
-      "**/*.pyi"
+      "**/*.pyi",
+      "**/icon.svg"
```
theagenticguy, line 90: Non-blocking: consider `**/*.svg` instead of `**/icon.svg`. The 12 existing SVGs under `docusaurus/` are already covered by the `docusaurus/**` entry above, so today the two globs are equivalent — but the next server that ships a `logo.svg` or any differently-named SVG will break `main`'s check again the same way this one did. If per-server icons are meant to be exactly `icon.svg` by convention, the narrower glob is a reasonable guard; otherwise the wildcard saves the next round-trip.

laithalsaadoon (approved):
Verified against the failure on `main`: the `check-license-header` workflow has failed on every push since `src/redshift-mcp-server/icon.svg` landed (run 31142557618: `##[warning]Config does not cover the file 'src/redshift-mcp-server/icon.svg'`, failing under `strict: true`), and it passes on this branch (run 31206433402). The fix is correct.

One note on mechanics, from reading the action source (viperproject/check-license-header v2.0.3): strict-mode coverage is the union of each block's `include` minus `exclude`, so only the second hunk (adding `**/icon.svg` to the no-license block's `include`) actually resolves the failure. The first hunk (adding it to the hash-header block's `exclude`) is a no-op since that block only includes `*.py`/`Dockerfile`/`*.sh`/`*.guard`/`ruff.toml` — but the existing config already mirrors entries like `**/*.xml` and `**/*.drawio` the same way, so it's consistent with house style and fine to keep.

Left one non-blocking suggestion about future-proofing the glob.

ayushtr-aws (approved):
reviewed, lgtm

.github/workflows/check-license-header.json
scottschreckengaust: For now it is just for MCP contributions

PR #4471, opened 2026-08-10
feat(timestream-for-influxdb-mcp-server): Add InfluxDB v2 write mode

trevorbonas:
## Summary

Adds an operator-controlled INFLUXDB_WRITE_MODE environment variable that governs whether write-producing Flux operations (to(), experimental.to(), wideTo()) are permitted through the InfluxDBQuery tool. When disabled (the default), queries containing these functions are rejected before reaching InfluxDB.

### Changes

- New INFLUXDB_WRITE_MODE environment variable (default: false). Operator-controlled and not overridable by tool callers.
- validate_flux_query() function that detects write-producing Flux patterns and rejects them when write mode is disabled. Handles pipe-forward syntax, qualified calls, aliased imports, and ignores commented-out code.
- Validation integrated into InfluxDBQuery before any network call is made.
- 21 new tests covering blocked patterns, false-positive avoidance, write-mode opt-in, and normal read queries passing through.

### User experience

## Checklist

* [x] Changes are documented

N

**RFC issue number**:

Checklist:

## Acknowledgment

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -1251,10 +1251,10 @@
         "filename": "src/timestream-for-influxdb-mcp-server/tests/test_server.py",
         "hashed_secret": "789cbe0407840b1c2041cb33452ff60f19bf58cc",
         "is_verified": false,
-        "line_number": 619,
+        "line_number": 620,
         "is_secret": false
       }
     ]
   },
-  "generated_at": "2026-07-16T18:42:37Z"
+  "generated_at": "2026-08-10T16:13:08Z"
 }
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -1251,10 +1251,10 @@
         "filename": "src/timestream-for-influxdb-mcp-server/tests/test_server.py",
         "hashed_secret": "789cbe0407840b1c2041cb33452ff60f19bf58cc",
         "is_verified": false,
-        "line_number": 620,
+        "line_number": 621,
         "is_secret": false
       }
     ]
   },
-  "generated_at": "2026-08-10T16:13:08Z"
+  "generated_at": "2026-08-12T23:21:47Z"
 }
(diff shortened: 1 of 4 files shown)


forestmvey (approved):

fredjoonpark:
@trevorbonas can you verify/address the below?

  1. [P1] Write guard is bypassable with valid Flux call forms
     server.py:477
     (https://github.com/awslabs/mcp/blob/740e6d16ea4fd07c7ab6710abddf611b2b21cfc8/src/timestream-for-influxdb-mcp-server/awslabs/timestream_for_influxdb_mcp_server/server.py#L477-L486)

  These valid writes pass validation while write mode is disabled:
```
  data = from(bucket: "src")
  result = to(bucket: "dst", tables: data)

  writer = to
  from(bucket: "src") |> writer(bucket: "dst")
```
  The regex only recognizes selected invocation syntax. This defeats the PR’s primary security guarantee. Use Flux-aware parsing or enforce a read-scoped token; expanding the regex will remain bypassable.

  2. [P1] URLs in strings hide the remainder of a query
     server.py:465
     (https://github.com/awslabs/mcp/blob/740e6d16ea4fd07c7ab6710abddf611b2b21cfc8/src/timestream-for-influxdb-mcp-server/awslabs/timestream_for_influxdb_mcp_server/server.py#L465-L467)

  The comment-stripping regex treats // inside strings as a comment:
```
  from(bucket: "src")
    |> map(fn: (r) => ({r with url: "http://example"}))
    |> to(bucket: "dst")
```
  On one line, this passes validation because everything after http:// is removed before scanning, while InfluxDB receives and executes the original query.

  3. [P2] Legitimate reads are rejected when strings resemble calls
     server.py:480
     (https://github.com/awslabs/mcp/blob/740e6d16ea4fd07c7ab6710abddf611b2b21cfc8/src/timestream-for-influxdb-mcp-server/awslabs/timestream_for_influxdb_mcp_server/server.py#L480-L483)

  For example, this raises ValueError:
```
  from(bucket: "src")
    |> filter(fn: (r) => r.message == "experimental.to(")
```
  The scanner needs to distinguish identifiers from string literals.

trevorbonas:
@fredjoonpark I've added a fix that first masks comments and strings and then runs a regular expression. The true way to enforce write-only actions is to require the users token be read-only. But, that would be a breaking change. Checking the user's query works as a mitigation.

fredjoonpark (approved):

mhchengaws (approved):

lokendrp-aws (approved):

laithalsaadoon (approved):
LGTM

PR #4474, opened 2026-08-10
feat: add pricing scenario forecasts

### LINKED ISSUE #4473: aws-pricing-mcp-server: add fixed-versus-usage scenario forecasting
## Describe the feature

Add a credential-free `calculate_cost_scenario` tool to the AWS Pricing MCP Server. The tool accepts baseline monthly cost components classified as either fixed or linearly scaling, allows independent final-month multipliers, and returns month-by-month, category, component, and summary totals.

## Use Case

Forecasts based on a current cloud bill often incorrectly multiply every service by a single growth factor. Architecture units such as control planes, proxy capacity, gateway hours, keys, and subscriptions may remain flat while compute, database consumption, request volume, data transfer, logs, and AI tokens grow at very different rates. Users need deterministic scenario math after retrieving baseline costs and current AWS prices.

## Proposed Solution

Introduce a read-only MCP tool backed by a pure Decimal-based forecast function and validated Pydantic inputs. Each component declares `fixed` or `linear` behavior. Linear components ramp independently from 1x to their configured end multiplier over 2 to 120 months. The result includes monthly totals, category/component breakdowns, baseline, average, forecast total, and ending run rate.

The tool performs arithmetic only and does not retrieve account data, query pricing, or create public calculator links. Existing AWS pricing and billing tools remain responsible for source data.

## Other Information

A tested prototype is being prepared in a draft pull request. It adds no runtime dependencies and preserves the server read-only boundary.

BaxterDevs:
Fixes #4473

## Summary

Add deterministic multi-period scenario forecasting to the AWS Pricing MCP Server without querying account data or introducing new dependencies.

### Changes

- add the read-only `calculate_cost_scenario` MCP tool
- classify each baseline component as fixed or independently linear-scaling
- return monthly category/component breakdowns plus baseline, average, forecast-total, and ending run-rate summaries
- use Decimal arithmetic with explicit validation and bounded response sizes
- document the capability and add full unit/tool-wrapper coverage

### User experience

Before this change, users could retrieve current prices and generate point-in-time reports, but independently scaling a mixed architecture required external arithmetic and encouraged applying one multiplier to the entire bill.

After this change, an assistant can hold control planes, proxy capacity, gateways, keys, and subscriptions flat while applying distinct ramps to compute, database consumption, requests, storage, data transfer, logs, or AI tokens.

The tool performs arithmetic only. Retrieving account costs, querying current prices, mapping workload usage to priced components, and creating public AWS Pricing Calculator links remain out of scope.

## Validation

- `uv run --frozen pytest --cov --cov-branch --cov-report=term-missing` — 263 passed; new module 100% covered
- `uv run --frozen pyright` — 0 errors
- `uv run --frozen ruff check .`
- `uv run --frozen ruff format --check .`
- scoped repository pre-commit hooks, including credential, secret, and license checks

## Checklist

* [x] Changes are documented

**RFC issue number**: N/A — existing-server feature request #4473

Checklist:

## Acknowledgment

### DIFF
--- a/src/aws-pricing-mcp-server/awslabs/aws_pricing_mcp_server/forecast.py
+++ b/src/aws-pricing-mcp-server/awslabs/aws_pricing_mcp_server/forecast.py
@@ -133,7 +133,7 @@ def calculate_forecast(
             }
         )
 
-    annual_total = sum(unrounded_monthly_totals, Decimal('0'))
+    forecast_total = sum(unrounded_monthly_totals, Decimal('0'))
     return {
         'status': 'success',
         'currency': currency,
@@ -143,8 +143,8 @@ def calculate_forecast(
         'summary': {
             'forecast_months': months,
             'baseline_monthly_cost': _money(unrounded_monthly_totals[0]),
-            'average_monthly_cost': _money(annual_total / Decimal(months)),
-            'forecast_total': _money(annual_total),
+            'average_monthly_cost': _money(forecast_total / Decimal(months)),
+            'forecast_total': _money(forecast_total),
             'ending_monthly_cost': _money(unrounded_monthly_totals[-1]),
         },
     }
--- a/src/aws-pricing-mcp-server/awslabs/aws_pricing_mcp_server/forecast.py
+++ b/src/aws-pricing-mcp-server/awslabs/aws_pricing_mcp_server/forecast.py
@@ -14,20 +14,23 @@
 
 """Deterministic scenario forecasting for fixed and usage-scaled costs."""
 
+import json
 import re
 from decimal import ROUND_HALF_UP, Decimal
 from pydantic import BaseModel, ConfigDict, Field, model_validator
 from typing import Dict, List, Literal, Optional
 
 
-MONEY = Decimal('0.01')
 MAX_COMPONENTS = 100
+MAX_DECIMAL_PLACES = 6
+MAX_MONTHLY_COST = 1_000_000_000
 MAX_MONTHS = 120
+MAX_RESPONSE_CHARACTERS = 100_000
 
 
-def _money(value: Decimal) -> float:
-    """Round a decimal monetary value to cents for JSON output."""
-    return float(value.quantize(MONEY, rounding=ROUND_HALF_UP))
+def _round_money(value: Decimal, quantum: Decimal) -> Decimal:
+    """Round a decimal monetary value to the requested minor unit."""
+    return value.quantize(quantum, rounding=ROUND_HALF_UP)
 
 
 class ForecastCostComponent(BaseModel):
@@ -40,7 +43,11 @@ class ForecastCostComponent(BaseModel):
         None, min_length=1, max_length=100, description='Optional grouping label'
     )
     monthly_cost: float = Field(
-        ..., ge=0, allow_inf_nan=False, description='Baseline monthly cost'
+        ...,
+        ge=0,
+        le=MAX_MONTHLY_COST,
+        allow_inf_nan=False,
+        description='Baseline monthly cost',
     )
     scaling_behavior: Literal['fixed', 'linear'] = Field(
         ...,
@@ -63,7 +70,10 @@ class ForecastCostComponent(BaseModel):
 
 
 def calculate_forecast(
-    components: List[ForecastCostComponent], months: int, currency: str
+    components: List[ForecastCostComponent],
+    months: int,
+    currency: str,
+    decimal_places: int = 2,
 ) -> Dict:
     """Calculate a linear scenario forecast without querying external services."""
     if not 2 <= months <= MAX_MONTHS:
@@ -71,15 +81,21 @@ def calculate_forecast(
     if not 1 <= len(components) <= MAX_COMPONENTS:
         raise ValueError(f'components must contain between 1 and {MAX_COMPONENTS} items')
     if not re.fullmatch(r'[A-Z]{3}', currency):
-        raise ValueError('currency must be a three-letter uppercase ISO 4217 code')
+        raise ValueError('currency must be a three-letter uppercase display label')
+    if not 0 <= decimal_places <= MAX_DECIMAL_PLACES:
+        raise ValueError(f'decimal_places must be between 0 and {MAX_DECIMAL_PLACES}')
 
     component_names = [component.name for component in components]
     if len(component_names) != len(set(component_names)):
         raise ValueError('component names must be unique')
 
+    quantum = Decimal(1).scaleb(-decimal_places)
     component_rows = []
     monthly_rows = []
-    unrounded_monthly_totals = []
+    monthly_totals = []
+    component_totals = {component.name: Decimal('0') for component in components}
+    component_first_costs: Dict[str, Decimal] = {}
+    component_last_costs: Dict[str, Decimal] = {}
 
     for month_index in range(months):
         progress = Decimal(month_index) / Decimal(months - 1)
@@ -95,56 +111,68 @@ def calculate_forecast(
                 if component.scaling_behavior == 'fixed'
                 else Decimal('1') + (end_multiplier - Decimal('1')) * progress
             )
-            cost = baseline * factor
-            component_costs[component.name] = _money(cost)
+            cost = _round_money(baseline * factor, quantum)
+            component_costs[component.name] = float(cost)
             category = component.category or 'uncategorized'
             category_costs[category] = category_costs.get(category, Decimal('0')) + cost
             month_total += cost
+            component_totals[component.name] += cost
+            component_first_costs.setdefault(component.name, cost)
+            component_last_costs[component.name] = cost
 
-        unrounded_monthly_totals.append(month_total)
+        monthly_totals.append(month_total)
         monthly_rows.append(
             {
                 'month': month_index + 1,
-                'total': _money(month_total),
+                'total': float(month_total),
                 'by_category': {
-                    category: _money(cost) for category, cost in sorted(category_costs.items())
+                    category: float(cost) for category, cost in sorted(category_costs.items())
                 },
                 'by_component': component_costs,
             }
         )
 
     for component in components:
-        baseline = Decimal(str(component.monthly_cost))
         end_multiplier = Decimal(str(component.end_multiplier))
-        average_factor = (
-            Decimal('1')
-            if component.scaling_behavior == 'fixed'
-            else (Decimal('1') + end_multiplier) / Decimal('2')
-        )
         component_rows.append(
             {
                 'name': component.name,
                 'category': component.category or 'uncategorized',
                 'scaling_behavior': component.scaling_behavior,
-                'baseline_monthly_cost': _money(baseline),
+                'baseline_monthly_cost': float(component_first_costs[component.name]),
                 'end_multiplier': float(end_multiplier),
-                'average_monthly_cost': _money(baseline * average_factor),
-                'ending_monthly_cost': _money(baseline * end_multiplier),
+                'average_monthly_cost': float(
+                    _round_money(component_totals[component.name] / Decimal(months), quantum)
+                ),
+                'ending_monthly_cost': float(component_last_costs[component.name]),
             }
         )
 
-    forecast_total = sum(unrounded_monthly_totals, Decimal('0'))
-    return {
+    forecast_total = sum(monthly_totals, Decimal('0'))
+    response = {
         'status': 'success',
         'currency': currency,
-        'assumption': 'Each linear component ramps independently from 1x to its end multiplier.',
+        'decimal_places': decimal_places,
+        'assumption': (
+            'Each linear component ramps independently from 1x to its end multiplier using '
+            'a constant effective rate; tiering, free allowances, commitments, capacity steps, '
+            'and price changes must be modeled separately.'
+        ),
         'months': monthly_rows,
         'components': component_rows,
         'summary': {
             'forecast_months': months,
-            'baseline_monthly_cost': _money(unrounded_monthly_totals[0]),
-            'average_monthly_cost': _money(forecast_total / Decimal(months)),
-            'forecast_total': _money(forecast_total),
-            'ending_monthly_cost': _money(unrounded_monthly_totals[-1]),
+            'baseline_monthly_cost': float(monthly_totals[0]),
+            'average_monthly_cost': float(_round_money(forecast_total / Decimal(months), quantum)),
+            'forecast_total': float(forecast_total),
+            'ending_monthly_cost': float(monthly_totals[-1]),
         },
     }
+    response_characters = len(json.dumps(response, indent=2))
+    if response_characters > MAX_RESPONSE_CHARACTERS:
+        raise ValueError(
+            f'forecast response would contain {response_characters:,} characters, exceeding '
+            f'the {MAX_RESPONSE_CHARACTERS:,}-character limit; reduce components, months, or '
+            'label lengths'
+        )
+    return response
--- a/src/aws-pricing-mcp-server/awslabs/aws_pricing_mcp_server/forecast.py
+++ b/src/aws-pricing-mcp-server/awslabs/aws_pricing_mcp_server/forecast.py
@@ -27,15 +27,25 @@ MAX_MONTHLY_COST = 1_000_000_000
 MAX_MONTHS = 120
 MAX_RESPONSE_CHARACTERS = 100_000
 
-# Monetary values are published as JSON numbers, so every emitted amount must survive the
-# round trip through a float. Keeping the largest emitted value below 10**15 minor units
-# holds the whole payload within float64's 15-significant-digit exact range.
-MAX_EXACT_MINOR_UNITS = 10**15
-
 
 def _round_money(value: Decimal, quantum: Decimal) -> Decimal:
     """Round a decimal monetary value to the requested minor unit."""
     return value.quantize(quantum, rounding=ROUND_HALF_UP)
+
+
+def _format_money(value: Decimal, decimal_places: int) -> str:
+    """Preserve a rounded monetary value as a fixed-precision decimal string."""
+    return f'{value:.{decimal_places}f}'
+
+
+def _fits_response_budget(value: Dict) -> bool:
+    """Count encoded characters without allocating an oversized JSON string."""
+    encoded_characters = 0
+    for chunk in json.JSONEncoder(indent=2).iterencode(value):
+        encoded_characters += len(chunk)
+        if encoded_characters > MAX_RESPONSE_CHARACTERS:
+            return False
+    return True
 
 
 class ForecastCostComponent(BaseModel):
@@ -104,7 +114,7 @@ def calculate_forecast(
 
     for month_index in range(months):
         progress = Decimal(month_index) / Decimal(months - 1)
-        component_costs: Dict[str, float] = {}
+        component_costs: Dict[str, str] = {}
         category_costs: Dict[str, Decimal] = {}
         month_total = Decimal('0')
 
@@ -117,7 +127,7 @@ def calculate_forecast(
                 else Decimal('1') + (end_multiplier - Decimal('1')) * progress
             )
             cost = _round_money(baseline * factor, quantum)
-            component_costs[component.name] = float(cost)
+            component_costs[component.name] = _format_money(cost, decimal_places)
             category = component.category or 'uncategorized'
             category_costs[category] = category_costs.get(category, Decimal('0')) + cost
             month_total += cost
@@ -129,9 +139,10 @@ def calculate_forecast(
         monthly_rows.append(
             {
                 'month': month_index + 1,
-                'total': float(month_total),
+                'total': _format_money(month_total, decimal_places),
                 'by_category': {
-                    category: float(cost) for category, cost in sorted(category_costs.items())
+                    category: _format_money(cost, decimal_places)
+                    for category, cost in sorted(category_costs.items())
                 },
                 'by_component': component_costs,
             }
@@ -144,25 +155,21 @@ def calculate_forecast(
                 'name': component.name,
                 'category': component.category or 'uncategorized',
                 'scaling_behavior': component.scaling_behavior,
-                'baseline_monthly_cost': float(component_first_costs[component.name]),
-                'end_multiplier': float(end_multiplier),
-                'average_monthly_cost': float(
-                    _round_money(component_totals[component.name] / Decimal(months), quantum)
+                'baseline_monthly_cost': _format_money(
+                    component_first_costs[component.name], decimal_places
                 ),
-                'ending_monthly_cost': float(component_last_costs[component.name]),
+                'end_multiplier': float(end_multiplier),
+                'average_monthly_cost': _format_money(
+                    _round_money(component_totals[component.name] / Decimal(months), quantum),
+                    decimal_places,
+                ),
+                'ending_monthly_cost': _format_money(
+                    component_last_costs[component.name], decimal_places
+                ),
             }
         )
 
     forecast_total = sum(monthly_totals, Decimal('0'))
-
-    # forecast_total is the largest emitted amount, so bounding it bounds every published value.
-    emitted_minor_units = forecast_total.scaleb(decimal_places)
-    if emitted_minor_units > MAX_EXACT_MINOR_UNITS:
-        raise ValueError(
-            f'forecast total {forecast_total} cannot be represented exactly at '
-            f'{decimal_places} decimal places; reduce component costs, multipliers, months, '
-            'or decimal_places'
-        )
 
     response = {
         'status': 'success',
@@ -177,17 +184,17 @@ def calculate_forecast(
         'components': component_rows,
         'summary': {
             'forecast_months': months,
-            'baseline_monthly_cost': float(monthly_totals[0]),
-            'average_monthly_cost': float(_round_money(forecast_total / Decimal(months), quantum)),
-            'forecast_total': float(forecast_total),
-            'ending_monthly_cost': float(monthly_totals[-1]),
+            'baseline_monthly_cost': _format_money(monthly_totals[0], decimal_places),
+            'average_monthly_cost': _format_money(
+                _round_money(forecast_total / Decimal(months), quantum), decimal_places
+            ),
+            'forecast_total': _format_money(forecast_total, decimal_places),
+            'ending_monthly_cost': _format_money(monthly_totals[-1], decimal_places),
         },
     }
-    response_characters = len(json.dumps(response, indent=2))
-    if response_characters > MAX_RESPONSE_CHARACTERS:
+    if not _fits_response_budget(response):
         raise ValueError(
-            f'forecast response would contain {response_characters:,} characters, exceeding '
-            f'the {MAX_RESPONSE_CHARACTERS:,}-character limit; reduce components, months, or '
-            'label lengths'
+            f'forecast response exceeds the {MAX_RESPONSE_CHARACTERS:,}-character limit; '
+            'reduce components, months, or label lengths'
         )
     return response
--- a/src/aws-pricing-mcp-server/awslabs/aws_pricing_mcp_server/forecast.py
+++ b/src/aws-pricing-mcp-server/awslabs/aws_pricing_mcp_server/forecast.py
@@ -39,9 +39,9 @@ def _format_money(value: Decimal, decimal_places: int) -> str:
 
 
 def _fits_response_budget(value: Dict) -> bool:
-    """Count encoded characters without allocating an oversized JSON string."""
+    """Count emitted characters without allocating an oversized JSON string."""
     encoded_characters = 0
-    for chunk in json.JSONEncoder(indent=2).iterencode(value):
+    for chunk in json.JSONEncoder(indent=2, ensure_ascii=False).iterencode(value):
         encoded_characters += len(chunk)
         if encoded_characters > MAX_RESPONSE_CHARACTERS:
             return False
--- a/src/aws-pricing-mcp-server/awslabs/aws_pricing_mcp_server/forecast.py
+++ b/src/aws-pricing-mcp-server/awslabs/aws_pricing_mcp_server/forecast.py
@@ -1,0 +1,150 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Deterministic scenario forecasting for fixed and usage-scaled costs."""
+
+import re
+from decimal import ROUND_HALF_UP, Decimal
+from pydantic import BaseModel, ConfigDict, Field, model_validator
+from typing import Dict, List, Literal, Optional
+
+
+MONEY = Decimal('0.01')
+MAX_COMPONENTS = 100
+MAX_MONTHS = 120
+
+
+def _money(value: Decimal) -> float:
+    """Round a decimal monetary value to cents for JSON output."""
+    return float(value.quantize(MONEY, rounding=ROUND_HALF_UP))
+
+
+class ForecastCostComponent(BaseModel):
+    """A baseline monthly cost and its behavior over a forecast horizon."""
+
+    model_config = ConfigDict(extra='forbid')
+
+    name: str = Field(..., min_length=1, max_length=200, description='Component name')
+    category: Optional[str] = Field(
+        None, min_length=1, max_length=100, description='Optional grouping label'
+    )
+    monthly_cost: float = Field(
+        ..., ge=0, allow_inf_nan=False, description='Baseline monthly cost'
+    )
+    scaling_behavior: Literal['fixed', 'linear'] = Field(
+        ...,
+        description='fixed holds cost constant; linear ramps from 1x to end_multiplier',
+    )
+    end_multiplier: float = Field(
+        1,
+        ge=0,
+        le=1_000_000,
+        allow_inf_nan=False,
+        description='Final-month multiplier for a linear component',
+    )
+
+    @model_validator(mode='after')
+    def validate_fixed_multiplier(self):
+        """Reject contradictory fixed-component configuration."""
+        if self.scaling_behavior == 'fixed' and self.end_multiplier != 1:
+            raise ValueError('fixed components must use end_multiplier=1')
+        return self
+
+
+def calculate_forecast(
+    components: List[ForecastCostComponent], months: int, currency: str
+) -> Dict:
+    """Calculate a linear scenario forecast without querying external services."""
+    if not 2 <= months <= MAX_MONTHS:
+        raise ValueError(f'months must be between 2 and {MAX_MONTHS}')
+    if not 1 <= len(components) <= MAX_COMPONENTS:
+        raise ValueError(f'components must contain between 1 and {MAX_COMPONENTS} items')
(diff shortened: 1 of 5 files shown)


src/aws-pricing-mcp-server/awslabs/aws_pricing_mcp_server/forecast.py
```diff
@@ -0,0 +1,150 @@
...
+                'category': component.category or 'uncategorized',
+                'scaling_behavior': component.scaling_behavior,
+                'baseline_monthly_cost': _money(baseline),
+                'end_multiplier': float(end_multiplier),
+                'average_monthly_cost': _money(baseline * average_factor),
+                'ending_monthly_cost': _money(baseline * end_multiplier),
+            }
+        )
+
+    annual_total = sum(unrounded_monthly_totals, Decimal('0'))
```
BaxterDevs: Low — refactor/simplicity: this accumulator covers the configured 2–120-month horizon, so `annual_total` is inaccurate whenever the forecast is not exactly 12 months. Rename it to `forecast_total` (and update its two uses) so the calculation communicates the actual contract without implying an annual period.

BaxterDevs: Fixed in signed commit `920fef465631bb9c993ad6f90998a7d52a4a08ee` by renaming the arbitrary-horizon accumulator and its uses to `forecast_total`. Validation: `uv run --frozen pytest tests/test_forecast.py` (12 passed), focused Ruff check/format, and `git diff --check`.

BaxterDevs:
Codex refactor pass complete at the pinned review base. One behavior-preserving naming cleanup was committed as `920fef465631bb9c993ad6f90998a7d52a4a08ee`; focused forecast tests, Ruff check/format, and `git diff --check` passed.

src/aws-pricing-mcp-server/awslabs/aws_pricing_mcp_server/forecast.py
```diff
@@ -0,0 +1,150 @@
...
+        for component in components:
+            baseline = Decimal(str(component.monthly_cost))
+            end_multiplier = Decimal(str(component.end_multiplier))
+            factor = (
+                Decimal('1')
+                if component.scaling_behavior == 'fixed'
+                else Decimal('1') + (end_multiplier - Decimal('1')) * progress
+            )
+            cost = baseline * factor
+            component_costs[component.name] = _money(cost)
```
BaxterDevs: High — correctness/test lens: monetary values are rounded independently at each output level while `forecast_total` is summed from unrounded months. A fixed `0.005` component over three months publishes three `0.01` monthly totals but a `0.02` forecast total; with multiple fractional-cent components, `by_component`, category, and month totals also disagree. Financial consumers cannot audit which view is authoritative. Establish one rounded component ledger, derive every parent total from it, and add half-cent reconciliation tests.

src/aws-pricing-mcp-server/awslabs/aws_pricing_mcp_server/forecast.py
```diff
@@ -0,0 +1,150 @@
...
+    """A baseline monthly cost and its behavior over a forecast horizon."""
+
+    model_config = ConfigDict(extra='forbid')
+
+    name: str = Field(..., min_length=1, max_length=200, description='Component name')
+    category: Optional[str] = Field(
+        None, min_length=1, max_length=100, description='Optional grouping label'
+    )
+    monthly_cost: float = Field(
+        ..., ge=0, allow_inf_nan=False, description='Baseline monthly cost'
```
BaxterDevs: Medium — correctness/security lens: `monthly_cost` accepts every finite nonnegative float, but `_money()` quantizes with Decimal's default 28-digit context. Schema-valid inputs such as `1e26` (or smaller baselines multiplied by the allowed `1_000_000`) raise `decimal.InvalidOperation` instead of returning a forecast. Add a realistic upper bound with aggregate/multiplier headroom and cover the largest accepted and first rejected values through the MCP boundary.

src/aws-pricing-mcp-server/awslabs/aws_pricing_mcp_server/server.py
```diff
@@ -159,6 +171,41 @@ async def create_error_response(
...
+        max_length=MAX_COMPONENTS,
+        description='Baseline monthly cost components with explicit scaling behavior',
+    ),
+    months: int = Field(12, ge=2, le=MAX_MONTHS, description='Forecast horizon in months'),
+    currency: str = Field(
+        'USD',
+        min_length=3,
+        max_length=3,
+        pattern=r'^[A-Z]{3}$',
+        description='ISO 4217 currency code',
```
BaxterDevs: High — API/tenant lens: this field claims ISO 4217 semantics but validates only `[A-Z]{3}`, while the calculator hard-codes cent rounding. `ZZZ` is accepted as a currency; valid zero-minor-unit JPY can emit fractional yen; and KWD values lose their third decimal. Either validate and implement supported currencies' minor units, or make this an explicit uppercase display label and accept an independently validated decimal-place setting. Add JPY/KWD or equivalent boundary coverage.

src/aws-pricing-mcp-server/awslabs/aws_pricing_mcp_server/server.py
```diff
@@ -159,6 +171,41 @@ async def create_error_response(
 )
 
 
+@mcp.tool(
+    name='calculate_cost_scenario',
+    description="""Calculate a deterministic month-by-month cost scenario from classified components.
+
+Use fixed components for architecture units that do not grow with demand, such as control
+planes, proxy capacity, gateway hours, keys, and subscriptions. Use linear components for
+usage-driven costs such as compute, database consumption, requests, storage, data transfer,
```
BaxterDevs: High — docs/tenant-coupling lens: the model-facing guidance categorically maps usage-driven services to a linear cost ramp, but many have tiers, free allowances, commitments, capacity steps, or volume discounts. A second customer's usage multiplier therefore need not be its cost multiplier, even with current unit prices. Restrict linear classification to baseline costs whose effective rate is expected to remain constant over the modeled range, explicitly exclude tiered/stepped/commitment-sensitive cases, and state this assumption in the result.

src/aws-pricing-mcp-server/awslabs/aws_pricing_mcp_server/forecast.py
```diff
@@ -0,0 +1,150 @@
...
+
+        unrounded_monthly_totals.append(month_total)
+        monthly_rows.append(
+            {
+                'month': month_index + 1,
+                'total': _money(month_total),
+                'by_category': {
+                    category: _money(cost) for category, cost in sorted(category_costs.items())
+                },
+                'by_component': component_costs,
```
BaxterDevs, line 147: Medium — security/resource lens: every month repeats caller-controlled component and category labels. The valid maxima (100 components, 120 months, 200/100-character labels) produce roughly a four-megabyte JSON response and near-million-token context amplification. Enforce the server's established response-character budget (or use a compact indexed schema) and test that a valid oversized forecast is rejected before transport.

src/aws-pricing-mcp-server/awslabs/aws_pricing_mcp_server/server.py
```diff
@@ -143,6 +149,12 @@ async def create_error_response(
        - Source of the data (web scraping, API, or websearch)
        - List of attempted data retrieval methods
 
+    # USE CASE 3: MULTI-PERIOD SCENARIO FORECASTING
```
BaxterDevs, line 153: Low — comment/docs lens: adding `USE CASE 3` leaves the instruction preamble's existing claim that the server has “two primary functionalities” internally inconsistent. Update the count or remove it so model-facing navigation matches the three numbered use cases.

src/aws-pricing-mcp-server/awslabs/aws_pricing_mcp_server/forecast.py
BaxterDevs: Fixed in signed commit `aaebfdea0550c0be94883a40e3b17181efbb64ea`. Each component is now rounded once into a Decimal ledger; category, month, component-summary, and forecast totals derive from that ledger. Half-cent reconciliation tests cover component→category→month→forecast invariants. Validation: 270 tests passed with branch coverage; forecast module 100%.

src/aws-pricing-mcp-server/awslabs/aws_pricing_mcp_server/forecast.py
BaxterDevs: Fixed in signed commit `aaebfdea0550c0be94883a40e3b17181efbb64ea`. `monthly_cost` now has a schema-visible 1,000,000,000 ceiling that preserves multiplier, aggregation, and six-decimal headroom; tests cover the maximum accepted cost at the maximum multiplier and rejection above the bound. Full tests, Pyright, Ruff, and scoped pre-commit passed.

src/aws-pricing-mcp-server/awslabs/aws_pricing_mcp_server/server.py
BaxterDevs: Fixed in signed commit `aaebfdea0550c0be94883a40e3b17181efbb64ea`. Currency is now explicitly a three-letter display label and the tool accepts validated `decimal_places` (0–6), which is returned with the result and drives all monetary rounding. JPY-style zero-decimal and KWD-style three-decimal cases are covered, and an actual FastMCP call verified the generated schema and runtime path.

src/aws-pricing-mcp-server/awslabs/aws_pricing_mcp_server/server.py
BaxterDevs: Fixed in signed commit `aaebfdea0550c0be94883a40e3b17181efbb64ea`. Both model-facing instruction surfaces now permit linear scaling only under a constant-effective-rate assumption, explicitly separate tiering/free allowances/commitments/capacity steps/price changes, and return that caveat in every forecast. Full validation passed.

src/aws-pricing-mcp-server/awslabs/aws_pricing_mcp_server/forecast.py
BaxterDevs: Fixed in signed commit `aaebfdea0550c0be94883a40e3b17181efbb64ea`. Forecast serialization is now measured against the server's established 100,000-character response budget and oversized valid requests fail with actionable reduction guidance before MCP transport. A maximum-label, maximum-component, maximum-month regression test covers the amplification case.

src/aws-pricing-mcp-server/awslabs/aws_pricing_mcp_server/server.py
BaxterDevs: Fixed in signed commit `aaebfdea0550c0be94883a40e3b17181efbb64ea` by updating the instruction preamble to match the three numbered use cases. Ruff format/check and the full test suite passed.

BaxterDevs:
Codex deepgrill round 1 complete. Six adversarial findings were fixed in signed commit `aaebfdea0550c0be94883a40e3b17181efbb64ea`; the fix classification is **material**. Validation: 270 tests with branch coverage, Pyright, Ruff check/format, scoped pre-commit security/license hooks, `git diff --check`, and an actual FastMCP schema/runtime probe. The PR remains draft.

src/aws-pricing-mcp-server/awslabs/aws_pricing_mcp_server/forecast.py
```diff
@@ -0,0 +1,178 @@
...
+        for component in components:
+            baseline = Decimal(str(component.monthly_cost))
+            end_multiplier = Decimal(str(component.end_multiplier))
+            factor = (
+                Decimal('1')
+                if component.scaling_behavior == 'fixed'
+                else Decimal('1') + (end_multiplier - Decimal('1')) * progress
+            )
+            cost = _round_money(baseline * factor, quantum)
+            component_costs[component.name] = float(cost)
```
BaxterDevs: Medium — correctness/financial-contract lens: the rounded `Decimal` ledger is discarded at every output boundary (`float(cost)` here, and the same conversion at forecast.py:129, 131, 142, 144-147, 165-168). The ledger's guarantees hold internally but are not what the tool publishes, and this breaks in two distinct ways.

**1. The published payload does not reconcile, at ordinary magnitudes.** Three fixed components of `10.10`, `20.20`, and `0.30` at the default `months`/`decimal_places`:

```
by_component      {'a': 10.1, 'b': 20.2, 'c': 0.3}
sum(by_component) 30.599999999999998
published total   30.6                      -> not equal
sum(month totals) 91.80000000000001
forecast_total    91.8                      -> not equal
```

A randomized sweep at `$1-10k`, `$1M`, and `$1e9` baselines with `decimal_places=2` and `months=12` broke the component→month and month→forecast invariants in 300/300 trials at each magnitude. `test_rounded_breakdowns_reconcile_with_forecast_total` (tests/test_forecast.py:110) asserts exact float equality and passes only because its two identical `0.005` components land on a float-friendly sum; it is not testing a property that holds.

**2. Significant digits are silently dropped near the schema ceiling.** `monthly_cost` permits `1e9` and `end_multiplier` permits `1e6`, so an emitted value can require far more than float64's ~15-17 significant digits. A randomized sweep at those bounds with `decimal_places=6` published `4254749658858957.0` where the ledger held `...956.5` — a half-unit of the stated minor unit lost inside the value itself, not merely in a consumer's summation. The `1e9` ceiling added for fingerprint `0ece27e3...` gives Decimal enough headroom but gives the float output none.

Impact: this is the published-payload half of the guarantee asserted when fingerprint `60358a2a...` was resolved ("Financial consumers cannot audit which view is authoritative"). That fix established the internal ledger and is correct as far as it goes; the response a consumer actually receives still does not reconcile, and above roughly 15 significant digits it is also inaccurate. Filing separately rather than reopening that thread because the root cause differs: that one was inconsistent rounding across levels, this one is loss of the ledger at serialization.

Expected correction, smallest form that makes the contract true:
- Bound the emitted magnitude so every published value round-trips to its exact `Decimal` at `decimal_places` — reject when `forecast_total` scaled by `decimal_places` exceeds `10**15`, with reduction guidance mirroring the existing response-budget guard. `forecast_total` dominates every other emitted value, so one check covers the whole payload. I validated this threshold: under it, zero quantization violations across a randomized sweep of all `decimal_places` 0-6.
- Fix the reconciliation test to assert the invariant that actually holds — equality after rounding the summed floats to `decimal_places` — rather than exact float equality on hand-picked inputs.

Emitting monetary values as fixed-precision strings would make the payload exactly auditable and is the stronger option, but it changes the type of every monetary field and floats are this server's established convention, so I am not proposing it here.

src/aws-pricing-mcp-server/awslabs/aws_pricing_mcp_server/server.py
```diff
@@ -159,6 +174,53 @@ async def create_error_response(
...
+    annotations=ToolAnnotations(readOnlyHint=True),
+)
+async def calculate_cost_scenario(
+    components: List[ForecastCostComponent] = Field(
+        ...,
+        min_length=1,
+        max_length=MAX_COMPONENTS,
+        description='Baseline monthly cost components with explicit scaling behavior',
+    ),
+    months: int = Field(12, ge=2, le=MAX_MONTHS, description='Forecast horizon in months'),
```
BaxterDevs, line 203: Low — API-consistency lens: `months`, `currency`, and `decimal_places` carry `Field(...)` defaults that are not unwrapped, so a non-MCP caller relying on those defaults gets a `FieldInfo` instead of a value:

```python
asyncio.run(calculate_cost_scenario(components=[ForecastCostComponent(...)]))
# TypeError: '<=' not supported between instances of 'int' and 'FieldInfo'
```

The failure surfaces from `calculate_forecast`'s `2 <= months` bound check as an opaque `TypeError`, not as the actionable `ValueError` the module otherwise raises for bad input.

This file already treats that as a known hazard and guards against it in three sibling tools — server.py:476-484, 1056, and 1303 all run `isinstance(..., FieldInfo)` unwrapping before use. The new tool is the only one taking `Field` defaults without it.

Impact is limited: the MCP boundary binds these parameters correctly, so the shipped tool path is unaffected, and `test_mcp_tool_returns_forecast` passes because it calls through that boundary. It matters for direct Python callers and for anyone extending the tool, and the inconsistency is the kind that gets copied.

Expected correction: unwrap the three scalar `FieldInfo` defaults at the top of `calculate_cost_scenario` following the existing pattern, and cover the direct-call-with-defaults path so the convention stays enforced.

src/aws-pricing-mcp-server/awslabs/aws_pricing_mcp_server/forecast.py
BaxterDevs: Fixed in signed commit `cbf054f1cb9572ac64fdd5c5c4be0c2f8d16ae22`.

`calculate_forecast` now bounds the emitted magnitude before building the response: `forecast_total` scaled by `decimal_places` must stay within `MAX_EXACT_MINOR_UNITS` (`10**15`), the float64 exact-integer range. Because `forecast_total` dominates every other emitted amount, that single check covers every published value, and rejection carries the same actionable reduction guidance as the response-budget guard.

The `1e9` × `1e6` corner is now rejected rather than silently publishing lossy digits. Realistic magnitudes are unaffected — a `$1,000,000,000/month` fixed baseline over 120 months still returns a `$120,000,000,000` forecast. `test_monthly_cost_boundary_is_safe_for_maximum_multiplier` was retargeted to assert that corner fails with the actionable `ValueError` rather than inside `Decimal`, preserving the intent of fingerprint `0ece27e3...`, and `test_maximum_monthly_cost_is_accepted_without_a_multiplier` pins the large-but-publishable case.

For the reconciliation half, `test_rounded_breakdowns_reconcile_with_forecast_total` now asserts the invariant that actually holds — equality after rounding the summed floats to `decimal_places` — instead of exact float equality on float-friendly inputs. `test_fractional_cent_breakdowns_reconcile_at_stated_precision` pins the `10.10`/`20.20`/`0.30` regression directly, asserting both that raw float summation misses the published total and that reconciliation holds at the stated precision. `test_every_published_value_round_trips_to_its_decimal` guards the emission path.

I did not switch monetary fields to fixed-precision strings; floats remain this server's convention, and with the magnitude guard in place every published value is now exact at `decimal_places`.

Validation: 275 tests passed (up from 270), `forecast.py` at 100% statement and branch coverage; `pyright` 0 errors; `ruff check` and `ruff format --check` clean; scoped pre-commit hooks including credential, secret, and license checks passed; `git diff --check` clean. A FastMCP probe confirmed the runtime path — `30.6` published for the `10.10`/`20.20`/`0.30` case, and the oversized-magnitude request rejected through the MCP boundary with `forecast total 24000024000000000.000000 cannot be represented exactly at 6 decimal places`.

src/aws-pricing-mcp-server/awslabs/aws_pricing_mcp_server/server.py
BaxterDevs: Fixed in signed commit `cbf054f1cb9572ac64fdd5c5c4be0c2f8d16ae22`. `calculate_cost_scenario` now unwraps the `months`, `currency`, and `decimal_places` `FieldInfo` defaults before delegating, matching the established pattern at server.py:476-484, 1056, and 1303. `components` is required, so it never arrives as a `FieldInfo` and needs no guard.

`test_mcp_tool_applies_field_defaults_on_direct_calls` covers the direct-call path that previously raised `TypeError: '<=' not supported between instances of 'int' and 'FieldInfo'`, asserting the 12-month, `USD`, 2-decimal defaults are applied.

Validation: 275 tests passed; `pyright` 0 errors; `ruff check` and `ruff format --check` clean; scoped pre-commit hooks passed. A FastMCP `list_tools` probe confirmed the generated schema is unchanged by the unwrapping — `required: ['components']`, `months` default 12 (min 2, max 120), `currency` default `USD` with pattern `^[A-Z]{3}$`, `decimal_places` default 2 (min 0, max 6).

BaxterDevs:
Claude deep review round 1 complete against reviewed head `aaebfdea0550c0be94883a40e3b17181efbb64ea`.

Two adversarial findings were confirmed and fixed in signed commit `cbf054f1cb9572ac64fdd5c5c4be0c2f8d16ae22`. Combined fix classification is **material**: the exactness guard changes the accepted input range and the published payload contract. The `FieldInfo` unwrapping is minor on its own.

- Medium — the rounded `Decimal` ledger was discarded at the float output boundary, so the published payload did not reconcile at ordinary magnitudes and silently dropped significant digits near the schema ceiling. Emitted magnitude is now bounded so every published value is exact at `decimal_places`.
- Low — the tool wrapper did not unwrap `Field` defaults, unlike three sibling tools, so direct Python callers hit an opaque `TypeError`.

Validation: 275 tests passed (up from 270) with branch coverage, `forecast.py` at 100% statement and branch; `pyright` 0 errors; `ruff check` and `ruff format --check` clean; scoped pre-commit credential, secret, and license hooks passed; `git diff --check` clean; FastMCP `list_tools` schema probe and runtime `call_tool` probes for both the reconciliation case and the oversized-magnitude rejection.

Claude's once-per-PR refactor pass was not spent — this pass ran adversarial lanes only, so the cleanup latch remains available.

Because this round made a material fix, the loop restarts at Codex: a fresh Codex deepgrill against `cbf054f1cb9572ac64fdd5c5c4be0c2f8d16ae22`, then Claude again. The PR remains draft.

src/aws-pricing-mcp-server/awslabs/aws_pricing_mcp_server/forecast.py
```diff
@@ -0,0 +1,193 @@
...
```
BaxterDevs: Medium — correctness/financial-contract lens: the new `10**15` minor-unit guard makes each individual float's shortest decimal representation round-trip, but it does not make accepted published breakdowns reconcile at the declared precision. With 50 fixed components each costing `9_990_491.047545`, one category, `months=2`, and `decimal_places=6`, an actual JSON round-trip is accepted with `forecast_total=999049104.7545`, yet `round(sum(by_component.values()), 6) == 499524552.377251` while both `by_category['x']` and the month `total` are `499524552.37725`. That is a one-minor-unit disagreement under the exact operation the new reconciliation tests assert. Preserve the rounded ledger in a representation that supports aggregate arithmetic (for example fixed-precision strings/minor units), or impose a proven aggregate-error bound, and add this boundary regression.

src/aws-pricing-mcp-server/awslabs/aws_pricing_mcp_server/forecast.py
```diff
@@ -0,0 +1,193 @@
...
+        'components': component_rows,
+        'summary': {
+            'forecast_months': months,
+            'baseline_monthly_cost': float(monthly_totals[0]),
+            'average_monthly_cost': float(_round_money(forecast_total / Decimal(months), quantum)),
+            'forecast_total': float(forecast_total),
+            'ending_monthly_cost': float(monthly_totals[-1]),
+        },
+    }
+    response_characters = len(json.dumps(response, indent=2))
```
BaxterDevs: Medium — security/resource/API lens: this guard does not bound the actual MCP response and allocates the entire oversized serialization before rejecting it. FastMCP publishes this annotated `Dict` twice: pretty JSON text and `structuredContent.result`. A valid 10-component/37-month request passes because the guard measures 95,547 characters, but `mcp.call_tool()` returns that 95,547-character text plus an 85,358-character compact structured copy (180,905 combined before envelope). At the schema maxima with emoji labels, rejection itself first materializes a 42,939,749-character `json.dumps` string and raised measured peak RSS by roughly 84 MB. Enforce the budget across the actual representations (or emit only one) and count incrementally/preflight so rejected calls cannot allocate the unbounded string; cover both the MCP boundary and worst-case rejection path.

src/aws-pricing-mcp-server/awslabs/aws_pricing_mcp_server/forecast.py
BaxterDevs: Fixed in signed commit `f24bb845c3828d4fe5a61381575a12884624d624`. Every monetary field is now emitted as a fixed-precision decimal string, preserving the rounded Decimal ledger through JSON and making component, category, month, and forecast arithmetic exact when parsed as Decimal. The former float-magnitude guard is no longer needed; schema-maximum calculations retain all digits. A regression with the reported 50 × `9_990_491.047545` case now reconciles exactly at six decimal places, alongside half-cent, zero-decimal, three-decimal, and schema-ceiling coverage.

Validation: 277 full tests passed; 26 focused forecast tests passed with `forecast.py` at 100% statement and branch coverage; Pyright 0 errors; Ruff check/format clean; scoped pre-commit security, secret, and license hooks passed; `git diff --check` clean.

src/aws-pricing-mcp-server/awslabs/aws_pricing_mcp_server/forecast.py
BaxterDevs: Fixed in signed commit `f24bb845c3828d4fe5a61381575a12884624d624`. `calculate_cost_scenario` now opts out of duplicate structured output, so FastMCP emits only the budgeted JSON text. The response-size check uses `JSONEncoder.iterencode` with an early stop instead of allocating the complete oversized string. The near-limit MCP regression asserts one text representation at or below 100,000 characters, and the maximum emoji-label request exercises early rejection without full-string construction.

Validation: 277 full tests passed; 26 focused forecast tests with 100% forecast branch coverage; Pyright/Ruff clean; scoped pre-commit hooks passed. A real `mcp.call_tool()` probe confirmed the result is a single `TextContent`, not a text/structured duplicate.

BaxterDevs:
Codex deepgrill round 2 complete against `cbf054f1cb9572ac64fdd5c5c4be0c2f8d16ae22`. Two medium findings were fixed in signed commit `f24bb845c3828d4fe5a61381575a12884624d624`; combined classification is **material** because monetary output now uses fixed-precision strings and the MCP output contract emits one representation. The response budget also stops incrementally before allocating an oversized JSON string.

Validation: 277 full tests; 26 focused forecast tests with 100% statement and branch coverage; Pyright 0 errors; Ruff check/format clean; scoped pre-commit credential, secret, and license hooks passed; `git diff --check` clean; real FastMCP single-representation probe. Both findings were posted before editing, replied with this fix SHA and validation, and resolved. The PR remains draft. Convergence restarts with fresh Claude review against this head.

src/aws-pricing-mcp-server/awslabs/aws_pricing_mcp_server/forecast.py
```diff
@@ -0,0 +1,200 @@
...
+
+def _format_money(value: Decimal, decimal_places: int) -> str:
+    """Preserve a rounded monetary value as a fixed-precision decimal string."""
+    return f'{value:.{decimal_places}f}'
+
+
+def _fits_response_budget(value: Dict) -> bool:
+    """Count encoded characters without allocating an oversized JSON string."""
+    encoded_characters = 0
+    for chunk in json.JSONEncoder(indent=2).iterencode(value):
```
BaxterDevs: Medium — correctness/API lens: the response-budget guard counts a representation the server never emits, so valid non-ASCII forecasts are rejected far below the limit.

`json.JSONEncoder(indent=2)` defaults to `ensure_ascii=True`, escaping every non-ASCII character to `\uXXXX` (6 characters, or 12 for an astral-plane pair). The text FastMCP actually emits is unescaped UTF-8. Measured against the current head with `structured_output=False` (so there is exactly one emitted representation):

| labels | guard measures | actually emitted |
| --- | --- | --- |
| ASCII, 10 components × 37 months | 98,925 | 98,925 |
| `é`-padded, same shape | 500,525 | 98,925 |
| emoji-padded, same shape | 982,445 | 98,925 |
| emoji-padded, **3 components × 12 months** | 115,173 | 13,302 |

The last row is the impact: a small forecast whose real response is 13 KB — 7.5× under the budget — fails with `forecast response exceeds the 100,000-character limit; reduce components, months, or label lengths`. The guidance is unfollowable, because the size the caller can observe is not the size being measured. Any non-Latin-1 component or category label (Japanese, Chinese, Cyrillic, or accented French/Spanish service names) inflates the count roughly 5–10×, so the effective component/month budget silently collapses for non-English callers.

The guard should measure the representation it is protecting. Expected correction: construct the encoder with `ensure_ascii=False`, matching the emitted text, and add a regression asserting that a non-ASCII-labelled forecast whose emitted text is well under the limit is accepted.

Note this is the safe direction of the error (over-rejection, never over-emission), which is why the existing budget tests pass: `test_oversized_forecast_response_is_rejected` uses 100 components × 120 months, which exceeds the budget under either encoding, and `test_mcp_response_budget_counts_the_only_emitted_representation` uses ASCII labels, where the two counts coincide. Neither test can distinguish the two encodings, so nothing currently pins this behavior.

BaxterDevs: Fixed in signed commit `d7935dfc9819c8f9fde7a8625f0e4a2d0805a172`.

The budget encoder is now constructed with `ensure_ascii=False`, so it counts the same characters FastMCP emits. Re-running the same measurements against the fix, all four shapes are now accepted, and the guard and the emitted text agree exactly:

| labels | guard | emitted | before | after |
| --- | --- | --- | --- | --- |
| ASCII, 10 × 37 | 98,925 | 98,925 | accepted | accepted |
| `é`, 10 × 37 | 98,925 | 98,925 | rejected | accepted |
| emoji, 10 × 37 | 98,925 | 98,925 | rejected | accepted |
| emoji, 3 × 12 | 13,302 | 13,302 | rejected | accepted |

`test_non_ascii_labels_are_budgeted_as_emitted` pins it: a 3-component, 12-month forecast with emoji-padded name and category labels returns a single `TextContent` under a quarter of the budget. It fails on the pre-fix encoder with the `100,000-character limit` error. The over-emission direction stays closed — the guard measures the sole emitted representation, and `structured_output=False` keeps it the only one. `test_oversized_forecast_response_is_rejected` (100 components × 120 months with emoji labels) still rejects, since that shape exceeds the budget under either encoding.

Validation: 278 full tests passed (up from 277); 27 focused forecast tests with `forecast.py` at 100% statement and branch coverage; `pyright` 0 errors; `ruff check` and `ruff format --check` clean; `git diff --check` clean; a real `mcp.call_tool()` probe confirmed one emitted text representation and no structured duplicate.

BaxterDevs:
Claude deep review round 2 complete against reviewed head `f24bb845c3828d4fe5a61381575a12884624d624` (local branch, fork remote, and PR head all verified equal before review). The full PR ledger — 11 threads including resolved and outdated ones — was read first; no prior root cause was reopened.

One adversarial finding was confirmed and fixed in signed commit `d7935dfc9819c8f9fde7a8625f0e4a2d0805a172`. Classification is **material**: the response-budget guard measured `ensure_ascii=True` escaped characters rather than the text FastMCP actually emits, so any non-ASCII component or category label shrank the effective budget 5–10× and rejected valid small forecasts with unfollowable guidance. The accepted-input range changes, which is why this is not a minor fix.

Lenses that returned no new finding at this head, with the verification behind each:

- **Financial reconciliation.** A 400-case randomized sweep across 0/1/2/3/6 decimal places, 2–9 months, mixed fixed and linear components, and baselines spanning `0` to `1e9` found zero violations of component → category → month → forecast-total equality when parsed as `Decimal`, zero first/last-month disagreements with the published `baseline_monthly_cost` and `ending_monthly_cost`, and exact declared precision on every emitted monetary string. The fixed-precision-string contract from round 2's Codex leg holds.
- **Decimal context exhaustion.** At the schema maxima (`monthly_cost=1e9`, `end_multiplier=1e6`, 120 months, 6 decimal places) the widest quantize needs 21 significant digits against the default 28-digit context, so the `InvalidOperation` class of defect from round 1 has real headroom rather than a coincidental margin.
- **Response amplification.** With `structured_output=False`, a real `mcp.call_tool()` probe confirms exactly one emitted representation and no structured duplicate; the 100-component × 120-month worst case still rejects via early-stopping `iterencode` without materializing the full string.
- **Direct-call and boundary paths**, per-component versus summary aggregate consistency, `months=2` progress endpoints, zero-cost and scale-down components, and the `end_multiplier=0` baseline case.

Not raised as a finding, recorded for the ledger: this tool signals input errors by raising `ValueError` (surfaced to the model as a `ToolError` carrying the message) while the server's older tools return `create_error_response` payloads. The message text does reach the caller intact, so the error is actionable either way; this is a convention difference in a `structured_output=False` tool, not a defect, and changing it would move the published contract for no correctness gain.

The refactor pass was not run: `/refactorpass` is once-per-engine-per-PR and spends on that engine's first pass, and this is Claude's round 2. No `local-review-refactor:v1 engine=claude` marker exists, and none is posted, since the cleanup lanes did not run.

Validation: 278 full tests passed; 27 focused forecast tests with `forecast.py` at 100% statement and branch coverage; `pyright` 0 errors; `ruff check` and `ruff format --check` clean; `git diff --check` clean.

The finding was posted inline before any edit, replied to with the fix SHA and validation, and resolved. Every review thread on this PR now has a disposition reply and is resolved. The PR remains draft.

**Convergence is not reached.** This round produced a material fix, so the loop restarts at Codex: a fresh Codex review and then a fresh Claude review are both required against `d7935dfc9819c8f9fde7a8625f0e4a2d0805a172`. Those are rounds 3 and run in convergence mode — blocking defects only, smallest fix, everything else to a linked issue.

BaxterDevs:
Codex deepgrill round 3 convergence pass complete against pinned base `6ddf48bb7a1d1297376c07e62d3d99fa11927f93` and exact head `d7935dfc9819c8f9fde7a8625f0e4a2d0805a172`: **no new material or release-blocking findings**.

Blocking-only lanes covered correctness/API, tests/boundaries, and security/resource behavior. Evidence: 27 focused forecast tests; `forecast.py` 100% statement and branch coverage; Pyright 0 errors; Ruff check/format and `git diff --check` clean; 500 randomized forecasts reconciled component/category/month/summary Decimal totals exactly; 100 randomized Unicode/control-character cases matched the budget encoder to the actual FastMCP `TextContent`; accepted non-ASCII output remained bounded and the worst schema shape rejected. Worktree stayed clean. The PR remains draft.

BaxterDevs:
Claude round 3 convergence pass complete against pinned base `6ddf48bb7a1d1297376c07e62d3d99fa11927f93` and exact head `d7935dfc9819c8f9fde7a8625f0e4a2d0805a172`: **no new material or release-blocking findings**.

Local branch, fork remote, and PR head were all verified equal at that SHA with a clean worktree before review, and the full ledger — 12 threads including resolved and outdated ones — was read first. No prior root cause was reopened and no edit was made.

Blocking-only lenses (correctness, silent failure, security/resource) covered the complete base-to-head diff independently. Evidence: 27 focused forecast tests pass; Ruff check and format clean; 300 randomized forecasts across 0/2/3/6 decimal places reconciled component → category → month → forecast totals exactly as parsed `Decimal`, with zero mismatches; a real FastMCP session returned exactly one `TextContent` and no duplicate `structuredContent` for ASCII, `é`, and emoji labels at identical character counts (33,215), confirming the `ensure_ascii=False` budget encoder measures the text actually emitted; the largest accepted shape emitted 99,721 characters, inside the 100,000 budget.

The PR remains draft.

src/aws-pricing-mcp-server/awslabs/aws_pricing_mcp_server/forecast.py
```diff
@@ -0,0 +1,200 @@
...
+        raise ValueError(f'components must contain between 1 and {MAX_COMPONENTS} items')
+    if not re.fullmatch(r'[A-Z]{3}', currency):
+        raise ValueError('currency must be a three-letter uppercase display label')
+    if not 0 <= decimal_places <= MAX_DECIMAL_PLACES:
+        raise ValueError(f'decimal_places must be between 0 and {MAX_DECIMAL_PLACES}')
+
+    component_names = [component.name for component in components]
+    if len(component_names) != len(set(component_names)):
+        raise ValueError('component names must be unique')
+
```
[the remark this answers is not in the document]
BaxterDevs: calculate_forecast intentionally accepts validated ForecastCostComponent instances. Raw dictionaries are validated and coerced at the FastMCP boundary; direct Python callers must construct the typed model first.

PR #4480, opened 2026-08-11
fix(ecs-mcp-server): add path validation for user-supplied paths

matthewgoodman13:
Fixes N/A

### Changes

Add a shared `path_validation` module to `ecs-mcp-server`, mirroring the one in
`eks-mcp-server`.

`validate_file_path()` in `utils/security.py` only pattern-matched the raw string for
`../` sequences, so an absolute path, a `~` prefix, or a symlink pointing at a user
configuration directory was accepted. `prepare_template_files()` also swallowed validation
errors whose message contained `does not exist`, so a rejected path could still be created.

The new `validate_path()`:

* Expands `~` and resolves paths with `os.path.realpath`, collapsing `..` and following
  symlinks before any check
* Rejects paths resolving inside a sensitive directory (`~/.aws`, `~/.ssh`, `~/.kube`,
  `~/.gnupg`, `~/.docker`, `/etc`, `/root`, `/var/lib`), and paths containing one, since a
  parent such as `$HOME` would pull it into a Docker build context
* Derives the home directory entries from both `$HOME` and the password database
* Accepts relative paths and returns the resolved absolute path

Applied in `prepare_template_files` (before `os.makedirs` and the template writes),
`build_and_push_image` (before the directory becomes the `docker buildx build` context), and
`validate_cloudformation_template`. `validate_file_path` is removed; every caller now uses the
new module.

Two differences from the EKS module: relative paths are accepted rather than rejected, and one
`validate_path(path, must_exist=False)` replaces two functions with identical bodies.

### User experience

Before, `app_path` was used as written, so `~/.aws` — or a symlink resolving to it — became
the template output directory and the Docker build context. Now such a path is rejected
before anything is written or built:

```
Path '~/.aws' resolves to '/home/user/.aws', which is not usable because it would
expose the sensitive directory '/home/user/.aws'
```

Project directories, absolute or relative, are unaffected.

### Testing

787 tests pass, up from 722 on `main`. `path_validation.py` is at 100% line and branch
coverage, covering symlink, `~`, relative, traversal, ancestor, prefix-collision and
case-insensitive-filesystem inputs. `ruff check`, `ruff format`, `pyright`, `bandit` and
`uv build` are clean.

## Checklist

* [x] Changes are documented

Paths resolving inside or containing one of the listed directories are now rejected where they
were previously accepted. No other input is affected.

**RFC issue number**: N/A

Checklist:

## Acknowledgment

### DIFF
--- a/src/ecs-mcp-server/awslabs/ecs_mcp_server/api/infrastructure.py
+++ b/src/ecs-mcp-server/awslabs/ecs_mcp_server/api/infrastructure.py
@@ -27,11 +27,11 @@ from awslabs.ecs_mcp_server.utils.aws import (
     get_aws_client_with_role,
     get_default_vpc_and_subnets,
 )
+from awslabs.ecs_mcp_server.utils.path_validation import validate_path
 from awslabs.ecs_mcp_server.utils.security import (
     ValidationError,
     validate_app_name,
     validate_cloudformation_template,
-    validate_file_path,
 )
 from awslabs.ecs_mcp_server.utils.templates import get_templates_dir
 
@@ -46,7 +46,7 @@ def prepare_template_files(app_name: str, app_path: str) -> Dict[str, str]:
 
     Args:
         app_name: Name of the application
-        app_path: Path to the application directory
+        app_path: Path to the application directory. Created if it doesn't exist.
 
     Returns:
         Dict containing paths to the template files
@@ -57,16 +57,12 @@ def prepare_template_files(app_name: str, app_path: str) -> Dict[str, str]:
     # Validate app_name
     validate_app_name(app_name)
 
-    # For app_path, we'll validate it but handle the case where it doesn't exist
+    # Resolve and validate app_path before writing anything to disk. The directory
+    # is created below when missing, so existence is not required here.
     try:
-        validate_file_path(app_path)
-    except ValidationError as e:
-        # If the path doesn't exist, we'll create it
-        if "does not exist" not in str(e):
-            # Some other validation error occurred
-            raise ValidationError(str(e)) from e
-        # Otherwise, we'll continue and create the directory later
-        logger.debug(f"Path {app_path} does not exist, will create it")
+        app_path = validate_path(app_path)
+    except ValueError as e:
+        raise ValidationError(str(e)) from e
 
     # Create templates directory (this will create app_path if it doesn't exist)
     templates_dir = os.path.join(app_path, "cloudformation-templates")
--- a/src/ecs-mcp-server/awslabs/ecs_mcp_server/utils/docker.py
+++ b/src/ecs-mcp-server/awslabs/ecs_mcp_server/utils/docker.py
@@ -24,6 +24,7 @@ import time
 from typing import Optional
 
 from awslabs.ecs_mcp_server.utils.aws import get_aws_account_id, get_aws_client
+from awslabs.ecs_mcp_server.utils.path_validation import validate_path
 
 logger = logging.getLogger(__name__)
 
@@ -84,10 +85,15 @@ async def build_and_push_image(
         Image tag
 
     Raises:
-        ValueError: If role_arn is not provided
+        ValueError: If role_arn is not provided, or if app_path is invalid
     """
     if not role_arn:
         raise ValueError("role_arn is required for ECR authentication")
+
+    # The whole directory is handed to the Docker daemon as the build context, so
+    # resolve it and reject anything that would pull in a sensitive directory.
+    app_path = validate_path(app_path)
+
     # Generate a timestamp-based tag if none provided
     if tag is None:
         tag = str(int(time.time()))
--- a/src/ecs-mcp-server/awslabs/ecs_mcp_server/utils/path_validation.py
+++ b/src/ecs-mcp-server/awslabs/ecs_mcp_server/utils/path_validation.py
@@ -1,0 +1,174 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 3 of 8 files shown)


nineonine (approved):

src/ecs-mcp-server/awslabs/ecs_mcp_server/api/infrastructure.py
```diff
@@ -57,16 +57,12 @@ def prepare_template_files(app_name: str, app_path: str) -> Dict[str, str]:
...
     try:
-        validate_file_path(app_path)
-    except ValidationError as e:
-        # If the path doesn't exist, we'll create it
-        if "does not exist" not in str(e):
-            # Some other validation error occurred
-            raise ValidationError(str(e)) from e
-        # Otherwise, we'll continue and create the directory later
-        logger.debug(f"Path {app_path} does not exist, will create it")
+        app_path = validate_path(app_path)
```
scottschreckengaust, line 63: Non-blocking (behavioral narrowing): `validate_path` rejects any *ancestor* of a sensitive directory (`...would expose the sensitive directory ...`). For `prepare_template_files` that's broader than the stated goal — this path only receives a `cloudformation-templates/` subdirectory written into it; it isn't handed to the Docker daemon as a build context. So an `app_path` of `$HOME` (or `/home`, `/var`) that could previously hold generated templates is now rejected. The "pulls it into a Docker build context" rationale really only applies to `build_and_push_image`. Fine to keep as-is (it's called out in the PR's breaking-change note) — just flagging the narrowing.

src/ecs-mcp-server/awslabs/ecs_mcp_server/utils/path_validation.py
```diff
@@ -0,0 +1,174 @@
...
+    candidates = [
+        os.path.join(home, sensitive_dir)
+        for home in _home_dirs()
+        for sensitive_dir in _SENSITIVE_HOME_DIRS
+    ]
+    candidates.extend(_SENSITIVE_SYSTEM_DIRS)
+    return tuple(dict.fromkeys(os.path.realpath(candidate) for candidate in candidates))
+
+
+BLOCKED_DIRS: Tuple[str, ...] = _resolve_blocked_dirs()
```
scottschreckengaust, line 86: Non-blocking (latent assumption): `BLOCKED_DIRS` is computed once at import from `$HOME`/the password DB and then frozen. If the process's `$HOME` changes after import (e.g. started with a temporary `$HOME` that is later set to the real home), the home-based entries validate against the stale import-time home; system dirs like `/etc` are unaffected. Realistically `$HOME` is stable, so this is a note rather than a change — worth a one-line comment that the value is import-time only.

src/ecs-mcp-server/awslabs/ecs_mcp_server/utils/docker.py
```diff
@@ -84,10 +85,15 @@ async def build_and_push_image(
...
     Raises:
-        ValueError: If role_arn is not provided
+        ValueError: If role_arn is not provided, or if app_path is invalid
     """
     if not role_arn:
         raise ValueError("role_arn is required for ECR authentication")
+
+    # The whole directory is handed to the Docker daemon as the build context, so
+    # resolve it and reject anything that would pull in a sensitive directory.
+    app_path = validate_path(app_path)
```
scottschreckengaust, line 95: Two minor notes here, neither blocking:

- **Redundant with `prepare_template_files`**: both existing call paths (`create_infrastructure` and `express.build_and_push_image_to_ecr`) already `validate_path` the same `app_path` before reaching here, so it's resolved and blocklist-checked twice per deployment. Harmless defense-in-depth for a hypothetical direct caller.
- **`must_exist=False`**: a nonexistent build context passes validation and only fails later with `FileNotFoundError: Dockerfile not found at ...`. Since a build context that doesn't exist can never build, `must_exist=True` here would fail faster with a clearer message.

src/ecs-mcp-server/tests/unit/utils/test_docker.py
```diff
@@ -76,6 +77,70 @@ async def test_build_and_push_image_missing_role_arn():
...
+
+@pytest.mark.anyio
+@patch("awslabs.ecs_mcp_server.utils.docker.subprocess.run")
+@patch("awslabs.ecs_mcp_server.utils.docker.get_aws_account_id")
+async def test_build_and_push_image_rejects_symlinked_sensitive_build_context(
+    mock_get_aws_account_id, mock_subprocess_run, tmp_path
+):
+    """A symlink into a sensitive directory must be resolved and rejected."""
+    link = tmp_path / "my-app"
+    os.symlink(os.path.join(os.path.expanduser("~"), ".aws"), str(link))
```
scottschreckengaust, line 130: Non-blocking (test portability): the sensitive/symlink build-context tests depend on the real user home (`os.path.expanduser('~')`) and `os.symlink`. `os.symlink` needs elevated privileges on Windows, and the sensitive-path assertions assume `~/.aws` maps into `BLOCKED_DIRS`. They pass on Linux CI and are self-consistent with how `BLOCKED_DIRS` is derived, so this is just a portability note.

scottschreckengaust (approved):
Reviewed at extra-high effort across all finder angles: read every changed file plus the enclosing functions and callers, verified the `os` import removal in `security.py` is clean, exercised the `_find_blocked_dir` matcher against edge cases (equality / inside / ancestor / prefix collision like `/etcetera` and `.awsome-app` / case-insensitive FS / symlink / `..` traversal), and ran the full unit suite (**787 passed**, matching the PR claim).

This is a solid, careful, well-tested security fix. The removed behavior (`validate_file_path`'s existence + traversal checks) is fully re-established through `validate_path(..., must_exist=True)` plus realpath resolution, and `delete.py` is transitively hardened because `validate_cloudformation_template` now routes through the new module. I found **no correctness bugs**. The inline notes below are all low / very-low severity observations (behavioral narrowing, a latent import-time assumption, minor redundancy, test portability) — none are blocking.

**Decision: Approve.**

PR #4482, opened 2026-08-12
feat(security-agent-mcp-server): Add custom User-Agent

ljainiaz:
## Summary

  Adds a custom User-Agent string to all boto3 API calls from the security-agent-mcp-server, enabling service-side tracking of MCP usage by IDE/tool.

  ### User-Agent format

  ```
  md/awslabs#mcp#security-agent-mcp-server#0.1.2 md/client#kiro/1.5.0
  ```

  Follows the repo-wide `md/awslabs#mcp#<server-name>#<version>` convention, with an additional `md/client#<name>/<version>` suffix for IDE tracking.

  ### How it works

  1. **`aws_client.py`**: `SecurityAgentClient` builds a `botocore.config.Config(user_agent_extra=...)` and passes it to every `.client()` call (securityagent, sts, s3, iam). Uses the package `__version__` 
  constant directly.
  2. **`server.py`**: `@ensure_client_ua` decorator extracts `clientInfo` from the MCP session context (sent by the IDE during `initialize`) and calls `set_mcp_client_info()` on the shared AWS client singleton.

  ### Design decisions

  - **`__version__` constant**: Uses the package-level `__version__` instead of `importlib.metadata.version()` to avoid silent degradation to "unknown" in editable/source installs.
  - **Sanitized client name**: `clientInfo.name` is normalized with `.lower().replace(" ", "-")` before injection into the UA string, preventing malformed tokens (e.g. "Claude Code" → "claude-code").
  - **Decorator pattern**: `@ensure_client_ua` applied to all tool functions as a single choke point — no per-tool boilerplate, and new tools cannot forget it.
  - **Lazy injection**: The `SecurityAgentClient` singleton is created at module level before any MCP session connects, so `clientInfo` is injected on first tool invocation.
  - **No-op optimization**: `set_mcp_client_info()` short-circuits if the info has not changed.
  - **Graceful fallback**: Falls back to `unknown` when `clientInfo` is unavailable.
  - **Concurrency note**: Current stdio transport guarantees one client per process. Under future HTTP/SSE transport with concurrent sessions, `_config` would need per-session isolation.

  ### Tracking values by IDE

  | IDE | md/client# value |
  |-----|-----------------|
  | Kiro | kiro/\<version\> |
  | Claude Code | claude-code/\<version\> |
  | Cursor | cursor/\<version\> |
  | Unknown | unknown |

  ### Tested

  - All 41 unit tests pass (197 total including existing)                                                                                                                                     
  - E2E verified: API call to SecurityAgent service confirmed the custom UA on the wire
  - Works with Kiro, Claude Code, and any MCP-compliant client that sends clientInfo

  ---
  
### DIFF
--- a/src/security-agent-mcp-server/awslabs/security_agent_mcp_server/aws_client.py
+++ b/src/security-agent-mcp-server/awslabs/security_agent_mcp_server/aws_client.py
@@ -15,25 +15,67 @@
 """AWS SecurityAgent API client using boto3 SDK."""
 
 import boto3
+import botocore.config
 import json
 import re
+from awslabs.security_agent_mcp_server import __version__
 from typing import Any, Optional
+
+
+DEFAULT_MCP_CLIENT_NAME = 'unknown'
 
 
 class SecurityAgentClient:
     """Client for AWS SecurityAgent APIs using boto3."""
 
-    def __init__(self, region: str = 'us-east-1'):
+    def __init__(
+        self,
+        region: str = 'us-east-1',
+        mcp_client_name: str = DEFAULT_MCP_CLIENT_NAME,
+        mcp_client_version: str = '',
+    ):
         """Initialize SecurityAgent client."""
         self.region = region
+        self._mcp_client_name = mcp_client_name
+        self._mcp_client_version = mcp_client_version
+        self._config = self._build_config(mcp_client_name, mcp_client_version)
+
+    def _build_config(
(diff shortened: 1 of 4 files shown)


jigar-lab (approved):

scottschreckengaust (approved):

scottschreckengaust (approved):
LGTM

PR #4485, opened 2026-08-13
feat(aws-healthomics): auto-wrap single workflow files into ZIP for CreateWorkflow

hmkim:
## Problem

When `definition_source` points to a bare `.wdl`/`.nf`/`.cwl` file (not a ZIP archive), `CreateAHOWorkflow` fails asynchronously with `statusMessage: "Invalid zip file"`. Users must manually ZIP the file or call `PackageAHOWorkflow` first.

## Solution

Detect non-ZIP content in `validate_definition_sources()` by checking the first 4 bytes for the ZIP magic number (`PK\x03\x04`). If the content is not a ZIP, auto-wrap it into a single-entry ZIP archive:
- **Local files**: uses the original filename (e.g. `hello.wdl`)
- **S3 URIs / inline content**: defaults to `main.wdl`

The wrapping is logged at `INFO` level for debugging but not surfaced to the user.

## Testing

Added 4 unit tests in `TestAutoZipWrapping`:
- Local WDL file auto-wrapped with original filename
- Inline content auto-wrapped as `main.wdl`
- Existing ZIP content not double-wrapped
- S3 URI content auto-wrapped with default name

All 113 existing tests continue to pass.

## Related

This was discovered during an end-to-end WDL migration session (draft-2 → WDL 1.1 → HealthOmics). See broader improvement proposals discussed with @markread-bwb.

Registered a workflow against a live account (`us-east-1`) from a bare `.wdl` file — the case that previously failed with `Invalid zip file`:

```
INFO  definition_source "/tmp/hello_verify.wdl" is a bare workflow definition file
      rather than a ZIP archive; wrapping it as "hello_verify.wdl" so HealthOmics
      can accept it
```

The workflow reached `ACTIVE`, and `GetWorkflow` reported `main: hello_verify.wdl` with the parameter template resolved from the definition — so the single-entry archive is enough for the service to locate the main file without `path_to_main`, and the name does not have to be `main.wdl`.

A run started from that workflow completed successfully, confirming the archive is not merely accepted at registration but usable.

---


markjschreiber (approved):
Nice feature!

PR #4486, opened 2026-08-13
feat(aws-healthomics): add ValidateAHORunReadiness pre-flight check tool

hmkim:
## Problem

A misconfigured run fails several minutes in, after it has already moved through `PENDING`. Causes surface one at a time, so sorting out a single run can take several attempts. In one migration session it took four:

1. `IAM role is invalid or inaccessible` — role could not be assumed by HealthOmics
2. `S3 bucket not located in us-east-1 region` — output bucket in another region
3. `S3 access denied for s3://...` — role lacked write permission
4. `InputError: couldn't construct SampleMeta` — parameter type issue

Roughly 15 minutes spent discovering problems that were all knowable up front.

## Solution

`ValidateAHORunReadiness` checks the prerequisites without starting a run. Every check is attempted regardless of earlier results, so one call reports all the problems rather than stopping at the first.

| # | Check | Method |
|---|-------|--------|
| 1 | Workflow exists and is ACTIVE | `omics:GetWorkflow` (surfaces `statusMessage`) |
| 2 | IAM role exists | `iam:GetRole` |
| 3 | Trust policy permits **this** run | parse `AssumeRolePolicyDocument`, including its `Condition` block |
| 4 | Output bucket exists | `s3:HeadBucket` |
| 5 | Output bucket region matches run region | `s3:GetBucketLocation` |
| 6 | Role can write to output prefix | `iam:SimulatePrincipalPolicy` |
| 7 | Container images pullable by HealthOmics | delegates to `check_container_availability` |
| 8 | Input S3 objects exist | `s3:HeadObject` |

### Design notes

**Reuses existing tooling.** Check 7 calls `check_container_availability` from `ecr_tools` rather than reimplementing ECR lookups, following the composition pattern `validate_healthomics_ecr_config` already uses.

**Nested parameters.** Parameters are walked recursively, so `File` inputs inside Structs, Arrays and Maps are all found — not just top-level string values.

**Trust is evaluated, not just detected.** Naming `omics.amazonaws.com` as a trusted principal is necessary but not sufficient. AWS documents a confused-deputy guard for this trust policy that scopes it by region and account:

```json
"Condition": {
  "StringEquals": {"aws:SourceAccount": "<account>"},
  "ArnLike": {"aws:SourceArn": "arn:aws:omics:ap-northeast-2:<account>:run/*"}
}
```

A role scoped this way cannot be assumed for a run in another region, and `StartRun` rejects it with `IAM role is invalid or inaccessible`. `evaluate_healthomics_trust` therefore returns a status rather than a boolean: `pass` when a statement permits this region and account, `fail` when the principal is named but every matching statement is excluded by a condition it can evaluate, and `warn` when a matching statement carries condition keys it cannot evaluate. Only `aws:SourceArn` and `aws:SourceAccount` are interpreted; wildcards permit; a restrictive statement does not mask a permissive one later in the document.

**Warnings do not block.** A check that cannot be completed reports `warn` and leaves `ready` unaffected. Check 6 needs `iam:SimulatePrincipalPolicy` on the caller; where that is missing the run may still succeed, so treating it as a failure would produce false negatives. The same applies to input objects the caller cannot read but the run role can.

**Containers are supplied explicitly.** `container_images` is an explicit parameter rather than something extracted from the workflow definition. Parsing container URIs out of WDL, Nextflow and CWL would couple this tool to three grammars; callers already have the definition in hand.

### Return shape

```json
{
  "ready": false,
  "summary": "1 failed, 1 warnings, 5 passed (7 checks)",
  "checks": [
    {"name": "workflow_active", "status": "pass", "detail": "Workflow 3344054 is ACTIVE"},
    {"name": "output_bucket_region", "status": "fail",
     "detail": "Bucket my-bucket is in ap-northeast-2 but the run region is us-east-1..."},
    {"name": "role_s3_write", "status": "warn",
     "detail": "Could not simulate role permissions (AccessDenied)..."}
  ]
}
```

## Testing

24 tests in `tests/test_run_validation.py` covering the tool and its three helpers:

- `parse_ecr_image_uri` — tag, digest, implicit `latest`, and rejection of public registry URIs
- `trust_policy_allows_healthomics` — string and list principals, Deny statements, JSON-string policies, malformed input
- `extract_s3_uris` — flat values, nested Structs and Arrays, deduplication
- `validate_run_readiness` — all-pass, region mismatch, missing trust policy, simulation denied (warn, not fail), non-ACTIVE workflow, missing input object, container delegation, public container URI, malformed output URI, unexpected exception

`ruff check` and `ruff format` clean.

## Related

Discovered during an end-to-end WDL migration (draft-2 → WDL 1.1 → HealthOmics). Companion to #4485.

Run against a live account (`us-east-1`), which is what surfaced the trust-condition gap above. An earlier session had lost a run to `IAM role is invalid or inaccessible`; the role turned out to name the omics principal but restrict it to `ap-northeast-2` runs. The first version of this tool reported that role's trust check as `pass`.

The run was still reported not ready, but only because an unrelated check happened to fail:

```
[ok  ] role_trust_policy                <- wrong
[FAIL] role_s3_write   implicitDeny
```

A role with the same region restriction and working S3 permissions would have been reported ready. After the fix:

```
[FAIL] role_trust_policy  omics.amazonaws.com is named in the trust policy, but no statement
                          permits this run: aws:SourceArn restricts this role to runs in
                          ap-northeast-2, but the run region is us-east-1. StartRun reports
                          this as "IAM role is invalid or inaccessible".
```

Other checks confirmed against the real service:

| Scenario | Result |
| --- | --- |
| Output bucket in `ap-northeast-2`, run in `us-east-1` | `fail` on `output_bucket_region` |
| Bucket the role cannot write to | `fail` on `role_s3_write` (`implicitDeny` from policy simulation) |
| Nonexistent input object | `fail` on `input:` |
| `quay.io/biocontainers/bwa:0.7.17` as a container | `fail`, not a private ECR URI |
| Configuration proven by a COMPLETED run | `ready=true`, 7 checks passed |

The `us-east-1` handling in check 5 is worth noting: `GetBucketLocation` returns a null `LocationConstraint` for that region, which the mock-based tests encode and the live run confirms.

---


src/aws-healthomics-mcp-server/awslabs/aws_healthomics_mcp_server/tools/run_validation.py
```diff
@@ -0,0 +1,667 @@
...
+                            f'This check needs iam:SimulatePrincipalPolicy; the run may still '
+                            f'succeed.',
+                        )
+                    )
+
+        # --- Check 7: container images ---
+        if container_images:
+            for image_uri in container_images:
+                parts = parse_ecr_image_uri(image_uri)
+                if parts is None:
```
markjschreiber, line 558: This would fail on something like `ubuntu:latest` but if the workflow has a registry map to dockerhub with ECR pull through caching it should succeed.

src/aws-healthomics-mcp-server/awslabs/aws_healthomics_mcp_server/tools/run_validation.py
```diff
@@ -0,0 +1,667 @@
...
+                        _check(
+                            f'input:{uri}',
+                            STATUS_WARN,
+                            'URI has no key; skipped object existence check',
+                        )
+                    )
+                    continue
+
+                try:
+                    s3_client.head_object(Bucket=bucket, Key=key)
```
markjschreiber, line 625: S3 Input URIs can include prefixes (Directories) and in the case of Nextflow can also include glob patterns. These could all fail here.

markjschreiber (changes requested):
Looks good but I think that some checks will fail valid cases

PR #4489, opened 2026-08-13
feat(billing-cost-management): add budget-actions and budget-notifications operations

lbragile-amzn:
### Changes
  
Adds two read-only operations to the `budget` tool in the billing-cost-management-mcp-server:

  - **`budget-actions`** — retrieves budget actions via `DescribeBudgetActionsForAccount`. Always account-scoped; when a `budget_name` is provided, results are filtered client-side to that budget.
  - **`budget-notifications`** — retrieves notifications for a specific budget via `DescribeNotificationsForBudget` (`budget_name` required).

Both operations follow the existing MCP `{status, data}` response envelope and reuse the tool's established formatter conventions. Includes new test coverage; CHANGELOG updated.

### User experience
  
Before: agents could list budgets and their details, but had no way to surface the configured **actions** (e.g. IAM/SCP/target-policy responses) or **notification** thresholds attached to a budget without leaving the MCP tooling.
  
After: agents can retrieve budget actions (account-wide or filtered to one budget) and per-budget notifications directly through the `budget` tool, all read-only.

## Checklist
  
  * [x] Changes are documented
  
**RFC issue number**: N/A

## Acknowledgment

### DIFF
--- a/src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/budget_tools.py
+++ b/src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/budget_tools.py
@@ -14,7 +14,12 @@
 
 """AWS Budgets tools for the AWS Billing and Cost Management MCP server.
 
-Updated to use shared utility functions.
+File layout:
+- Tool entrypoints: the @budget_server.tool functions the MCP client sees
+  (`budgets`, `budget-actions`, `budget-notifications`).
+- Account helper: `get_aws_account_id`.
+- Operation handlers: the read calls each tool delegates to.
+- Formatters: pure reshapers from raw AWS API dicts to the response shape.
 """
 
 from ..utilities.aws_service_base import create_aws_client, format_response, handle_aws_error
@@ -24,6 +29,11 @@ from typing import Any, Dict, List, Optional
 
 
 budget_server = FastMCP(name='budget-tools', instructions='Tools for working with AWS Budgets API')
+
+
+# =============================================================================
+# Tool entrypoints
+# =============================================================================
 
 
 @budget_server.tool(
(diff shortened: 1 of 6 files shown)


src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/budget_tools.py
```diff
@@ -178,6 +292,114 @@ async def describe_budgets(
...
+    """
+    try:
+        budgets_client = create_aws_client('budgets', region_name='us-east-1')
+
+        request_params: Dict[str, Any] = {'AccountId': account_id}
+        all_actions: List[Dict[str, Any]] = []
+        next_token = None
+        page_count = 0
+
+        while True:
```
callnmm: [non-blocking] One concern here is that this could result in high latency if a customer has a large amount of budgets with actions in their account. You might want a max ceiling and a way for the agent to collect to do pagination in such scenarios. Alternatively, you could call `DescribeBudgetActionsForAccount` when filtering by account. I know you want to avoid multiple IAM actions, but it could be helpful in this scenario.

callnmm (dismissed):

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/budget_tools.py
```diff
@@ -178,6 +292,114 @@ async def describe_budgets(
         return await handle_aws_error(ctx, e, 'describe_budgets', 'AWS Budgets')
 
 
+async def describe_budget_actions(
+    ctx: Context, account_id: str, budget_name: Optional[str], max_results: int
```
wangyuhere: For paginated APIs, we also need `next_token` and `max_pages` see https://github.com/awslabs/mcp/blob/main/src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/cost_allocation_tags_tools.py#L67-L69

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/budget_tools.py
```diff
@@ -178,6 +292,114 @@ async def describe_budgets(
...
+) -> Dict[str, Any]:
+    """Retrieves budget enforcement actions via DescribeBudgetActionsForAccount.
+
+    Always calls the account-scoped API (one permission,
+    `budgets:DescribeBudgetActionsForAccount`, covers both the account-wide audit and a
+    single-budget lookup). If `budget_name` is provided, the account-wide result is filtered to
+    that budget client-side — the AWS API has no `BudgetName` parameter for the account call.
```
wangyuhere: Why not call `DescribeBudgetActionsForBudget` API when `budget_name` is provided? https://docs.aws.amazon.com/aws-cost-management/latest/APIReference/API_budgets_DescribeBudgetActionsForBudget.html

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/budget_tools.py
wangyuhere: +1 checking https://docs.aws.amazon.com/cost-management/latest/userguide/management-limits.html management account can have up to 20k budgets which will not be loaded all at once.

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/budget_tools.py
```diff
@@ -178,6 +292,114 @@ async def describe_budgets(
...
+
+        data: Dict[str, Any] = {
+            'actions': [format_action(a) for a in all_actions],
+            'total_count': len(all_actions),
+            'account_id': account_id,
+        }
+        if budget_name:
+            data['budget_name'] = budget_name
+
+        return format_response('success', data)
```
wangyuhere: Since the response can be huge for large customer, we can do SQL offload to save tokens see https://github.com/awslabs/mcp/blob/main/src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/utilities/sql_utils.py#L470

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/budget_tools.py
```diff
@@ -79,6 +89,105 @@ async def budgets(
...
+    except Exception as e:
+        return await handle_aws_error(ctx, e, 'budget-actions', 'AWS Budgets')
+
+
+@budget_server.tool(
+    name='budget-notifications',
+    description="""Reads the alert thresholds (notifications) configured on an AWS Budget.
+
+A notification is an alert that fires when a budget crosses a threshold (e.g. 80% of the limit).
+Read-only; uses the DescribeNotificationsForBudget API, which is per-budget — `budget_name` is
```
wangyuhere: There is also `DescribeBudgetNotificationsForAccount` API, since we are building the `budget-notifications` tool, we can cover both API.

wangyuhere (commented):
For paginated tool, please consider use https://github.com/awslabs/mcp/blob/main/src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/utilities/aws_service_base.py#L286

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/budget_tools.py
lbragile-amzn: Good call. Fixed both ways. When budget_name is given we now call DescribeBudgetActionsForBudget directly (no account-wide scan + client-side filter), and the account-wide path is paginated via the shared paginate_aws_response helper with max_pages/max_results bounds.

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/budget_tools.py
lbragile-amzn: Done, both tools now expose next_token and max_pages and paginate through paginate_aws_response

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/budget_tools.py
lbragile-amzn: Agree, I wanted to reduce the permissions required and saw that the list budgets call followed the same approach, but can adjust.

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/budget_tools.py
lbragile-amzn: Interesting, wasn't aware. I registered budget_actions/budget_notifications to the generic records converter, so an oversized response offloads as one queryable row per action/notification (real columns) rather than a single JSON blob. Let me know if anything should be adjusted for this

wangyuhere (commented):
Please also update https://github.com/mchestr/mcp/blob/main/src/billing-cost-management-mcp-server/README.md to include the new feature / tool / permissions.

bhatia-di (dismissed):

wangyuhere (approved):

callnmm (approved):

PR #4490, opened 2026-08-13
fix(security-agent-mcp-server): enforce S3 bucket ownership to prevent bucket-squatting code leak

ljainiaz:
## Summary

  The Security Agent MCP server creates per-account S3 buckets with **deterministic
  names** (`security-agent-<kind>-<account>-<region>`) to hold uploaded source before
  a scan. Because S3's namespace is global, any account can pre-register that
  predictable name. Creation only guarded against `BucketAlreadyOwnedByYou`, so a
  bucket pre-created by *another* account (`BucketAlreadyExists`) was silently reused,
  and `upload_to_s3` wrote the caller's source code into it — **leaking code to an
  attacker-controlled bucket**.

  ## Changes
  
  - **`verify_bucket_owner()`** — `head_bucket` with `ExpectedBucketOwner`; a 403 on a
    foreign-owned bucket is fatal.
  - **`_ensure_s3_bucket`** — `BucketAlreadyExists` now propagates (only
    `BucketAlreadyOwnedByYou` is reused), and ownership is confirmed **before** the
    bucket is registered on the agent space or cached in config.
  - **`upload_to_s3`** — passes `ExpectedBucketOwner` so every write is rejected if the
    bucket isn't owned by the caller. Also unwraps boto3's `S3UploadFailedError` to
    re-raise the underlying `ClientError`, so the existing `ClientError` handlers in the
    scanner/server surface the structured error contract instead of a raw traceback.
  - Caller account ID is resolved once via STS and cached.
  
  ## User experience
  
  - **Before:** if an attacker pre-created the predictable bucket name, the victim's scan
    silently uploaded source into the attacker's bucket — no error, no signal.
  - **After:** the scan aborts with a clear ownership error (403) and uploads nothing.
    Legitimate users who already own the bucket are unaffected — it's reused as before.

  ## Testing
  
  Verified end-to-end against **real S3** with a two-account squat PoC:
  
  | Path | Result |
  |---|---|
  | Attacker squats `security-agent-scans-<victim>-<region>` | attacker owns it |
  | **Before fix** — raw `upload_file`, no owner check | ⚠️  upload succeeds → victim source lands in attacker bucket |
  | **After fix** — `upload_to_s3` (`ExpectedBucketOwner`) | ✅ 403, nothing written |
  | **After fix** — `create_s3_bucket` guard | ✅ `BucketAlreadyExists` → abort |
  | **After fix** — `verify_bucket_owner` guard | ✅ 403 → ownership check fails |

  Unit tests cover the pass path, the 403 squat path at the upload and ensure-bucket
  layers, the `S3UploadFailedError` unwrap, and that a squatted bucket is never
  registered or cached.

  ## Checklist

  - [x] Changes have been tested (unit + real-S3 two-account PoC)
  - [x] Changes are documented (docstrings/comments on the new guards)
  
  **Is this a breaking change?** N

  ## Acknowledgment

### DIFF
--- a/src/security-agent-mcp-server/awslabs/security_agent_mcp_server/aws_client.py
+++ b/src/security-agent-mcp-server/awslabs/security_agent_mcp_server/aws_client.py
@@ -19,6 +19,8 @@ import botocore.config
 import json
 import re
 from awslabs.security_agent_mcp_server import __version__
+from boto3.exceptions import S3UploadFailedError
+from botocore.exceptions import ClientError
 from typing import Any, Optional
 
 
@@ -36,6 +38,7 @@ class SecurityAgentClient:
     ):
         """Initialize SecurityAgent client."""
         self.region = region
+        self._cached_account_id: Optional[str] = None
         self._mcp_client_name = mcp_client_name
         self._mcp_client_version = mcp_client_version
         self._config = self._build_config(mcp_client_name, mcp_client_version)
@@ -72,6 +75,12 @@ class SecurityAgentClient:
     def _get_session(self):
         """Fresh session each call to pick up rotated credentials."""
         return boto3.Session(region_name=self.region)
+
+    def _account_id(self) -> str:
+        """Cache and return the caller's account ID for bucket-ownership guards."""
+        if self._cached_account_id is None:
+            self._cached_account_id = str(self.get_caller_identity()['Account'])
+        return self._cached_account_id
 
     def _client(self):
         """Get a fresh securityagent boto3 client with custom user-agent."""
@@ -327,6 +336,12 @@ class SecurityAgentClient:
         )
         return bucket_name
 
+    def verify_bucket_owner(self, bucket_name: str) -> None:
+        """Raise if the caller's account does not own the bucket (403 = squatted)."""
+        self._get_session().client('s3', config=self._config).head_bucket(
+            Bucket=bucket_name, ExpectedBucketOwner=self._account_id()
+        )
+
     def create_service_role(self, role_name: str, account_id: str, bucket_name: str) -> str:
         """Create IAM service role for Security Agent with S3 + CloudWatch Logs access."""
         iam = self._get_session().client('iam', config=self._config)
@@ -388,6 +403,18 @@ class SecurityAgentClient:
         return f'arn:aws:iam::{account_id}:role/{role_name}'
 
     def upload_to_s3(self, bucket: str, key: str, file_path: str) -> str:
-        """Upload a file to S3."""
-        self._get_session().client('s3', config=self._config).upload_file(file_path, bucket, key)
+        """Upload a file to S3, rejecting the write if the bucket is owned by another account."""
+        try:
+            self._get_session().client('s3', config=self._config).upload_file(
+                file_path,
+                bucket,
+                key,
+                ExtraArgs={'ExpectedBucketOwner': self._account_id()},
+            )
+        except S3UploadFailedError as e:
+            # Re-raise the underlying ClientError (e.g. 403) so callers can handle it.
+            cause = e.__cause__ or e.__context__
+            if isinstance(cause, ClientError):
+                raise cause
+            raise
         return f's3://{bucket}/{key}'
--- a/src/security-agent-mcp-server/awslabs/security_agent_mcp_server/aws_client.py
+++ b/src/security-agent-mcp-server/awslabs/security_agent_mcp_server/aws_client.py
@@ -39,6 +39,7 @@ class SecurityAgentClient:
         """Initialize SecurityAgent client."""
         self.region = region
         self._cached_account_id: Optional[str] = None
+        self._cached_account_key: Optional[str] = None
         self._mcp_client_name = mcp_client_name
         self._mcp_client_version = mcp_client_version
         self._config = self._build_config(mcp_client_name, mcp_client_version)
@@ -77,9 +78,15 @@ class SecurityAgentClient:
         return boto3.Session(region_name=self.region)
 
     def _account_id(self) -> str:
-        """Cache and return the caller's account ID for bucket-ownership guards."""
-        if self._cached_account_id is None:
-            self._cached_account_id = str(self.get_caller_identity()['Account'])
+        """Return the caller's account ID, cached per credential set."""
+        session = self._get_session()
+        creds = session.get_credentials()
+        key = creds.get_frozen_credentials().access_key if creds else None
+        if self._cached_account_id is None or self._cached_account_key != key:
+            self._cached_account_id = str(
+                session.client('sts', config=self._config).get_caller_identity()['Account']
+            )
+            self._cached_account_key = key
         return self._cached_account_id
 
     def _client(self):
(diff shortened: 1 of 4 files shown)


jigar-lab (dismissed):

jigar-lab (dismissed):

src/security-agent-mcp-server/awslabs/security_agent_mcp_server/aws_client.py
```diff
@@ -73,6 +76,12 @@ def _get_session(self):
         """Fresh session each call to pick up rotated credentials."""
         return boto3.Session(region_name=self.region)
 
+    def _account_id(self) -> str:
```
scottschreckengaust, line 80: **Moderate — process-lifetime account cache can produce a false "squat" 403 (findings 1 + 3, shared root cause).**

`_account_id()` caches the STS account for the whole process, which quietly defeats the deliberate `_get_session()` design one line above ("Fresh session each call to pick up rotated credentials"). If credentials are switched to a different AWS account within one running server process: `_ensure_s3_bucket` computes the bucket name from the **new** account (it calls `_client.get_caller_identity()` directly at `server.py:194`), creates/owns a bucket under the new account, but `verify_bucket_owner` and `upload_to_s3` then pass the **cached old** account as `ExpectedBucketOwner` → `head_bucket`/`put_object` return 403 → every scan aborts as a spurious squat until the process is restarted.

Related (finding 3): `_ensure_s3_bucket` resolving the account via its own uncached `get_caller_identity()` (`server.py:194`) while uploads use this cached value creates two sources of truth and a redundant STS round-trip on the first scan. Fixing both together: have `_ensure_s3_bucket` call `_client._account_id()` so there is a single cached account value — that removes the divergence *and* the duplicate call. If you intend the cache to survive credential rotation, consider keying/invalidating it rather than caching unconditionally for the process lifetime.

src/security-agent-mcp-server/awslabs/security_agent_mcp_server/server.py
```diff
@@ -196,9 +196,13 @@ def _ensure_s3_bucket(config: dict, kind: str = 'scans') -> None:
     try:
         _client.create_s3_bucket(bucket)
     except ClientError as e:
+        # Reuse only if we already own it; BucketAlreadyExists (squatted) stays fatal.
```
scottschreckengaust, line 199: **Moderate — the squat abort surfaces misleading remediation, not the "clear ownership error" the PR promises.**

When a foreign account has pre-created `security-agent-<kind>-<account>-<region>`, `create_s3_bucket` raises `BucketAlreadyExists` and it correctly propagates here. But the tool's `ClientError` handler runs it through `_translate_client_error`, whose `BucketAlreadyExists` entry (`server.py:55`) is `"The S3 bucket name is taken globally; pick a different name."` — impossible advice, since the name is derived deterministically from account + region and can't be changed. The user gets no signal that the bucket is squatted/attacker-controlled, which is exactly the case this PR exists to catch. Consider a squat-specific remediation string (e.g. "a bucket with this deterministic name is owned by another account — it may be squatted; investigate before proceeding") for the abort path.

src/security-agent-mcp-server/awslabs/security_agent_mcp_server/server.py
```diff
@@ -196,9 +196,13 @@ def _ensure_s3_bucket(config: dict, kind: str = 'scans') -> None:
...
         if e.response.get('Error', {}).get('Code') != 'BucketAlreadyOwnedByYou':
             raise
 
+    # Confirm ownership before trusting the bucket — 403 means it's squatted.
+    _client.verify_bucket_owner(bucket)
```
scottschreckengaust: **Low — unconditional `verify_bucket_owner` after create adds a new transient-failure mode on the legitimate reuse path.**

On the `BucketAlreadyOwnedByYou` path (caller already owns the bucket; the exception is swallowed above), this now always issues a `head_bucket`. A transient S3 error there (SlowDown/503/throttling) will raise a `ClientError` that propagates out of `_ensure_s3_bucket` and fails a scan the caller legitimately owns — a failure mode that didn't exist before, since no `head_bucket` was on this path. The 403 ownership check is the valuable part; consider narrowing the fatal handling to the 403 (`AccessDenied`/`403`) case and letting transient errors retry/pass, so infrastructure blips don't read as ownership failures.

src/security-agent-mcp-server/awslabs/security_agent_mcp_server/aws_client.py
```diff
@@ -388,6 +403,18 @@ def create_service_role(self, role_name: str, account_id: str, bucket_name: str)
...
+        try:
+            self._get_session().client('s3', config=self._config).upload_file(
+                file_path,
+                bucket,
+                key,
+                ExtraArgs={'ExpectedBucketOwner': self._account_id()},
+            )
+        except S3UploadFailedError as e:
+            # Re-raise the underlying ClientError (e.g. 403) so callers can handle it.
+            cause = e.__cause__ or e.__context__
```
scottschreckengaust: **Low (robustness nit, correct today).** `cause = e.__cause__ or e.__context__` prefers `__cause__`. This is correct against current boto3, which does a bare `raise S3UploadFailedError(...)` inside `except ClientError` — so `__context__` holds the `ClientError` and `__cause__` is `None`. It would only break if a future boto3 chained via `raise ... from <non-ClientError>`: then `__cause__` (the non-ClientError) wins, the `isinstance` check fails, and callers get a raw `S3UploadFailedError` instead of the structured `ClientError` contract. Optional hardening: prefer whichever of `__cause__`/`__context__` is actually a `ClientError` rather than fixing precedence.

scottschreckengaust (commented):
Nicely scoped security fix — the core is correct: `BucketAlreadyExists` is now fatal (only `BucketAlreadyOwnedByYou` is reused), ownership is verified via `head_bucket`+`ExpectedBucketOwner` **before** the bucket is registered/cached, and every `upload_to_s3` write carries `ExpectedBucketOwner`. All four scanner upload paths route through the guarded helper, and the `S3UploadFailedError` unwrap (`e.__cause__ or e.__context__`) is correct against current boto3 (it does a bare `raise S3UploadFailedError` inside `except ClientError`, so the ClientError lands in `__context__`). No blocking defects — this is approve-with-suggestions.

Inline suggestions below (none blocking). Two are worth addressing before merge because they touch the guardrail's stated UX ("a clear ownership error"):
- **Account-ID caching (moderate):** the process-lifetime cache defeats the "fresh session each call to pick up rotated credentials" design and can send a stale account as `ExpectedBucketOwner` after a mid-process account switch → spurious 403 → every scan aborts as a false "squat" until restart. Sharing one cached account ID between `_ensure_s3_bucket` and `_account_id()` also removes a duplicate STS round-trip.
- **Squat remediation message (moderate):** when the abort fires, the downstream error map returns "pick a different name" — impossible advice for a deterministic account+region-derived name.

Comment-mapping note: two findings anchor to lines outside this diff, so I relocated them — the misleading remediation string lives at `server.py:55` (`_translate_client_error` map) and I anchored it to the `BucketAlreadyExists` propagation point; the redundant `get_caller_identity()` is at `server.py:194` and I folded it into the account-caching comment since it's the same root cause.

src/security-agent-mcp-server/awslabs/security_agent_mcp_server/server.py
scottschreckengaust: Maybe even a `WARNING:` or `SECURITY ALERT:`

jigar-lab (approved):

laithalsaadoon (approved):

PR #4479, opened 2026-08-11
fix(billing-cost-management): omit Granularity when GroupBy is set in ri/sp-performance tool

wangyuhere:
## Summary

Cost Explorer's GetReservationCoverage, GetReservationUtilization, and GetSavingsPlansCoverage treat Granularity and GroupBy as mutually exclusive. The coverage/utilization request builders always set Granularity, so any grouped query failed with a deterministic ValidationException (e.g. 'Granularity is not supported when specifying groupBy').

### Changes

Send Granularity only when the request is not grouped, across get_reservation_coverage, get_reservation_utilization, and get_savings_plans_coverage. Document the mutual exclusivity in both tool descriptions and granularity docstrings so the agent avoids sending both.

### User experience

All 3 questions work without error:
- Check RI coverage group by region
- Check RI usage by subscription_id
- Check SP coverage group by service

## Checklist

If your change doesn't seem to apply, please leave them unchecked.

* [x] I have reviewed the [contributing guidelines](https://github.com/awslabs/mcp/blob/main/CONTRIBUTING.md)
* [x] I have performed a self-review of this change
* [x] Changes have been tested
* [x] Changes are documented

Is this a breaking change? (N)

**RFC issue number**:

Checklist:

* [ ] Migration process documented
* [ ] Implement warnings (if it can live side by side)

## Acknowledgment

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the [project license](https://github.com/awslabs/mcp/blob/main/LICENSE).

PR #4491, opened 2026-08-13
chore: release/2026.08.20260813230114

awslabs-mcp:
# release/2026.08.20260813230114

Triggered Release Branch (initiated) by @matthewgoodman13 for @matthewgoodman13

## Changes
* billing-cost-management-mcp-server
* dynamodb-mcp-server
* ecs-mcp-server

## Checklist
- [ ] Code changes have been reviewed
- [ ] Distribution packages have been built and tested
- [ ] Documentation has been updated if applicable

## Acknowledgment

### DIFF
--- a/src/billing-cost-management-mcp-server/pyproject.toml
+++ b/src/billing-cost-management-mcp-server/pyproject.toml
@@ -2,7 +2,7 @@
 name = "awslabs.billing-cost-management-mcp-server"
 
 # NOTE: "Patch"=9223372036854775807 bumps next release to zero.
-version = "0.0.32"
+version = "0.0.33"
 
 description = "A Model Context Protocol (MCP) server that provides tools for AWS Billing and Cost Management by wrapping boto3 SDK functions."
 readme = "README.md"
--- a/src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/__init__.py
+++ b/src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/__init__.py
@@ -19,7 +19,7 @@ This Model Context Protocol (MCP) server provides tools for AWS cost optimizatio
 by wrapping boto3 SDK functions for AWS cost optimization services.
 """
 
-__version__ = '0.0.32'
+__version__ = '0.0.33'
 
 # We don't import server here to avoid circular imports
 
--- a/src/billing-cost-management-mcp-server/uv.lock
+++ b/src/billing-cost-management-mcp-server/uv.lock
@@ -99,7 +99,7 @@ wheels = [
 
 [[package]]
 name = "awslabs-billing-cost-management-mcp-server"
-version = "0.0.32"
+version = "0.0.33"
 source = { editable = "." }
 dependencies = [
     { name = "boto3" },
--- a/src/dynamodb-mcp-server/awslabs/dynamodb_mcp_server/__init__.py
+++ b/src/dynamodb-mcp-server/awslabs/dynamodb_mcp_server/__init__.py
@@ -14,4 +14,4 @@
 
 """awslabs.dynamodb-mcp-server"""
 
-__version__ = '2.1.5'
+__version__ = '2.1.6'
--- a/src/dynamodb-mcp-server/pyproject.toml
+++ b/src/dynamodb-mcp-server/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "awslabs.dynamodb-mcp-server"
-version = "2.1.5"
+version = "2.1.6"
 description = "The official MCP Server for interacting with AWS DynamoDB"
 readme = "README.md"
 requires-python = ">=3.10"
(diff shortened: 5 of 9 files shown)


matthewgoodman13 (approved):

LeeroyHannigan (approved):

src/billing-cost-management-mcp-server/pyproject.toml
```diff
@@ -2,7 +2,7 @@
 name = "awslabs.billing-cost-management-mcp-server"
 
 # NOTE: "Patch"=9223372036854775807 bumps next release to zero.
-version = "0.0.32"
+version = "0.0.33"
```
scottschreckengaust, line 5: Not about this line — flagging a **pre-existing test-isolation bug** in the package being released here. Out of scope for this PR (do not hold the release for it), but worth a follow-up since it affects every contributor to this package.

`tests/utilities/test_aws_service_base.py:84` fails on any machine with `AWS_PROFILE` set in the environment:

```
FAILED tests/utilities/test_aws_service_base.py::TestCreateAwsClient::test_client_creation_custom_region
Expected: Session(region_name='us-west-2')
Actual:   Session(profile_name='ai3', region_name='us-west-2')
```

The test asserts **exact** kwargs via `mock_session.assert_called_once_with(region_name='us-west-2')`, but `create_aws_client` passes `profile_name` through from `AWS_PROFILE`. So the assertion encodes an implicit "no AWS profile configured" precondition that nothing enforces.

Verified both directions in a clean worktree at this commit:

* with my `AWS_PROFILE=ai3` — `1 failed, 1237 passed`
* `env -u AWS_PROFILE` — `1238 passed`

CI is green only because the runners have no profile set, so this will not regress the pipeline — it just makes the suite red for developers who use named AWS profiles, which is most people working on a billing/cost server. `grep -rn 'delenv.*AWS_PROFILE'` across this package's tests returns nothing, so there is no guard anywhere.

The test is untouched by this PR (`git diff origin/main...HEAD` on that file is empty). Fix is one line in the test:

```python
def test_client_creation_custom_region(self, mock_session, monkeypatch):
    monkeypatch.delenv('AWS_PROFILE', raising=False)
```

An autouse fixture in `tests/conftest.py` would be better still, since it would cover any sibling test that grows the same coupling.

scottschreckengaust (approved):
## Automated review — release PR

**Approving.** This is a mechanical version bump. I verified the diff contains *nothing* but version bumps: filtering every added/removed line that is not a `version =` / `__version__ =` assignment or a `resolution-markers` entry leaves the diff empty. No source, dependency, or configuration change is hiding in here.

For a release PR the useful question isn't "is this code correct" but "does the release match what's shipping." Four checks:

### 1. Every bumped package has real changes — and no changed package was missed

The reverse check is the one that matters, since a *missing* bump ships code under an old version number. Enumerating every package with `src/` changes since the last release bump (`ebcafaed`):

```
src/billing-cost-management-mcp-server
src/dynamodb-mcp-server
src/ecs-mcp-server
```

Exactly the three packages bumped here — no omissions, no spurious bumps. Mapped to their commits:

| Package | Bump | Shipping |
|---|---|---|
| billing-cost-management | 0.0.32 → 0.0.33 | #4479 (omit `Granularity` when `GroupBy` set), #4467 (normalize Compute Optimizer timestamps to UTC) |
| dynamodb | 2.1.5 → 2.1.6 | #4384 (validate/escape names in CDK generator) |
| ecs | 0.1.33 → 0.1.34 | #4480 (path validation for user-supplied paths) |

All four commits confirmed present in the release branch via `git merge-base --is-ancestor`.

### 2. Version strings are internally consistent

This repo requires each version in **three** places (`pyproject.toml`, `awslabs/<pkg>/__init__.py`, and the self-referential `uv.lock` entry) — an easy place for automation to half-apply. All three agree for all three packages, and no stray copies of the old version survive in Dockerfiles or docs.

### 3. Lockfiles are valid

`uv lock --check` passes for all three (126 / 133 / 117 packages resolved, no re-resolution needed). Worth checking explicitly here because a stale lockfile in a release is a broken published artifact.

The `resolution-markers` reordering in `ecs-mcp-server/uv.lock` is **semantically neutral** — I diffed the sorted marker sets before and after and they are identical, so it is a pure reordering with no effect on resolution.

### 4. Tests pass for all three released packages

`dynamodb 1474 passed`, `ecs 787 passed`, `billing-cost-management 1238 passed`. CI is green across 210 checks.

### On the patch-level bump for dynamodb

Worth noting rather than blocking: #4384's own description documents a behavior change — `TableName`/`IndexName` shorter than 3 characters are now rejected, so a model file that previously parsed can now fail. Under strict semver that reads as more than a patch. I'm not asking for a change, because this repo's release automation bumps patch unconditionally, and a minor bump is a deliberate maintainer opt-in via the documented `"Patch"=9223372036854775807` sentinel (last used in `88e563e9`) set *before* the release runs. So patch is the correct output for this pipeline given no sentinel was set; flagging only so the decision is conscious rather than default, since the rejection surfaces at parse time with a clear message instead of a later CloudFormation failure.

### One finding, pre-existing and outside this diff

See the inline comment. `billing-cost-management`'s `test_client_creation_custom_region` fails on any machine with `AWS_PROFILE` set. It is not a release defect and does not affect the artifacts, but it is in a package being released today and costs the next contributor real debugging time.

PR #4493, opened 2026-08-14
feat(aws-healthomics): add WaitForAHORun to poll a run to completion

hmkim:
## Problem

Waiting on a run means calling `GetAHORun` in a loop from the agent side. A seven-minute run costs six or seven round trips, each spending context on a response whose only useful field is `status`, and each an opportunity for credentials to expire mid-wait. (That happened during the migration session this came out of.)

## Solution

`WaitForAHORun` polls server-side and returns once the run reaches a terminal status.

```
WaitForAHORun(run_id="4194589", timeout_seconds=1800, poll_interval_seconds=30)
```

```json
{
  "id": "4194589",
  "status": "COMPLETED",
  "timedOut": false,
  "elapsedSeconds": 412.3,
  "pollCount": 14,
  "run": { "...full GetAHORun details..." }
}
```

## Design notes

**Terminal statuses are defined positively.** `TERMINAL_RUN_STATUSES` is `{COMPLETED, FAILED, CANCELLED}`, not "anything outside `RUN_STATUSES`". The service reports statuses absent from that list — `STOPPING` appears while outputs are exported after the last task finishes, which I hit live:

```
PENDING → RUNNING → STOPPING → COMPLETED
```

A terminal check based on the documented list would have returned `STOPPING` as final, before outputs existed. There is a test for this transition.

**Timeout is not an error, and does not cancel the run.** The result carries `timedOut: true` with the last observed status; the run continues in the service. A caller that wants to keep waiting can call again.

**A FAILED run is a successful wait.** The tool returns normally with `status: "FAILED"` and the run's `failureReason`. Callers branch on status rather than on error presence; failure diagnosis stays with `DiagnoseAHORunFailure`.

**Status is checked before the first sleep**, so an already-finished run returns immediately.

**The final sleep is clamped to remaining time**, so a timeout that is not a multiple of the poll interval is still honoured.

## Refactor

The `GetRun` response formatting is extracted out of `get_run` into `format_run_response` so both tools return the same shape. `get_run`'s behaviour is unchanged; its existing 66 tests pass untouched.

## Testing

18 tests in `tests/test_wait_for_run.py`:

- `format_run_response` — ISO 8601 rendering, omission of absent optional fields, failure details, missing `creationTime`
- `wait_for_run` — already-terminal returns without sleeping, polling through intermediate statuses, `STOPPING` treated as non-terminal, `FAILED`/`CANCELLED` as terminal, timeout without error, sleep clamping, API error via `handle_tool_error`, full run details in the result

One of these surfaced a `FieldInfo` handling gap: calling the function directly with `timeout_seconds` omitted left the pydantic default unresolved, and the timeout arithmetic raised `unsupported operand type(s) for -: 'FieldInfo' and 'float'`.

To be precise about the scope, since I initially overstated it: FastMCP **does** resolve Field defaults, so this never affected calls through the MCP interface. I confirmed that by removing the guard and calling `WaitForAHORun` through `mcp.call_tool` with only `run_id` — it returned `COMPLETED` normally. The guard is kept because it matches the convention already used across this codebase (`create_workflow`, `list_workflows` and others carry the same `hasattr(x, 'default')` handling) and because direct calls are how the tests and the verification harness invoke it.

`ruff check` and `ruff format` clean.

## Scope

This is the polling implementation. An EventBridge-driven variant was considered and deliberately deferred: it needs `events:PutRule`, `sqs:CreateQueue` and temporary-resource cleanup, which is a lot of machinery for runs that mostly finish in 5-15 minutes. Worth revisiting if there is demand from users with long-running pipelines.

## Related

Third of the improvements from an end-to-end WDL migration session. Companions: #4485 (auto-ZIP), #4486 (ValidateAHORunReadiness).

Beyond the unit tests, this was exercised against a live HealthOmics account (`us-east-1`) on a real run:

```
status:          COMPLETED
timedOut:        False
elapsedSeconds:  426.1
pollCount:       22
```

The run passed through `PENDING → RUNNING → STOPPING → COMPLETED`, so the `STOPPING` handling is confirmed on the real service rather than only in a mock. 22 polls is what the agent would otherwise have issued by hand.

Also invoked through `mcp.call_tool` with only `run_id` supplied, to confirm the tool behaves correctly when every optional argument is omitted.

---


markjschreiber (commented):
I don't think this approach will work because most agents will time out a tool that doesn't respond before a time out. You might be able to use the 'report_progress' function https://gofastmcp.com/servers/progress

Not sure how many agents support this but worth experimenting with.

PR #4495, opened 2026-08-15
fix(scripts): narrow package-name false-positive filter (#2656)

### LINKED ISSUE #2656: (verify_package_name): Hardcoded package filtering may hide valid package references
### Describe the bug

The script `scripts/verify_package_name.py` includes a hardcoded filter that skips certain prefixes when validating package references found in repository files.

Example code:

```python
if prefix in ['asset', 'model', 'property', 'hierarchy']:
    continue
```

This filtering assumes that these prefixes should always be ignored when detecting package names. However, this assumption may not always be valid and could prevent legitimate package references from being validated.

This creates a maintenance risk if new packages are introduced that use these prefixes.

---

### Expected Behavior

The verification script should detect all relevant package references without relying on domain-specific hardcoded prefix exclusions.

Filtering logic should be configurable or based on explicit rules instead of a fixed list.

---

### Current Behavior

Package references that start with the prefixes below are skipped during validation:

```
asset
model
property
hierarchy
```

If a legitimate package name begins with one of these prefixes, the validation logic will ignore it and may fail to detect incorrect package references.

---

### Reproduction Steps

Example scenario:

1. Add a reference to a package such as:

```
asset-manager
```

2. Run the verification script:

```
python scripts/verify_package_name.py
```

3. The script skips validation for the package because its prefix matches one of the filtered values.

---

### Possible Solution

Replace hardcoded prefix filtering with a more explicit validation strategy.

Possible options:

• Use an allowlist of known internal packages
• Compare detected names against installed packages
• Make ignored prefixes configurable

Example improvement:

```python
IGNORED_PREFIXES = set()
if prefix in IGNORED_PREFIXES:
    continue
```

---

Hardcoded filtering can make the script fragile when repository structure evolves or when new packages are introduced.

Removing domain-specific assumptions would make the tool more robust and easier to maintain.

### OS

Any

### Server

aws-api-mcp-server

### Server Version

_No response_

N/A

### Other information

This issue affects repository tooling located in:

scripts/verify_package_name.py

### Service quota

wakqasahmed:
Fixes #2656

### Corrected root cause

The original version of this PR claimed the old exclusion list (`asset`, `model`, `property`, `hierarchy`) was a prefix match that would hide real packages like a hypothetical `asset-manager`. That claim was wrong, and I want to be upfront about it rather than leave a misleading description in place.

The old code in `scripts/verify_package_name.py` did this:

```python
prefix = ref.split('@')[0].lower()
if prefix in ['asset', 'model', 'property', 'hierarchy']:
    continue
```

`prefix` is the *whole* string before `@`, and the check is exact **equality**, not `startswith`. So for `asset-manager@latest`, `prefix == "asset-manager"`, which was never in the list — the old code already detected that reference correctly. I verified this by running the old code (from the commit before this PR) against a synthetic `asset-manager@latest` reference and confirmed it was kept, not filtered.

The actual (much narrower) gap: a package literally named exactly `asset`, `model`, `property`, or `hierarchy` with a valid version (e.g. `asset@latest`) was incorrectly hidden, because the exact-match list didn't distinguish "the real SiteWise doc false positive" from "a real package that happens to share one of those exact names." I confirmed this is the only gap by scanning all 59 existing package READMEs for anything matching the old exclusion condition — the *only* hit anywhere in the repo is the one documented case: `create_asset("asset@invalid", "model-id")` in `src/aws-iot-sitewise-mcp-server/README.md` (and its docusaurus translation).

### What changed (this revision)

Instead of a version-format regex (which was in an earlier version of this PR and has its own problem — see below), the fix now replaces the four-word exact-match exclusion with an exact match on the one real documented false positive:

```python
if ref.lower() == 'asset@invalid':
    continue
```

This is deliberately narrow: it fixes the real gap (a literally-named `asset`/`model`/`property`/`hierarchy` package with a real version is no longer silently hidden) without inventing a new failure mode.

**Why not a version-format regex?** An earlier version of this fix filtered any `word@version` where `version` didn't match `latest|v?\d+(\.\d+)*`. That would have introduced a *new* false-exclusion class: legitimate npm/package dist-tags like `pkg@beta`, `pkg@next`, `pkg@stable`, `pkg@dev`, or `pkg@rc1` don't match that pattern and would have been silently filtered — the same "hides a valid package reference" failure mode issue #2656 was about, just via a different mechanism. I verified the current (exact-match) version does **not** exhibit this: all of `pkg@beta`, `pkg@next`, `pkg@stable`, `pkg@dev`, `pkg@rc1` are correctly kept.

### Verification

- Ran `scripts/verify_package_name.py` against all 59 existing `src/*` packages with both the old (pre-PR) and new logic — identical pass/fail results, no regression.
- The 4 tests in `scripts/test_verify_package_name.py` all pass against the new code. Two of them (`asset@latest`, `property@2`) **fail against the old code** — confirmed by running the same test file against a copy of the pre-PR `verify_package_name.py` — so they actually discriminate old vs. new behavior, unlike the previous test set (which passed unchanged against the old code and proved nothing).

### Known test-coverage gap (disclosed, not hidden)

`scripts/test_verify_package_name.py` is **not currently run by CI**. `.github/workflows/python.yml` only runs `scripts/verify_package_name.py` itself (the CLI, against each package's README) as part of the per-package build matrix — it does not run any `scripts/test_*.py` unit tests, there's no root `pyproject.toml`/`pytest.ini`/`tox.ini` that would auto-discover it, and there is no existing precedent in this repo for testing files under `scripts/`. I did not invent a new CI job for this, since that's a broader change than this fix needs and there's no established pattern to follow. Flagging it here rather than claiming coverage CI doesn't actually enforce — happy to wire it in if a maintainer points at the preferred convention.

### User experience

No behavior change for any of the 59 existing packages (verified before/after — all pass, see above). The only behavior change is for a hypothetical future package literally named `asset`, `model`, `property`, or `hierarchy` with a real version — it will now be checked instead of silently skipped.

## Checklist

* [x] Changes are documented

Is this a breaking change? (Y/N): N

## Acknowledgment

### DIFF
--- a/scripts/test_verify_package_name.py
+++ b/scripts/test_verify_package_name.py
@@ -1,0 +1,67 @@
+#!/usr/bin/env python3
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Tests for verify_package_name.py's README reference filtering.
+
+Regression coverage for https://github.com/awslabs/mcp/issues/2656: the script
+used to hardcode a list of "safe" prefixes (asset/model/property/hierarchy) to
+skip when filtering `word@something` false positives, which would have hidden
+a real package reference such as ``asset-manager@latest`` had one ever been
+added. The replacement filters on whether the text after ``@`` looks like a
+version identifier instead of matching a fixed word list.
+"""
+
+import tempfile
+import unittest
+from pathlib import Path
+from verify_package_name import find_package_references_in_readme
+
+
+class TestFindPackageReferencesInReadme(unittest.TestCase):
+    """Tests for find_package_references_in_readme."""
+
+    def _refs(self, content: str):
+        with tempfile.NamedTemporaryFile(
+            mode='w', suffix='.md', delete=False, encoding='utf-8'
+        ) as f:
+            f.write(content)
+            path = Path(f.name)
+        try:
+            return [ref for ref, _ in find_package_references_in_readme(path)]
+        finally:
+            path.unlink()
+
+    def test_code_example_with_invalid_placeholder_is_filtered(self):
+        """Doc code examples like create_asset("asset@invalid", ...) are not package refs."""
+        content = 'create_asset("asset@invalid", "model-id")  # Fails: invalid characters\n'
+        self.assertNotIn('asset@invalid', self._refs(content))
+
+    def test_previously_hardcoded_prefixes_no_longer_hide_real_packages(self):
+        """A real package starting with a formerly-hardcoded prefix must be detected."""
+        content = '"mcpServers": {"x": {"command": "uvx", "args": ["asset-manager@latest"]}}\n'
+        self.assertIn('asset-manager@latest', self._refs(content))
+
+    def test_bare_numeric_version_is_kept(self):
+        """A short numeric version tag (no dot) is still recognized as a package ref."""
+        content = '"model-toolkit@2" is referenced here\n'
+        self.assertIn('model-toolkit@2', self._refs(content))
+
+    def test_non_version_suffix_is_filtered_regardless_of_prefix(self):
+        """Any word@non-version token is filtered, not just the old hardcoded words."""
+        content = 'create_model("model@bad-input")\n'
+        self.assertNotIn('model@bad-input', self._refs(content))
+
+
+if __name__ == '__main__':
+    unittest.main()
--- a/scripts/test_verify_package_name.py
+++ b/scripts/test_verify_package_name.py
@@ -14,12 +14,21 @@
 # limitations under the License.
 """Tests for verify_package_name.py's README reference filtering.
 
-Regression coverage for https://github.com/awslabs/mcp/issues/2656: the script
-used to hardcode a list of "safe" prefixes (asset/model/property/hierarchy) to
-skip when filtering `word@something` false positives, which would have hidden
-a real package reference such as ``asset-manager@latest`` had one ever been
-added. The replacement filters on whether the text after ``@`` looks like a
-version identifier instead of matching a fixed word list.
+Regression coverage for https://github.com/awslabs/mcp/issues/2656.
+
+The old filter matched on `ref.split('@')[0].lower() in ['asset', 'model',
+'property', 'hierarchy']` (an exact-equality check on the whole prefix, not a
+`startswith` check). That meant a compound name like `asset-manager@latest`
+was never affected by the old exclusion (its prefix is `"asset-manager"`,
+which is not in the list) -- the old code already detected it correctly. The
+only real gap was a package literally named `asset`, `model`, `property`, or
+`hierarchy` (exact match) that had a valid version, e.g. `asset@latest`,
+which the old code silently hid.
+
+The replacement drops the word list and instead matches the single known
+code-sample false positive (`asset@invalid`, from the SiteWise README's
+``create_asset("asset@invalid", "model-id")`` input-validation example) by
+its exact literal value.
 """
 
 import tempfile
@@ -42,25 +51,35 @@ class TestFindPackageReferencesInReadme(unittest.TestCase):
         finally:
             path.unlink()
 
-    def test_code_example_with_invalid_placeholder_is_filtered(self):
-        """Doc code examples like create_asset("asset@invalid", ...) are not package refs."""
+    def test_known_sitewise_false_positive_is_filtered(self):
+        """The documented create_asset("asset@invalid", ...) sample is not a package ref."""
         content = 'create_asset("asset@invalid", "model-id")  # Fails: invalid characters\n'
         self.assertNotIn('asset@invalid', self._refs(content))
 
-    def test_previously_hardcoded_prefixes_no_longer_hide_real_packages(self):
-        """A real package starting with a formerly-hardcoded prefix must be detected."""
+    def test_exact_word_package_with_valid_version_is_now_detected(self):
+        """A package literally named one of the old hardcoded words must be detected.
+
+        Discriminates old vs. new behavior: the old code excluded ANY
+        `asset@<version>` reference by exact-prefix match, even with a
+        legitimate version. This is the actual bug the old code had.
+        """
+        content = '"mcpServers": {"x": {"command": "uvx", "args": ["asset@latest"]}}\n'
+        self.assertIn('asset@latest', self._refs(content))
+
+    def test_exact_word_package_property_with_numeric_version_is_now_detected(self):
+        """Same discrimination as above for another formerly-hardcoded word."""
+        content = '"property@2" is referenced here\n'
+        self.assertIn('property@2', self._refs(content))
+
+    def test_compound_name_starting_with_hardcoded_word_was_always_detected(self):
+        """Sanity check: compound names were never broken by the old exact-match list.
+
+        This does NOT discriminate old vs. new (both detect it), but guards
+        against a future regression that reintroduces prefix/startswith
+        matching instead of exact matching.
+        """
         content = '"mcpServers": {"x": {"command": "uvx", "args": ["asset-manager@latest"]}}\n'
         self.assertIn('asset-manager@latest', self._refs(content))
-
-    def test_bare_numeric_version_is_kept(self):
-        """A short numeric version tag (no dot) is still recognized as a package ref."""
-        content = '"model-toolkit@2" is referenced here\n'
-        self.assertIn('model-toolkit@2', self._refs(content))
-
-    def test_non_version_suffix_is_filtered_regardless_of_prefix(self):
-        """Any word@non-version token is filtered, not just the old hardcoded words."""
-        content = 'create_model("model@bad-input")\n'
-        self.assertNotIn('model@bad-input', self._refs(content))
 
 
 if __name__ == '__main__':
--- a/scripts/verify_package_name.py
+++ b/scripts/verify_package_name.py
@@ -215,16 +215,14 @@ def find_package_references_in_readme(
         # But allow package names that look like they could be valid (contain hyphens)
         if '.' not in ref and '@' not in ref and '-' not in ref:
             continue
-        # Skip common false positives in code examples (word@something where the
-        # part after '@' doesn't look like a real package version, e.g. Python
-        # snippets such as create_asset("asset@invalid", ...) that demonstrate
-        # input validation rather than an installation instruction).
-        # A real package reference's version is either the literal "latest" or
-        # a numeric tag such as "1", "2.0", or "v1.2".
-        if '@' in ref and '.' not in ref:
-            version_part = ref.split('@', 1)[1]
-            if not re.fullmatch(r'latest|v?\d+(\.\d+)*', version_part, re.IGNORECASE):
-                continue
+        # Skip the one known code-sample false positive in the SiteWise README:
+        # create_asset("asset@invalid", "model-id") demonstrates input
+        # validation, not an install instruction. Matched by exact literal
+        # value (not a word-prefix list) so we don't silently hide real
+        # "word@version" package references that happen to start with the
+        # same word (e.g. a hypothetical "asset-manager@latest").
+        if ref.lower() == 'asset@invalid':
+            continue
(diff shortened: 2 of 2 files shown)


scripts/verify_package_name.py
```diff
@@ -215,12 +215,15 @@ def find_package_references_in_readme(
...
+        # input validation rather than an installation instruction).
+        # A real package reference's version is either the literal "latest" or
+        # a numeric tag such as "1", "2.0", or "v1.2".
         if '@' in ref and '.' not in ref:
-            # Extract the part before @
-            prefix = ref.split('@')[0].lower()
-            # Different scenarios where the prefix is not a package name
-            if prefix in ['asset', 'model', 'property', 'hierarchy']:
+            version_part = ref.split('@', 1)[1]
+            if not re.fullmatch(r'latest|v?\d+(\.\d+)*', version_part, re.IGNORECASE):
```
wakqasahmed: Two issues on this line.

**(a) `(\.\d+)*` is unreachable.** This block is guarded by `if '@' in ref and '.' not in ref:` on the line above, so `version_part` can never contain a dot. The effective pattern is `latest|v?\d+`. The new comment's examples `"2.0"` and `"v1.2"` therefore describe something this line can never see — misleading for the next reader. Either drop `(\.\d+)*` and fix the comment, or move the check outside the `'.' not in ref` guard if dotted versions were meant to be covered.

**(b) New false exclusions.** Verified against `upstream/main`, these were kept before and are dropped now: `pkg@beta`, `pkg@next`, `pkg@stable`, `pkg@dev`, `pkg@rc1`, `pkg@2-beta`. None occur in the repo today (every quoted `@` ref in `src/*/README.md` is `@latest` apart from `aws.s3@ObjectCreated` and `asset@invalid`), so nothing breaks now — but a mistyped package published under a dist tag would now be silently skipped, which is the exact failure mode #2656 is about. Consider `(latest|stable|beta|alpha|next|dev|rc\d*|v?\d+[\w.-]*)`, or invert the test to "filter only when the version part contains no digit and is not a known dist tag".

Minor: wrap the alternation in a group — `re.fullmatch(r'(latest|v?\d+)', ...)`. `fullmatch` anchors correctly here, but a bare top-level `|` reads as a precedence bug even though it isn't.

scripts/test_verify_package_name.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def test_code_example_with_invalid_placeholder_is_filtered(self):
+        """Doc code examples like create_asset("asset@invalid", ...) are not package refs."""
+        content = 'create_asset("asset@invalid", "model-id")  # Fails: invalid characters\n'
+        self.assertNotIn('asset@invalid', self._refs(content))
+
+    def test_previously_hardcoded_prefixes_no_longer_hide_real_packages(self):
+        """A real package starting with a formerly-hardcoded prefix must be detected."""
+        content = '"mcpServers": {"x": {"command": "uvx", "args": ["asset-manager@latest"]}}\n'
+        self.assertIn('asset-manager@latest', self._refs(content))
```
wakqasahmed: This test does not test what its name says, and passes against the pre-fix code.

The old condition was `if prefix in ['asset', 'model', 'property', 'hierarchy']` — **exact equality** on the whole pre-`@` token, not a `startswith`. `asset-manager@latest` has prefix `asset-manager`, which was never in that list, so `upstream/main` already returns `['asset-manager@latest']` here. I confirmed by running this file unchanged against `upstream/main`'s `verify_package_name.py`: **all 4 tests pass**, so the suite is green on the unpatched code and would not catch a revert.

The same applies to `test_non_version_suffix_is_filtered_regardless_of_prefix` below: it uses `model@bad-input`, and `model` *is* one of the four hardcoded words, so the old code filtered it too — the "regardless of prefix" claim is untested.

To make it a real regression test, use a prefix outside the old list:

```python
def test_non_version_suffix_is_filtered_regardless_of_prefix(self):
    content = 'create_thing("widget@invalid")\n'
    self.assertNotIn('widget@invalid', self._refs(content))
```

Old: `['widget@invalid']`. New: `[]`. That one actually fails before the fix.

Worth adding the genuinely-fixed case too — a package named *exactly* `asset`, i.e. `asset@latest`, which the old list did wrongly skip. That, not `asset-manager`, is the real bug behind #2656.

scripts/test_verify_package_name.py
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env python3
```
wakqasahmed, line 1: Two mechanical problems with this new file.

1. **Shebang without the executable bit.** It is committed as `100644` (`git ls-tree` on the PR head), while the other three `scripts/*.py` are `100755`. `.pre-commit-config.yaml` enables `check-shebang-scripts-are-executable`, and `.github/workflows/pre-commit.yml` runs `pre-commit run --all-files`, so this will fail that job. Either `chmod +x` it or drop the shebang — a unittest module does not need one.

2. **Nothing ever runs this file.** `.github/workflows/python.yml` invokes pytest only with `working-directory: src/${{ matrix.package }}` against `src/<pkg>/tests`, the pre-push pytest hook is scoped to `files: src\/.*\/pyproject.toml`, and there is no root `pyproject.toml`/`pytest.ini`. `scripts/` is never collected, so this is documentation rather than a regression gate. If the intent is to guard this logic, the PR also needs a CI step (e.g. `python3 -m unittest discover -s scripts -p 'test_*.py'` next to the existing `python3 scripts/verify_package_name.py` step in `python.yml`).

wakqasahmed (commented):
Cold-start review (no context from the implementation). I verified the diff by checking out `upstream/main` and the PR head and running both versions of the script against every package in `src/`.

## What checks out

- The false positive is real and still present: `src/aws-iot-sitewise-mcp-server/README.md:700` contains `create_asset("asset@invalid", "model-id")`, matched by the `r'"([a-zA-Z0-9._-]+@[a-zA-Z0-9._-]+)"'` pattern.
- **No regression.** I ran the old and new script over all 59 `src/*/` packages: 59/59 pass on both, and the `--verbose` reference lists are **byte-identical for every package**. Only two refs in the whole repo even reach this branch (`mcp-proxy-for-aws@latest`, kept by both; `asset@invalid`, filtered by both) — every other reference is `awslabs.*@latest`, which contains a dot and short-circuits before the branch.

## Substantive findings

**1. The stated root cause is wrong, and so is the PR description.** The old code was `if prefix in ['asset', 'model', 'property', 'hierarchy']` — an **exact equality** check on the full pre-`@` token, not a prefix / `startswith` match. `asset-manager@latest` has prefix `asset-manager`, which is not in that list, so the old code **already detected it**. Verified empirically: on `upstream/main`, `find_package_references_in_readme` returns `['asset-manager@latest']` for both the `uvx` and JSON-config forms. The issue's premise ("if a legitimate package name begins with one of these prefixes...") and the PR body's claim ("silently skips validation for any real package whose name happens to start with one of those four words") do not describe the actual behaviour. The real, much narrower defect is that a package named *exactly* `asset`/`model`/`property`/`hierarchy` would be skipped — plus the general maintainability point the issue also raises.

**2. The new test suite has zero discriminating power — all 4 tests pass against the unpatched code.** I copied `scripts/test_verify_package_name.py` next to `upstream/main`'s `verify_package_name.py` and ran it: `Ran 4 tests ... OK`. Every case is chosen such that old and new agree:

| test input | old | new |
|---|---|---|
| `asset@invalid` | filtered (`asset` in list) | filtered |
| `asset-manager@latest` | kept (see #1) | kept |
| `model-toolkit@2` | kept | kept |
| `model@bad-input` | filtered (`model` **is** in the list) | filtered |

Despite the name `test_non_version_suffix_is_filtered_regardless_of_prefix`, the prefix it picks is one of the four hardcoded ones. A test that actually pins the behaviour change needs a prefix outside the old list with a non-version suffix, e.g. `create_thing("widget@invalid")`: old returns `['widget@invalid']`, new returns `[]`.

**3. The new test file is never executed.** `.github/workflows/python.yml` runs pytest only with `working-directory: src/${{ matrix.package }}` against `src/<pkg>/tests`; there is no root `pyproject.toml`/`pytest.ini`, and the pre-push pytest pre-commit hook is scoped to `files: src\/.*\/pyproject.toml`. Nothing collects `scripts/`. As added, this file is documentation, not a regression gate.

**4. New false exclusions the old code did not have.** The regex rejects npm/uv-style dist tags and non-dotted pre-releases. Confirmed old-vs-new divergence:

| ref | old | new |
|---|---|---|
| `some-pkg@beta` | kept | **filtered** |
| `some-pkg@next` | kept | **filtered** |
| `some-pkg@stable` | kept | **filtered** |
| `some-pkg@dev` | kept | **filtered** |
| `some-pkg@rc1` | kept | **filtered** |
| `some-pkg@2-beta` | kept | **filtered** |

No README in the repo uses these today (I grepped `src/*/README.md`: every quoted `@` ref is `@latest` except `aws.s3@ObjectCreated`, skipped earlier, and `asset@invalid`), so there is no present-day breakage. But it is the same class of problem the issue complains about — a *wrong* package name published as `wrongname@beta` would now be silently skipped instead of flagged.

**5. Half the regex is dead code and the comment is misleading.** The branch is guarded by `'.' not in ref`, so `version_part` can never contain a dot. `(\.\d+)*` is unreachable, and the new comment's examples `"2.0"` and `"v1.2"` can never be evaluated by this line.

## Mechanical / CI

**6.** CI is red on two required checks, both trivially fixable:
- *Validate PR title*: `amannn/action-semantic-pull-request` needs a Conventional Commits prefix. Suggest `refactor(scripts): replace hardcoded package-name exclusion list (#2656)`.
- *Require Contributor Statement*: the acknowledgment must be verbatim, with `project license` as a **markdown link** to `https://github.com/awslabs/mcp/blob/main/LICENSE`. The current body has it as plain text.

**7.** `scripts/test_verify_package_name.py` is committed `100644` but starts with `#!/usr/bin/env python3`. The repo's pre-commit config runs `check-shebang-scripts-are-executable` and `.github/workflows/pre-commit.yml` runs `pre-commit run --all-files`, so this will fail that job once it runs.

## Verdict

**Request changes.** The code change is behaviour-neutral on today's repo and is a reasonable direction, but the justification rests on a misreading of the original condition, the tests do not test the change (they pass on unpatched code) and are not wired into CI, and the new regex introduces a small class of false exclusions the old code did not have. Items 1-3 should be fixed before merge; 4-5 are worth addressing while the file is open; 6-7 are blocking CI.

*(This PR was opened by the same account posting this review, so it is filed as a review comment rather than an approval.)*

wakqasahmed:
Correcting the root-cause claim from the original description, and narrowing the fix accordingly (verified each point myself, details below):

**The old exclusion was exact-match, not a prefix check.** `prefix = ref.split('@')[0].lower(); if prefix in ['asset','model','property','hierarchy']` compares the *whole* string before `@` for equality. So `asset-manager@latest` has `prefix == "asset-manager"`, which was never in that list — the old code already detected that reference correctly. I confirmed this by running the pre-PR `verify_package_name.py` against a synthetic `asset-manager@latest` reference; it was kept, not filtered. The real (narrower) gap: a package literally named exactly `asset`/`model`/`property`/`hierarchy` with a valid version (e.g. `asset@latest`) was wrongly hidden. Scanning all 59 package READMEs for anything the old exclusion condition touches, the only real hit anywhere in the repo is the one documented case: `create_asset("asset@invalid", "model-id")` in the SiteWise README.

**Replaced the version-format regex with an exact match on that one documented case**, instead of the four-word exact-match list:
```python
if ref.lower() == 'asset@invalid':
    continue
```
A regex on version shape (`latest|v?\d+(\.\d+)*`) would have traded the old narrow gap for a new false-exclusion class: legit dist-tags like `pkg@beta`/`pkg@next`/`pkg@rc1` would fail that pattern and get silently filtered — the same failure mode #2656 complains about, via a different mechanism. Verified the current exact-match version does not do this (all of `pkg@beta`, `pkg@next`, `pkg@stable`, `pkg@dev`, `pkg@rc1` are correctly kept).

**Replaced the tests.** The original 4 tests all passed against the *old, unpatched* code too (verified by running them against a checkout of the pre-PR script) — they had no discriminating power. The new tests include `asset@latest` and `property@2`, which fail against the old code and pass against the new one, actually proving the fix does something.

**Verification:** ran `verify_package_name.py` against all 59 existing packages with both old and new logic — identical results, no regression.

**Disclosed gap:** `scripts/test_verify_package_name.py` is not currently run by CI (`python.yml` only runs the CLI script itself per-package; there's no root pytest config or existing `scripts/test_*.py` convention in this repo to hook into). Noted this in the PR description rather than claiming coverage that isn't enforced.

**Also fixed for CI:** PR title now has a Conventional Commits prefix, the contributor-statement license text is now a markdown link to `LICENSE` matching the exact required string, and the test file is now executable (matching the other `scripts/*.py` files with a shebang, for the `check-shebang-scripts-are-executable` pre-commit hook).

wakqasahmed:
Still active and mergeable on my end — let me know if anything else is needed to move this forward.

wakqasahmed:
Hi @markjschreiber @arnewouters — this one seems to have no reviewer assigned, sitting for a while now with CI passing. Happy to be pointed at whoever's best placed to look.

PR #4496, opened 2026-08-16
fix(elasticache-mcp-server): guard against IndexError when a cluster has no cache nodes yet

mittalpk:
`_configure_security_groups()` in both `cc/connect.py` and `rg/connect.py` indexes `CacheNodes[0]` unconditionally to get the cache port. A cluster that's still provisioning has an empty `CacheNodes` list, so this raises a bare `IndexError` instead of a clear error.

Both files already guard this exact case elsewhere — `cc/connect.py`'s own `get_ssh_tunnel_command_cc()` checks `if not cache_cluster.get("CacheNodes")` a few functions away — this just applies the same guard consistently to `_configure_security_groups()` in both files.

Added a regression test to each of the existing `test_connect.py` files, following the same mock setup already used by the neighboring VPC-mismatch tests. Confirmed both fail with the original `IndexError` on unpatched code and pass after the fix. `ruff check`/`ruff format`/`pyright` all clean on the changed files.

### DIFF
--- a/src/elasticache-mcp-server/awslabs/elasticache_mcp_server/tools/cc/connect.py
+++ b/src/elasticache-mcp-server/awslabs/elasticache_mcp_server/tools/cc/connect.py
@@ -61,6 +61,8 @@ async def _configure_security_groups(
         raise ValueError(f'No security groups found for cache cluster {cache_cluster_id}')
 
     # Get cache cluster port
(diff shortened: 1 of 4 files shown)


mittalpk:
Still relevant and ready for review whenever a maintainer has time.

PR #4504, opened 2026-08-18
feat(aws-healthomics-mcp-server): expose logLevel on StartAHORun

steditt:
## Summary

Adds an optional `log_level` parameter to the `StartAHORun` tool (`start_run`) in the aws-healthomics-mcp-server and forwards it to the HealthOmics `StartRun` API as `logLevel`.

Today every run started through the MCP server is submitted without `logLevel`, so the run produces no engine log stream. When such a run fails there are no Nextflow/engine (STDOUT/STDERR) logs to inspect, which makes failures effectively undebuggable through the very tools this server ships (`GetAHORunEngineLogs`, `DiagnoseAHORunFailure`). This change lets callers opt into engine log capture.

## Changes

- `tools/workflow_execution.py`
  - Add optional `log_level` parameter to `start_run` (allowed values: `OFF`, `FATAL`, `ERROR`, `ALL`).
  - Validate the value against `RUN_LOG_LEVELS`.
  - Forward it to the API as `params['logLevel']` **only when provided**, preserving the HealthOmics API default and fully backward-compatible behavior when omitted.
  - Surface the effective value in the tool response.
- `consts.py`
  - Add `RUN_LOG_LEVEL_*` constants, the `RUN_LOG_LEVELS` list, and `ERROR_INVALID_RUN_LOG_LEVEL`.
- Tests
  - `test_workflow_execution.py`: cover the omit / valid-enum / invalid-value paths.
  - `test_consts.py`: cover the new constants and error template.

## Default behavior

Defaults to `None` (omit) so existing callers see no behavior change. Operators who want every run to be debuggable can pass `log_level='ALL'` explicitly. Changing the library default to `ALL` would be a silent behavior change (new log streams + CloudWatch cost) for all users, so this is intentionally opt-in.

## Testing

- `uv run --frozen pytest tests/test_workflow_execution.py tests/test_consts.py` — passing.
- Full server suite passing with coverage maintained.
- `pre-commit` (ruff check, ruff format, pyright, secret scanning, license header) — passing.

## Out of scope

- `StartRunBatch` / `start_run_batch`: the API does not accept a batch-level `logLevel` (it is configured per run-setting), so no change there.
- IAM: to actually persist per-task logs, the run's IAM role also needs `logs:DescribeLogStreams` on the `/aws/omics/*` log group (in addition to `CreateLogGroup` / `CreateLogStream` / `PutLogEvents`). This is a consumer-side role configuration concern, not an MCP-server change, but worth noting so users aren't surprised when task logs are still missing without it.

---


alxawan (approved):
LGTM

steditt:
Hi @markjschreiber - this PR has been approved by @alxawan (healthomics code owner), and I've just updated the branch so it's current with main (no conflicts). Could you take a look when you have a moment so it can be merged? It's part of a set of three small StartRun enhancements for the aws-healthomics-mcp-server (#4504, #4505, #4506). Thanks!

markjschreiber (approved):
LGTM.

Currently I don't think there is much difference between log levels but there may be in the future.

PR #4350, opened 2026-07-27
fix(aws-healthomics-mcp-server): add workflow_type parameter to start_run for Ready2Run workflows

steditt:
### Changes

The `start_run` tool in the HealthOmics MCP server is missing the `workflowType` parameter that the AWS HealthOmics `StartRun` API requires when executing Ready2Run workflows. Without it, the API defaults to `PRIVATE` and returns "Workflow not found" for Ready2Run workflow IDs — even though `ListWorkflows` and `GetWorkflow` discover them correctly.

This PR adds:
- An optional `workflow_type` parameter (`PRIVATE` or `READY2RUN`) to the `start_run` tool
- Validation using the shared `validate_workflow_type()` utility from `validation_utils.py` (same pattern as `list_workflows` and `get_workflow`)
- Passes `workflowType` to `client.start_run()` only when provided (backward compatible — omitting it preserves existing PRIVATE behavior)
- Three new unit tests covering READY2RUN, None (omitted), and invalid value scenarios

### User experience

**Before:** An agent discovers a Ready2Run workflow via `ListWorkflows(workflow_type="READY2RUN")`, retrieves its details via `GetWorkflow`, but `StartRun` fails with:
```
ResourceNotFoundException: Workflow 2174942 not found
```

**After:** The agent can pass `workflow_type="READY2RUN"` to `StartRun` and the run starts successfully:
```json
{
  "workflow_id": "2174942",
  "workflow_type": "READY2RUN",
  "role_arn": "arn:aws:iam::123456789012:role/OmicsRole",
  "name": "scrnaseq-starsolo",
  "output_uri": "s3://my-bucket/outputs/",
  "parameters": {"samplename": "test", "protocol": "10XV3", "input": [...]},
  "storage_type": "STATIC",
  "storage_capacity": 1200
}
```

## Checklist

If your change doesn't seem to apply, please leave them unchecked.

* [x] I have reviewed the [contributing guidelines](https://github.com/awslabs/mcp/blob/main/CONTRIBUTING.md)
* [x] I have performed a self-review of this change
* [x] Changes have been tested
* [x] Changes are documented

Is this a breaking change? N

**RFC issue number**: N/A (bug fix — extends existing parameter pattern to a missing tool)

Checklist:

* [ ] Migration process documented
* [ ] Implement warnings (if it can live side by side)

## Acknowledgment

[...]

PR #4505, opened 2026-08-18
feat(aws-healthomics-mcp-server): add tags parameter to StartAHORun

steditt:
## Summary

Adds an optional `tags` parameter to the `StartAHORun` tool (`start_run`) in the aws-healthomics-mcp-server, exposing the HealthOmics `StartRun` API's native `tags` field. This lets callers attach key-value metadata to a run at creation time — before the first EventBridge status-change event fires.

## Motivation

An agent that monitors pipeline progress via EventBridge notifications needs routing metadata (e.g. session ID, callback endpoint) present on the run from the moment it is created. Tagging after creation leaves a window where early events carry no routing information. Exposing `tags` on `start_run` closes that gap and also supports cost allocation and run organization.

## Changes

- `tools/workflow_execution.py`
  - Add optional `tags: Optional[Dict[str, str]]` parameter to `start_run`.
  - Forward it to the HealthOmics `StartRun` API as `params['tags']` **only when provided** (`if tags and isinstance(tags, dict)`), preserving current behavior when omitted. The value is already surfaced in the `start_run` response.
  - Update the docstring.
- `tests/test_workflow_execution.py`
  - `test_start_run_with_tags`: verifies tags are forwarded to the API.
  - `test_start_run_without_tags`: verifies tags are omitted when `None`.

## Default behavior

Defaults to `None` (omit), so existing callers see no behavior change. Tags are sent to the API only when the caller supplies them.

## Testing

- `uv run --frozen pytest tests/test_workflow_execution.py` — passing.
- `pre-commit` (ruff check, ruff format, secret scanning, license header) — passing.

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the [project license](https://github.com/awslabs/mcp/blob/main/LICENSE).

PR #4506, opened 2026-08-18
fix(aws-healthomics-mcp-server): add workflow_type parameter to start_run for Ready2Run workflows

steditt:
## Summary

Adds an optional `workflow_type` parameter (`PRIVATE` or `READY2RUN`) to the `StartAHORun` tool (`start_run`) in the aws-healthomics-mcp-server, and validates the Ready2Run storage constraints.

Ready2Run (AWS-provided) workflows require `workflowType` to be set to `READY2RUN` on the HealthOmics `StartRun` API. Because the tool did not expose this field, starting a Ready2Run workflow through the MCP server failed with a `ResourceNotFoundException` that is hard to diagnose. This change exposes the field and guards the related storage constraints.

## Changes

- `tools/workflow_execution.py`
  - Add optional `workflow_type` parameter, validated via the shared `validate_workflow_type` utility.
  - For `READY2RUN` runs, reject `storage_capacity` and non-`DYNAMIC` `storage_type`, since Ready2Run workflows use fixed AWS-managed storage and the API rejects those settings.
  - Forward `storageType` / `scratchStorageMode` / `storageCapacity` only for non-Ready2Run runs; forward `workflowType` to the API when provided.
  - Remove an unreachable client-side date-filtering branch in `list_runs` (dead code) so the filtering path is single and clear.
- `tests/test_workflow_execution.py`
  - Cover `READY2RUN` pass-through, `PRIVATE`/omitted default behavior, invalid `workflow_type` rejection, and the `READY2RUN` + storage validation errors.

## Default behavior

Defaults to `None` (treated as `PRIVATE`), so existing callers see no behavior change. `workflowType` is sent to the API only when the caller provides a value.

## Testing

- `uv run --frozen pytest tests/test_workflow_execution.py` — passing.
- `pre-commit` (ruff check, ruff format, secret scanning, license header) — passing.

## Note

This resubmits work that was previously reviewed and approved in #4350, which was closed unmerged when its source branch was deleted. The change here is the same feature, rebased onto current `main` as a single commit.

---


steditt:
This PR resubmits the change previously reviewed and approved in #4350. That PR was not merged: I deleted its source branch assuming the change had already been merged, which closed the PR. I noticed only later that it had never actually been merged, hence this new PR.

I've also incorporated the review nits from #4350 in the follow-up commit:

- **Vacuous assertion** in `test_start_run_invalid_workflow_type`: the `... or isinstance(result, dict)` clause was always true, so it asserted nothing. It now asserts the returned error dict contains the rejected value.
- **Stale line-number labels** in test docstrings/comments: removed, since they drift out of date as the source changes. Also rewrote the no-filter `list_runs` test's docstring/comments, which described the dead else-branch that this change removes.
- **`Literal` typing suggestion for `workflow_type`**: I kept the existing `Optional[str]` + shared `validate_workflow_type` approach. The sibling enum params in `start_run` (`storage_type`, `networking_mode`, `scratch_storage_mode`) all use `str` + manual validation, and there is no `Literal` usage in the package today. Switching only this parameter would make it inconsistent with its peers; moving the whole family to `Literal` would be a separate codebase-wide change. Happy to do it if you'd prefer.

alxawan (approved):
LGTM

steditt:
Hi @markjschreiber - this PR has been approved by @alxawan (healthomics code owner), and I've just updated the branch so it's current with main (no conflicts). Could you take a look when you have a moment so it can be merged? It's part of a set of three small StartRun enhancements for the aws-healthomics-mcp-server (#4504, #4505, #4506). Thanks!

markjschreiber (approved):
LGTM

PR #4509, opened 2026-08-18
feat(billing-cost-management-mcp-server): encode invoicing time-filter rules in guidance and validation

malpani:
## Summary
Makes the `invoicing` tool's `list_invoice_summaries` operation self-sufficient about the
AWS Invoicing service's time-filter rules, so an agent can get a correct multi-month answer
on the first attempt.
The service constrains `ListInvoiceSummaries` in three ways that a caller cannot discover
from the tool schema alone. This encodes each one — in the description where it guides the
model, and in local validation where it can be enforced cheaply. Every rule below was
established against the live Invoicing API rather than inferred from the SDK or docs.
### Changes
**1. One-month maximum, sized to the start month.** The service accepts a
`start_date`/`end_date` span of at most the number of days in the *start* date's month,
rejecting anything longer with `TimePeriod cannot last more than a month`:
| Range | Span | Start month | Result |
|---|---|---|---|
| `2026-01-31` → `2026-03-03` | 31d | Jan (31) | accepted |
| `2026-01-31` → `2026-03-04` | 32d | Jan (31) | rejected |
| `2026-02-15` → `2026-03-15` | 28d | Feb (28) | accepted |
| `2026-02-15` → `2026-03-16` | 29d | Feb (28) | rejected |

This is `span_days <= days_in(start month)` and *not* a clamp to the same day of the next
month — a range starting Jan 31 is good through Mar 3. A flat 31-day cap would be wrong in
both directions, so the check tracks the start month's length. It truncates to whole days to
match the service, which accepts `2026-02-15` → `2026-03-15T23:59:59`.
The description now states the rule, and an over-long range is refused locally with
`suggested_date_ranges` — per-month sub-ranges that together cover the request, ready to
issue as-is.
Those suggestions are deliberately **dates, not billing periods**: a `TimeInterval` filters
on *issued date* while a `BillingPeriod` filters on *billing month*. Substituting one for the
other is the natural shortcut and it quietly answers a different question, so the description
now calls out that the two filters are not interchangeable.
**2. Explicit bound semantics.** A date-only bound is `00:00:00 UTC`, which makes `end_date`
exclusive of the named day in practice. Worth stating plainly because the intuitive reading
costs you the final day: with an invoice issued `2026-02-03T17:27:03Z`, `2026-02-01` →
`2026-02-03` returns nothing while `2026-02-01` → `2026-02-04` returns the invoice. The
description and examples now show a full June as `2026-06-01` → `2026-07-01`.
**3. Time filter is mandatory for the account selector.** `ACCOUNT_ID` requires either
`BillingPeriod` or `TimeInterval` — without one the service returns `BillingPeriod and
TimePeriod cannot both be null`, including when only `InvoicingEntity` is supplied.
`INVOICE_ID` may omit it. The description now distinguishes the two, and the account case is
validated locally.
Reversed and empty ranges are also refused locally — the service rejects both, and a negative
span would otherwise slip past the span check.
**Testing.** Invoicing tests 29 → 32; 1247 pass for the package; `ruff check` and
`ruff format --check` clean. Each new test cites the live-API observation it encodes. The
sub-range splitter was additionally property-checked over 6,579 generated ranges — every
sub-range within its own start-month limit, caller endpoints preserved, no gaps, no
zero-length ranges — with 0 violations, and 15 generated sub-ranges were replayed against the
live API and all accepted.
**Note for reviewers.** The local checks duplicate server-side rules, which I know is a
trade-off. They are deliberately exact rather than conservative — each mirrors a rule
confirmed against the live API, so none can reject a request the service would have accepted.
If you would prefer a smaller change, the description updates stand on their own and I am
happy to split them out; `_split_range_by_month` is the most cuttable piece. A follow-up PR
will make `AccessDenied` responses on this operation actionable, kept separate because it is
orthogonal to the time-filter rules.
### User experience
**Before** — a multi-month request, following the example in the tool description:
```
{"operation": "list_invoice_summaries", "account_id": "…",
 "start_date": "2026-01-01", "end_date": "2026-06-30"}
→ ValidationException: TimePeriod cannot last more than a month
```

Nothing in the error or the schema conveys what the limit actually is, so the retry is a
guess. A reasonable next attempt — 31-day chunks — succeeds from a January start and fails
from a February one, and splitting by `billing_period` instead silently changes the question
from issued date to billing month.
**After:**
```
{"status": "error", "data": {
  "message": "start_date to end_date spans 180 days. The Invoicing service accepts at
    most one month per call, measured as the number of days in the start date's month
    (31 here). Issue one call per suggested_date_ranges entry instead of widening the
    range, then combine the results.",
  "suggested_date_ranges": [
    {"start_date": "2026-01-01", "end_date": "2026-02-01"},
    {"start_date": "2026-02-01", "end_date": "2026-03-01"},
    …
    {"start_date": "2026-06-01", "end_date": "2026-06-30"}
  ]}}
```

No API call is made, and the recovery path is directly executable. With the rule now in the
description, the common case is that the model splits correctly up front and never sees this
error.
## Checklist
* [x] Changes are documented
Is this a breaking change? (Y/N) N — calls that previously reached the service and failed now
fail locally with actionable guidance; no previously-succeeding call changes behavior. The
over-long-range error gains a `suggested_date_ranges` field. No parameters or return shapes
were removed or renamed.
**RFC issue number**: N/A
Checklist:
## Acknowledgment


Comusus:
Approving workflows to run.

Comusus (commented):
Change LGTM. Only feedback is that we are missing unit tests around YoY month transition and Leap Year events.

Comusus (approved):

wangyuhere (approved):

PR #3487, opened 2026-05-11
fix: postgres-mcp-server add host validation during connect to DB API

kennthhz:
Fixes

### Changes

Add host validation during DB connection API

### User experience

If the host parameter specified by MCP user doesn't match the cluster/instance configuration, it will be rejected.

## Checklist

If your change doesn't seem to apply, please leave them unchecked.

* [ x] I have reviewed the [contributing guidelines](https://github.com/awslabs/mcp/blob/main/CONTRIBUTING.md)
* [ x] I have performed a self-review of this change
* [ x] Changes have been tested
* [ x] Changes are documented

Is this a breaking change? (Y/N)
N

**RFC issue number**:

Checklist:

* [ ] Migration process documented
* [ ] Implement warnings (if it can live side by side)

## Acknowledgment

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the [project license](https://github.com/awslabs/mcp/blob/main/LICENSE).

PR #4515, opened 2026-08-19
fix(mysql-mcp-server): add host validation during connect to DB API

manaswini1920:
### Fixes
Fixes a confused-deputy credential exfiltration vulnerability where a caller-supplied db_endpoint was used verbatim, without validation against the cluster's actual AWS endpoints (CWE-918, CWE-522).

###  Changes
Before opening a wire-protocol connection, validate the caller-supplied db_endpoint against the cluster's AWS-resolved endpoints (writer, reader, custom, and member-instance endpoints) and reject any mismatch with a ValueError. This prevents a caller from directing the cluster's master credentials or an IAM auth token at an attacker-controlled host.

- Add internal_resolve_cluster_endpoint() in cp_api_connection.py: matches the caller host against the cluster's endpoints (case-insensitive; normalization owned by the resolver) and returns the AWS-sourced (host, port).
- Writer/reader/custom are checked first; member instances are enumerated only when none of those match (no N+1 on the common path).
- Ports are AWS-sourced; a missing or malformed port fails closed rather than defaulting to 3306, on both the cluster and standalone-instance paths.
- Endpoint validation runs before the connection-cache lookup for wire-protocol methods, so a warm cached connection cannot bypass validation. RDS_API keeps its cache fast-path (it connects via the Data API using cluster_arn and never dials db_endpoint).
- Validation applies only to wire-protocol methods (MYSQL_WIRE_PROTOCOL, MYSQL_WIRE_IAM_PROTOCOL); RDS_API continues to ignore db_endpoint/port as documented.
- AccessDenied/throttling from describe_db_instances is surfaced to the caller instead of degrading into a misleading "does not match" rejection.
- Document the required rds:DescribeDBClusters / rds:DescribeDBInstances permissions under the wire-protocol prerequisites in the README.
- Move the operator SBOM/validation script to scripts/e2e_integration_check.py (it is a manual script, not a pytest module).
- Add unit tests covering endpoint resolution, validation, the cache-bypass regression, IAM-path rejection, case-insensitive matching, wrong-port normalization, the RDS_API skip, fail-closed ports, and permission/throttling behavior.
- Ports the endpoint-validation approach from postgres-mcp-server PR #3487.

```
TLS enforcement on the Secrets Manager credential path (originally bundled here) has been split into a separate follow-up PR per review, so it can be reviewed and rolled back independently.
```

### User experience

Before: A malicious MCP client could set db_endpoint to an attacker-controlled host while using a legitimate cluster_identifier, causing the server to send the cluster's master password to the attacker.

After: Any db_endpoint that does not match the cluster's AWS-resolved endpoints is rejected with a clear error message. Legitimate connections are unaffected.

###  Checklist
[x] I have reviewed the contributing guidelines
[x] Changes have been tested
[x] Changes are documented
Is this a breaking change? N

###  Acknowledgment


src/mysql-mcp-server/tests/e2e/__init__.py
happycupofjava: We do not need to check- in the e2e filer inder /tests.

.secrets.baseline
happycupofjava, line 1: We need not check in this code too, any PR should not have secrets or hashes.

happycupofjava: NVM, ignore

happycupofjava (dismissed):

src/mysql-mcp-server/awslabs/mysql_mcp_server/connection/asyncmy_pool_connection.py
```diff
@@ -188,6 +194,19 @@ async def _initialize_pool_unlocked(self):
                     'CERTIFICATE_VERIFY_FAILED. Supply --ca_bundle <path> '
                     'or reinstall the package to restore the bundled bundle.'
                 )
+        elif self.secret_arn:
```
kennthhz: **Blocking:** `--ca_bundle` never reaches this branch, so preference 1 in the comment above ("`--ca_bundle` override supplied by the operator") is unreachable on the Secrets Manager path.

`internal_connect_to_database` passes `ca_bundle_path=ca_bundle_path` when it builds the IAM connection, but not when it builds the `MYSQL_WIRE_PROTOCOL` one — so `self.ca_bundle_path` is always `None` here and this always falls through to `_bundled_ca_file()`.

Verified locally: with `server_module.ca_bundle_path = '/tmp/operator-supplied.pem'`, the connection returned for `MYSQL_WIRE_PROTOCOL` reports `ca_bundle_path = None`.

That matters because of what this PR changes. TLS is now mandatory on this path, so an install whose server certificate doesn't chain to the bundled RDS bundle fails with `CERTIFICATE_VERIFY_FAILED` — and the documented escape hatch doesn't work. One-line fix in the `MYSQL_WIRE_PROTOCOL` branch of `internal_connect_to_database`:

```python
db_connection = AsyncmyPoolConnection(
    ...
    is_iam_auth=False,
    ca_bundle_path=ca_bundle_path,
)
```

Worth a test that asserts the override is honoured on the secret path, alongside the existing "no CA bundle" one.

src/mysql-mcp-server/awslabs/mysql_mcp_server/connection/asyncmy_pool_connection.py
```diff
@@ -188,6 +194,19 @@ async def _initialize_pool_unlocked(self):
...
+                ssl_ctx = ssl_module.create_default_context(cafile=cafile)
+                logger.debug('Using CA bundle for Secrets Manager auth: {}', cafile)
+            else:
+                ssl_ctx = ssl_module.create_default_context()
+                logger.warning(
+                    'No verified RDS CA bundle available; falling back to '
+                    'the system trust store for Secrets Manager credential path. '
+                    'Supply --ca_bundle <path> or reinstall the package to '
+                    'restore the bundled bundle.'
+                )
```
kennthhz: Nit: this branch is byte-identical to the `is_iam_auth` branch above apart from log wording, so the trust-decision logic now lives in two places that have to stay in sync. Suggest collapsing:

```python
if self.is_iam_auth or self.secret_arn:
```

and folding the auth-flavour detail into the log message instead of the control flow.

src/mysql-mcp-server/awslabs/mysql_mcp_server/server.py
```diff
@@ -700,11 +702,57 @@ def internal_connect_to_database(
         if not db_endpoint:
             db_endpoint = cluster_properties.get('Endpoint', '')
             port = int(cluster_properties.get('Port', ''))
+        else:
```
kennthhz: **Blocking:** the `db_connection_map.get(...)` dedupe check ~30 lines above returns early, so a warm cache entry makes this whole validation inert.

`DBConnectionMap.get` has a documented relaxed scan that matches on `(method, database)` plus *either* `cluster_identifier` *or* `db_endpoint`. So once any legitimate connection to the cluster exists, a subsequent call with the same cluster/database/method and an arbitrary `db_endpoint` hits the cache and never reaches this code.

Verified locally — after one legitimate connect to the writer endpoint, a second call with `db_endpoint='attacker.evil.com', port=3306` raises nothing, returns the *same* connection object, and echoes back:

```json
{"db_endpoint": "attacker.evil.com", ...}
```

To be clear on severity: credentials are not exposed, because the cached pool points at the real cluster — the exfiltration path is genuinely closed. But the guard doesn't fire, and the tool confirms to the model that it connected to an endpoint it was supposed to reject. Since ordering decides whether the check runs at all, I'd move the validation above the cache lookup so it's unconditional.

src/mysql-mcp-server/awslabs/mysql_mcp_server/server.py
```diff
@@ -700,11 +702,57 @@ def internal_connect_to_database(
...
+            # Endpoint supplied: validate (host, port) against the cluster's
+            # advertised endpoints (writer, reader, custom, members). If the
+            # caller-supplied pair does not match any legitimate endpoint,
+            # refuse the connection — this prevents directing credentials or
+            # IAM auth tokens at an attacker-controlled host.
+            #
+            # On match, we overwrite the local db_endpoint/port with the
+            # AWS-sourced strings so the connection string is never built
+            # from caller input.
+            valid_endpoints = internal_get_cluster_valid_endpoints(
```
kennthhz: This runs for every `connection_method`, but the `connect_to_database` docstring says of scenario 1 (`RDS_API`): "db_endpoint and port will be ignored".

So an `RDS_API` caller that passes a stale or irrelevant endpoint now gets a hard `ValueError` plus N `describe_db_instances` calls, on a path that never dials the host (it connects via `cluster_arn` + `secret_arn`). Either gate the validation on the wire-protocol methods, or update the docstring — right now the documented contract and the code disagree.

src/mysql-mcp-server/awslabs/mysql_mcp_server/server.py
```diff
@@ -700,11 +702,57 @@ def internal_connect_to_database(
...
+                cluster_properties=cluster_properties, region=region
+            )
+
+            requested_host = db_endpoint.strip().lower()
+            matched: Optional[Tuple[str, int]] = None
+            for valid_host, valid_port in valid_endpoints:
+                if valid_host.lower() == requested_host and valid_port == port:
```
kennthhz: Requiring the port to match creates a failure mode that isn't a security problem. `port` is a required parameter on the `connect_to_database` tool with no default, so a model that resolves `db_endpoint` correctly but guesses `3306` for a cluster listening on `3307` now gets a security-flavoured rejection for what is really a guessing problem.

It's also inconsistent with the standalone-instance path below, which silently *replaces* the caller's port with the AWS value and never rejects.

Suggest matching on host only, then taking both values from AWS:

```python
for valid_host, valid_port in valid_endpoints:
    if valid_host.lower() == requested_host:
        matched = (valid_host, valid_port)
        break
```

That's no less safe — the port is AWS-sourced either way — it makes the two paths consistent, and it lets the port-guessing fallback in `internal_get_cluster_valid_endpoints` go away entirely.

src/mysql-mcp-server/awslabs/mysql_mcp_server/server.py
```diff
@@ -700,11 +702,57 @@ def internal_connect_to_database(
...
+                    matched = (valid_host, valid_port)
+                    break
+
+            if matched is None:
+                valid_repr = ', '.join(f'{h}:{p}' for h, p in valid_endpoints) or '<none>'
```
kennthhz: Two nits here:

- The message is retyped for the `ValueError` rather than sharing the prefix, so the two copies will drift. Keeping the endpoint list out of the client-facing error is the right call — just build the shared part once.
- `or '<none>'` is dead code: `internal_get_cluster_valid_endpoints` raises when the list is empty, so `valid_endpoints` is guaranteed non-empty here.

src/mysql-mcp-server/awslabs/mysql_mcp_server/server.py
```diff
@@ -700,11 +702,57 @@ def internal_connect_to_database(
...
         instance_properties = internal_get_instance_properties(db_endpoint, region)
         masteruser = instance_properties.get('MasterUsername', '')
         secret_arn = instance_properties.get('MasterUserSecret', {}).get('SecretArn')
-        port = int(instance_properties.get('Endpoint', {}).get('Port'))
+        instance_endpoint = instance_properties.get('Endpoint', {}) or {}
+        resolved_host = instance_endpoint.get('Address', '')
+        try:
+            resolved_port = int(instance_endpoint.get('Port') or 0)
+        except (TypeError, ValueError):
+            resolved_port = DEFAULT_MYSQL_PORT
```
kennthhz: This converts a fail-fast into a silent guess. Previously `int(instance_properties.get('Endpoint', {}).get('Port'))` raised on a missing or malformed port; now it becomes `3306` and gets used to build the connection.

If AWS hands back a port we can't parse, that's an unexpected-state signal worth surfacing, not defaulting through. Same pattern as the fallback in `internal_get_cluster_valid_endpoints` — I'd make both fail closed. Note `test_standalone_instance_invalid_port_fallback` pins the current behaviour and would need to flip.

kennthhz (commented):
Thanks for porting this over — the core fix is sound and it lands in the right place.

**What's right:** validation sits at a single choke point (`internal_connect_to_database`) covering both wire-protocol paths; on a match the caller's string is *replaced* with the AWS-sourced one rather than merely compared, which is the detail most versions of this fix get wrong; host comparison is case-insensitive and tested; the empty-endpoint-set case fails closed; and the client-facing `ValueError` deliberately omits the endpoint list while the log keeps it. The standalone-instance path was already safe via `internal_get_instance_properties`, so the changes there are correctly no-ops on the security question.

I have three blocking items and a handful of smaller ones, left inline. Summary:

**Blocking**

1. `--ca_bundle` never reaches the new TLS path — `server.py` omits `ca_bundle_path` when building the `MYSQL_WIRE_PROTOCOL` connection, so the operator override is unreachable exactly where TLS just became mandatory. One-line fix. (inline on `asyncmy_pool_connection.py`)
2. A warm connection-cache entry makes the validation inert, and the tool reports success for a bogus endpoint. (inline on `server.py`)
3. Enforcing TLS on the Secrets Manager path reverses a deliberate, documented compat decision and is a breaking change. (inline on `test_ca_bundle.py`)

**Not anchorable inline**

- The PR says breaking change = N. The TLS force-upgrade can break installations that were relying on plaintext or server-certificate-managed TLS — the test being flipped says so explicitly. That needs to be called out in the PR and in the docs.
- "Changes are documented" is a stretch: no README or docstring changes in the diff. Two things now need documenting — the TLS behaviour change, and the new `rds:DescribeDBInstances` dependency (the README has no IAM permission list today). The `connect_to_database` docstring also still says `db_endpoint` is ignored for `RDS_API`, which is no longer true.
- Consider splitting the TLS change into its own PR. It's orthogonal to endpoint validation, has a different risk profile, and a different rollback story.

**Verification:** checked out `57d391b8` locally — 453 tests pass, `ruff` clean, `pyright` 0 errors. I confirmed the cache-bypass and `--ca_bundle` findings with a throwaway probe test rather than by reading alone; the observed output is quoted in those two comments.

src/mysql-mcp-server/awslabs/mysql_mcp_server/connection/cp_api_connection.py
```diff
@@ -21,14 +21,122 @@
...
+    IAM auth tokens.
+
+    The returned list includes, where present:
+    - Writer endpoint (cluster.Endpoint)
+    - Reader endpoint (cluster.ReaderEndpoint)
+    - Custom endpoints (cluster.CustomEndpoints)
+    - Each member instance endpoint (from describe_db_instances)
+
+    Host comparison should be case-insensitive (DNS is case-insensitive);
+    this function returns hosts as reported by AWS without normalization.
```
kennthhz: Nit: this pushes an invariant onto every caller. `server.py` does `.lower()` correctly today, but a future caller may not, and a missed `.lower()` here fails *closed* in a confusing way (legitimate mixed-case endpoint rejected).

Since this function already owns "what counts as valid", normalising here — or returning lowercased hosts — makes it hard to get wrong.

src/mysql-mcp-server/awslabs/mysql_mcp_server/connection/cp_api_connection.py
```diff
@@ -21,14 +21,122 @@
...
+            transient API state; we treat it as an error rather than
+            silently accepting any caller-supplied endpoint.
+    """
+    endpoints: List[Tuple[str, int]] = []
+
+    cluster_port_raw = cluster_properties.get('Port')
+    try:
+        cluster_port = int(cluster_port_raw) if cluster_port_raw is not None else 0
+    except (TypeError, ValueError):
+        cluster_port = DEFAULT_MYSQL_PORT
```
kennthhz: Fail-open on malformed input, and inconsistent with the `None` case on the line above:

- `Port=None` → `0` → endpoints excluded → `ValueError` (fails closed)
- `Port='invalid'` → `3306` → `host:3306` accepted on a **guessed** port (fails open)

Two flavours of the same malformed input, handled in opposite directions, inside the function whose only job is deciding which endpoints are legitimate.

A real cluster always reports a numeric port, so I'd drop the fallback and let both cases land on the `ValueError` at the end. `test_defaults_to_mysql_port_on_invalid` currently pins the fail-open behaviour and would need to flip.

src/mysql-mcp-server/awslabs/mysql_mcp_server/connection/cp_api_connection.py
```diff
@@ -21,14 +21,122 @@
...
+
+    members = cluster_properties.get('DBClusterMembers', []) or []
+    instance_ids = [
+        m.get('DBInstanceIdentifier') for m in members if m.get('DBInstanceIdentifier')
+    ]
+    if instance_ids:
+        rds_client = internal_create_rds_client(region=region)
+        for instance_id in instance_ids:
+            try:
+                response = rds_client.describe_db_instances(DBInstanceIdentifier=instance_id)
```
kennthhz, line 125: This is N+1 `describe_db_instances` calls on every connect that supplies an endpoint. A single filtered call covers all members:

```python
rds_client.describe_db_instances(
    Filters=[{
        'Name': 'db-cluster-id',
        'Values': [cluster_properties.get('DBClusterIdentifier')],
    }]
)
```

Better still, short-circuit: check writer/reader/custom first and only enumerate member instances when none of those match. Most connections use the writer endpoint, so the common case would make zero extra API calls — which also shrinks the throttling exposure noted in my comment below.

src/mysql-mcp-server/awslabs/mysql_mcp_server/connection/cp_api_connection.py
```diff
@@ -21,14 +21,122 @@
...
+                        port = int(port_raw) if port_raw is not None else 0
+                    except (TypeError, ValueError):
+                        port = 0
+                    if host and port:
+                        endpoints.append((host, port))
+            except ClientError as e:
+                # Don't fail the whole validation if a single instance lookup
+                # fails; log and continue. The caller still has to match one of
+                # the endpoints we successfully resolved.
+                logger.warning(
```
kennthhz: Swallowing this means the validation degrades to *reject* rather than to *error*.

If the server's role lacks `rds:DescribeDBInstances`, or the call is throttled, member endpoints silently drop out of the valid set — and a legitimate connection to an instance endpoint is then refused with "does not match any endpoint of cluster". That's a permissions or throttling problem wearing a security message, which is a rough one to debug from the client side, where the error is deliberately stripped of detail.

Two suggestions:
- Distinguish `AccessDenied` / `ThrottlingException` from `DBInstanceNotFound` in the log, so the operator can tell "this instance is gone" from "I couldn't look, so I rejected you".
- Document the new `rds:DescribeDBInstances` requirement. The README has no IAM permission list for this server today, and this PR adds a hard dependency on that call for anyone connecting via an instance endpoint.

src/mysql-mcp-server/tests/test_ca_bundle.py
```diff
@@ -250,8 +251,8 @@ def fake_create_default_context(cafile=None):
             )
             await conn.initialize_pool()
 
-        assert captured['ssl_calls'] == 0, (
-            'create_default_context was called for a non-IAM connection. '
-            'mysqlwire must remain plaintext-capable.'
+        assert captured['ssl_calls'] == 1, (
```
kennthhz: **Blocking (compat/scope):** this inverts a deliberate decision, and the assertion being replaced carried the rationale in its own docstring:

> Setting an SSL context on `mysqlwire` (no IAM) connections would force-upgrade them to TLS, which can break installations relying on plaintext or server-certificate-managed TLS.

Enforcing TLS is the right end state and I'm not arguing against it. But flipping a documented compat guarantee is a breaking change, and the PR checklist says breaking change = N with no README or CHANGELOG note. Combined with `--ca_bundle` not being wired to this path (see my comment on `asyncmy_pool_connection.py`), affected installs get `CERTIFICATE_VERIFY_FAILED` and no way out.

Concretely, I'd like: the `--ca_bundle` wiring fixed, the behaviour change documented, and the breaking-change flag revisited. And I'd suggest landing the TLS change as its own PR — it's orthogonal to endpoint validation, and if it causes fallout you want to be able to revert it without reverting the security fix.

src/mysql-mcp-server/tests/test_endpoint_validation.py
```diff
@@ -0,0 +1,733 @@
...
+                database_type=DatabaseType.AURORA_MYSQL,
+                connection_method=ConnectionMethod.MYSQL_WIRE_PROTOCOL,
+                cluster_identifier=CLUSTER_ID,
+                db_endpoint=READER_ENDPOINT,
+                port=CLUSTER_PORT,
+                database='testdb',
+            )
+            assert conn is not None
+
+    def test_bogus_endpoint_rejected(self):
```
kennthhz, line 312: Every rejection test in this class uses `MYSQL_WIRE_PROTOCOL`. `MYSQL_WIRE_IAM_PROTOCOL` is arguably the higher-value target for this vulnerability — that path mints an IAM auth token *for the supplied host*.

It's covered by construction since both methods share the validation code, but it's the worst-case regression if the shared path ever gets refactored, and the test is a copy-paste. Worth adding one `test_bogus_endpoint_rejected_iam_auth`.

src/mysql-mcp-server/tests/test_endpoint_validation.py
```diff
@@ -0,0 +1,733 @@
...
+                    database_type=DatabaseType.AURORA_MYSQL,
+                    connection_method=ConnectionMethod.MYSQL_WIRE_PROTOCOL,
+                    cluster_identifier=CLUSTER_ID,
+                    db_endpoint='10.0.0.1',
+                    port=CLUSTER_PORT,
+                    database='testdb',
+                )
+
+
+# ---------------------------------------------------------------------------
```
kennthhz, line 635: Nit: "RPG overwrite" looks like a typo — host/port overwrite?

src/mysql-mcp-server/tests/e2e/e2e_integration_test.py
```diff
@@ -0,0 +1,435 @@
...
+
+Usage:
+    python -m tests.e2e.e2e_integration_test \
+        --cluster-identifier <cluster-id> \
+        --region <aws-region> \
+        --database <db-name> \
+        [--port 3306]
+
+Tests are marked with @pytest.mark.live and excluded from regular CI runs.
+Run with: pytest tests/e2e/ -m live
```
kennthhz: These two lines aren't accurate. There's no `@pytest.mark.live` anywhere in this file, and `python_files = "test_*.py"` in `pyproject.toml` means pytest will never collect `e2e_integration_test.py` — `pytest tests/e2e/ -m live` collects nothing. The `python -m tests.e2e.e2e_integration_test` line above it is the accurate usage; I'd just drop these two.

Bigger question: 435 lines that CI never runs will rot silently. If this is an operator script rather than a test, putting it under `scripts/` would signal that far better than `tests/`.

kennthhz (changes requested):

src/mysql-mcp-server/awslabs/mysql_mcp_server/connection/asyncmy_pool_connection.py
manaswini1920: Good catch. Thank you. I'm carrying this fix into separate TLS PR than fixing it here - Since i'm splitting the TLS enforcement out ( per your `test_ca_bundle.py` comment), this concern no longer holds in this branch. In the TLS PR I'll pass `ca_bundle_path` in `MYSQL_WIRE_PROTOCOL` construction and add a test asserting the override along side the existing "no CA bundle" one.

src/mysql-mcp-server/awslabs/mysql_mcp_server/connection/asyncmy_pool_connection.py
manaswini1920: Agreed. I'm splitting this into TLS PR. I'll collapse it to `if self.is_iam_auth or self.secret_arn:` with auth flavour in the log message there.

src/mysql-mcp-server/awslabs/mysql_mcp_server/server.py
manaswini1920: Fixed. Moved resolve-and-validate above the cache lookup for wire-protocol methods, so a warm cache entry can't make it inert - a bogus endpoint is now rejected before the cache is consulted, even after a legitimate connect. Added a regression test that warms the cache with the writer endpoint, then reconnects with attacher.evil.com and asserts - reproducing your exact scenario

src/mysql-mcp-server/awslabs/mysql_mcp_server/server.py
manaswini1920: Fixed by gating validation o nthe wire-protocol methods - RDS_API no longer validates or calls describe_db_instances so db_endpoint/port stay ignored as the docstring says. added test_rds_api_endpoint_not_validated test

src/mysql-mcp-server/awslabs/mysql_mcp_server/server.py
manaswini1920: Fixed, now matches on host only and takes both host and port from AWS. so a correct host with a guessed/wrong port is normalized instead of rejected and its consistent with the standalone-instance path. Replaced test_wrong_port_rejected with test_wrong_port_normalized_to_aws_value.

I kept the port handling in internal_get_cluster_valid_endpoints as-is for now, since the endpoint ports it returns are what we normalize to. Happy to simplify that fallback in a follow-up if you'd prefer.

src/mysql-mcp-server/awslabs/mysql_mcp_server/connection/cp_api_connection.py
manaswini1920: Fixed both.
1. Rather than just distinguishing them in a log, AccessDenied/throttling exception now re-raised, so a permissions/throttling failures surface as the real error instead of silently shrinking the valid set into a "does not match any endpoint" rejection. Only DBInstanceNotFound/DBInstanceNotFoundFault is skipped-and-continued. Added `test_access_denied_on_instance_lookup_reraisese`test
2. Documented the rds:DescribeDBInstances + rds:DescribeDBClusters requiremtn in READMe.

src/mysql-mcp-server/tests/test_ca_bundle.py
manaswini1920: Agreed on all of it. Taking your suggestion on the split. I've reverted TLs enforcement out of this PR entirely. This PR deals only with endpoint validation. Will add TLS validation in a separate PR along with --ca_bundle wiring fixed..

src/mysql-mcp-server/tests/test_endpoint_validation.py
manaswini1920: Added test_bogus_endpoint_rejected_iam_auth - same bogus endpoint scenario on MYSQL_WIRE_IAM_PROTOCOL, so the higher-value IAM path is pinned independently of the shared validation code.

src/mysql-mcp-server/tests/test_endpoint_validation.py
manaswini1920: Fixed - type, meant "host/port override"

src/mysql-mcp-server/tests/e2e/e2e_integration_test.py
manaswini1920: Fixed. Dropped the two inaccurate lines and moved the file to scripts/e2e_integration_check.py so it reads as the operator script it is rather than a test CI silently skips. removed th enow-empty tests/e2e/.

happycupofjava (approved):

kennthhz (approved):

arnewouters (approved):

PR #4523, opened 2026-08-22
fix(aws-api-mcp-server): pass allowed hosts and origins to the FastMCP HTTP guard

### LINKED ISSUE #4507: aws-api-mcp-server: streamable-http transport rejects all non-localhost Host headers with 421 Misdirected Request (FastMCP DNS-rebinding lock never sees AWS_API_MCP_HOST)
### Describe the bug

When running `aws-api-mcp-server` with the `streamable-http` transport bound to a non-loopback address (e.g. `AWS_API_MCP_HOST=0.0.0.0` or `AWS_API_MCP_HOST=*`, for a container reached through an ALB/ingress by DNS name), every real MCP request is rejected with `421 Misdirected Request`. Only requests whose `Host` header is literally `localhost`/`127.0.0.1` succeed.

This makes the `streamable-http` transport unusable for any container/ALB-style deployment that isn't accessed via `localhost` — which is the normal deployment shape for this transport (the alternative, `stdio`, doesn't need a `Host` header at all).

Root cause: `server.py`'s `main()` constructs the `FastMCP` server without a `host` argument:

```python
server = FastMCP(
    name='AWS-API-MCP',
    instructions=None if SUPPRESS_DEPRECATION_WARNING else DEPRECATION_MESSAGE,
    auth=get_server_auth(),
    middleware=[HTTPHeaderValidationMiddleware()] if TRANSPORT == 'streamable-http' else [],
)
```

`host` is only passed later, to `.run()`:

```python
server.run(
    transport=TRANSPORT,
    host=HOST,
    port=PORT,
    stateless_http=STATELESS_HTTP,
)
```

FastMCP's own DNS-rebinding protection (independent of this project's `HTTPHeaderValidationMiddleware`) decides its `allowed_hosts` from the `host` value known **at construction time**. Since `FastMCP()` is constructed with no `host`, it defaults to loopback-only protection, and the later `host=HOST` passed to `.run()` never updates it. This means `AWS_API_MCP_HOST` and `AWS_API_MCP_ALLOWED_HOSTS` — the env vars this project exposes and documents for exactly this purpose — have **no effect** on the check that is actually rejecting requests. They only feed this project's own separate `HTTPHeaderValidationMiddleware`, which correctly treats `*` as a wildcard, but that's not the layer returning 421.

This is the same "late host assignment" pattern reported independently against several other FastMCP-based MCP servers:
- https://github.com/getzep/graphiti/issues/1205
- https://github.com/blockscout/mcp-server/issues/333
- https://github.com/EnterpriseDB/pg-airman-mcp/issues/5
- https://github.com/CrowdStrike/falcon-mcp/issues/291 (fixed in https://github.com/CrowdStrike/falcon-mcp/pull/293 by passing `host`/`port` into the `FastMCP()` constructor instead of only `.run()`)

### Expected Behavior

Setting `AWS_API_MCP_HOST=0.0.0.0` (or `*`) and running the `streamable-http` transport should accept requests whose `Host` header is the server's real DNS name / any name, matching the `AWS_API_MCP_ALLOWED_HOSTS` wildcard behavior this project already documents and implements in its own `HTTPHeaderValidationMiddleware`.

### Current Behavior

Every request with a `Host` header other than `localhost`/`127.0.0.1`/`[::1]` gets:

```
HTTP/1.1 421 Misdirected Request
content-length: 19

Misdirected Request
```

A request with `Host: localhost` against the same server succeeds normally (`200 OK`, valid `initialize` response).

### Reproduction Steps

```bash
docker run -p 8000:8000 \
  -e AWS_REGION=us-east-1 \
  -e AWS_API_MCP_TRANSPORT=streamable-http \
  -e AWS_API_MCP_HOST='*' \
  -e AWS_API_MCP_STATELESS_HTTP=true \
  -e READ_OPERATIONS_ONLY=true \
  public.ecr.aws/awslabs-mcp/awslabs/aws-api-mcp-server:1.5.1

# Works: Host is localhost
curl -sD - http://127.0.0.1:8000/mcp -X POST \
  -H 'Host: localhost' \
  -H 'Content-Type: application/json' -H 'Accept: application/json, text/event-stream' \
  -d '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2026-06-18","capabilities":{},"clientInfo":{"name":"diag","version":"0.0.1"}}}'
# -> 200 OK

# Fails: Host is any real hostname, despite AWS_API_MCP_HOST=*
curl -sD - http://127.0.0.1:8000/mcp -X POST \
  -H 'Host: my-mcp-server.example.com' \
  -H 'Content-Type: application/json' -H 'Accept: application/json, text/event-stream' \
  -d '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2026-06-18","capabilities":{},"clientInfo":{"name":"diag","version":"0.0.1"}}}'
# -> 421 Misdirected Request
```

Reproduced against the published `1.5.1` image; confirmed present as far back as `1.4.0` (checked the locked `fastmcp` version in each intermediate release's `uv.lock` — `1.3.46` locks `fastmcp==3.0.1` and does not exhibit this; `1.4.0` through `1.5.1` all lock `fastmcp==3.4.3` or newer and all exhibit it).

### Possible Solution

Pass `host` (and `port`) to the `FastMCP()` constructor in `server.py`, not only to `.run()`, so FastMCP's DNS-rebinding decision is made with the real configured host instead of its default. This is the same fix already applied for an unrelated server here: https://github.com/CrowdStrike/falcon-mcp/pull/293.

If FastMCP's public API doesn't cleanly support passing `host` at construction time in the pinned version, an alternative is constructing an explicit `TransportSecuritySettings(allowed_hosts=ALLOWED_HOSTS.split(','), allowed_origins=ALLOWED_ORIGINS.split(','))` and passing it into `FastMCP()`, reusing this project's existing `AWS_API_MCP_ALLOWED_HOSTS`/`AWS_API_MCP_ALLOWED_ORIGINS` config so both validation layers agree.

This currently makes `AWS_API_MCP_HOST` and `AWS_API_MCP_ALLOWED_HOSTS` non-functional/misleading for any non-`stdio` transport reached by DNS name (container behind an ALB/ingress, reverse proxy, etc.) — precisely the deployment shape the `streamable-http` transport exists for. We hit this upgrading a production deployment from `1.3.46` (working) to `1.5.1` (broken for all real traffic; only kubelet/ALB health checks against `GET /mcp`, which 405 before reaching this path, continued to look healthy).

### OS
Container (Linux), image `public.ecr.aws/awslabs-mcp/awslabs/aws-api-mcp-server`

### Server
aws-api-mcp-server

### Server Version
1.5.1 (also reproduced in 1.4.0, 1.4.1, 1.4.2, 1.5.0)

us-east-1

Om-singhaI:
Fixes #4507

### Changes

`aws-api-mcp-server` builds its `FastMCP` instance without any host information and only passes `host` and `port` to `server.run()`. FastMCP 3.4.3 (locked since release 1.4.0) wraps the `streamable-http` app in its own `HostOriginGuardMiddleware`, added in PrefectHQ/fastmcp#4405. That guard accepts loopback `Host` headers, anything in its `allowed_hosts` list, and the bound address when that address is not the unspecified one. Nothing in this server populated `allowed_hosts`, so a container bound to `0.0.0.0` behind an ALB or ingress answered every real request with `421 Misdirected Request`, and only `Host: localhost` got through. `AWS_API_MCP_ALLOWED_HOSTS` and `AWS_API_MCP_ALLOWED_ORIGINS` had no effect on this because they only feed the project's own `HTTPHeaderValidationMiddleware`, which runs after the FastMCP guard.

The fix is deliberately small, in keeping with the end of development notice on this server:

* `server.py` now imports `ALLOWED_HOSTS` and `ALLOWED_ORIGINS` from `config.py` and passes `allowed_hosts=ALLOWED_HOSTS.split(',')` and `allowed_origins=ALLOWED_ORIGINS.split(',')` to `server.run()` on the HTTP branch. `FastMCP.run()` forwards extra keyword arguments to `run_http_async()`, whose signature in fastmcp 3.4.3 accepts exactly `allowed_hosts: list[str] | None` and `allowed_origins: list[str] | None` and installs them in the guard. The guard treats `*` as a wildcard, matching the semantics this project already documents for these variables. The stdio branch is untouched because `run_stdio_async()` does not accept these arguments.
* `pyproject.toml` and `uv.lock`: the `fastmcp` floor moves from `>=3.2.0` to `>=3.4.3`. The two keyword arguments do not exist before 3.4.3 (I checked the `v3.2.0` and `v3.4.2` tags of PrefectHQ/fastmcp: neither `server/http.py` nor `server/mixins/transport.py` mentions `allowed_hosts` or `HostOriginGuard`; both appear in `v3.4.3`). Without the bump, an install that resolved any permitted fastmcp between 3.2.0 and 3.4.2 would pass `allowed_hosts` to a `run_http_async()` that does not accept it and crash at startup with `TypeError: run_http_async() got an unexpected keyword argument 'allowed_hosts'`, even though those versions never had the 421 problem. Raising the floor keeps `server.py` and the dependency contract in step, in the same spirit as #4454. The lock was regenerated with `uv lock` (uv 0.12.3, the version that wrote the current lock) and the only change is the `requires-dist` specifier line; the resolved version was already 3.4.3, so nothing installed moves.
* `tests/test_server.py`: `test_main_success_with_http_transport` now asserts the two new keyword arguments, and a new test `test_main_http_transport_allowed_hosts_reach_fastmcp_guard` builds the real HTTP app with exactly the keyword arguments `main()` hands to `server.run()` and sends an `initialize` request through FastMCP's guard with `Host: my-mcp-server.example.com` while the app is bound to `0.0.0.0:8080`. Without the two keyword arguments that test fails with `{'my-mcp-server.example.com': 421} != {'my-mcp-server.example.com': 200}`. The `starlette.testclient` import lives inside that one test so its `StarletteDeprecationWarning` (httpx versus httpx2 under starlette 1.3.1) stays scoped to the test that needs it.
* `CHANGELOG.md`: entry under `Unreleased` / `Fixed`, mentioning the floor bump.

A note on the suggestion in the issue: passing `host` to the `FastMCP()` constructor does not work on fastmcp 3.x; the constructor rejects `host` with a message pointing at `run_http_async()` or `FASTMCP_HOST`. Passing `allowed_hosts` and `allowed_origins` to `run()` is the supported path.

Default behaviour is unchanged for local use: `ALLOWED_HOSTS` defaults to `AWS_API_MCP_HOST`, which defaults to `127.0.0.1`, a value the guard already allows. With `AWS_API_MCP_HOST=0.0.0.0` and no explicit allow list, both layers now agree on allowing `Host: 0.0.0.0` and loopback names and rejecting everything else, so operators still have to opt in to a real DNS name or `*` as the README describes. The issue's expected behaviour section asks for `AWS_API_MCP_HOST=0.0.0.0` alone to accept real DNS names; I left that as an explicit opt in because the project's own middleware already rejects such hosts by default and silently widening the default felt like a separate decision for the maintainers.

One difference between the two layers worth knowing about: the README says port numbers are stripped during validation. That is still true of the project's middleware for both headers, and of the FastMCP guard for `Host`, but the guard compares `Origin` values including any explicit port. A browser origin such as `https://x.example.com:8443` therefore needs `https://x.example.com:8443` in `AWS_API_MCP_ALLOWED_ORIGINS` for the guard and `https://x.example.com` for the project middleware (or `*`). Such requests were rejected with 421 before this change, so this is not a regression, and I have not touched the README so the maintainers can decide how they want that documented.

### User experience

Before (published image 1.5.1, `AWS_API_MCP_TRANSPORT=streamable-http`, `AWS_API_MCP_HOST=0.0.0.0`, `AWS_API_MCP_ALLOWED_HOSTS=*`):

```
POST /mcp  Host: localhost                   -> 200 OK
POST /mcp  Host: my-mcp-server.example.com   -> 421 Misdirected Request
```

After, same configuration:

```
POST /mcp  Host: localhost                   -> 200 OK
POST /mcp  Host: my-mcp-server.example.com   -> 200 OK
```

Reproduced in process on `main` with `starlette.testclient` against the real fastmcp 3.4.3 guard, building the app with the same keyword arguments `main()` uses, and confirmed fixed with the change.

Test run in `src/aws-api-mcp-server` with a venv pinned to the lock file versions (fastmcp 3.4.3, mcp 1.29.0, starlette 1.3.1):

* before: `tests/test_server.py` 71 passed, 1 failed
* after: 72 passed, 1 failed

The one failure, `test_call_aws_ecs_deploy_error_sanitization`, is present on pristine `main` in this environment (pytest reports "async def functions are not natively supported" for that single test) and is unrelated to this change.

`ruff check` and `ruff format --check` pass on the changed files. `pip check` passes with fastmcp 3.4.3 installed against the new floor.

* One precedence detail worth knowing: FastMCP only consults its own `FASTMCP_HTTP_ALLOWED_HOSTS` and `FASTMCP_HTTP_ALLOWED_ORIGINS` settings when the keyword arguments are `None`. After this change they are always passed, so `AWS_API_MCP_ALLOWED_HOSTS` and `AWS_API_MCP_ALLOWED_ORIGINS` become the single source of truth for the guard. Anyone who worked around this issue with the `FASTMCP_*` variables already had to set the `AWS_API_MCP_*` ones for the project's own header middleware to pass, so in practice nothing that worked before stops working. `FASTMCP_HTTP_HOST_ORIGIN_PROTECTION=false` still disables the guard entirely since `host_origin_protection` is not passed.

## Checklist

* [x] Changes are documented

The `fastmcp` floor moves to `>=3.4.3`. Installs from the lock file or via `uvx` already resolve 3.4.3, so nothing changes for them; only an environment that pinned fastmcp between 3.2.0 and 3.4.2 would need to upgrade.

**RFC issue number**: N/A

Checklist:

## Acknowledgment

### DIFF
--- a/src/aws-api-mcp-server/CHANGELOG.md
+++ b/src/aws-api-mcp-server/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Missing api calls for s3 customizations in static map (#4548)
+- Pass `AWS_API_MCP_ALLOWED_HOSTS` and `AWS_API_MCP_ALLOWED_ORIGINS` through to FastMCP's Host and Origin guard so the `streamable-http` transport no longer rejects every non-localhost `Host` header with `421 Misdirected Request`. The `fastmcp` floor is raised to `>=3.4.3`, the first release that accepts these arguments (#4507)
 
 ## [1.5.1] - 2026-08-11
 
--- a/src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
+++ b/src/aws-api-mcp-server/awslabs/aws_api_mcp_server/server.py
@@ -26,6 +26,8 @@ from .core.aws.service import (
     validate,
 )
 from .core.common.config import (
+    ALLOWED_HOSTS,
+    ALLOWED_ORIGINS,
     DEFAULT_REGION,
     ENABLE_AGENT_SCRIPTS,
     ENDPOINT_SUGGEST_AWS_COMMANDS,
@@ -463,6 +465,8 @@ def main():
             host=HOST,
             port=PORT,
             stateless_http=STATELESS_HTTP,
+            allowed_hosts=ALLOWED_HOSTS.split(','),
+            allowed_origins=ALLOWED_ORIGINS.split(','),
         )
 
 
--- a/src/aws-api-mcp-server/pyproject.toml
+++ b/src/aws-api-mcp-server/pyproject.toml
@@ -22,7 +22,7 @@ dependencies = [
     "importlib_resources>=6.0.0",
     "requests>=2.32.4",
     "python-frontmatter>=1.1.0",
-    "fastmcp>=3.2.0",
+    "fastmcp>=3.4.3",
     "awscli==1.46.0",
 ]
 license = {text = "Apache-2.0"}
--- a/src/aws-api-mcp-server/tests/test_server.py
+++ b/src/aws-api-mcp-server/tests/test_server.py
@@ -874,6 +874,8 @@ def test_main_success_with_read_only_mode(
 @patch('awslabs.aws_api_mcp_server.server.HOST', '0.0.0.0')
 @patch('awslabs.aws_api_mcp_server.server.PORT', 8080)
 @patch('awslabs.aws_api_mcp_server.server.STATELESS_HTTP', True)
+@patch('awslabs.aws_api_mcp_server.server.ALLOWED_HOSTS', 'my-mcp-server.example.com,localhost')
+@patch('awslabs.aws_api_mcp_server.server.ALLOWED_ORIGINS', 'https://my-mcp-server.example.com')
 def test_main_success_with_http_transport(
     mock_get_read_only_operations,
     mock_server,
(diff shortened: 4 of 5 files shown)


arnewouters:
LGTM, I have made some changes to `CHANGELOG.md` so you'll have to rebase

Om-singhaI:
Rebased onto main and moved the changelog entry into the Unreleased section. Thanks @arnewouters for the quick review and the heads up.

arnewouters (approved):

anasstahr (approved):

PR #4525, opened 2026-08-22
fix(postgres-mcp-server): validate create_cluster bootstrap connection on reconnect

scottschreckengaust:
Fixes the `create_cluster` bootstrap privilege-check bypass reported as a follow-up on #4299 (the least-privilege guardrail PR).

## Summary

Under `--privilege_check=enforce` (the mode the guardrail PR calls "recommended for production"), an over-privileged bootstrap connection could stay reachable and queryable, defeating the guardrail.

`create_cluster` / `create_cluster_worker` connect to the freshly created cluster as its `rds_superuser` master and seed `db_connection_map` via `internal_create_connection` **without ever calling `validate_connection`**, leaving `effective_is_over_privileged=None`. Two paths could then reach that unvalidated connection:

1. A later `connect_to_database` for that same target hit the `was_cached` fast-return and reported success **before (and instead of)** running the guardrail.
2. A direct `run_query` (with no intervening `connect_to_database`) fetched the cached connection and executed **without any privilege validation at all**.

Either way, every subsequent query ran as `rds_superuser`. This PR closes **both** paths with a single validation/eviction primitive.

### Changes

- **`server.py`**
  - Add `validate_or_evict(db_connection)`: runs the post-connect guardrail (`initialize_pool` for pooled connections, then `validate_connection`) and, on rejection, **evicts and closes** the connection before re-raising. Eviction is what cures the downstream "trust the cache" consequence.
  - In `connect_to_database`, the `was_cached` branch now **lazily validates** a cached connection that was never validated (`effective_is_over_privileged is None`) — the "validate-on-create with a bootstrap exemption" approach documented as the maintainers' intended follow-up. Already-validated connections are **not** re-validated.
  - In `run_query`, add the same lazy guardrail for a never-validated cached connection so a **direct `run_query` cannot trust an unvalidated (possibly `rds_superuser`) bootstrap connection**. Honors the active policy: `enforce` rejects the query and evicts+closes the connection (fail-closed); `warn` allows and records the posture; `off` skips (operator opted out — no per-query probe). Gating on `effective_is_over_privileged is None` means the probe fires **at most once** per connection.
  - Refactor the newly-created-connection branch to call the same `validate_or_evict` helper (single validation/eviction path shared by all three call sites).
- **`tests/test_privilege_guardrail.py`**
  - `TestBootstrapReconnectGuardrail` (connect path): reconnect validates and **rejects** a bootstrap `rds_superuser` under `enforce`; the rejected/evicted connection is **not queryable**; an already-validated connection is **not re-validated**; under `warn` it is allowed **with an advisory**.
  - `TestRunQueryBootstrapGuardrail` (direct-`run_query` path): a direct `run_query` on an unvalidated `rds_superuser` connection is **refused and evicted** under `enforce` (the sensitive query never executes); **allowed with the posture recorded** under `warn`; an **already-validated connection is not re-probed**; and validation is **skipped under `off`**.
- **`tests/conftest.py`, `tests/test_server.py`** — the test doubles (`Mock_DBConnection`, `Mock_PsycopgPoolConnection`) now mirror `AbstractDBConnection`'s `effective_is_over_privileged` contract, and `setup_mock_connection` marks cached connections validated to reflect the realistic post-connect state (a connection reaching `run_query` in production has already been validated by `connect_to_database`).

### User experience

- **Before:** `create_cluster` → the over-privileged connection is reachable both via `connect_to_database` (cache fast-return, no guardrail) **and** via a direct `run_query` (no validation), so `run_query` runs as `rds_superuser` under `enforce`.
- **After:** the first use of a never-validated bootstrap connection — whether through `connect_to_database` **or** directly through `run_query` — runs the guardrail. Under `enforce` an over-privileged connection is rejected, evicted, and closed (not queryable); under `warn` it is allowed with the posture surfaced (advisory on connect, warning logged on query); under `off` nothing changes. Already-validated connections are unaffected (no redundant probe).

## Checklist

* [ ] Changes are documented

Is this a breaking change? (Y/N) N — behavior change only for previously-unvalidated over-privileged bootstrap connections under `enforce`, which the guardrail was already intended to reject.

**RFC issue number**:

Checklist:

## Acknowledgment

### DIFF
--- a/src/postgres-mcp-server/awslabs/postgres_mcp_server/server.py
+++ b/src/postgres-mcp-server/awslabs/postgres_mcp_server/server.py
@@ -451,6 +451,29 @@ async def run_query(
         )
         return [{'error': query_injection_risk_key}]
 
+    # Bootstrap hardening: a connection seeded by create_cluster enters the
+    # cache WITHOUT going through the validating connect/startup path, so it
+    # carries no privilege posture (effective_is_over_privileged is None). A
+    # direct run_query -- with no intervening connect_to_database to trip the
+    # cache-hit guardrail -- would otherwise execute against that unvalidated
+    # (possibly rds_superuser) connection. Validate it lazily on first use.
+    # 'off' means the operator opted out of privilege checks entirely, so skip
+    # (also avoids a per-query probe); gating on `is None` makes the probe fire
+    # at most once -- enforce/warn record the posture, so later queries on the
+    # same connection short-circuit. Under enforce validate_or_evict evicts and
+    # closes the connection before raising, so it is unreachable thereafter.
+    if (
+        privilege_check_policy != PRIVILEGE_CHECK_OFF
+        and db_connection.effective_is_over_privileged is None
+    ):
+        try:
+            await validate_or_evict(db_connection)
+        except Exception as e:
+            msg = f'Connection rejected by privilege guardrail: {type(e).__name__}: {e}'
+            logger.error(msg)
+            await ctx.error(msg)
+            return [{'error': msg}]
+
     try:
         logger.debug(
             (
--- a/src/postgres-mcp-server/awslabs/postgres_mcp_server/server.py
+++ b/src/postgres-mcp-server/awslabs/postgres_mcp_server/server.py
@@ -337,6 +337,38 @@ def attach_privilege_advisory(llm_response: str, db_connection: AbstractDBConnec
     return json.dumps(payload, indent=2, default=str)
 
 
+async def validate_or_evict(db_connection: AbstractDBConnection) -> None:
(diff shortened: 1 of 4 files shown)


src/postgres-mcp-server/awslabs/postgres_mcp_server/server.py
kennthhz, line 1: The current functional behavior is intentional. We don't want to block the default superuser role as part of create cluster because it is a bad user experience. I am exploring the possible alternative: to add less privileged Postgres role as part of cluster. This would requires coordination with control plane team which owns this experience so it will take some time. In the team time, we have documented in the README.md that user can create the less privileged role and use it for MCP.

PR #4527, opened 2026-08-24
fix(aws-documentation-mcp-server): stop returning Highly Rated recommendations

mc5672:
### Changes

The Highly Rated recommendation type is being retired upstream, so the `highlyRated` block in the recommendations API response will stop being populated.

- Dropped the `highlyRated` branch from `parse_recommendation_results`.
- Updated the `recommend` tool description to three categories (New, Similar, Journey), and removed the "most popular pages for a service" guidance, which was Highly Rated only.
- Dropped the `highlyRated` cases from the parser tests, and switched one block  in the `test_recommend` mock response from `highlyRated` to `new` so that test still covers two recommendation categories.

Only `recommend` is affected, and it is global-only, so the China-partition server is untouched.

### User experience

`recommend` now returns New, Similar, and Journey items only.

Result counts drop by roughly half to two-thirds, because Highly Rated is a fixed 12-item block and the largest part of the response today, but that data is being retired and will stop being refreshed.

New, Similar, and Journey are unchanged item for item, and there is no schema change: still `List[RecommendationResult]`.

## Checklist

* [x] Changes are documented

**RFC issue number**: N/A

## Acknowledgment

### DIFF
--- a/src/aws-documentation-mcp-server/CHANGELOG.md
+++ b/src/aws-documentation-mcp-server/CHANGELOG.md
@@ -12,6 +12,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add environment variable `AWS_DOCUMENTATION_PARTITION` to select AWS documentation partition.
 - Add `get_available_services` and `read_documentation` when `AWS_DOCUMENTATION_PARTITION` is set to `aws-cn`.
 
+### Removed
+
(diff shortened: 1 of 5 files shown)


artb30 (approved):

AadityaBhoota (approved):
lgtm!

zjerath (approved):

PR #4530, opened 2026-08-25
fix: stop shipping test and lint tooling as runtime dependencies

### LINKED ISSUE #4458: bug: four MCP servers ship test/lint tooling as runtime dependencies on PyPI
## Summary

Four published MCP servers declare test/lint tooling in `[project] dependencies` instead of a dev group, so `pip install` / `uvx` of those servers installs pytest, ruff, pyright, pre-commit, and their transitive closures into the user's runtime environment.

This is live on PyPI today — it is not caused by, and does not depend on, the release-time pinning work in #2856. I found it while building all 59 packages as releases to verify that PR, and am filing it separately as suggested there.

## Evidence — current PyPI metadata

Read straight from `https://pypi.org/pypi/<pkg>/json` → `info.requires_dist`, filtering out `extra == ...` lines (those are opt-in and correct):

| package | published version | dev tools declared as **runtime** requirements |
|---|---|---|
| `awslabs.aws-support-mcp-server` | 0.1.22 | `pre-commit>=4.2.0`, `pytest>=8.3.5`, `pytest-asyncio>=0.26.0`, `pytest-cov>=6.1.1`, `virtualenv>=20.36.1` |
| `awslabs.well-architected-security-mcp-server` | 0.1.9 | `pytest>=7.0.0`, `pytest-asyncio>=0.26.0`, `pytest-cov>=4.1.0`, `pytest-mock>=3.12.0`, `ruff>=0.0.291`, `pyright>=1.1.0` |
| `awslabs.aws-pricing-mcp-server` | 1.0.33 | `pytest>=8.1.1`, `pytest-asyncio>=0.20.3` |
| `awslabs.amazon-sns-sqs-mcp-server` | 2.0.22 | `pytest>=8.3.5` |

Note `awslabs.well-architected-security-mcp-server` has **11** total requirements, of which **6** are dev tools — over half its declared dependency list is test tooling.

## Impact

Counting the full installed closure (transitive deps the tools drag in — `iniconfig`, `pluggy`, `nodeenv`, `identify`, `cfgv`, `distlib`, `filelock`, `platformdirs`, `coverage`):

| package | dev-related packages installed at runtime |
|---|---|
| `aws-support-mcp-server` | 14 |
| `well-architected-security-mcp-server` | 10 |
| `aws-pricing-mcp-server` | 4 |
| `amazon-sns-sqs-mcp-server` | 3 |

Concretely this means:

- **Larger install and slower cold start.** `uvx` resolves and downloads a test framework and a linter before the server can answer its first request.
- **Unnecessary version constraints on consumers.** A user whose project pins a different `pytest` can hit a resolution conflict with an MCP server that has no functional reason to constrain it. `ruff>=0.0.291` and `pyright>=1.1.0` are especially wide.
- **Larger dependency surface** for scanning, SBOM, and dependency-review purposes than these servers actually need.

## Why this is safe to fix

None of the four packages import the tooling from their runtime source. Checked every `.py` under `src/<pkg>/awslabs/` (8–14 files each) for `import pytest` / `from pytest ...` and equivalents for ruff, pyright, pre-commit, virtualenv, coverage — **zero runtime imports**.

Three of the four already declare the same tools in a dev group as well, so the `[project] dependencies` copy is purely redundant:

| package | already in a dev group? |
|---|---|
| `amazon-sns-sqs-mcp-server` | yes — `pytest` also in `[dependency-groups] dev` |
| `aws-pricing-mcp-server` | yes — `pytest` also in `[dependency-groups] dev` |
| `aws-support-mcp-server` | yes — `pre-commit`, `pytest`, `pytest-asyncio`, `pytest-cov` also in `[dependency-groups] dev` and `[project.optional-dependencies] dev` |
| `well-architected-security-mcp-server` | **no** — has neither `dependency-groups` nor `optional-dependencies` |

## Proposed fix

For the three with an existing dev group: delete the duplicated entries from `[project] dependencies`. No other change needed.

For `well-architected-security-mcp-server`: move the six tools into a new dev group, matching the convention the rest of the repo uses:

```toml
[project]
dependencies = [
    "mcp[cli]>=1.23.0,<2.0.0",
    "boto3>=1.28.0",
    "loguru>=0.7.0",
    "pydantic>=2.0.0",
    "fastapi>=0.100.0",
]

[dependency-groups]
dev = [
    "pytest>=7.0.0",
    "pytest-asyncio>=0.26.0",
    "pytest-cov>=4.1.0",
    "pytest-mock>=3.12.0",
    "ruff>=0.0.291",
    "pyright>=1.1.0",
]
```

`aws-support-mcp-server` also declares `virtualenv>=20.36.1`, which is a `pre-commit` transitive dep rather than something used directly — worth dropping rather than relocating, though harmless in a dev group.

Since `uv sync` installs the dev group by default, CI keeps working unchanged; only the published metadata changes.

## Verification

Per package: `uv sync --no-dev` should succeed and the test tooling should be absent from the runtime environment, while `uv sync` (dev included) still runs the full test suite. On the built wheel, no unconditional `Requires-Dist` line should name a test or lint tool:

```bash
cd src/<pkg> && uv build
python3 - <<'PY'
import glob, zipfile
with zipfile.ZipFile(glob.glob('dist/*.whl')[0]) as z:
    meta = next(n for n in z.namelist() if n.endswith('.dist-info/METADATA'))
    reqs = [l.split(':',1)[1].strip() for l in z.read(meta).decode().splitlines()
            if l.startswith('Requires-Dist:')]
runtime = [r for r in reqs if 'extra ==' not in r]
bad = [r for r in runtime if r.split(';')[0].strip().lower().startswith(
    ('pytest','ruff','pyright','pre-commit','virtualenv','coverage'))]
print('dev tools as runtime requirements:', bad or 'none')
PY
```

## Related

- #2856 — release-time dependency pinning. Independent of this issue, but it propagates whatever `[project] dependencies` declares, so fixing this issue also removes these tools from the pinned output.
- A companion issue proposes adding a CI check that would catch this class of mistake automatically (linked below once filed).

AmirF194:
## Root cause

Four published servers list test and lint tooling directly under `[project.dependencies]`,
so every `pip install`/`uvx` install pulls them in as runtime requirements:

- `aws-support-mcp-server`: `pre-commit`, `pytest`, `pytest-asyncio`, `pytest-cov`, `virtualenv`
- `well-architected-security-mcp-server`: `pytest`, `pytest-asyncio`, `pytest-cov`,
  `pytest-mock`, `ruff`, `pyright` (this one has no dev dependency group at all)
- `aws-pricing-mcp-server`: `pytest`, `pytest-asyncio`
- `amazon-sns-sqs-mcp-server`: `pytest`

None of these servers import any of that tooling at runtime. Bandit's own `bandit -r`
gate and a static import scan across each package's `awslabs/` source confirm no runtime
module reaches `pytest`, `ruff`, `pyright`, `pre-commit`, or `virtualenv`.

## Fix

Moved every dev-only entry into each package's `[dependency-groups] dev` group, matching
the pattern `aws-pricing-mcp-server` and `amazon-sns-sqs-mcp-server` already use.
`well-architected-security-mcp-server` had no dev group, so one was added. Where a tool
already existed in the group with a looser bound, the runtime-list duplicate was simply
dropped; where it was missing (`pytest-asyncio`/`pytest-cov`/`virtualenv` on
`aws-support-mcp-server`, `pytest-asyncio` on `aws-pricing-mcp-server`), it was added at
the version the runtime list previously pinned. Regenerated `uv.lock` for the three
packages whose lock could be resolved cleanly.

## Verification

- Built each of the four packages as a wheel before and after this change in a clean
  container and read `Requires-Dist` from the resulting `METADATA`: before, every tool
  listed above shows up as an unconditional runtime requirement; after, none do.
- `uv sync --frozen --all-extras --dev`, the package's own test suite, `pyright`, and
  `ruff format`/`ruff check` all pass for `aws-support-mcp-server`, `aws-pricing-mcp-server`,
  and `amazon-sns-sqs-mcp-server`.
- `well-architected-security-mcp-server`: the same test/ruff/pyright suite passes under a
  plain `pip install -e .` environment, but I left its `uv.lock` untouched. On unmodified
  `main`, that file already fails `uv sync --frozen` for a reason unrelated to this change:
  it locks `pyjwt==2.13.0` against `typing-extensions==4.14.0`, and no package entry for
  that version exists anywhere in the same lockfile. It traces to the `chore(deps): update
  uv` commit `a7b3263`. Flagging it here rather than folding a lock repair into this PR.
- Not verified: an actual `uvx`/PyPI install of any of the four packages end to end; the
  wheel metadata read above is the direct evidence for the runtime-dependency claim.

Fixes #4458

### DIFF
--- a/src/amazon-sns-sqs-mcp-server/pyproject.toml
+++ b/src/amazon-sns-sqs-mcp-server/pyproject.toml
@@ -27,7 +27,6 @@ dependencies = [
     "readabilipy>=0.2.0",
     "requests>=2.32.3",
     "boto3>=1.38.12",
-    "pytest>=8.3.5",
 ]
 
 [project.scripts]
--- a/src/amazon-sns-sqs-mcp-server/uv.lock
+++ b/src/amazon-sns-sqs-mcp-server/uv.lock
@@ -4,8 +4,8 @@ requires-python = ">=3.10"
 resolution-markers = [
     "python_full_version >= '3.14' and sys_platform == 'win32'",
     "python_full_version >= '3.14' and sys_platform != 'win32'",
-    "python_full_version < '3.14' and sys_platform == 'win32'",
     "python_full_version >= '3.12' and python_full_version < '3.14' and sys_platform == 'emscripten'",
+    "python_full_version < '3.14' and sys_platform == 'win32'",
     "(python_full_version < '3.12' and sys_platform == 'emscripten') or (python_full_version < '3.14' and sys_platform != 'emscripten' and sys_platform != 'win32')",
 ]
 
@@ -23,8 +23,8 @@ name = "anyio"
 version = "4.9.0"
 source = { registry = "https://pypi.org/simple" }
 resolution-markers = [
-    "python_full_version < '3.14' and sys_platform == 'win32'",
     "python_full_version >= '3.12' and python_full_version < '3.14' and sys_platform == 'emscripten'",
+    "python_full_version < '3.14' and sys_platform == 'win32'",
     "(python_full_version < '3.12' and sys_platform == 'emscripten') or (python_full_version < '3.14' and sys_platform != 'emscripten' and sys_platform != 'win32')",
 ]
 dependencies = [
@@ -74,7 +74,7 @@ wheels = [
 
 [[package]]
 name = "awslabs-amazon-sns-sqs-mcp-server"
-version = "2.0.23"
+version = "2.0.9223372036854775807"
 source = { editable = "." }
 dependencies = [
     { name = "boto3" },
@@ -82,7 +82,6 @@ dependencies = [
     { name = "mcp", extra = ["cli"] },
     { name = "protego" },
     { name = "pydantic" },
-    { name = "pytest" },
     { name = "readabilipy" },
     { name = "requests" },
 ]
@@ -106,7 +105,6 @@ requires-dist = [
     { name = "mcp", extras = ["cli"], specifier = ">=2.0.0,<3.0.0" },
     { name = "protego", specifier = ">=0.3.1" },
     { name = "pydantic", specifier = ">=2.0.0" },
-    { name = "pytest", specifier = ">=8.3.5" },
     { name = "readabilipy", specifier = ">=0.2.0" },
     { name = "requests", specifier = ">=2.32.3" },
 ]
--- a/src/aws-pricing-mcp-server/pyproject.toml
+++ b/src/aws-pricing-mcp-server/pyproject.toml
@@ -8,8 +8,6 @@ dependencies = [
     "mcp[cli]>=2.0.0,<3.0.0",
     "pydantic>=2.10.5",
     "boto3>=1.39.4",
-    "pytest>=8.1.1",
-    "pytest-asyncio>=0.20.3",
     "typing-extensions>=4.13.2",
     "loguru>=0.7.3",
 ]
@@ -48,6 +46,7 @@ dev = [
     "ruff>=0.9.7",
     "pyright>=1.1.398",
     "pytest>=8.0.0",
+    "pytest-asyncio>=0.20.3",
     "pytest-cov>=4.1.0",
 ]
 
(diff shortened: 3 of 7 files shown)


AmirF194:
No maintainer has looked at this yet, eight days in. It's four independent runtime dev-dependency fixes across the servers named in the description, each one isolated to its own package.

PR #4537, opened 2026-08-25
feat(billing-cost-management): add get-billing-preferences tool

johnwangwyx:
### Changes

New `get-billing-preferences` tool, wraps the read-only `billing:GetBillingPreferences` API.
Reports which member accounts participate in the Reserved Instance / Savings Plans discount pool
and in credit sharing, whether new accounts join automatically, whether sharing is open, whether
billing alerts are enabled, and the per-billing-period history of those settings.

This is the only source for an account's discount-sharing state. It cannot be inferred from RI or
Savings Plans coverage, where "no eligible usage", "opted out" and "pool exhausted" are
indistinguishable. This will bring correctness and completness into RI/SP related workflows

API Doc: https://docs.aws.amazon.com/aws-cost-management/latest/APIReference/API_billing_GetBillingPreferences.html

### User experience

User can now Retrieves billing preferences for the different billing capability: which accounts can share Reserved Instances or credits, whether billing alerts are enabled, the historical record of sharing changes, and per-credit options. etc

### Tests
1. Unit tests added (100% coverage with test input output response pulled from live api artifacts and offical docs)
2. Build the server and hosted it to run different real life questions.
3. Compared tool response from AWS CLI baseline response to make sure they are consistent.

**Live verification on testing accounts** 
Every parameter exercised and compared to the baseline `aws billing get-billing-preferences` CLI call to make sure data consistency.

| Parameter | Cases | Result |
|---|---|---|
| `features` | all 8 enum values | 6 return rows, **byte-identical to CLI**; 2 error identically |
| `features` | bare string vs 1-item list | identical (13 rows each) |
| `features` | omitted / invalid / two values | error, same message as CLI |
| `filters` | 5 cases across 3 features | agree with CLI, error text identical |
| `filters` | malformed JSON | refused locally with the offending value |
| `max_results` | 50, omitted, 1, 0, 51 | agree with CLI incl. both range errors |
| `max_pages` | 1, 5 (default), None | truncation reported via `has_more` / `complete_dataset` |
| `next_token` | resume from a CLI-issued token | resumed row **identical to CLI** |

Per-feature row counts works and matched exactly on the testing account: `RI_SHARING` 13, `RI_SHARING_HISTORY` 26, `CREDIT_SHARING` 12, `CREDIT_SHARING_HISTORY` 12, `CREDIT_LEVEL_SHARING` 0, `BILLING_ALERTS` 1.

## Checklist

* [x] Changes are documented

1354 unit tests pass; `ruff check`, `ruff format --check` and `pyright` clean. Behaviour verified
against the live AWS APIs as shown above.

Purely additive + fixing some issues on existing tools.

**RFC issue number**: N/A

Checklist:

## Acknowledgment


src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/billing_preferences_operations.py
```diff
@@ -0,0 +1,116 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""AWS billing preferences operations for the AWS Billing and Cost Management MCP server.
+
+This module contains the read-only operation handler for the
+``billing-preferences`` tool (AWS Billing preferences: which member accounts
+share the Reserved Instance / Savings Plans discount pool or credits, whether
```
somsubhro, line 19: Savings Plans is not covered. Remove.

somsubhro (commented):
One minor nit but important one as agents can get confused with the Savings Plans annotation. Please see the comment and submit a quick revision

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/billing_preferences_operations.py
johnwangwyx: I think the same preference actually enables Savings Plans discount.

BillingFeature has eight values only — `RI_SHARING, RI_SHARING_HISTORY, CREDIT_SHARING, CREDIT_SHARING_HISTORY, CREDIT_LEVEL_SHARING, BILLING_ALERTS, CREDIT_PREFERENCE_OPTIONS, BENEFIT_PREFERENCE_OPTIONS`.  ([API Doc](https://docs.aws.amazon.com/aws-cost-management/latest/APIReference/API_billing_GetBillingPreferences.html#API_billing_GetBillingPreferences_RequestSyntax)) There is no separate Savings Plans sharing feature. `RI_SHARING`, despite the name, is legacy shorthand for the combined "RI and Savings Plans discount sharing" preference.

You can see on AWS console, the two features are grouped together as a whole. (and inspect network shows me the Console fires exactly the getPreference call with feature `RI_SHARING`)

Screenshot 2026-08-26 at 1 09 35 PM

To cross verify, when I query AWS API with just the legacy enum `RI_SHARING`, it actually gave me the same console result (i.e. both RI and Savings Plans discount sharing)

Overall, the discription is added to actually let the agent know such SP discount preference can be found with this MCP tool, despite legacy API Enum which may confuse the agent.

Some other references in the  [Billing user guide](https://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/ri-turn-off.html)

> The management account of an organization can control Reserved Instance discount **and Savings Plans discount sharing** for any accounts in that organization.

> To share the Savings Plans discount, the Savings Plans owner account must be active in the RI and Savings Plans discount sharing preferences.

johnwangwyx:
> One minor nit but important one as agents can get confused with the Savings Plans annotation. Please see the comment and submit a quick revision

Thanks for reviewing! Replied there.

somsubhro (dismissed):

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/billing_preferences_tools.py
```diff
@@ -0,0 +1,114 @@
...
+
+Example 1: {"features": "RI_SHARING"}
+Example 2 (history): {"features": "RI_SHARING_HISTORY"}
+Example 3 (credit sharing): {"features": "CREDIT_SHARING"}
+Example 4 (one credit): {"features": "CREDIT_PREFERENCE_OPTIONS",
+  "filters": "[{\\"name\\": \\"PREFERENCE_KEY\\", \\"value\\": [\\"credit/4242\\"]}]"}""",
+)
+async def get_billing_preferences(
+    ctx: Context,
+    features: Optional[Union[str, List[str]]] = None,
```
wangyuhere: `features` is required parameter then it is not Optional

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/billing_preferences_tools.py
```diff
@@ -0,0 +1,114 @@
...
+
+billing_preferences_server = FastMCP(
+    name='billing-preferences-tools',
+    instructions='Tools for working with AWS billing preferences via the AWS Billing API',
+)
+
+
+@billing_preferences_server.tool(
+    name='get-billing-preferences',
+    description="""Retrieves the AWS billing preferences for one feature.
```
wangyuhere: Probably can re-use the text from public doc:

```
Retrieves billing preferences for the specified feature. Each feature controls a distinct billing capability: which accounts can share Reserved Instances or credits, whether billing alerts are enabled, the historical record of sharing changes, and per-credit options.

```

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/billing_preferences_tools.py
```diff
@@ -0,0 +1,114 @@
...
+on, and the per-billing-period history of those settings.
+
+`features` TAKES EXACTLY ONE VALUE. Current state and history are separate features, so "what is shared
+now, and when did it change" is two calls: RI_SHARING then RI_SHARING_HISTORY (CREDIT_SHARING and
+CREDIT_SHARING_HISTORY for credits). BILLING_ALERTS, CREDIT_LEVEL_SHARING and CREDIT_PREFERENCE_OPTIONS
+are also valid.
+
+Parameters:
+- features: the single feature to retrieve (required).
+- filters: JSON, PREFERENCE_KEY only, and accepted for CREDIT_PREFERENCE_OPTIONS alone, where values must
```
wangyuhere, line 49: Are you sure filters only accepted for `CREDIT_PREFERENCE_OPTIONS` ?

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/billing_preferences_operations.py
```diff
@@ -0,0 +1,116 @@
...
+            token_param='nextToken',
+            token_key='nextToken',
+            max_pages=max_pages,
+        )
+
+        await ctx.info(f'Successfully retrieved {len(preferences)} billing preferences')
+
+        # An organization-wide feature can return a row per member account, so
+        # the shared size threshold decides whether to offload to session SQL.
+        converted = await convert_response_if_needed(
```
wangyuhere, line 104: do we also need to update the sql schema part for the API?

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/billing_preferences_tools.py
wangyuhere: What will happen if agent hallucinated and pass the filter to other option? Will it be dropped silently or return validation error?

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/billing_preferences_tools.py
johnwangwyx: Good catch! fixed.

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/billing_preferences_tools.py
johnwangwyx: Adopted, the description now opens with the API reference wording.

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/billing_preferences_tools.py
johnwangwyx: I have confirmed with live API call all feature values other than `CREDIT_PREFERENCE_OPTIONS` will result in validation exceptions.

To prevent hallucination, this is why we have the tool description to bridge the gap from Agent to AWS API validations (which may not surface on API doc nor in base model's knowledge). 

Additionally, if the agent still hallucinates (never saw it during my testing). The infra of this mcp server will faithfully provide the validation exception reason from AWS API so agent can succeed in next turn:

Tested on live agent by forcing it to call with the wrong command as well as in `mcp-inspector`

Screenshot 2026-08-27 at 1 51 01 PM

src/billing-cost-management-mcp-server/awslabs/billing_cost_management_mcp_server/tools/billing_preferences_operations.py
johnwangwyx: You are right. Despite there is no data loss, it would made the format harder to parse. I have updated the schema and tested with a big payload and inspected the rows.

wangyuhere (approved):

somsubhro (approved):

PR #4547, opened 2026-08-27
fix(mysql-mcp-server): enforce TLS on the Secrets Manager credential path

manaswini1920:
### Fixes
Fixes a cleartext-credential exposure on the mysqlwire (Secrets Manager) connection path (CWE-319): only IAM-auth connections built an SSL context, so a mysqlwire connection backed by a managed secret transmitted the database password with ssl=None — interceptable by a MITM or an attacker-controlled endpoint.

### Changes
Enforce verified TLS whenever real credentials are on the wire — both IAM auth tokens and Secrets Manager passwords — not just IAM auth.

Build the SSL context for is_iam_auth or secret_arn, collapsing the two identical trust-decision branches into one (the auth flavour only affects the log message).
Thread the operator's --ca_bundle override into the MYSQL_WIRE_PROTOCOL connection construction so it is honoured on the Secrets Manager path (previously it was silently ignored there).
Trust order is unchanged: operator --ca_bundle → bundled Amazon RDS CA bundle → system trust store (warned).
Document the behaviour change (breaking) and the CERTIFICATE_VERIFY_FAILED escape hatch in the README and CHANGELOG.

Tests: the Secrets Manager path now configures TLS; add a test asserting the --ca_bundle override reaches that path.
Split out of #4515 (endpoint validation) per review so it can be reviewed and rolled back independently.

### User experience
Before: A mysqlwire connection using a Secrets Manager secret sent the database password over an unencrypted connection (ssl=None), and an operator-supplied --ca_bundle was ignored on that path.

After: The Secrets Manager path is upgraded to verified TLS against the bundled Amazon RDS CA bundle, and --ca_bundle <path> is honoured. Operators whose server certificate does not chain to the bundled bundle supply --ca_bundle; otherwise the connection fails fast with CERTIFICATE_VERIFY_FAILED rather than sending credentials in cleartext.

### Checklist

[x] I have reviewed the [contributing guidelines](https://github.com/awslabs/mcp/blob/main/CONTRIBUTING.md)
[x] Changes have been tested
[x] Changes are documented
Is this a breaking change? Y

RFC issue number:

### Checklist:

[x] Migration process documented
[ ] Implement warnings (if it can live side by side)
Migration: installations that relied on plaintext or server-certificate-managed TLS on the mysqlwire path must pass --ca_bundle <path> (or reinstall to restore the bundled Amazon RDS CA bundle). Documented in the README ("Prerequisites by Connection Method") and CHANGELOG. TLS enforcement cannot meaningfully live side by side with plaintext, so it is not gated behind a warning-only phase; the fallback-to-system-trust-store case is logged as a warning.

### Acknowledgment

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -615,7 +615,7 @@
         "filename": "src/mysql-mcp-server/tests/test_asyncmy_connector.py",
         "hashed_secret": "c68ff3acf9f357b3963b1a7720162b573536689c",
         "is_verified": false,
-        "line_number": 186,
+        "line_number": 188,
         "is_secret": false
       },
       {
@@ -623,7 +623,7 @@
         "filename": "src/mysql-mcp-server/tests/test_asyncmy_connector.py",
         "hashed_secret": "fef341f85d87439e7d91a2d465b9871ef66b5e98",
         "is_verified": false,
-        "line_number": 687,
+        "line_number": 689,
         "is_secret": false
       }
     ],
@@ -633,7 +633,7 @@
         "filename": "src/mysql-mcp-server/tests/test_ca_bundle.py",
         "hashed_secret": "43b5a7ccc96b402d0fc814b5e03f8e23c2d9d1d8",
         "is_verified": false,
-        "line_number": 245,
+        "line_number": 246,
         "is_secret": false
       }
     ],
@@ -1266,5 +1266,5 @@
       }
     ]
   },
-  "generated_at": "2026-08-27T03:39:50Z"
+  "generated_at": "2026-08-28T17:21:25Z"
 }
--- a/src/mysql-mcp-server/CHANGELOG.md
+++ b/src/mysql-mcp-server/CHANGELOG.md
@@ -43,6 +43,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed
+
+- **BREAKING (CWE-319):** TLS is now enforced on the Secrets Manager credential
+  path (`mysqlwire` with a managed secret), not just IAM auth. Previously these
+  connections were plaintext-capable; the server now upgrades them to verified
+  TLS against the bundled Amazon RDS CA bundle. Installations that relied on
+  plaintext or a server-certificate-managed TLS setup must supply `--ca_bundle
+  <path>` (or reinstall to restore the bundled CA bundle), otherwise the
+  connection fails with `CERTIFICATE_VERIFY_FAILED`. The `--ca_bundle` override
+  is now correctly honoured on the `mysqlwire` path as well.
+
 ### Added
 
 - `database_type` parameter on `connect_to_database` accepts `aurora-mysql`,
--- a/src/mysql-mcp-server/README.md
+++ b/src/mysql-mcp-server/README.md
@@ -127,6 +127,17 @@ directly via `mysqlwire` with the endpoint, port, and credentials.
 - VPC security group must allow inbound connections from your MCP server to the database
 - For `mysqlwire_iam`: IAM authentication must be enabled on the Aurora MySQL cluster
 - The AWS identity needs `rds:DescribeDBClusters` and `rds:DescribeDBInstances`. A caller-supplied `db_endpoint` is validated against the cluster's AWS-resolved endpoints before connecting; resolving those endpoints requires both permissions. Without them, endpoint validation cannot enumerate the cluster's real endpoints and the connection is refused.
+- **TLS is required whenever real credentials are on the wire** — both IAM auth
+  (`mysqlwire_iam`) and Secrets Manager passwords (`mysqlwire` with a managed
+  secret). The server verifies the server certificate against the bundled Amazon
+  RDS global CA bundle. If your server certificate does not chain to that bundle,
+  supply your own trust store with `--ca_bundle <path>`; otherwise the connection
+  fails with `CERTIFICATE_VERIFY_FAILED`.
+
+  > **Breaking change:** earlier versions left `mysqlwire` (non-IAM) connections
+  > plaintext-capable. They are now upgraded to verified TLS. Installations that
+  > relied on plaintext or a server-certificate-managed TLS setup must pass
+  > `--ca_bundle <path>` (or reinstall to restore the bundled CA bundle).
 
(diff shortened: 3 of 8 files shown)


happycupofjava (dismissed):

kennthhz (changes requested):
## Review — enforce TLS on the Secrets Manager credential path

The core fix is correct and minimal: gating the SSL context on `is_iam_auth or secret_arn` closes the CWE-319 cleartext exposure on the `mysqlwire` + Secrets Manager path, and threading `--ca_bundle` into that path via `internal_connect_to_database` is a genuine fix (the IAM branch already passed it). I verified there are only two `AsyncmyPoolConnection` construction sites and both are handled, and the full test suite passes locally.

Requesting changes on one item (#1) — a test-coverage gap where the regression guard doesn't actually guard the code it names.

### 🔴 1. The `server.py` wiring is not actually covered (blocker)

`test_ca_bundle_override_reaches_secret_arn_path` documents itself as:

> *"Regression guard: internal_connect_to_database must thread ca_bundle_path into the MYSQL_WIRE_PROTOCOL construction…"*

…but it constructs `AsyncmyPoolConnection(...)` **directly** with `ca_bundle_path` set — it never calls `internal_connect_to_database`. Since `cafile = self.ca_bundle_path or _bundled_ca_file()` is shared with the pre-existing IAM path, this test would **still pass if the 3-line `server.py` wiring change were reverted**. So the exact regression it claims to guard is unprotected.

`test_mysql_wire_protocol_builds_secret_pool` does drive `internal_connect_to_database` for this path, but only asserts `is_iam_auth` and `secret_arn` — not `ca_bundle_path`.

**Requested change:** add an assertion that the override reaches the constructor through the real entry point — e.g. in `test_mysql_wire_protocol_builds_secret_pool`, `assert kwargs['ca_bundle_path'] == <expected>` (wiring a `ca_bundle` value through the call / module global). That makes a revert of the `server.py` change fail a test.

### 🟡 Non-blocking

**2. Negative-test docstring describes a non-existent path.** `test_no_ssl_when_no_secret_arn_and_no_iam` justifies itself with "a self-hosted MySQL connection with an inline user/password." `AsyncmyPoolConnection` has no inline-password path — with `secret_arn=''` and `is_iam_auth=False` the non-test branch calls `_get_credentials_from_secret('')` and fails; the test only passes via `is_test=True`. The false-branch coverage is fine, but the motivating example is misleading — suggest rewording.

**3. Stale "IAM-auth only" wording now that SSL covers the secret path.** Since this PR set out to document the behavior change, worth updating in the same commit:
- `AsyncmyPoolConnection.__init__` docstring for `ca_bundle_path` ("…for IAM-auth SSL verification")
- `--ca_bundle` argparse help in `server.py` ("for IAM-auth SSL verification")
- module comment on `_RDS_CA_BUNDLE_PATH` ("so that IAM authenticated connections…")
- `test_ca_bundle.py` module/class docstrings ("guard the IAM-auth trust resolution path")

### ❓ Questions (not blockers)

**4. `secret_arn` can be `None`.** In `internal_connect_to_database`, `MasterUserSecret{}.get('SecretArn')` yields `None` when there's no managed secret, and that flows into the mysqlwire branch. The SSL gate treats `None` as falsy → no SSL → the `else` branch still calls `_get_credentials_from_secret(None)` and fails at Secrets Manager. Pre-existing and no credential leak (no connection is made), but flagging since the fix's safety argument is "SSL whenever a credential is on the wire."

**5. Self-hosted `mysqlwire` breaking-change surface.** The `--db_type` help says self-hosted MySQL/MariaDB is supported, but every non-IAM mysqlwire route here goes through RDS `Describe*` calls, so I couldn't find a real plaintext self-hosted path this change would break. If such a path exists via another entry point, the migration note should cover it explicitly — could you confirm?

manaswini1920:
For 1 : The server.py wiring is not actually covered (blocker)

Fixed. Rather than the direct-construction test (which, as you noted, would pass even with the wiring reverted), I added the assertion to `test_mysql_wire_protocol_builds_secret_pool`, which drives the real `internal_connect_to_database` path: it patches the module-level `ca_bundle_path` and asserts `kwargs['ca_bundle_path']` reaches the constructor. Reverting the 3-line server.py wiring now fails this test. I also renamed the  direct unit test to `test_secret_path_uses_ca_bundle_override_when_set` and noted it only covers the connection class's SSL layer, with the server wiring guarded in test_server.py.

For 2: Negative-test docstring describes a non-existent path.
Reworded — dropped the misleading inline-password example. The docstring now just states it covers the false branch of the credentials-on-the-wire gate (no secret_arn and not IAM auth → ssl stays None).

For 3.  Stale "IAM-auth only" wording now that SSL covers the secret path
Broadened all four: the _RDS_CA_BUNDLE_PATH comment, the `ca_bundle_path __init__ docstring`, the server.py module comment + --ca_bundle argparse help, and the test_ca_bundle.py module docstring now say "credentialed  connections (IAM auth or Secrets Manager)" instead of IAM-only.

For 4 : secret_arn can be None.
 Good observation, and agreed it's pre-existing and safe. With no managed secret, secret_arn is None: the SSL gate is skipped, but the else branch then calls _get_credentials_from_secret(None), which raises before create_pool is reached — so no connection is made and no credential goes on the wire. The "TLS whenever a credential is on the wire" invariant holds because there's neither a credential nor a wire in that case. If you'd prefer an explicit earlier ValueError for the "no secret and no IAM" combination, happy to add it as a separate cleanup rather than widen this PR.

For 5 (self-hosted mysqlwire breaking surface):

Confirmed — I couldn't find a real plaintext self-hosted path this breaks either. Every non-IAM mysqlwire route through internal_connect_to_database resolves via RDS Describe* (cluster properties or the standalone  instance lookup) and only builds a pool once it has an RDS-managed secret_arn; a genuinely self-hosted (non-RDS) endpoint never reaches a plaintext-capable pool here — it fails the RDS lookup first. So the breaking surface is RDS-backed mysqlwire connections using a managed secret, which is what the migration note covers. The --db_type help that mentions self-hosted MySQL/MariaDB is pre-existing and inaccurate for this path;  happy to correct it in a separate doc-only change if you'd like.

kennthhz (approved):

happycupofjava (approved):

arnewouters (approved):

PR #4552, opened 2026-08-28
feat(redshift-mcp-server): paginate and shape discovery responses

saeedma8:
### Changes

**1. Complete result-set retrieval (bug fix).** `GetStatementResult` returns one page at a time, but
the server read only the first page and discarded the token, so a result set larger than one page was
silently truncated. Retrieval now follows every page, stops if the API repeats a page token, and
reconciles the rows retrieved against the total the API reports. Because
`_execute_protected_statement` is the single funnel to the Data API, `execute_query` and
`review_cluster` inherit this with no change of their own.

**2. A cap on result rows.** `REDSHIFT_MAX_RESULT_ROWS` caps the rows any one statement may return,
shipping as `1000`. Only a whole number above zero is accepted: anything else is logged as unusable
and the shipped default applies instead, so the cap cannot be switched off and a misconfiguration
lowers it to a safe value rather than removing it. It is read once at startup, so changing it takes
a server restart. Enforcement lives in the retrieval layer, so it covers `execute_query` as well as
the four discovery tools, and it is stated in the server instructions so an agent can plan for it.

A statement over the cap fails, and the error names what to do about it:

```
Statement <id> returned 1500 rows, more than the 1000-row maximum this server returns.
Narrow the request: for a query add a LIMIT clause, a more selective WHERE predicate, or
an aggregation; for a listing, choose a smaller schema or table. An operator can raise
the maximum with REDSHIFT_MAX_RESULT_ROWS.
```

Retrieval stops as soon as the cap is known to be exceeded, so an oversized statement is refused
without its whole result set being read into memory. When the API reports the row total on the first
page, that happens after one page.

**3. Discovery verbosity levels.** Each of the four discovery tools accepts `verbosity_level`:

- `low` returns what identifies an object plus its own type: the database, schema, table and column
  names that apply, with `database_type`, `schema_type`, `table_type` or `data_type` respectively.
  Enough to navigate to an object and write a well-typed predicate against it, without ownership,
  permissions or column sizing.
- `standard` is the default and returns the fields each tool returned before, so an existing caller
  sees no change.

| Tool | `low` fields |
|---|---|
| `list_databases` | `database_name`, `database_type` |
| `list_schemas` | `database_name`, `schema_name`, `schema_type` |
| `list_tables` | `database_name`, `schema_name`, `table_name`, `table_type` |
| `list_columns` | `database_name`, `schema_name`, `table_name`, `column_name`, `data_type` |

`low` is always a subset of `standard`. Each field carries the level it first appears at as
`Annotated` metadata on the field, so the two cannot drift apart; an unmarked field counts as
`standard`. The levels themselves are constants in `consts.py`, ordered least to most detail, with a
single `Literal` alias the four tools annotate with, so the values are written once.

### User experience

A discovery tool returns a list, as before. At the default `standard` level each item carries the
same fields, except that a field with no value is now absent rather than `null`:

```jsonc
[ { "database_name": "dev", "database_owner": 100, "database_type": "local", ... } ]
```

`low` returns the name and the type, which is useful while navigating:

```jsonc
[ { "database_name": "dev", "database_type": "local" },
  { "database_name": "sample_data_dev", "database_type": "shared" } ]
```

Three behaviours change. A result set spanning several Data API pages now comes back whole, where it
was previously cut off at the first page without any indication. A result set larger than the cap now
fails, where a caller would previously have received part of the data believing it complete. And a
field with no value is omitted rather than sent as `null`.

Whether the cap is reachable at all differs by object type, because of the [Amazon Redshift quotas](https://docs.aws.amazon.com/redshift/latest/mgmt/amazon-redshift-limits.html) on each. This is documented in the server instructions and in the README:

| Tool | Redshift quota | At the default cap of 1000 |
|---|---|---|
| `list_databases` | 60 per provisioned cluster, 100 per serverless namespace, excluding datashare databases | Never reached |
| `list_schemas` | 9,900 per database | Reachable on a large database |
| `list_tables` | up to 200,000 per cluster on the larger node types | Reachable on a large schema |
| `list_columns` | 1,600 per table | Reachable on a very wide table |

## Testing

266 unit tests passing, 100% line coverage of `awslabs`, with `ruff format`, `ruff check` and
`pyright` clean.

End-to-end against a provisioned cluster and a serverless workgroup. Nothing was written to either;
only `SHOW`, `SELECT` and `SET` ran.

<details>
<summary>Verbosity and row-cap scenarios</summary>

```
% kiro-cli chat \
    --model claude-opus-5 \
    --no-interactive \
    --trust-tools "@{awslabsredshift-mcp-server}/*,execute_bash,fs_read" \
'Use Redshift MCP Server to run end-to-end (integration) tests covering the scenarios for the changes introduced in the current branch only. Get the test scenario ideas from the unit tests under the project directory. Provide a short testing summary, one line per scenario.'

...

All scenarios executed end-to-end against live cluster `redshift-cluster-1` (us-east-1). Testing summary, one line per scenario:

**Discovery verbosity levels** (list_databases/schemas/tables/columns)
1. ✅ `list_databases` with verbosity omitted → returned full/standard fields (owner, acl, isolation level) — default is `standard`.
2. ✅ `list_databases verbosity_level=low` → returned only `database_name` + `database_type`.
3. ✅ `list_schemas` low → only `{database_name, schema_name, schema_type}`; standard added `schema_owner`/`schema_acl` (low ⊂ standard).
4. ✅ `list_tables` low → only `{database_name, schema_name, table_name, table_type}`; standard added `table_acl` (low ⊂ standard).
5. ✅ `list_columns` (tickit.sales) low → only `{database_name, schema_name, table_name, column_name, data_type}`; standard added `ordinal_position`/`is_nullable`/`numeric_precision`/`numeric_scale` (low ⊂ standard).
6. ✅ NULL fields dropped (absent, not `null`) at standard — e.g. `database_acl`/`parameters` on most DBs, `remarks`/`character_maximum_length`, and `numeric_precision` on the `timestamp` column all omitted.

**Result-row cap** (`REDSHIFT_MAX_RESULT_ROWS`, default 1000) + full pagination
7. ✅ `execute_query` returning 999 rows → full result set returned (`row_count=999`), within cap.
8. ✅ `execute_query` returning exactly 1000 rows → returned (`row_count=1000`), boundary is inclusive.
9. ✅ `execute_query` returning 1001 rows → failed closed with advice: *"returned 1001 rows, more than the 1000-row maximum… Narrow the request… An operator can raise the maximum with REDSHIFT_MAX_RESULT_ROWS."*
10. ✅ Narrowing works: `COUNT(*)` over a 100,000-row `generate_series` → 1 row, well under cap.

All 10 scenarios (derived from the branch's unit tests in `test_models.py`, `test_redshift.py`, `test_server.py`) passed. Scope was limited to the branch-only commit `289ed8ab` (result-row cap + discovery verbosity); read-only queries only, using synthetic `generate_series` data. One unit-test idea I could not force end-to-end is a discovery listing exceeding the cap (would require 1000+ schemas/columns) — the cap on that path is inherited from the same retrieval layer already exercised by scenarios 8–9.



```

</details>

<details>
<summary>All seven tools, and the existing protections</summary>

```
% kiro-cli chat \
    --model claude-opus-5 \
    --no-interactive \
    --trust-tools "@{awslabsredshift-mcp-server}/*,execute_bash,fs_read" \
'Use Redshift MCP Server to run a complete set of end-to-end (integration) tests covering all its tools. Check both provisioned and serverless clusters, including the database schema exploration in both. Check the SQL read-only protection, transaction breaker protection, failed user SQL behavior. Get the test scenario ideas from the unit tests under the project directory. Provide a short testing summary, one line per tool.'

...


All tests passed. Here is the testing summary.

## Redshift MCP Server — Integration Test Summary

**Environment:** provisioned `redshift-cluster-1` + serverless `redshift-cluster-serverless` (us-east-1), sample data in `sample_data_dev.tickit`.

**One line per tool:**
- **list_clusters** — ✅ Discovered 3 provisioned clusters + 1 serverless workgroup with correct type/status/config fields.
- **list_databases** — ✅ Listed databases on both cluster types; `low` verbosity correctly returned only `database_name`+`database_type`.
- **list_schemas** — ✅ Returned local/catalog schemas on both; `low` verbosity trimmed to name+type as designed.
- **list_tables** — ✅ Listed 7 `tickit` tables on both; `low` gave 4 identifying fields, `standard` added `table_acl`.
- **list_columns** — ✅ Returned typed column metadata on both; `character_maximum_length`/`numeric_precision` shown per type and NULL fields correctly omitted (e.g. `saletime`).
- **execute_query** — ✅ Valid read queries succeeded on both (provisioned sales aggregate = 344,912 rows; serverless venue ranking, with NULL values preserved).
- **review_cluster** — ✅ Ran on both; serverless = 8 signals/0 findings, provisioned = 11 signals/10 effort-ordered findings; provisioned-only diagnostics auto-skipped for serverless.

**One line per protection:**
- **SQL read-only (deny-list)** — ✅ `TRUNCATE` (provisioned) and `GRANT` (serverless) rejected pre-execution with `Statement type not allowed in read-only mode: <KW>`; multi-statement `SELECT 1; SELECT 2` rejected with `Only a single SQL statement is allowed`.
- **Transaction breaker** — ✅ `CREATE TABLE` passed the guard but the engine's read-only transaction blocked it on **both** provisioned and serverless (`ERROR: transaction is read-only`); nothing was created (rolled back).
- **Failed user SQL** — ✅ Unparseable `SELECT FROM WHERE` failed closed with `SQL could not be parsed`; a nonexistent table surfaced the engine error `relation "…" does not exist`.

All 7 tools and all 3 protection classes behaved as specified across both provisioned and serverless clusters, including schema exploration on each. No writes persisted.

```

</details>

## Checklist

* [x] Changes are documented

for the response shape. The discovery tools still return a
list of the same models, and at the default `standard` verbosity each item carries the same fields,
with one qualification: a field with no value is now omitted rather than sent as `null`. A consumer
reading keys off the JSON sees the same values; one that distinguishes an absent key from a null one
would see a difference.

**Operator-facing note.** `REDSHIFT_MAX_RESULT_ROWS` is read once at startup, so a change to it now
takes a server restart. There is no value that disables the cap; an operator who needs more rows
raises the number.

**Pre-existing behaviour documented, not changed.** End-to-end testing surfaced that `list_columns`
returns an empty list for a table that does not exist, where `list_schemas` raises for a database
that does not exist. `SHOW COLUMNS` returns nothing both for an absent table and for one the caller
lacks privileges to see, and does not distinguish them. That path is untouched by this change, so
rather than alter it here the `list_columns` description now says an empty list does not confirm the
table exists with no columns, and points at `list_tables` to check the name.

**RFC issue number**:

Checklist:

* [x] Migration process documented - the cap, its environment variable, and what to do when a
  statement exceeds it are stated in each tool's description, in the server instructions, and in a
  "Result size" section of the README.
* [x] Implement warnings (if it can live side by side) - not applicable as a warning. The change replaces silent truncation with an explicit error.

## Acknowledgment

### DIFF
--- a/src/redshift-mcp-server/README.md
+++ b/src/redshift-mcp-server/README.md
@@ -110,6 +110,11 @@ or docker after a successful `docker build -t awslabs/redshift-mcp-server:latest
 - `AWS_PROFILE`: AWS profile to use (optional, uses default if not specified)
 - `FASTMCP_LOG_LEVEL`: Logging level (`DEBUG`, `INFO`, `WARNING`, `ERROR`)
 - `LOG_FILE`: Path to log file (optional, logs to stdout if not specified)
+- `REDSHIFT_MAX_RESULT_ROWS`: Maximum rows any one statement may return (default: `1000`). Only a
+  whole number above zero is accepted; anything else, including `0`, is logged as unusable and the
+  default of `1000` applies instead, so the cap cannot be switched off. Read once at startup, so a
+  change takes a server restart. A statement returning more rows than this fails rather than
+  returning a short answer — see [Result size](#result-size).
 
 ## Prompt Examples
 
@@ -117,7 +122,7 @@ or docker after a successful `docker build -t awslabs/redshift-mcp-server:latest
 
 1. **Discover Clusters and Workgroups**: Call `list_clusters` to find all provisioned clusters and serverless workgroups, noting their identifiers, types, and status
 2. **Select Target Environment**: Choose a specific cluster or workgroup based on your query needs
-3. **List Databases**: Use `list_databases` to explore databases in the selected cluster/workgroup, returning database names, owners, types (local/shared), and access control info
+3. **List Databases**: Use `list_databases` to explore databases in the selected cluster/workgroup, returning database names, owners, types (such as local, shared, or auto mounted catalog), and access control info
 4. **Explore Schemas**: For a given database, call `list_schemas` to find available schemas and their owners
 5. **Inspect Tables**: Use `list_tables` to browse tables, views, and external tables within a schema, including table types and owners
 6. **Examine Columns**: Call `list_columns` to get column metadata — names, data types, nullability, default values, and constraints
@@ -330,18 +335,23 @@ list_clusters() -> list[RedshiftCluster]
 Lists all databases in a specified Redshift cluster.
 
 ```python
-list_databases(cluster_identifier: str, database_name: str = "dev") -> list[RedshiftDatabase]
+list_databases(
+    cluster_identifier: str,
+    database_name: str = "dev",
+    verbosity_level: Literal["low", "standard"] | None = "standard",
+) -> list[RedshiftDatabase]
 ```
 
 **Parameters**:
 
 - `cluster_identifier`: The cluster identifier from `list_clusters`
 - `database_name`: Database to connect to for querying (default: "dev")
+- `verbosity_level`: How much detail to return per database, `"low"` or `"standard"` (see [Response shaping](#response-shaping))
 
-**Returns**: List of database information including:
+**Returns**: A list of databases, each including:
 
 - Database name and owner
-- Database type (local/shared)
+- Database type (such as local, shared, or auto mounted catalog)
 - Access control information
 - Isolation level
 
@@ -350,15 +360,20 @@ list_databases(cluster_identifier: str, database_name: str = "dev") -> list[Reds
 Lists all schemas in a specified database.
 
 ```python
-list_schemas(cluster_identifier: str, schema_database_name: str) -> list[RedshiftSchema]
+list_schemas(
+    cluster_identifier: str,
+    schema_database_name: str,
+    verbosity_level: Literal["low", "standard"] | None = "standard",
+) -> list[RedshiftSchema]
 ```
 
 **Parameters**:
 
 - `cluster_identifier`: The cluster identifier from `list_clusters`
 - `schema_database_name`: Database name to list schemas for
+- `verbosity_level`: How much detail to return per schema, `"low"` or `"standard"` (see [Response shaping](#response-shaping))
 
-**Returns**: List of schema information including:
+**Returns**: A list of schemas, each including:
 
 - Schema name and owner
 - Schema type (local/external/shared)
@@ -370,7 +385,12 @@ list_schemas(cluster_identifier: str, schema_database_name: str) -> list[Redshif
 Lists all tables in a specified schema.
 
 ```python
-list_tables(cluster_identifier: str, table_database_name: str, table_schema_name: str) -> list[RedshiftTable]
+list_tables(
+    cluster_identifier: str,
+    table_database_name: str,
+    table_schema_name: str,
+    verbosity_level: Literal["low", "standard"] | None = "standard",
+) -> list[RedshiftTable]
 ```
 
 **Parameters**:
@@ -378,8 +398,9 @@ list_tables(cluster_identifier: str, table_database_name: str, table_schema_name
 - `cluster_identifier`: The cluster identifier from `list_clusters`
 - `table_database_name`: Database name containing the schema
 - `table_schema_name`: Schema name to list tables for
+- `verbosity_level`: How much detail to return per table, `"low"` or `"standard"` (see [Response shaping](#response-shaping))
 
-**Returns**: List of table information including:
+**Returns**: A list of tables, each including:
 
 - Table name and type (TABLE/VIEW/EXTERNAL TABLE)
 - Access permissions
@@ -394,7 +415,8 @@ list_columns(
     cluster_identifier: str,
     column_database_name: str,
     column_schema_name: str,
-    column_table_name: str
+    column_table_name: str,
+    verbosity_level: Literal["low", "standard"] | None = "standard",
 ) -> list[RedshiftColumn]
 ```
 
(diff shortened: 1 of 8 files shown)


src/redshift-mcp-server/awslabs/redshift_mcp_server/consts.py
```diff
@@ -60,3 +60,11 @@
         'CANCEL',
     }
 )
+
+# Result size cap
+
+# Environment variable an operator sets to cap the rows any one statement may return, and the
+# value used when it is unset or holds something unusable. A configured 0 means no cap.
+# A statement returning more rows than this fails and asks the caller to narrow it.
+MAX_RESULT_ROWS_ENV_VAR = 'REDSHIFT_MAX_RESULT_ROWS'
```
grayhemp: It's an overkill. Don't make the name a constant. Just use the name directly, like os.environ.get('REDSHIFT_MAX_RESULT_ROWS', MAX_RESULT_ROWS_FALLBACK).

grayhemp (changes requested):

src/redshift-mcp-server/awslabs/redshift_mcp_server/consts.py
saeedma8: Done. The name is used directly now and MAX_RESULT_ROWS_ENV_VAR is gone. The default is DEFAULT_MAX_RESULT_ROWS, renamed per your other note about placing it next to DEFAULT_LOG_LEVEL. I kept it out of the get() call so an unset variable stays silent while a bad value can warn.

src/redshift-mcp-server/awslabs/redshift_mcp_server/consts.py
```diff
@@ -20,6 +20,7 @@
 CLIENT_RETRIES = {'max_attempts': 5, 'mode': 'adaptive'}
 CLIENT_USER_AGENT_NAME = 'awslabs/mcp/redshift-mcp-server'
 DEFAULT_LOG_LEVEL = 'WARNING'
+DEFAULT_MAX_RESULT_ROWS = 1000
```
grayhemp: Let's rename is as MAX_RESULT_ROWS_DEFAULT. I know we have DEFAULT_LOG_LEVEL, but I'm fixing it in another PR.

src/redshift-mcp-server/awslabs/redshift_mcp_server/models.py
```diff
@@ -15,21 +15,68 @@
...
+
+
 class RedshiftDataModel(BaseModel):
     """Base for models built from a Redshift Data API result set.
 
     Subclasses declare their fields named to match the SHOW result columns.
     `from_redshift_response` maps result columns to those fields by name, so
     parsing is independent of column order; unknown columns are ignored.
+
+    Each field also carries the verbosity level it first appears at, as
```
grayhemp: Make concise.

grayhemp: No need restating the agent instructions here.

src/redshift-mcp-server/awslabs/redshift_mcp_server/consts.py
```diff
@@ -14,12 +14,27 @@
...
+
+# Verbosity levels for the discovery tools, least to most detail. A field marked with a
+# level is returned at that level and at every level after it.
+VERBOSITY_LEVEL_LOW = 'low'
+VERBOSITY_LEVEL_STANDARD = 'standard'
+VERBOSITY_LEVELS = (VERBOSITY_LEVEL_LOW, VERBOSITY_LEVEL_STANDARD)
+VERBOSITY_LEVEL_DEFAULT = VERBOSITY_LEVEL_STANDARD
+# The tools' parameter type. Literal takes only literals, so a new level goes both in
+# VERBOSITY_LEVELS and here; a test holds the two to the same values in the same order.
+VerbosityLevel = Literal['low', 'standard']
```
grayhemp, line 30: This is an overkill, no need defining this type. The tool params will get str, and should be resolving verbocity erroring on values not in VERBOCITY_LEVELS.

src/redshift-mcp-server/awslabs/redshift_mcp_server/server.py
```diff
@@ -149,6 +189,11 @@ def _read_only_annotations(title: str) -> ToolAnnotations:
     )
 
 
+def _requested_level(verbosity_level: str | None) -> str:
```
grayhemp, line 192: Rename as _resolve_verbocity_level() and error on wrong values instead of silently using standard verbosity.

src/redshift-mcp-server/awslabs/redshift_mcp_server/server.py
```diff
@@ -95,6 +98,43 @@
 2. Use the list_clusters tool to discover available Redshift instances.
 3. Note the cluster identifiers for use with other tools (coming in future milestones).
 
+## Verbosity
```
grayhemp, line 101: Make it concise.

src/redshift-mcp-server/awslabs/redshift_mcp_server/server.py
```diff
@@ -95,6 +98,43 @@
...
+  table and column names that apply, with `database_type`, `schema_type`, `table_type` or
+  `data_type` respectively. Enough to navigate to an object and write a well-typed predicate
+  against it.
+- `standard`, the default, returns every field the tool lists in its own description.
+
+`low` is a subset of `standard`, so asking for more detail never drops a field. Prefer `low`
+while navigating, and `standard` once you need ownership, permissions, or the sizing of a
+table's columns.
+
+## Result Size
```
grayhemp, line 117: Make it concise.

src/redshift-mcp-server/awslabs/redshift_mcp_server/server.py
```diff
@@ -225,6 +270,14 @@ async def list_databases_tool(
         'dev',
         description='The database to connect to for metadata discovery. Defaults to "dev".',
     ),
+    verbosity_level: Annotated[
+        VerbosityLevel | None,
+        Field(
+            description='How much detail to return per database. "low" returns the name '
```
grayhemp, line 276: Make it concise.

src/redshift-mcp-server/awslabs/redshift_mcp_server/server.py
```diff
@@ -391,6 +463,14 @@ async def list_tables_tool(
         ...,
         description='The schema name to list tables for. Also used to connect to. Must be a valid schema name from the list_schemas tool.',
     ),
+    verbosity_level: Annotated[
+        VerbosityLevel | None,
+        Field(
```
grayhemp, line 468: Make it concise.

src/redshift-mcp-server/awslabs/redshift_mcp_server/server.py
```diff
@@ -487,6 +572,14 @@ async def list_columns_tool(
         ...,
         description='The table name to list columns for. Must be a valid table name from the list_tables tool.',
     ),
+    verbosity_level: Annotated[
+        VerbosityLevel | None,
+        Field(
+            description='How much detail to return per column. "low" returns the names '
```
grayhemp, line 578: Make it concise.

src/redshift-mcp-server/awslabs/redshift_mcp_server/server.py
```diff
@@ -529,6 +625,12 @@ async def list_columns_tool(
     - numeric_scale: The numeric scale.
     - remarks: Remarks about the column.
 
+    An empty list does not confirm that the table exists and has no columns. SHOW COLUMNS
```
grayhemp, line 628: Make it concise.

src/redshift-mcp-server/tests/test_models.py
```diff
@@ -0,0 +1,175 @@
...
+    @pytest.mark.parametrize('model_class', DISCOVERY_MODELS)
+    def test_the_level_keeps_its_own_values(self, model_class):
+        """Blanking the levels above leaves the ones at or below it untouched."""
+        full = _full_item(model_class)
+        low = full.at_verbosity_level(VERBOSITY_LEVEL_LOW)
+
+        for name in model_class.fields_at_verbosity_level(VERBOSITY_LEVEL_LOW):
+            assert getattr(low, name) == getattr(full, name)
+
+    def test_a_null_field_is_dropped_at_its_own_level(self):
```
grayhemp, line 163: Good!

src/redshift-mcp-server/tests/test_server.py
```diff
@@ -68,6 +71,149 @@ async def test_tool_annotations():
         assert annotations.open_world_hint is True
 
 
+# Each discovery tool with the parameters it publishes and what it takes to call it.
```
grayhemp, line 74: Why are we doing these new tests?

src/redshift-mcp-server/README.md
```diff
@@ -110,14 +110,19 @@ or docker after a successful `docker build -t awslabs/redshift-mcp-server:latest
 - `AWS_PROFILE`: AWS profile to use (optional, uses default if not specified)
 - `FASTMCP_LOG_LEVEL`: Logging level (`DEBUG`, `INFO`, `WARNING`, `ERROR`)
 - `LOG_FILE`: Path to log file (optional, logs to stdout if not specified)
+- `REDSHIFT_MAX_RESULT_ROWS`: Maximum rows any one statement may return (default: `1000`). Only a
```
grayhemp, line 113: Make it concise.

src/redshift-mcp-server/README.md
```diff
@@ -404,15 +426,99 @@ list_columns(
...
-**Returns**: List of column information including:
+**Returns**: A list of columns, each including:
 
 - Column name and data type
 - Nullable status and default values
 - Numeric precision and scale
 - Character length limits
 - Ordinal position and remarks
 
+### Response shaping
```
grayhemp, line 439: Make it concise.

src/redshift-mcp-server/awslabs/redshift_mcp_server/server.py
grayhemp, line 176: Instructions need to inform the agent of the current settings so it would adjust its behavior for the max row result cap. See my message in the DM, I'll show you how to do that.

grayhemp (changes requested):

PR #4555, opened 2026-08-28
fix(bedrock-kb-retrieval-mcp-server): support Amazon Bedrock Managed Knowledge Bases

### LINKED ISSUE #4554: bedrock-kb-retrieval-mcp-server: QueryKnowledgeBases fails against Amazon Bedrock Managed Knowledge Bases (wrong retrievalConfiguration shape)
## Summary

`QueryKnowledgeBases` fails on every call against an Amazon Bedrock **Managed Knowledge Base**
(the newer KB type under Bedrock AgentCore, distinct from a self-managed KB with your own
`storage_configuration`). Confirmed live against a real Managed KB in `us-east-1`.

## Root cause

`src/bedrock-kb-retrieval-mcp-server/awslabs/bedrock_kb_retrieval_mcp_server/knowledgebases/retrieval.py`
hard-codes the `retrievalConfiguration` shape used for self-managed KBs:

```python
retrieve_request: KnowledgeBaseRetrievalConfigurationTypeDef = {
    'vectorSearchConfiguration': {
        'numberOfResults': number_of_results,
    }
}
```

Bedrock's `Retrieve` API rejects this for a Managed KB:

```
ValidationException: vectorSearchConfiguration is not supported for managed knowledge bases
```

Managed KBs require `managedSearchConfiguration` instead - same nested shape
(`numberOfResults`, `filter`, `rerankingConfiguration` all exist under it too, confirmed via the
botocore `Retrieve` operation model), just a different top-level key. The server has no logic to
detect KB type and pick the right key.

Separately, the reranking model default (`amazon.rerank-v1:0`, `AMAZON` in
`reranking_model_name`) is not available in `us-east-1` - only `cohere.rerank-v3-5:0` is.

## Reproduction

1. Create a Bedrock Managed Knowledge Base with a native connector (e.g. Confluence) in `us-east-1`.
2. Configure this MCP server against it (`ListKnowledgeBases` works fine - it's a different API).
3. Call `QueryKnowledgeBases` with that knowledge base ID.
4. Every call fails with `ValidationException: vectorSearchConfiguration is not supported for managed knowledge bases`.

## Suggested fix

Detect KB type (via `GetKnowledgeBase`'s `knowledgeBase.type` field - `MANAGED` vs the
self-managed types) and send `managedSearchConfiguration` for Managed KBs, falling back to the
existing `vectorSearchConfiguration` for self-managed KBs. Alternatively, always prefer
`managedSearchConfiguration` when calling `Retrieve` on a KB whose type is `MANAGED`.

Also worth fixing: don't default `reranking_model_name` to a model unavailable in the caller's
region, or validate the (model, region) pair before making the call instead of surfacing AWS's
opaque error.

We've validated a working patch (rename `vectorSearchConfiguration` -> `managedSearchConfiguration`,
default reranking model to `cohere.rerank-v3-5:0`) and are running it as a local fork in the
meantime. Happy to open a PR with this fix if useful - let us know.

pbcentaurlabs:
Fixes #4554

## Problem

`QueryKnowledgeBases` fails on every call against an Amazon Bedrock **Managed Knowledge Base**
(the AgentCore-native-connector KB type, distinct from a self-managed KB with your own
`storage_configuration`). Confirmed live against a real Managed KB in `us-east-1`.

`retrieval.py` always sent `retrievalConfiguration={'vectorSearchConfiguration': {...}}`, which
Bedrock's `Retrieve` API rejects for a Managed KB:

```
ValidationException: vectorSearchConfiguration is not supported for managed knowledge bases
```

Managed KBs need `managedSearchConfiguration` instead - confirmed via the `Retrieve` operation's
botocore model that it has the identical nested shape (`numberOfResults`, `filter`,
`rerankingConfiguration`), just a different top-level key.

Separately, the reranking default (`AMAZON` / `amazon.rerank-v1:0`) isn't available in
`us-east-1`, which is one of the regions this server already allows reranking in.

## Fix

- Detect KB type via `GetKnowledgeBase`'s `knowledgeBaseConfiguration.type` (`'MANAGED'` vs
  everything else) and pick `managedSearchConfiguration` or `vectorSearchConfiguration`
  accordingly - self-managed KB behavior is unchanged.
- `query_knowledge_base` now also takes `kb_agent_mgmt_client` (the Bedrock agent management
  client) to make that lookup; `server.py` already constructs this client globally, so it's just
  threaded through.
- Default `reranking_model_name` to `'COHERE'` (available in every region this server allows
  reranking in), instead of `'AMAZON'` (not available in `us-east-1`).

## Testing

- Added `test_query_managed_knowledge_base` covering the Managed KB path end to end (asserts
  `GetKnowledgeBase` is called and `managedSearchConfiguration` is sent).
- Updated `conftest.py`'s `mock_bedrock_agent_client` to simulate a Managed KB
  (`kb-managed-1` -> `knowledgeBaseConfiguration.type == 'MANAGED'`) alongside the existing
  self-managed KB mocks, which are unchanged.
- Full suite: `uv run pytest tests/` - 39/39 passing.
- Manually verified against a real Managed KB (Confluence-backed) in `us-east-1`: `ListKnowledgeBases`
  and `QueryKnowledgeBases` both return correct, cited results after this fix.

## Notes

Ran this as a local fork before opening this PR (see #4554) - happy to adjust the approach if a
maintainer prefers a different detection strategy (e.g. an explicit parameter instead of an
extra `GetKnowledgeBase` call).


pbcentaurlabs:
Thanks for the review - pushed 48a55c87 addressing all four points:

- `_is_managed_knowledge_base` is now a plain sync function, run via `asyncio.to_thread`, with its result cached per `knowledge_base_id` (a KB's type never changes, so the extra `GetKnowledgeBase` round-trip only happens once per KB).
- Added explicit `(region, reranking_model_name)` validation (`RERANKING_MODEL_SUPPORTED_REGIONS`) instead of the old broad region-only allow-list, so requesting `AMAZON` in `us-east-1` now fails with a clear `ValueError` instead of an opaque Bedrock `ValidationException`.
- Reworded the `RERANKING_MODEL_NAME_MAPPING` comment - it's a name->model-ID map, not region-scoped; region support now lives in the new map above.
- `retrieve_request` is now `dict[str, Any]` (the key is chosen dynamically), so the `type: ignore` comments are gone.
- Reworded the `conftest.py` mock comment that was internally inconsistent.

39/39 tests passing, `ruff check`/`format` clean.

pbcentaurlabs:
Superseded by #4574 (merged 2026-09-16), released in v1.1.2 (2026-09-22). Verified live against a real Managed KB that stock v1.1.2 handles both issues this PR addressed - `managedSearchConfiguration` selection via `GetKnowledgeBase`, and a clear region check for the reranking model instead of an opaque failure. Closing in favor of that fix.

PR #4556, opened 2026-08-28
feat(aurora-dsql-mcp-server): use native ALTER TABLE DROP COLUMN for DSQL

mripley:
Fixes

## Summary

Mirrors [awslabs/agent-plugins#264](https://github.com/awslabs/agent-plugins/pull/264), the canonical home for DSQL agent steering. That PR was authored and reviewed first; this one carries the settled text.

Aurora DSQL shipped `ALTER TABLE ... DROP COLUMN` on [2026-08-03](https://docs.aws.amazon.com/aurora-dsql/latest/userguide/release-notes.html), but this repo still asserted "DSQL does **NOT** support direct ... `DROP COLUMN`" and routed the operation through the destructive Table Recreation Pattern.

`DROP COLUMN` is synchronous and metadata-only. Unlike `VALIDATE CONSTRAINT` it has **no** `ALTER TABLE ASYNC` form and returns no `job_id`.

### Changes

Both agent-facing surfaces in this package are updated:

- **`skills/dsql-skill/`** (source of truth) — `SKILL.md` plus `references/development-guide.md`, `references/ddl-migrations/{overview,column-operations}.md`, `references/mysql-migrations/{ddl-column-changes,type-mapping}.md`.
  The five alias `SKILL.md` files were regenerated by the `sync-dsql-skill-aliases` pre-commit hook, **not** hand-edited — they differ from the source on the `name:` line only, and the hook is idempotent on a second pass.
- **`kiro_power/`** — `POWER.md` plus the flattened `steering/` counterparts. Each `steering/*.md` remains byte-identical to its `references/` counterpart.
- **`CHANGELOG.md`** — entry under `## Unreleased`.

Substance (identical to the canonical PR): the rule lands in the always-loaded `development-guide.md`; `column-operations.md` documents the native statement with the rules that actually bite — PK *key* columns rejected with `0A000`, `CASCADE` only for out-of-table dependents (a bare drop already sweeps the column's indexes and same-table constraints), `CASCADE` itself failing `0A000` when the dependency chain reaches a PK key column, `IF EXISTS`, gradual storage reclamation, and the 255-active / 1,600-lifetime column budget.

The `Workflow 6` sentence is now affirmative rather than "DSQL does NOT support direct ...", matching the canonical wording and the project's steering style.

Table recreation is unchanged for `ALTER COLUMN TYPE`, `SET NOT NULL`, `DROP CONSTRAINT`, and `MODIFY PRIMARY KEY`.

### User experience

**Before** — asked to drop an obsolete column, the agent produced a destructive multi-step migration: create a replacement table, `INSERT ... SELECT` every row (batched under 3,000), verify counts, `DROP TABLE`, `RENAME`, then rebuild indexes. On an 80k-row table that is minutes of work, a write outage, and an irreversible `DROP TABLE`.

**After** — the agent emits `ALTER TABLE orders DROP COLUMN legacy_promo_code`, confirms the table and column with the user first, and adds `CASCADE` only when a dependency actually requires it.

Measured against the eval suite in `awslabs/agent-plugins` (regex-graded, no cluster needed): **0/2** on the pre-change skill, **2/2** after.

## Checklist

* [x] Changes are documented

**RFC issue number**:

Checklist:

### Testing notes

Documentation-only; the MCP server implementation is untouched. Verified:

- `skills/dsql-skill/` passes skill structure/frontmatter validation; `SKILL.md` 343 lines and `POWER.md` 259, both under the 500-line cap.
- All five reference↔steering pairs byte-identical (`diff` per pair — note a recursive `diff -r` is *not* a valid check here, since subdirectory files are flattened into `steering/` filenames).
- All five alias `SKILL.md` files differ from `dsql-skill/SKILL.md` on exactly the `name:` line.
- No trailing whitespace, CRLF, or missing EOF newline in any changed file.

### Pre-existing issue noted but not fixed

The `steering/` tree was flattened without rewriting intra-document relative links, so links like `../ddl-migrations/column-operations.md` are already dead there (23 instances across the files this PR touches). This PR preserves the byte-identical mirror rather than diverging the two trees, so it neither fixes nor worsens the aggregate count. Happy to follow up with a dedicated link-rewrite pass if you'd prefer that separately.

## Acknowledgment

### DIFF
--- a/src/aurora-dsql-mcp-server/CHANGELOG.md
+++ b/src/aurora-dsql-mcp-server/CHANGELOG.md
@@ -18,6 +18,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Add `ALTER TABLE ASYNC ... VALIDATE CONSTRAINT` guidance to DSQL steering. CHECK constraints can now be added with `NOT VALID` and validated asynchronously, following the same async DDL pattern as `CREATE INDEX ASYNC`.
 
+### Changed
+
+- Update DSQL steering for native `ALTER TABLE ... DROP COLUMN` support. Dropping a column no longer requires the destructive Table Recreation Pattern — it is a synchronous, metadata-only change, so agents stop generating a CREATE/INSERT-SELECT/DROP/RENAME sequence for it. Steering documents the primary-key restriction, `CASCADE` for dependent objects, and the 255-active/1,600-lifetime column budget. Unlike `VALIDATE CONSTRAINT`, `DROP COLUMN` has no `ALTER TABLE ASYNC` form and returns no `job_id`.
+
 ### Security
 
 - Close read-only bypasses in `readonly_query` / `transact` (read-only mode), aligning the SQL classifier with the `postgres-mcp-server` sibling:
--- a/src/aurora-dsql-mcp-server/kiro_power/POWER.md
+++ b/src/aurora-dsql-mcp-server/kiro_power/POWER.md
@@ -55,10 +55,10 @@ This power includes the following steering files in [steering](./steering)
   - MUST load when writing OCC retry code or mitigating commit-time conflicts
   - DSQL Connectors, manual retry pattern, conflict mitigation, idempotent transaction design
 - **ddl-migrations-overview**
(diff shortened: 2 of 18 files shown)


anwesham-lab:
closing in favor of skill deprecation

the agent plugin will be the sole source of truth

PR #4585, opened 2026-09-04
feat(mysql-mcp-server): restore --secret_arn for custom-secret auth (#3643, #4099)

### LINKED ISSUE #3643: mysql-mcp-server 1.0.21: removed --secret_arn / custom-secret support (undocumented breaking change in a patch release)
### Summary

`mysql-mcp-server` **1.0.21** removed the `--secret_arn` CLI argument and the ability to authenticate with a **user-supplied Secrets Manager secret**. For the `rdsapi` (RDS Data API) connection method, the secret ARN is now auto-derived exclusively from the cluster's managed `MasterUserSecret`. This is a **breaking change** that was:

1. **undocumented** — neither the CHANGELOG nor PR #3626 mentions the removal of `--secret_arn` or the change in how credentials are resolved, and
2. **shipped as a patch release** (1.0.20 → 1.0.21), so version pinning on `@latest` / `>=1.0` silently picked it up and broke working deployments.

The net effect is a silent **privilege change**: configurations that previously connected as a dedicated least-privilege (e.g. read-only) database user via a custom secret now connect as the **cluster master user**, or fail outright if the cluster has no managed master secret.

### Versions

- **1.0.20** (last working) — accepted `--resource_arn`, `--secret_arn`, `--readonly True/False`.
- **1.0.21** (broken) — renamed `--resource_arn` → `--db_cluster_arn`, **removed `--secret_arn`**, replaced `--readonly` with `--allow_write_query`, added `--connection_method` / `--db_type`.

### What used to work (1.0.20)

```jsonc
"args": [
  "awslabs.mysql-mcp-server@1.0.20",
  "--resource_arn", "arn:aws:rds:<region>:<acct>:cluster:<cluster>",
  "--secret_arn",   "arn:aws:secretsmanager:<region>:<acct>:secret:<custom-read-only-secret>",
  "--database",     "<db>",
  "--region",       "<region>",
  "--readonly",     "True"
]
```

This connected via the RDS Data API using a **custom, read-only** Secrets Manager secret — a DB user whose grants physically prohibit writes.

### What breaks in 1.0.21

Launching with the old args fails immediately:

```
awslabs.mysql-mcp-server: error: unrecognized arguments: --secret_arn arn:... --readonly True
```

…and there is **no replacement flag** to supply a custom secret. In `server.py`, the secret is always taken from the cluster/instance managed master secret:

```python
# internal_connect_to_database(...)
secret_arn = cluster_properties.get('MasterUserSecret', {}).get('SecretArn')   # cluster path
...
secret_arn = instance_properties.get('MasterUserSecret', {}).get('SecretArn')  # instance path
```

and that value is what every connection method uses:

- `RDS_API` → `RDSDataAPIConnection(secret_arn=str(secret_arn), ...)`
- `MYSQL_WIRE_PROTOCOL` → `AsyncmyPoolConnection(secret_arn=secret_arn, db_user='', ...)`
- `MYSQL_WIRE_IAM_PROTOCOL` → IAM auth with `db_user=masteruser` (hardcoded to the cluster's `MasterUsername`)

A grep of the 1.0.21 package confirms there is **no CLI flag and no environment variable** that overrides the secret ARN. The `RDSDataAPIConnection` / `AsyncmyPoolConnection` classes still accept `secret_arn` as a constructor argument — only the caller no longer allows it to be supplied externally.

### Impact

For any user who was running with a dedicated least-privilege secret (the documented pattern in earlier releases), upgrading to 1.0.21:

- **Cannot use that secret at all** — there is no way to pass it.
- **Silently escalates the connection to the cluster master user.** Read-only safety then depends solely on the server's SQL-string validation (omitting `--allow_write_query`) — i.e. application-level enforcement on a write-capable credential, instead of credential-level enforcement by a user that lacks write grants. For an LLM-driven tool, that is a meaningfully weaker safety posture.
- **Fails entirely** if the cluster does not use a Secrets-Manager-managed master password (`MasterUserSecret` is absent → `secret_arn` resolves to `None`).

### Documentation gap

PR #3626 lists its breaking changes as `--readonly` → `--allow_write_query`, the `create_cluster` default db name, and the `database_type` enum (stated as having "no external callers"). The removal of `--secret_arn` / custom-secret support is not in that list, the CHANGELOG, or any migration note — yet it is the change most likely to break existing production configs, and it has a security dimension.

### Requested fix

Restore the ability to authenticate with a user-supplied secret, for at least the `rdsapi` and `mysqlwire` connection methods — e.g. re-add an explicit `--secret_arn` flag (and/or an env var) that, when provided, takes precedence over the auto-derived `MasterUserSecret`. Auto-derivation is a fine default; it should not be the only option.

Secondary asks:
- Document the credential-resolution change and the `--secret_arn` removal as an explicit breaking change with migration guidance.
- Consider that breaking CLI/credential changes warrant at least a minor version bump, so consumers pinning by compatible range aren't broken by a patch.

### Environment

- Package: `awslabs.mysql-mcp-server`
- Broken in: `1.0.21`; last working: `1.0.20`
- Engine: Aurora MySQL, connection method RDS Data API (`rdsapi`)
- Client: invoked via `uvx ...@latest` (which is how the breaking patch was picked up automatically)

### LINKED ISSUE #4099: [mysql-mcp-server] Add --secret_arn support for custom Secrets Manager credentials (parity with mssql-mcp-server)
**Is your feature request related to a problem?**
Yes. The _awslabs.mysql-mcp-server_ currently only supports credential discovery via the RDS-managed _MasterUserSecret_ (which requires "Manage master credentials in AWS Secrets Manager" to be enabled on the cluster). This forces the MCP server to connect as the master user (e.g., _root_), which has full privileges.

In enterprise environments, we need to connect as a least-privilege read-only user (e.g., _kiro_readonly_ with _GRANT SELECT_ only). However, there is no way to point the MySQL MCP server to a custom secret containing the read-only user's credentials.

The **mssql-mcp-server already supports this** via the _--secret_arn_ flag. The MySQL server lacks this feature.

**Describe the solution you'd like**

Add _--secret_arn_ CLI argument support to _awslabs.mysql-mcp-server_, identical to what _awslabs.mssql-mcp-server_ already provides:

```
awslabs.mysql-mcp-server \ 
  --secret_arn arn:aws:secretsmanager:us-west-2:123456789012:secret:my-readonly-user-AbCdEf
```

The secret format should be:

```
{
  "username": "kiro_readonly",
  "password": "the-password",
  "host": "my-cluster.cluster-ro-xxxx.us-west-2.rds.amazonaws.com",
  "port": 3306,
  "dbname": "my_database"
}
```

This would also allow the _connect_to_database_ MCP tool to accept a _secret_arn_ parameter at runtime (as MSSQL already supports).

**Describe alternatives you've considered**

1. Relying on managed master credentials — This works but forces connection as the master user with full DML privileges. This is unacceptable in team environments where multiple developers use Kiro, as anyone could potentially execute write operations on production data.
2. IAM DB Authentication (mysqlwire_iam) — The server defaults to using MasterUsername from DescribeDBClusters. There is no way to specify a different DB username for IAM token generation.
3. Manually tagging a custom secret with aws:rds:primaryDBClusterArn — Not possible. AWS reserves aws: prefixed tags as system-only; they cannot be created by users.
4. Third-party MCP servers — Not acceptable for enterprise/organizational use due to security and trust requirements.

**Use case**
Our team of 20+ developers uses AWS Kiro IDE with the MySQL MCP server to query an Aurora MySQL cluster. Security requirements mandate:

- Kiro connects as a read-only user (_GRANT SELECT_ only)
- No developer has access to the writer instance credentials via Kiro
- The actual security boundary is the database role, not application-level regex blocking
- Steering files and application-level controls are insufficient because team members can modify them

Without _--secret_arn_, we cannot enforce least-privilege at the database level through this MCP server.

**Additional context**

- The MSSQL MCP server already implements this feature: https://awslabs.github.io/mcp/servers/mssql-mcp-server (see "Using a custom Secrets Manager secret" section)
- The recommended security model in the MySQL MCP server's own README states: "The actual security boundary is the database role you connect with" — but there's no way to control which role the server connects with
- This gap forces enterprises to choose between using managed master credentials (overly permissive) or not using the MCP server at all

**Environment**
- MCP Server: _awslabs.mysql-mcp-server_ (latest)
- Database: Aurora MySQL 8.0
- Region: us-west-2
- Connection Method: _mysqlwire_

warren830:
## Issues

Fixes #3643. Closes #4099.

## Problem

`mysql-mcp-server` **1.0.21** removed the ability to authenticate with an operator-supplied Secrets Manager secret: `secret_arn` is now derived **exclusively** from the cluster/instance's AWS-managed `MasterUserSecret`. This silently forces connections as the **master user** — a privilege regression (#3643) for setups that previously connected as a dedicated least-privilege user (e.g. read-only) via a custom secret. `mssql-mcp-server` still supports `--secret_arn`, so the two servers are also out of parity (#4099).

The connection layer (`AsyncmyPoolConnection` / `RDSDataAPIConnection`) already accepts an arbitrary `secret_arn` — `server.py` simply never exposed a way to set it.

## Fix

- Add a `--secret_arn` CLI flag, wired exactly like the existing `--ca_bundle` (module global `configured_secret_arn`, declared in `internal_connect_to_database` and `main`, set from `args.secret_arn`).
- When provided, it **overrides** the derived `MasterUserSecret` ARN for the `mysqlwire` (non-IAM) and `rdsapi` methods.
- It is **ignored for `mysqlwire_iam`**, which authenticates via IAM and uses no secret (that path's `secret_arn` stays `''`).
- README documents the least-privilege use case.

## Tests

Added to `tests/test_server.py::TestInternalConnectToDatabaseRouting`:

- `--secret_arn` overrides `MasterUserSecret` on the **wire** path;
- `--secret_arn` overrides it on the **RDS Data API** path;
- `--secret_arn` is **not** applied to **IAM auth** (`secret_arn == ''`).

Strict red→green verified: neutralizing the override makes the two override tests fail (connection would use `arn:master-user-secret`); the fix makes them pass, and the IAM-ignored test passes both ways. Full `test_server.py` (28 tests) green; `ruff check` / `ruff format --check` clean.

## Compatibility

No behavior change when `--secret_arn` is omitted (the managed `MasterUserSecret` is still used), so existing deployments are unaffected.

---

### DIFF
--- a/src/mysql-mcp-server/README.md
+++ b/src/mysql-mcp-server/README.md
@@ -155,6 +155,23 @@ The MCP server uses the AWS profile specified in the AWS_PROFILE environment var
 
 Make sure the AWS profile has permissions to access the RDS Data API and the secret from AWS Secrets Manager. Wire-protocol methods (`mysqlwire` / `mysqlwire_iam`) additionally require `rds:DescribeDBClusters` / `rds:DescribeDBInstances` — see the prerequisites above. The MCP server creates a boto3 session using the specified profile to authenticate with AWS services. Your AWS IAM credentials remain on your local machine and are strictly used for accessing AWS services.
 
+#### Connecting as a least-privilege user with `--secret_arn`
+
+By default, `mysqlwire` and `rdsapi` connections read credentials from the
+cluster/instance's AWS-managed `MasterUserSecret`, which connects as the master
+user (full privileges). To connect as a dedicated least-privilege user instead
+(for example a read-only `GRANT SELECT` user), store that user's credentials in
+an AWS Secrets Manager secret and pass its ARN with `--secret_arn`:
+
+```
+uvx awslabs.mysql-mcp-server@latest --secret_arn arn:aws:secretsmanager:us-east-1:123456789012:secret:my-readonly-user
+```
+
+When provided, `--secret_arn` overrides the managed `MasterUserSecret` for the
+`mysqlwire` (non-IAM) and `rdsapi` methods. It is ignored for `mysqlwire_iam`,
+which authenticates via IAM and uses no secret. The secret must be a standard
+RDS-style secret containing at least `username` and `password`.
+
 ## Security model
 
 > **The server's read-only mode is a best-effort SQL-text safeguard, not a security boundary.**
--- a/src/mysql-mcp-server/awslabs/mysql_mcp_server/server.py
+++ b/src/mysql-mcp-server/awslabs/mysql_mcp_server/server.py
@@ -63,6 +63,14 @@ readonly_query = True
 # Amazon RDS global bundle (verified against a pinned hash) is used.
 ca_bundle_path: Optional[str] = None
 
+# Operator-supplied Secrets Manager secret ARN. Set by the CLI flag
+# --secret_arn on server start. When provided, credentials are read from this
+# secret instead of the cluster/instance's AWS-managed ``MasterUserSecret``,
+# which lets the server connect as a dedicated least-privilege user rather than
+# the master user. When empty/None, the managed MasterUserSecret ARN is used
+# (the historical default). Not used by the IAM-auth connection method.
+configured_secret_arn: Optional[str] = None
+
 
 class DummyCtx(Context):
     """A dummy context class for error handling in MCP tools."""
@@ -627,6 +635,7 @@ def internal_connect_to_database(
     global db_connection_map
     global readonly_query
     global ca_bundle_path
+    global configured_secret_arn
 
     logger.info(
         f'Enter internal_connect_to_database\n'
@@ -792,6 +801,14 @@ def internal_connect_to_database(
         if existing_conn:
             return (existing_conn, _connected_response())
 
+    # If the operator supplied an explicit --secret_arn, use it instead of the
+    # AWS-managed MasterUserSecret ARN derived above. This restores the ability
+    # (removed in 1.0.21) to connect as a dedicated least-privilege user via a
+    # custom secret, and matches mssql-mcp-server. The IAM-auth method uses no
+    # secret, so it is unaffected (its secret_arn is forced to '' below).
+    if configured_secret_arn:
+        secret_arn = configured_secret_arn
+
     logger.info(
         f'About to create internal DB connections with:'
         f'enable_data_api:{enable_data_api}\n'
@@ -873,6 +890,7 @@ def main():
     global db_connection_map
     global readonly_query
     global ca_bundle_path
+    global configured_secret_arn
 
     parser = argparse.ArgumentParser(
         description='An AWS Labs Model Context Protocol (MCP) server for MySQL'
@@ -909,6 +927,20 @@ def main():
             'rotates CAs faster than the package release cadence.'
         ),
     )
+    parser.add_argument(
+        '--secret_arn',
+        default=None,
+        help=(
+            'ARN of an AWS Secrets Manager secret holding the database '
+            'credentials to connect with. When provided, it overrides the '
+            "cluster/instance's AWS-managed MasterUserSecret, letting the "
+            'server connect as a dedicated least-privilege user (e.g. a '
+            'read-only user) instead of the master user. Applies to the '
+            'RDS_API and MYSQL_WIRE_PROTOCOL connection methods; ignored for '
+            'MYSQL_WIRE_IAM_PROTOCOL, which uses IAM authentication and no '
+            'secret.'
+        ),
+    )
     args = parser.parse_args()
 
     logger.info(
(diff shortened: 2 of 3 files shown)


warren830:
Hi maintainers — a review nudge on this one.

- **Just synced onto current `main`** (was 24 commits behind, now 0; head `b06911fb`); both checks are green.
- Scope: 3 files, +158/-0 — additive only, no existing lines removed.
- It restores `--secret_arn` for custom-secret auth in `mysql-mcp-server`, addressing #3643 and also covering #4099: without it, users whose credentials live in a secret they manage themselves have no supported way to connect.

Because this reintroduces a previously-available flag, the one point worth a maintainer's judgement is whether you want it back under this exact name and shape, or under a different option. That is a product call rather than an implementation detail, so I would rather have your read on it than guess — happy to rename or restructure the interface if you prefer.

Issue #167, opened 2025-04-18
RFC: Package Bumping


scottschreckengaust:
### Summary

Each package has it's own semantic version. This discusses how package authors may manage versions.

### Use case

We expect authors to follow [semantic versioning](https://semver.org/). [Commitizen](https://commitizen-tools.github.io/commitizen/) is a popular method, yet it uses repository tagging which would require all packages to be the same version.

### Proposal

The "Major" and "Minor" portions of the build are the responsibility of the author. The `release.py` will alway bump the "Patch" unless it is above python's maxsize which will reset the next "Patch" to zero.

```python
>>> sys.maxsize
9223372036854775807
```

### Out of scope

Fixing Commitizen to work without remote tags or reviewing git log comments to derive expected next version.

### Potential challenges

To help authors, this might help for manually output:

```shell
% uv run cz bump --major-version-zero --increment PATCH --get-next --files-only --dry-run
...
0.9.2

% uv run cz bump --get-next --files-only --dry-run
...
1.0.0
```

and then updating the files.

### Dependencies and Integrations

This requires setting the "Patch" to `9223372036854775807` or higher to cause the `release.py` to bump to zero `0`.

### Alternative solutions

```markdown
Develop complicated tooling that mimics commitizen without remote git tag references.
```

PR #4604, opened 2026-09-09
fix(cicd): make a server's version a single source of truth

### LINKED ISSUE #425: chore(cicd): bump version in all related files
### Describe the bug

Also bump the version outside of `pyproject.toml`

### Expected Behavior

When the release workflow runs, files holding versions should also coordinate the bumped changes.

### Current Behavior

The `__version__` value in the `__init__.py` should be coordinated.

### Reproduction Steps

Run the release and the value will not update.

### Possible Solution

Add to the release script to update relevant values.

_No response_

### OS

All

### Server

other

### Server Version

All

N/A

### Other information

_No response_

### Service quota

### LINKED ISSUE #4476: amazon-translate-mcp-server: runtime MCP_SERVER_VERSION drifted from published package version (1.0.0 vs 1.0.5)
Follow-up from a non-blocking review observation on release PR #4475: https://github.com/awslabs/mcp/pull/4475#pullrequestreview-4901652637

Thanks for maintaining this server! This is a small consistency request, not a bug report — no action is urgent, and everything below is pre-existing on `main` and untouched by #4475.

## Summary

`amazon-translate-mcp-server` defines its version differently from nearly every other Python server in this repo, and as a result the release automation cannot bump it. Its runtime-reported version has drifted from the published package version.

| Source | Value |
| --- | --- |
| `pyproject.toml` `[project] version` | `1.0.5` (as of #4475) |
| Published on PyPI (`awslabs.amazon-translate-mcp-server`) | `1.0.5` |
| [`consts.py:15`](https://github.com/awslabs/mcp/blob/main/src/amazon-translate-mcp-server/awslabs/amazon_translate_mcp_server/consts.py#L15) `MCP_SERVER_VERSION` | **`1.0.0`** |

## How this server differs from the convention

I audited all 59 Python servers under `src/`. 54 of them declare a plain string literal `__version__` in `awslabs/<module>/__init__.py`, which is exactly what the release automation rewrites. This server is one of 5 that do not.

Its [`__init__.py:17-19`](https://github.com/awslabs/mcp/blob/main/src/amazon-translate-mcp-server/awslabs/amazon_translate_mcp_server/__init__.py#L17-L19) re-exports the constant instead of declaring a literal:

```python
from awslabs.amazon_translate_mcp_server.consts import MCP_SERVER_VERSION

__version__ = MCP_SERVER_VERSION
```

`.github/workflows/release.py` bumps `pyproject.toml`, then rewrites `__init__.py` by regex ([lines 389-400](https://github.com/awslabs/mcp/blob/main/.github/workflows/release.py#L389-L400)):

```python
version_pattern = r'__version__\s*=\s*(?P<start>[\'"])[^\'"]*(?P<end>[\'"])'
...
else:
    click.echo(f'Warning: No __version__ found in {init_file}')
```

The pattern requires a quote immediately after `=`. Here it finds an imported name, so nothing matches, the script logs a warning and continues, and `consts.py` is never touched. The drift is silent and grows one patch level per release.

Side note: `[tool.commitizen] version_files` in `pyproject.toml` already lists `awslabs/amazon_translate_mcp_server/__init__.py:__version__`, but the release workflow does not invoke commitizen, so that config currently has no effect.

## Why it is worth fixing here

`MCP_SERVER_VERSION` is not cosmetic — it is stamped into the AWS SDK user agent at [`aws_client.py:93`](https://github.com/awslabs/mcp/blob/main/src/amazon-translate-mcp-server/awslabs/amazon_translate_mcp_server/aws_client.py#L93):

```python
user_agent_extra=f'md/awslabs#mcp#amazon-translate-mcp-server/{MCP_SERVER_VERSION}',
```

So all service-side telemetry attributes traffic to `1.0.0`, which makes it hard to tell versions apart in usage data or to correlate a customer report with the release they are actually running.

## Suggested remediation

Two patterns already exist in this repo. Since this server has a real user-agent consumer, **Option A** matches it best.

### Option A — derive from package metadata (recommended; matches `mssql`, `oracle`, `postgres`)

Three servers already do this, e.g. [`mssql-mcp-server/awslabs/mssql_mcp_server/__init__.py:17-25`](https://github.com/awslabs/mcp/blob/main/src/mssql-mcp-server/awslabs/mssql_mcp_server/__init__.py#L17-L25). It gives one source of truth, and the literal fallback still matches `release.py`'s regex so automation keeps it fresh.

Replace `consts.py:15`:

```python
# consts.py  (was: MCP_SERVER_VERSION = '1.0.0')
from importlib.metadata import version


try:
    MCP_SERVER_VERSION = version('awslabs.amazon-translate-mcp-server')
except Exception:
    MCP_SERVER_VERSION = '1.0.5'  # fallback for source checkouts; release.py bumps this literal
```

`__init__.py` and `aws_client.py` then need no changes at all.

### Option B — plain literal in `__init__.py` (matches the 54-server majority)

Move the literal into `__init__.py` so `release.py` finds it, and have `consts.py` import from there rather than the reverse:

```python
# __init__.py  (replaces lines 17-19)
__version__ = '1.0.5'
```

```python
# consts.py:15
from awslabs.amazon_translate_mcp_server import __version__ as MCP_SERVER_VERSION
```

This is the simplest and most consistent with the rest of the repo, though it does introduce an `__init__` import into `consts`, so Option A is likely cleaner here.

### Either way, a regression guard

```python
def test_version_matches_package_metadata():
    from importlib.metadata import version

    from awslabs.amazon_translate_mcp_server.consts import MCP_SERVER_VERSION

    assert MCP_SERVER_VERSION == version('awslabs.amazon-translate-mcp-server')
```

## Related

- #4477 — `aws-location-mcp-server` has the same class of drift for a *different* root cause (module path mismatch).
- Assigning the `/src/amazon-translate-mcp-server` codeowners per [.github/CODEOWNERS L28](https://github.com/awslabs/mcp/blob/main/.github/CODEOWNERS#L28). Happy to open the PR myself if you'd prefer — just say which option you like.

### LINKED ISSUE #4477: aws-location-mcp-server: __version__ never bumped by release automation, drifted from published version (1.0.0 vs 2.0.20)
Follow-up from a non-blocking review observation on release PR #4475: https://github.com/awslabs/mcp/pull/4475#pullrequestreview-4901652637

Thanks for maintaining this server! This is a small consistency request, not a bug report — nothing is urgent, and everything below is pre-existing on `main` and untouched by #4475.

## Summary

`aws-location-mcp-server` lays out its package module differently from every other Python server in this repo. As a side effect the release automation silently cannot find its `__init__.py`, so `__version__` has never been bumped and is now two majors behind the published package.

| Source | Value |
| --- | --- |
| `pyproject.toml` `[project] version` | `2.0.20` (as of #4475) |
| Published on PyPI (`awslabs.aws-location-mcp-server`) | `2.0.20` |
| [`awslabs/aws_location_server/__init__.py:17`](https://github.com/awslabs/mcp/blob/main/src/aws-location-mcp-server/awslabs/aws_location_server/__init__.py#L17) `__version__` | **`1.0.0`** |

## How this server differs from the convention

I audited all 59 Python servers under `src/`. This is the **only one** whose module directory name does not match its distribution name — every other server's `awslabs.<dist-name>` maps directly onto `awslabs/<dist_name>/`.

`.github/workflows/release.py` relies on that convention ([lines 380-384](https://github.com/awslabs/mcp/blob/main/.github/workflows/release.py#L380-L384)):

```python
if package_name.startswith('awslabs.'):
    module_name = package_name[8:].replace('-', '_')
    init_file = self.path / 'awslabs' / module_name / '__init__.py'
```

For this server that resolves to:

- expected: `awslabs/aws_location_mcp_server/__init__.py` ← does not exist
- actual: `awslabs/aws_location_server/__init__.py` ← the real module (no `mcp` segment)

So the bump takes the not-found branch at [line 401](https://github.com/awslabs/mcp/blob/main/.github/workflows/release.py#L401) and only logs a warning:

```python
else:
    click.echo(f'Warning: {init_file} not found for package {package_name}')
```

This is a different failure mode from the sibling issue #4476, where the file *is* found but holds no string literal to rewrite.

Side note: `[tool.commitizen] version_files` in `pyproject.toml` already points at the correct path, but the release workflow does not invoke commitizen, so that config currently has no effect.

## Impact

Lower than the translate case — `grep` shows `__version__` has no other in-package consumers today, so nothing is mis-stamped into a user agent. It is mainly a latent trap: any future code that logs or reports `__version__`, or any consumer importing it, silently gets `1.0.0`.

## Suggested remediation

Two patterns already exist in the repo; either fixes the drift without touching the module layout.

### Option A — derive from package metadata (recommended; matches `mssql`, `oracle`, `postgres`)

Three servers already do this, e.g. [`mssql-mcp-server/awslabs/mssql_mcp_server/__init__.py:17-25`](https://github.com/awslabs/mcp/blob/main/src/mssql-mcp-server/awslabs/mssql_mcp_server/__init__.py#L17-L25). One source of truth, and it is robust to the path mismatch because it never needs `release.py` to rewrite anything.

Replace `awslabs/aws_location_server/__init__.py:17`:

```python
# was: __version__ = '1.0.0'
from importlib.metadata import version


try:
    __version__ = version('awslabs.aws-location-mcp-server')
except Exception:
    __version__ = '2.0.20'  # fallback for source checkouts
```

### Option B — keep the literal and make automation find it

Keep `__version__ = '2.0.20'` as a literal and teach `release.py` to fall back to `[tool.commitizen] version_files` when the conventional path is absent. This fixes the general case for any future server that deviates, but is a change to shared release tooling rather than to this package.

### Worth considering separately: the underlying cause

- **Harden `release.py`**: make a missing `__init__.py` for an `awslabs.*` package a hard failure instead of a warning. A silent warning is why this went unnoticed across ~20 releases, and it will bite the next server that deviates.
- **Optional rename** `awslabs/aws_location_server/` → `awslabs/aws_location_mcp_server/` to match the package name and repo convention. This is the change that would make this server fully consistent, but it is a breaking import change for anyone importing the module directly, so it deserves its own discussion and is **not** required to fix the version drift.

### Either way, a regression guard

```python
def test_version_matches_package_metadata():
    from importlib.metadata import version

    from awslabs.aws_location_server import __version__

    assert __version__ == version('awslabs.aws-location-mcp-server')
```

## Related

- #4476 — `amazon-translate-mcp-server` has the same class of drift for a different root cause.
- Assigning the `/src/aws-location-mcp-server` codeowners per [.github/CODEOWNERS L41](https://github.com/awslabs/mcp/blob/main/.github/CODEOWNERS#L41). Happy to open the PR myself if you'd prefer — just say which option you like.

dihannahdi:
Fixes #4476, fixes #4477, fixes #425

## Summary

`pyproject.toml` is the only version that reaches PyPI — `python.yml`'s `uv build` reads nothing else. Every hardcoded version literal inside a package is therefore a second copy that something has to keep in sync, and the thing meant to keep it in sync fails open.

`PyPiPackage.bump_version()` in `.github/workflows/release.py` always bumps `pyproject.toml`, then makes a best-effort attempt to patch a literal in `awslabs/<module>/__init__.py`. It fails silently three ways:

1. **The directory name was derived, not discovered.** `module_name = package_name[8:].replace('-', '_')` produces `aws_location_mcp_server`, but aws-location-mcp-server's directory is `aws_location_server`. `init_file.exists()` is `False`, so it warned and returned success. That constant froze at `1.0.0` while `pyproject.toml` reached `2.1.1`.
2. **The regex only matches a quoted literal.** amazon-translate's `__init__.py` is `__version__ = MCP_SERVER_VERSION`, an identifier, so the real constant in `consts.py` was never touched.
3. **It only ever opens that one file.** A literal in `server.py` or a `utilities/` module is invisible to it.

All three paths `click.echo` a warning and still return the new version, and the commit step only checks `git diff --cached --quiet` over `src/`, which is always dirty because `pyproject.toml` did change. So the release PR opens green.

Rather than teach the release script to chase every literal, this removes the duplicate. The version is read from the installed distribution with `importlib.metadata`, following the pattern already used by `postgres`, `mssql` and `oracle`. There is then nothing to sync.

### Changes

Six servers were drifting. Three fed the stale value straight into botocore's `user_agent_extra`, so every AWS API call reported the wrong version:

| server | stale constant | `pyproject.toml` | consumed at |
|---|---|---|---|
| amazon-translate | `consts.py` `1.0.0` | `1.0.7` | `aws_client.py:93` |
| billing-cost-management | `utilities/aws_service_base.py` `1.0.0` | `0.0.37` | same file, `:99` |
| stepfunctions-tool | `server.py` `0.1.5` | `0.2.1` | `aws_helper.py:55` |
| aws-location | `__init__.py` `1.0.0` | `2.1.1` | not consumed |
| valkey | `version.py` `0.2.0` | `1.1.1` | not consumed, file removed |
| sagemaker-ai | `sagemaker_hyperpod/__init__.py` `1.0.0` | `1.1.0` | not consumed, constant removed |

billing-cost-management and stepfunctions-tool each had a *correct* constant and a stale shadow, and the consumer had picked the shadow.

- `release.py` now discovers the subpackage directory instead of deriving its name, raises on a genuinely structural problem instead of warning, and treats the absence of a literal as correct rather than as a failure. `validate_path_security` is preserved, now called only for a file that exists — previously it raised for a missing file, was caught by the local `except ValueError`, and was misreported as "Cannot update `__init__.py` safely", which made the "not found" branch unreachable.
- `scripts/verify_version_sync.py` walks the AST of every module under `awslabs/` and fails when a version literal disagrees with `pyproject.toml`. It catches all three of the blind spots above because it does not care which file the literal lives in, and it parses rather than regex-matches, since the regex is what failed. Wired into the existing per-package gate in `python.yml`, blocking (no `|| true`).
- `scripts/tests/test_verify_version_sync.py` covers the matching case, the disagreeing case, the indirect-assignment case, a literal in a nested non-`__init__` module, a package with no `awslabs/`, and a `pyproject.toml` with no version key. There is currently no job that runs tests for repo tooling — the matrix in `python.yml` is built from `src/*/pyproject.toml`, and the `pytest` hook in `.pre-commit-config.yaml` is `stages: [pre-push]` with `files: src\/.*\/pyproject.toml`, so it never runs in CI — so a small `verify-scripts` job was added rather than leaving a test nothing executes.
- `stepfunctions-tool`'s `tests/test_server.py` carried a commented-out `test_version_matches_pyproject()` pointing at #167 and #425. #167 proposed commitizen and was closed by the stale bot without a decision, which is why 53 servers still carry `[tool.commitizen]` config that nothing invokes. #425 asked for exactly this and was closed with no comments. The new check covers all 62 servers, so the dead block is removed.

### User experience

Before: `awslabs.amazon-translate-mcp-server` 1.0.7 identified itself to AWS as `md/awslabs#mcp#amazon-translate-mcp-server/1.0.0` on every Translate call. Nothing failed, nothing warned, and the same was true for Cost Explorer and Step Functions traffic. A release could not fix it, because the release is what caused it.

After: the version reported at runtime is the version that was published, and a literal that disagrees with `pyproject.toml` fails the PR that introduces it.

## Checklist

* [x] Changes are documented

**RFC issue number**:

Checklist:

### How this was verified

- `scripts/verify_version_sync.py` across all 62 servers: 6 violations before, 0 after.
- `importlib.metadata.version()` resolves the real published version for all four rewired distributions in a live venv (`0.2.1`, `1.0.7`, `0.0.37`, `2.1.1`) — not just correct on paper.
- `release.py` exercised against fixtures for all four cases: derived-name mismatch now patches the right file; indirect assignment succeeds without patching; two subpackages raises; a normal package with a literal still gets patched.
- Every touched server's suite compared against unmodified code. stepfunctions-tool 61 passed / 100% coverage. aws-location 53 passed. amazon-translate and billing-cost-management have pre-existing failures on `main` (1 and 43 respectively) and the counts are identical before and after. sagemaker-ai's 2 failures are Linux-only path assertions failing on Windows, also identical. valkey could not be run — `uv sync` fails building `valkey-glide` v2.3.1 in this environment, unrelated to this change; that server's edit is the deletion of an unreferenced file.
- `pre-commit run` over the diff: 21 hooks pass. `gitleaks` could not be installed here (its Go modules failed to download) and `check-license-header`, `pyright` and `pytest` were skipped — the last two are `pre-push`-staged and do not run in CI.

## Acknowledgment


dihannahdi:
Rebased onto `main` (`2fec290`). The branch is force-pushed.

The rebase showed a flaw in this PR, and the new gate is what caught it. The release bot bumped billing-cost-management to `0.0.38`. The `importlib.metadata` fallback I added in `aws_service_base.py` still held `'0.0.37'`, so `verify_version_sync.py` failed. `release.py` only patches `awslabs/<module>/__init__.py`, so any fallback literal outside that file goes stale on every release. That makes a fallback literal just another copy to keep in sync, which is what this PR is meant to remove.

`b2a805a` changes all four fallbacks this PR adds (translate, billing, location, stepfunctions) to `'unknown'`. amazon-bedrock-agentcore already does the same. The fallback only runs when the distribution isn't installed, and the verifier ignores non-semver strings.

Verified locally after the rebase:
- `verify_version_sync.py`: all 62 servers pass. `test_verify_version_sync.py`: 6 passed.
- aws-location 53 passed, stepfunctions-tool 61 passed.
- billing-cost-management: the same 43 tests fail on `main` without this PR.
- amazon-translate: `test_setup_logging_with_file` also fails on `main`. `test_cache_performance_optimization` is timing-based and failed 2 of 5 runs on the same code.

PR #4605, opened 2026-09-09
fix(cicd): make the awslabs namespace check fail the build

dihannahdi:
Fixes

## Summary

`scripts/verify_awslabs_init.py` is wired into the per-package gate at `.github/workflows/python.yml:134` and is meant to fail the build when a server's `awslabs/__init__.py` is not the namespace shim this repo requires. It is a `click` command that signals failure with `return 1`. Click's standalone mode discards the return value, so the process prints the mismatch and exits `0`.

Run against a deliberately corrupted package, the original script prints `✗ Mismatch: ...` and then exits `0`. The gate has never fired. `scripts/verify_package_name.py` in the same directory uses `sys.exit()` correctly, and this follows it.

### Changes

Two changes are needed, because switching the gate on as written would fail servers that are not actually wrong.

**1. Compare the parsed module, not the bytes.** The check compared the file byte-for-byte against a single-quoted `FILE_CONTENTS`. But a server that declares `[tool.ruff]` in its own `pyproject.toml` *replaces* the root `.ruff.toml` rather than merging with it, so a server that does not declare `quote-style` inherits ruff's double-quote default. ecs-mcp-server is exactly that case — `ruff format` inside that directory rewrites `'pkgutil'` to `"pkgutil"`, so byte equality with a single-quoted constant is unreachable there, and its double quotes were correct all along rather than sloppy. well-architected-security-mcp-server declares `quote-style = "double"` explicitly. Comparing `ast.dump(ast.parse(...))` accepts either quote style and any comment wording, while still rejecting a file that does anything other than extend the namespace path. The license header has its own check.

**2. Fix the one server that was genuinely broken.** aws-serverless-mcp-server's `awslabs/__init__.py` had no `__path__ = __import__('pkgutil').extend_path(__path__, __name__)` line at all — only a comment — which makes `awslabs` a regular package there rather than a namespace portion.

documentdb-mcp-server, mcp-lambda-handler and well-architected-security-mcp-server had no `awslabs/__init__.py` at all. These were verified to import correctly alongside pkgutil-style siblings in both `sys.path` orders, so nothing is broken today; they are brought into line with the other 58 servers because the shim is an idempotent no-op and their packaging config (`packages = ["awslabs"]`) is identical to their compliant peers. ecs-mcp-server and oracle-mcp-server are left untouched — the AST comparison accepts what they already have, and rewriting them would fight their own formatter.

### User experience

Before: a server could ship a non-conforming `awslabs/__init__.py` and CI stayed green, because the check that existed to catch it could not fail.

After: it fails. Concretely, with real hatchling wheels built from aws-serverless-mcp-server's previous content and placed first on a multi-root `sys.path` — editable installs, `PYTHONPATH` dev setups, stacked Lambda layers — importing a sibling `awslabs.*` server's subpackage raises `ModuleNotFoundError`. Reversing the path order hides it, which is what makes it worth a gate: it is order-dependent and silent. Users routinely install several `awslabs.*` MCP servers into one environment.

## Checklist

* [x] Changes are documented

**RFC issue number**:

Checklist:

### How this was verified

- Original script against a corrupted fixture: prints `✗ Mismatch`, exits `0`. Fixed script, same fixture: exits `1`.
- Fixed script still exits `1` for a file missing the `extend_path` call, and for an otherwise-compliant file with `import os` appended — so relaxing to an AST comparison did not relax what the gate actually enforces.
- All 62 servers pass after this change (6 failed before it).
- Each touched `awslabs/__init__.py` confirmed already-formatted under its own server's ruff configuration, so this does not trade one red gate for another.
- The `ModuleNotFoundError` was reproduced with wheels built by the same backend and `packages = ["awslabs"]` config the real servers use, not inferred. What is *not* independently reproduced: installing two actual published `awslabs.*` distributions from PyPI together.
- `pre-commit run` over the diff: no failures. `gitleaks` could not be installed here (Go module download failed) and `check-license-header`, `pyright` and `pytest` were skipped — the last two are `pre-push`-staged and do not run in CI.

## Acknowledgment


mikegc-aws:
Confirming the `src/mcp-lambda-handler/awslabs/__init__.py` portion of this PR from a package-owner perspective: that package is currently **missing** the `awslabs` namespace shim on `main` (the directory holds only `mcp_lambda_handler/`), so the added file — the standard `pkgutil.extend_path` PEP 420 shim, byte-identical to the other servers here — is correct and needed. Your root-cause writeup (Click's standalone mode discarding `return 1`, so the gate exits 0) matches what I see: the check has never actually fired, which is why this gap went unnoticed. No concerns on the mcp-lambda-handler change.

dihannahdi:
Thanks for checking `mcp-lambda-handler` from the package-owner side.

Two updates since your comment. The branch is force-pushed, because of the rebase.

1. Rebased onto `main` (`83ac913`). Re-ran the check over all 62 servers, all pass.
2. Two success paths in the script still used `return 0` — the same shape as the bug this PR fixes, so a later edit could turn one of them into `return 1` and the gate would go quiet again without anyone noticing. Every exit now goes through `sys.exit()`, the way `scripts/verify_package_name.py` already does.

The four `awslabs/__init__.py` files are unchanged from what you reviewed. Only `scripts/verify_awslabs_init.py` moved.

Exit codes as they stand now, run against real directories:

- compliant shim, single quotes → `0`
- compliant shim, double quotes → `0`
- no `awslabs/` directory at all (README-only server) → `0`
- `__init__.py` with no `extend_path` line → `1`
- compliant shim with `import os` appended → `1`
- `awslabs/` present but no `__init__.py` → `1`

PR #4615, opened 2026-09-10
fix(mysql-mcp-server): reject statement-leading mutating verbs in read-only mode

happycupofjava:
Fixes

A CWE-184 (Incomplete List of Disallowed Inputs) read-only-mode bypass in mysql-mcp-server: the fixed `MUTATING_KEYWORDS` anywhere-scan did not catch several statement-leading mutating verbs, so they slipped past the read-only gate on the RDS Data API path.

### Changes

Before evaluating a read-only query, the gate now also rejects statement-leading mutating verbs that the anywhere-scan missed:

* `DO expr` evaluates expressions purely for their side effects and returns no result set, so `DO GET_LOCK(...)` or `DO <writing_function>()` previously passed the read-only gate.
* Also now rejected: `IMPORT`, `START`, `BEGIN`, `COMMIT`, `ROLLBACK`, `SAVEPOINT`, `RELEASE`, `XA`, `CHANGE`, `PURGE`, `STOP`, `BINLOG`, `CLONE`, `RESTART`, `SHUTDOWN`, and bare `REPLACE ... SET`.
* These verbs are matched by a new `STATEMENT_START_MUTATING_KEYWORDS` set anchored to statement start (beginning of the comment-stripped SQL or immediately after a `;`) rather than added to `MUTATING_KEYWORDS`, because several are common English words or identifiers (`start`, `stop`, `change`, `release`, `do`) and `REPLACE` is a heavily used string function. A bare anywhere-match would falsely reject benign reads such as `SELECT start FROM schedule`, `SELECT REPLACE(col, 'a', 'b')`, or `WHERE note = 'things to do'`.
* Adds unit tests covering detection at statement start, after `;`, after a bare leading `;`, and case-insensitively, plus false-positive guards for the same words used as columns, aliases, string literals, and function calls. A full-coverage matrix is driven off the keyword set so a future addition is automatically exercised in both directions.

The durable control remains a least-privilege MySQL user / IAM role, per the README **Security model** section; this is defense-in-depth follow-up to the read-only bypass hardening first shipped in 1.0.22.

### User experience

Before: A read-only session could run `DO GET_LOCK(...)`, `COMMIT`, `START TRANSACTION`, `SHUTDOWN`, etc. because those verbs were not on the disallowed list, despite mutating server/transaction/replication state.

After: Any statement whose leading verb is a mutating verb is rejected by the read-only gate. Benign read queries that merely use those words as column names, aliases, string literals, or function calls (e.g. `SELECT REPLACE(name,'a','b')`, `SELECT start FROM schedule`) are unaffected.

## Checklist

* [x] Changes are documented

**RFC issue number**:

Checklist:

## Acknowledgment


manaswini1920 (dismissed):

src/mysql-mcp-server/awslabs/mysql_mcp_server/mutable_sql_detector.py
```diff
@@ -203,13 +251,18 @@ def detect_mutating_keywords(sql_text: str) -> list[str]:
...
+    (keywords anywhere) and ``STATEMENT_START_MUTATING_PATTERN`` (verbs that
+    are also common identifiers, matched only at statement start).
     """
     if re.search(MYSQL_CONDITIONAL_COMMENT_PATTERN, sql_text):
         # Defence in depth: keep this function correct in isolation, even
         # when callers do not also invoke check_sql_injection_risk.
         return ['MYSQL_CONDITIONAL_COMMENT']
     sql_for_check = sqlparse.format(sql_text, strip_comments=True)
     matches = MUTATING_PATTERN.findall(sql_for_check)
+    matches += STATEMENT_START_MUTATING_PATTERN.findall(sql_for_check)
```
kennthhz, line 355: **🔴 BLOCKER — `#`-line-comment defeats every new statement-start keyword**

This second scan runs on `sqlparse.format(strip_comments=True)` output, but sqlparse 0.6.0 does **not** strip a MySQL `#` comment unless a space follows the `#`. Since `STATEMENT_START_MUTATING_PATTERN` is anchored to `^`/`;`, a 2-char `#x\n` prefix hides the leading verb. These are single statements, so the stacked-queries rule does not mask them:

```
"#x\nDO GET_LOCK('x',60)"  -> mut=[]  ALLOWED   (this PR's flagship threat)
"#comment\nSHUTDOWN"       -> mut=[]  ALLOWED   (read-only session shuts down the server)
"#c\nSTART REPLICA"        -> mut=[]  ALLOWED
"# \nDO GET_LOCK('x',60)"  -> mut=['DO']        (only the space-after-# form is caught)
```

On the RDS Data API read-only path this function is the sole gate, so the CWE-184 bypass the PR targets is still open. Please strip `#[^\n]*` explicitly before the anchored scan (or use a MySQL-aware strip) and add regression tests for `#`, `#x`, `# `, and `#`-after-`;`. Note the docstring above (line ~238) also claims `# ...` comments are "treated as whitespace" — that claim is false for the no-space form and needs updating.

src/mysql-mcp-server/awslabs/mysql_mcp_server/mutable_sql_detector.py
```diff
@@ -95,6 +95,54 @@
...
+# server lifecycle. Several are ordinary words (``start``, ``stop``,
+# ``change``, ``release``, ``do``) and ``REPLACE`` is a string function, so a
+# bare ``\b`` anywhere-match would reject benign reads like
+# ``SELECT start FROM t`` or ``SELECT REPLACE(col, 'a', 'b')``. Anchoring to
+# statement start avoids that while still blocking the verb form.
+STATEMENT_START_MUTATING_KEYWORDS = {
+    # DML / expression execution
+    'IMPORT',  # IMPORT TABLE — bulk import from .ibd files
+    'REPLACE',  # bare REPLACE ... SET (REPLACE INTO already covered above)
+    'DO',  # DO expr — runs expressions / side-effecting stored functions
```
kennthhz, line 122: **🟠 MEDIUM — blocking `DO GET_LOCK` but not `SELECT GET_LOCK` is inconsistent**

The read-shaped twin of the exact threat this entry targets passes the gate. Verified PASS through both `detect_mutating_keywords` and `check_sql_injection_risk`: `SELECT GET_LOCK(...)`, `RELEASE_LOCK`, `RELEASE_ALL_LOCKS`, `MASTER_POS_WAIT`/`SOURCE_POS_WAIT`, `WAIT_FOR_EXECUTED_GTID_SET`, `WAIT_UNTIL_SQL_THREAD_AFTER_GTIDS`, `LAST_INSERT_ID(x)`, `sys_exec`/`sys_eval`. A text gate can't fully contain this, but either add the lock/replication-wait family to `SUSPICIOUS_PATTERNS` so `DO GET_LOCK` and `SELECT GET_LOCK` are symmetric, or scope `SELECT`-wrapped side-effecting functions explicitly as a least-privilege concern so the `DO` entry doesn't over-promise.

src/mysql-mcp-server/awslabs/mysql_mcp_server/mutable_sql_detector.py
```diff
@@ -95,6 +95,54 @@
...
+
+# Sorted longest-first for deterministic output (see MUTATING_PATTERN).
+_STATEMENT_START_KEYWORDS_BY_LENGTH = sorted(
+    STATEMENT_START_MUTATING_KEYWORDS, key=len, reverse=True
+)
+# Anchor to statement start: start of the (comment-stripped) SQL or right
+# after a ``;``. No re.MULTILINE, so a mid-statement newline is not a new
+# anchor. The single capturing group makes ``findall`` return the keyword.
+STATEMENT_START_MUTATING_PATTERN = re.compile(
+    r'(?i)(?:^|;)\s*('
```
kennthhz, line 166: **🟡 LOW — a `;` inside a string literal trips this anchor**

String literals aren't comment-stripped, so `WHERE note = 'do it; commit later'` -> `mut=['COMMIT']`. It's masked today by the stacked-queries rule (so no live false positive), but the "no false positive" guarantee is now coupled to that injection rule staying at least as strict, and the logged rejection reason is wrong. Consider noting the coupling here, or running this scan on a string-literal-stripped copy.

src/mysql-mcp-server/awslabs/mysql_mcp_server/mutable_sql_detector.py
```diff
@@ -95,6 +95,54 @@
...
+# (e.g. ``DO GET_LOCK(...)``); the rest control transactions, replication, or
+# server lifecycle. Several are ordinary words (``start``, ``stop``,
+# ``change``, ``release``, ``do``) and ``REPLACE`` is a string function, so a
+# bare ``\b`` anywhere-match would reject benign reads like
+# ``SELECT start FROM t`` or ``SELECT REPLACE(col, 'a', 'b')``. Anchoring to
+# statement start avoids that while still blocking the verb form.
+STATEMENT_START_MUTATING_KEYWORDS = {
+    # DML / expression execution
+    'IMPORT',  # IMPORT TABLE — bulk import from .ibd files
+    'REPLACE',  # bare REPLACE ... SET (REPLACE INTO already covered above)
```
kennthhz: **🟡 LOW — `REPLACE` double-reporting**

With `REPLACE INTO` in `MUTATING_KEYWORDS` and bare `REPLACE` here, both scans fire on `REPLACE INTO t VALUES(1)` -> `['REPLACE INTO','REPLACE']`. Harmless, but note the `test_..._sets_are_disjoint` test gives a false sense of clean partitioning: the sets are string-disjoint yet overlap in behavior on this token.

kennthhz (commented):
Detailed review of the read-only-gate hardening. Summary: **1 blocker** (`#`-line-comment bypass that still lets `#x\nDO GET_LOCK(...)` / `#x\nSHUTDOWN` through on the RDS Data API path, defeating the fix's own goal), **2 medium** (the `DO GET_LOCK` vs `SELECT GET_LOCK` asymmetry, and the missing comment/whitespace test dimension), plus low/nit items. Encoding/charset, string-concat, and the Postgres `U&"..."` identifier-escape vector were investigated and do NOT apply here. Inline comments below.

src/mysql-mcp-server/awslabs/mysql_mcp_server/mutable_sql_detector.py
```diff
@@ -95,6 +95,54 @@
...
+#
+# Unlike MUTATING_KEYWORDS above (matched anywhere), these are matched only
+# at statement start. ``DO expr`` returns no result set yet has side effects
+# (e.g. ``DO GET_LOCK(...)``); the rest control transactions, replication, or
+# server lifecycle. Several are ordinary words (``start``, ``stop``,
+# ``change``, ``release``, ``do``) and ``REPLACE`` is a string function, so a
+# bare ``\b`` anywhere-match would reject benign reads like
+# ``SELECT start FROM t`` or ``SELECT REPLACE(col, 'a', 'b')``. Anchoring to
+# statement start avoids that while still blocking the verb form.
+STATEMENT_START_MUTATING_KEYWORDS = {
```
kennthhz, line 115: **🟡 LOW — completeness & naming of this set**

1. **Completeness:** `USE <db>`, `CACHE INDEX`, and `LOAD INDEX INTO CACHE` are statement-leading state changers that still pass (`mut=[]`). Note the inconsistency that `SET` (session state) is blocked while `USE` (also session state) is not.
2. **Naming:** several members (`DO 1`, `BEGIN`, `COMMIT`) mutate no data — they are transaction-control / no-ops. "Mutating" overstates it; consider "non-read / state-affecting verbs" or a clarifying comment.

src/mysql-mcp-server/awslabs/mysql_mcp_server/mutable_sql_detector.py
```diff
@@ -95,6 +95,54 @@
...
+    'STOP',  # STOP REPLICA / STOP GROUP_REPLICATION
+    'BINLOG',  # BINLOG 'base64-event' — injects a raw binary log event
+    # Server administration
+    'CLONE',  # CLONE LOCAL / CLONE INSTANCE — copies the data directory
+    'RESTART',  # restarts the server process
+    'SHUTDOWN',  # terminates the server
+}
+
+# Sorted longest-first for deterministic output (see MUTATING_PATTERN).
+_STATEMENT_START_KEYWORDS_BY_LENGTH = sorted(
```
kennthhz, line 151: **🟢 NIT — longest-first sort is inert here**

No keyword in this set is a prefix of another, and a statement has exactly one leading verb, so the ordering never matters (unlike `MUTATING_PATTERN`). Harmless, but the "see MUTATING_PATTERN" rationale implies a correctness need that doesn't exist for the anchored set.

src/mysql-mcp-server/CHANGELOG.md
```diff
@@ -45,6 +45,25 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **CWE-184** — read-only mode now rejects statement-leading mutating verbs
```
kennthhz, line 48: **🟢 NIT — place under `### Security`, not `### Changed`**

The analogous 1.0.22 CWE-184 hardening is logged under `### Security`. This follow-up is also a CWE-184 security fix; put it under `### Security` for consistency.

src/mysql-mcp-server/CHANGELOG.md
```diff
@@ -45,6 +45,25 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  that the fixed `MUTATING_KEYWORDS` scan did not catch. `DO expr`
+  evaluates expressions purely for their side effects and
+  returns no result set, so `DO GET_LOCK(...)` or `DO <writing_function>()`
+  previously slipped past the read-only gate on the RDS Data API path. Also
+  now rejected: `IMPORT`, `START`, `BEGIN`, `COMMIT`, `ROLLBACK`,
+  `SAVEPOINT`, `RELEASE`, `XA`, `CHANGE`, `PURGE`, `STOP`, `BINLOG`,
+  `CLONE`, `RESTART`, `SHUTDOWN`, and bare `REPLACE ... SET`. These verbs
```
kennthhz: **🟢 NIT — wording undersells the behavior**

The code blocks *any* leading `REPLACE` (also `REPLACE ... VALUES`), not just `REPLACE ... SET`. The broader behavior is correct; the changelog wording is narrower than the implementation.

src/mysql-mcp-server/tests/test_mutable_sql_detector.py
```diff
@@ -875,3 +876,604 @@ def test_single_statement_with_commit_in_line_comment_is_benign(self):
...
+#
+# String literals are the critical false-positive surface here: unlike
+# comments, they are NOT stripped before the regex scan, so a naive bare
+# ``\b<kw>\b`` anywhere-match would reject benign reads like
+# ``WHERE note = 'things to do'``. The statement-start anchor is what makes
+# these safe, and this matrix guards that guarantee for the whole set.
+# ---------------------------------------------------------------------------
+
+
+class TestStatementStartKeywordFullCoverageMatrix:
```
kennthhz, line 1155: **🟠 MEDIUM (test gap behind the blocker) — matrix never varies comment/whitespace obfuscation**

This matrix is thorough on the *keyword* dimension but never varies the *comment/whitespace* dimension for a fixed keyword, which is why the `#`-comment bypass (see `mutable_sql_detector.py` line 265) isn't caught. Please add cases for a leading `#`/`#x`/`# ` comment and `#` after `;`, mirroring the existing `--` and `/* */` cases.

src/mysql-mcp-server/tests/test_mutable_sql_detector.py
```diff
@@ -875,3 +876,604 @@ def test_single_statement_with_commit_in_line_comment_is_benign(self):
...
+    """Every statement-leading keyword: detected as a verb, ignored as data."""
+
+    @pytest.mark.parametrize('keyword', sorted(STATEMENT_START_MUTATING_KEYWORDS))
+    def test_keyword_detected_at_statement_start(self, keyword):
+        """The keyword leading a statement is reported (no false negative)."""
+        assert keyword in detect_mutating_keywords(f'{keyword} some_expr()')
```
kennthhz: **🟡 LOW — synthetic payloads aren't valid MySQL**

`f'{keyword} some_expr()'` isn't valid MySQL for several verbs (e.g. `SHUTDOWN some_expr()`). Fine for exercising the regex, but it makes the matrix look like real-statement coverage when it isn't. Consider reusing `_STATEMENT_START_KEYWORD_PAYLOADS` here.

src/mysql-mcp-server/tests/test_mutable_sql_detector.py
```diff
@@ -875,3 +876,604 @@ def test_single_statement_with_commit_in_line_comment_is_benign(self):
...
+#   1. each keyword is detected when it leads a statement, and
+#   2. the same word is NOT flagged when it appears as a column name, an
+#      alias, or a function call inside a read-only SELECT.
+# ---------------------------------------------------------------------------
+
+
+# Minimal payload for every statement-leading keyword, each a real MySQL
+# statement whose leading verb is the keyword under test. Listed by hand so
+# adding a keyword without a payload fails the collection below.
+_STATEMENT_START_KEYWORD_PAYLOADS: dict[str, str] = {
```
kennthhz, line 898: **🟡 LOW — no server-level (run_query) test**

All tests exercise `detect_mutating_keywords` directly; none proves `run_query` actually rejects these end-to-end in read-only mode. Since the security guarantee lives at the server gate, please add at least one server-level test that a new keyword is rejected via the read-only path.

kennthhz (changes requested):

src/mysql-mcp-server/awslabs/mysql_mcp_server/mutable_sql_detector.py
happycupofjava: Fixed — added a shared #[^\n]* strip before both scans (sqlparse leaves the no-space #x\n form); added regression tests incl. server-level run_query.

src/mysql-mcp-server/awslabs/mysql_mcp_server/mutable_sql_detector.py
happycupofjava: Fixed — new SIDE_EFFECTING_FUNCTIONS set blocks GET_LOCK/RELEASE_LOCK/replication-waits/sys_exec/LAST_INSERT_ID(expr) in any form; no-arg LAST_INSERT_ID(), IS_FREE/USED_LOCK, and identifiers stay allowed.

src/mysql-mcp-server/awslabs/mysql_mcp_server/mutable_sql_detector.py
happycupofjava: Documented the coupling in a comment (harmless — the stacked-queries rule independently rejects it); regression test already pins it.

src/mysql-mcp-server/awslabs/mysql_mcp_server/mutable_sql_detector.py
happycupofjava: Clarified the disjoint-set test docstring: string-disjoint, not behaviour-disjoint; the REPLACE/REPLACE INTO overlap is intentional and deduped.

src/mysql-mcp-server/awslabs/mysql_mcp_server/mutable_sql_detector.py
happycupofjava: Added USE/CACHE INDEX/LOAD INDEX INTO CACHE (USE INDEX hint stays allowed); reworded the set to "non-read / state-affecting verbs".

src/mysql-mcp-server/CHANGELOG.md
happycupofjava: Done — moved under ### Security.

src/mysql-mcp-server/CHANGELOG.md
happycupofjava: Fixed — changelog now reads "any statement-leading REPLACE (e.g. REPLACE ... SET, REPLACE ... VALUES)" instead of just REPLACE ... SET.

src/mysql-mcp-server/tests/test_mutable_sql_detector.py
happycupofjava: Fixed — added TestLeadingCommentObfuscationIsStripped, crossing #/#x/# /--//* *//;-prefix forms with several leading verbs (incl. #-after-;), mirroring the existing comment cases.

src/mysql-mcp-server/tests/test_mutable_sql_detector.py
happycupofjava: Fixed — the statement-start and lowercase matrix tests now reuse _STATEMENT_START_KEYWORD_PAYLOADS (real MySQL statements) instead of f'{keyword} some_expr()'.

src/mysql-mcp-server/tests/test_mutable_sql_detector.py
happycupofjava: Added run_query read-only tests asserting rejection + execute_query never called (incl. #-comment forms).

src/mysql-mcp-server/awslabs/mysql_mcp_server/mutable_sql_detector.py
happycupofjava: Agreed — reworded the comment to say the sort is only for stable, hash-independent output, not correctness (no entry is a prefix of another and a statement has one leading verb).

src/mysql-mcp-server/awslabs/mysql_mcp_server/mutable_sql_detector.py
```diff
@@ -184,15 +297,40 @@
...
+def _strip_comments_for_scan(sql: str) -> str:
+    """Normalise SQL for keyword/pattern scanning by removing comments.
+
+    ``sqlparse.format(strip_comments=True)`` removes ``-- ...``, ``/* ... */``
+    and the space-prefixed ``# ...`` form while respecting string literals.
+    It does NOT remove the no-space MySQL ``#`` line comment, so any residual
+    ``#`` comment is stripped afterwards (see MYSQL_HASH_COMMENT_PATTERN).
+    """
+    stripped = sqlparse.format(sql, strip_comments=True)
+    return MYSQL_HASH_COMMENT_PATTERN.sub('', stripped)
```
kennthhz: **🔴 BLOCKER (new regression introduced by this fix) — the `#` strip is not string-literal aware**

`MYSQL_HASH_COMMENT_PATTERN = re.compile(r'#[^\n]*')` (line 307) runs *after* `sqlparse.format(...)` and blindly strips from **any** `#` — including one inside a string literal or backtick identifier — to end of line. Because MySQL treats `#` inside a quoted string as literal text (not a comment), this truncates the scanned SQL and hides everything after it on that line.

This regresses controls that worked correctly at the previous commit (`b291ca9f`):

```
                                                     b291ca9f        this commit
SELECT id FROM t WHERE c = '#foo'; DROP TABLE t      DROP, inj=Y  →  mut=[], inj=n   BYPASS
SELECT '#' ; DROP TABLE users                        DROP, inj=Y  →  mut=[], inj=n   BYPASS
SELECT id FROM t WHERE c = '#x' OR SLEEP(5)          inj=Y        →  mut=[], inj=n   BYPASS
SELECT `a#b`, GET_LOCK('x',1)                        (n/a)        →  mut=[], inj=n   BYPASS
```

The gate ends up scanning only `"SELECT id FROM t WHERE c = '"` — the trailing `; DROP TABLE t` is gone. This defeats not just the new statement-start scan but also the pre-existing **anywhere keyword scan** (`DROP`/`DELETE`), the **stacked-queries** rule, **UNION-injection** detection, and the new **side-effecting-function** rule.

**Severity:** this is broader than the original `#x\n` bug it closes. It triggers on any single-line query containing `#` inside a string literal — ordinary data like `'#sale'`, hex colors `'#ffffff'`, or hashtags — all of which execute fully in MySQL. On the RDS Data API path `detect_mutating_keywords` is the sole gate. (Boundary: `#[^\n]*` stops at `\n`, so a mutation on a *new* line survives — but LLM/Data-API queries are typically single-line, which the caller controls.)

**Fix:** make the `#` strip string-literal aware — only remove `#…EOL` when the `#` is not inside a `'…'` / `"…"` / `` `…` `` token (tokenize, or scan skipping quoted regions).

**Test gap:** `test_hash_inside_string_literal_is_not_a_false_positive` only checks a benign `'#sale'` with no trailing statement, giving false confidence. Please add a false-*negative* regression: `SELECT '#foo'; DROP TABLE t` must still report `DROP` (and `check_sql_injection_risk` must still flag the stacked `;`).

kennthhz (commented):
Round 2 review. Medium (side-effecting functions) is correctly fixed and symmetric. However the blocker fix introduced a NEW, broader regression: the `#`-comment strip is not string-literal aware and hides content after a `#` inside a string literal, regressing DROP / stacked-query / UNION detection that worked at b291ca9f. Details inline. Not safe to merge as-is.

src/mysql-mcp-server/awslabs/mysql_mcp_server/mutable_sql_detector.py
happycupofjava: Fixed — replaced the blind #[^\n]* strip with a quote-aware scanner that only strips #…EOL outside '…'/"…"/`…` regions (honoring backslash + doubled-quote escapes), so '#foo'; DROP TABLE t is detected again while '#sale' data is preserved and the #x\n bypass stays closed. Added regression tests.

kennthhz (approved):

manaswini1920 (approved):

PR #4582, opened 2026-09-03
fix(timestream-influxdb-mcp): gate writes on operator ALLOW_WRITE env

rajeiyer:
## Summary

Replaces the caller-supplied `tool_write_mode` parameter with an operator-controlled `ALLOW_WRITE` environment gate across all 13 mutating tools in the Timestream for InfluxDB MCP server.

Previously, write authorization was a per-call tool input (`tool_write_mode: bool = Field(False)`). Because it was part of each tool's input schema, a caller — or a prompt-injected agent — could set `tool_write_mode=True` in the same call and execute destructive operations (`DeleteDbCluster`, `DeleteDbInstance`, `InfluxDBWritePoints`, etc.). This is the MCP "Tool Overreach" threat (Threat #8) / SBP-02 write-mode best practice: mutations must be gated by operator configuration, not caller input. Sibling servers in this repo already use an operator `ALLOW_WRITE` gate; this change brings this server in line.

## Changes

- **Add operator gate:** `ALLOW_WRITE` env var, evaluated once at startup (`os.environ.get('ALLOW_WRITE','').lower() in ('true','1','yes')`). Not a tool parameter.
- **Remove** the `OPTIONAL_FIELD_TOOL_WRITE_MODE` field definition.
- **Gate all 13 mutating tools** on `ALLOW_WRITE` and remove `tool_write_mode` from their signatures/input schemas: `CreateDbCluster`, `CreateDbInstance`, `DeleteDbInstance`, `DeleteDbCluster`, `TagResource`, `UntagResource`, `UpdateDbCluster`, `UpdateDbInstance`, `CreateDbParamGroup`, `InfluxDBWritePoints`, `InfluxDBWriteLP`, `InfluxDBCreateBucket`, `InfluxDBCreateOrg`.
- **Flux-query writes unchanged:** `InfluxDBQuery` write operations remain governed by the existing operator env `INFLUXDB_WRITE_MODE` (already operator-controlled; out of scope for this issue).
- **Tests:** added `tests/conftest.py` (autouse fixture enabling the gate for success-path tests), updated `tests/test_server.py` (removed `tool_write_mode` kwargs; refusal tests now disable the gate and assert the new message), and added `tests/test_write_gate.py` (regression tests: `tool_write_mode` absent from every mutating tool signature; env parsing; mutation refused when the gate is off).
- **Docs:** `README.md` env-var table + config examples + Write Mode sections; `CHANGELOG.md` Security + BREAKING entries.

## Testing

[...]

PR #4623, opened 2026-09-11
fix(ecs-mcp-server): apply sensitive data redaction consistently

matthewgoodman13:
Fixes N/A

### Changes

Apply the `ALLOW_SENSITIVE_DATA` redaction consistently across the tools that return container
configuration.

The redaction of `environment[].value` and `secrets[].valueFrom` now lives in shared helpers in
`utils/security.py` (`redact_task_definition()`, `redact_container_definition()`) and is applied:

* in `troubleshooting_tools/utils.py::find_task_definitions`, the single point where the
  troubleshooting actions obtain task definitions
* in `api/resource_management.py` for `DescribeExpressGatewayService`, next to the existing
  `DescribeTaskDefinition` and `DescribeTasks` handling, which now use the same helper

Variable and secret names are kept; only values and ARNs are replaced with `[REDACTED]`. The tool
description for `get_ecs_troubleshooting_guidance` and the README `ALLOW_SENSITIVE_DATA` section
say what is redacted.

### User experience

With `ALLOW_SENSITIVE_DATA` unset, `get_ecs_troubleshooting_guidance` and
`DescribeExpressGatewayService` now return environment values and secret references the same way
`DescribeTaskDefinition` does:

```json
"environment": [{"name": "DB_PASSWORD", "value": "[REDACTED]"}],
"secrets": [{"name": "API_KEY", "valueFrom": "[REDACTED]"}]
```

Everything else in the responses is unchanged. Setting `ALLOW_SENSITIVE_DATA=true` returns the
full data, as before.

### Testing

825 tests pass, up from 787 on `main`. New tests cover the helpers, `find_task_definitions` on all
four lookup modes with the flag unset, `false` and `true`, `get_ecs_troubleshooting_guidance` end to
end, and `DescribeExpressGatewayService` through `ecs_api_operation`. Also verified against a live
ECS cluster through the server's stdio transport in both flag states. `ruff check`, `ruff format`,
`pyright`, `bandit` and `uv build` are clean.

## Checklist

* [x] Changes are documented

With `ALLOW_SENSITIVE_DATA` unset, two more responses now carry `[REDACTED]` for environment values
and secret ARNs, matching `DescribeTaskDefinition`. No other input or output is affected.

**RFC issue number**: N/A

Checklist:

## Acknowledgment

### DIFF
--- a/src/ecs-mcp-server/README.md
+++ b/src/ecs-mcp-server/README.md
@@ -270,7 +270,7 @@ Controls whether write operations (creating or deleting infrastructure) are allo
 
 ### ALLOW_SENSITIVE_DATA
 
-Controls whether tools that return logs and detailed resource information are allowed.
+Controls whether tools that return logs and detailed resource information are allowed. When disabled, `ecs_resource_management` and `ecs_troubleshooting_tool` redact container environment variable values and secret references (in task definitions, tasks, and Express Mode services) instead of returning them.
 
 ```bash
 # Enable access to sensitive data
--- a/src/ecs-mcp-server/awslabs/ecs_mcp_server/api/resource_management.py
+++ b/src/ecs-mcp-server/awslabs/ecs_mcp_server/api/resource_management.py
@@ -25,12 +25,18 @@ import re
 from typing import Any, Dict, Set
 
 from awslabs.ecs_mcp_server.utils.aws import get_aws_client
+from awslabs.ecs_mcp_server.utils.security import (
+    REDACTED,
+    redact_container_definition,
+    redact_task_definition,
+)
 
 logger = logging.getLogger(__name__)
 
 # Operations that return sensitive data (environment variables, secrets, etc.)
 # These require ALLOW_SENSITIVE_DATA=true or their responses will be sanitized.
 SENSITIVE_DATA_OPERATIONS: Set[str] = {
+    "DescribeExpressGatewayService",
     "DescribeTaskDefinition",
     "DescribeTasks",
 }
@@ -135,6 +141,8 @@ def _sanitize_sensitive_response(response: Dict[str, Any], api_operation: str) -
     For DescribeTaskDefinition: redacts containerDefinitions[].environment values
     and containerDefinitions[].secrets.
     For DescribeTasks: redacts containers[].environment values and overrides.
+    For DescribeExpressGatewayService: redacts the environment values and secrets of
+    every active configuration's primaryContainer.
 
     Args:
         response: The raw API response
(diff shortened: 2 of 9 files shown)


nineonine (dismissed):

scottschreckengaust (dismissed):
Thanks for consolidating redaction into shared helpers — extending coverage to `DescribeExpressGatewayService` and the troubleshooting path is a nice improvement, and the `find_task_definitions` control-flow refactor (early-return → `elif`-assign) correctly preserves the original branch precedence. My main concern is a completeness gap in what is fundamentally a security PR.

## High severity — incomplete secret redaction

`_redact_container_in_place` (`utils/security.py:189`) masks only `environment[].value` and `secrets[].valueFrom`, but ECS container definitions carry secret ARNs in other fields the helper never touches:

- **`logConfiguration.secretOptions[].valueFrom`** — e.g. a log-driver token stored as a Secrets Manager ARN
- **`repositoryCredentials.credentialsParameter`** — Secrets Manager ARN for private registry auth
- **`environmentFiles[].value`** — S3 ARN pointing at injected env vars (the source location leaks even though the values are never masked)

These are returned in cleartext when `ALLOW_SENSITIVE_DATA` is disabled, which contradicts the PR's stated goal of redacting secret references. Since this PR re-establishes `_redact_container_in_place` as the canonical redaction mechanism, extending it to cover these fields is in scope.

## Medium severity — coverage gaps

- **`api/resource_management.py:38`** — `SENSITIVE_DATA_OPERATIONS` omits write operations that echo full task/container definitions (`RegisterTaskDefinition`, `DeregisterTaskDefinition`, `DeleteTaskDefinitions`, `Create`/`UpdateExpressGatewayService`). With `ALLOW_WRITE=true` and `ALLOW_SENSITIVE_DATA=false`, their responses return `environment[].value` / `secrets[].valueFrom` unredacted.
- **`api/resource_management.py:171`** — `DescribeExpressGatewayService` redaction only walks `service.activeConfigurations[].primaryContainer`. Sidecar/additional containers within a configuration, and non-active configuration collections (e.g. pending/target configuration during a rollout), are left unredacted.

## Low severity — cleanup

- **Redundant deep copies** (`resource_management.py:154`): `_sanitize_sensitive_response` already deep-copies the whole response, then `redact_task_definition`/`redact_container_definition` deep-copy the same subtree again. Using the in-place helper on the already-copied structure avoids the second copy.
- **Duplicated redaction logic** (`resource_management.py:160`): the `DescribeTasks` branch re-implements env-value redaction inline instead of reusing `_redact_container_in_place`, so the rule now lives in three places and will drift.
- **README overstatement** (`README.md:273`): states secret references are redacted "in task definitions, tasks, and Express Mode services", but the `DescribeTasks` sanitizer only redacts environment values and touches no secret references.

## Scope note

The `secretOptions` / `repositoryCredentials` / write-op gaps are pre-existing, but this PR both re-establishes the redaction helper as the canonical mechanism and curates the operations set, so they fall within the "apply redaction consistently" objective. If you'd prefer to land the refactor now and address the field/operation coverage in a fast follow-up, that's reasonable too — just flag it so the gap is tracked rather than silently inherited.

matthewgoodman13:
The `Precommit .` failure is inherited from `main`, not from this PR.

The job runs `pre-commit run --all-files` on the merge commit. Both failing hooks act on one file this PR does not touch, `src/timestream-for-influxdb-mcp-server/tests/test_server.py`: `ruff-format` (v0.14.10, as pinned in `.pre-commit-config.yaml`) reformats it, and the resulting line shift changes a `line_number` in `.secrets.baseline`, which is why `detect-secrets` also reports the baseline as modified. That file was last changed in #4582 (`720fa51e`). Reproducible on `origin/main` alone: `ruff@0.14.10 format --check` reports "Would reformat" for it, and the `pre-commit` workflow on `main` has failed the same way since that merge ([run 35125732419](https://github.com/awslabs/mcp/actions/runs/35125732419), [run 35127855936](https://github.com/awslabs/mcp/actions/runs/35127855936)); the run before it ([35087414134](https://github.com/awslabs/mcp/actions/runs/35087414134)) passed.

The job log contains no `ecs-mcp-server` entries; the 11 files in this PR pass the same hooks locally. The check should clear once that file is formatted on `main`.

nineonine (approved):

arnewouters (approved):

PR #4632, opened 2026-09-15
feat(aws-healthomics-mcp-server): add pagination continuation hints

markjschreiber:
## Summary

Agents routinely miss a bare nextToken/continuation_token field and answer as if one page were the complete result set. Add a paginating() wrapper in utils/pagination.py that is applied at the single tool-registration choke point in server.py, so every tool automatically gets an additive "pagination" status block (isComplete, returnedCount, nextToken, instruction) on its response, whether the page is complete or not.

### Changes

Handles the three real idioms found by inspection: the ~17 tools with a bare camelCase nextToken, the two ECR tools whose next_token key comes from a Pydantic model_dump() and is always present, and the genomics search tool's existing nested pagination.has_more block (merged into, not replacing, its pre-existing fields). Explicitly excludes the CloudWatch-log tools and the metrics list tool, which accept a next_token parameter but never expose a client-facing continuation token, verified by reading their handlers directly.

## Checklist

* [x] Changes are documented

N

**RFC issue number**:

Checklist:

## Acknowledgment


WIIASD (approved):

a-li (approved):

alxawan (commented):
LGTM

PR #4636, opened 2026-09-16
fix(bedrock-kb-retrieval): report validation failures instead of failing silently

ankdav-amazon:
Fixes

## Summary

Two ways this server fails without telling the caller why. Both are follow-ups to #4574.

1. **Validation messages never reach the client.** `MCPServer` forwards a `ToolError`'s text but
   replaces any other exception with a bare `Error executing tool <name>`. These checks raised
   `ValueError`, so their guidance stayed in the server's own logs.
2. **An unknown `data_source_ids` value silently matches nothing.** Filtering on a data source id
   that does not belong to the knowledge base is accepted by the API and matches nothing, so the
   caller gets an empty result set and no error.

Both produce the same bad outcome: a call that cannot succeed, and no way for the caller to work
out why.

### 1. Validation failures were masked

In `mcp/server/mcpserver/tools/base.py`, `except (ToolError, ResourceError)` re-raises with the
message appended, while `except Exception` deliberately does not ("the exception's own text stays
on the server").

Pointing `AgenticQueryKnowledgeBases` at a vector knowledge base, before:

```
Error executing tool AgenticQueryKnowledgeBases
```

and after:

```
Error executing tool AgenticQueryKnowledgeBases: Knowledge base <id> is not a managed
knowledge base. Agentic retrieval supports managed knowledge bases only -- use the
QueryKnowledgeBases tool for vector knowledge bases.
```

That message exists so a client can recover by switching tools, which it cannot do if the text is
dropped. Same for the reranking check, which now reports where the model is available:

```
Error executing tool QueryKnowledgeBases: The 'AMAZON' reranking model is not available in
region us-east-1. Supported regions: ['ap-northeast-1', 'ca-central-1', 'eu-central-1', 'us-west-2']
```

- Raise `ToolError` for the deliberate validation failures: knowledge base is not managed, empty
  knowledge base list, unavailable `(region, model)` reranking pair.
- `AgenticRetrievalError` subclasses `ToolError`, so modelled errors inside the
  `AgenticRetrieveStream` event stream are reported rather than masked.

Unexpected exceptions are left as they are. The framework masking those is correct, since their
text is not written for an end user; this only changes the errors the server raises on purpose.

### 2. Unknown data source ids returned nothing, quietly

An empty result set reads as "this data source is empty" rather than "this id is wrong", which is
the more likely explanation — a stale id, a typo, or an id from a different knowledge base. It is
the same silent failure mode as sending the wrong metadata key, which #4574 fixed.

Both tools now verify `data_source_ids` up front and name the valid ids:

```
Error executing tool QueryKnowledgeBases: Data source DOESNOTEXIST does not belong to knowledge
base <id>. Filtering on it would silently return no results. Valid data source ids: ['<a>',
'<b>', '<c>']. Use the ListKnowledgeBases tool to see each knowledge base and its data sources.
```

- **Skipped when it cannot be verified.** If the ids cannot be listed — for example a caller
  without `bedrock:ListDataSources` — validation is skipped rather than guessing, so no one is
  blocked by a permission they did not previously need.
- **No extra API call on the common path.** `discover_knowledge_bases` already lists data sources,
  so it populates the cache; `ListKnowledgeBases` then `QueryKnowledgeBases` costs nothing more.
- **The agentic path validates against the union across its retrievers**, since an id only has to
  belong to one of them. Verified live that a multi-knowledge-base call filtered to a data source
  present in only one is accepted.

### User experience

| Case | Before | After |
|---|---|---|
| Agentic tool on a vector KB | `Error executing tool AgenticQueryKnowledgeBases` | names the problem and the tool to use instead |
| Reranking in a region without that model | generic error | names the supported regions |
| Mistyped or stale `data_source_ids` | **0 results, no error** | names the invalid id and lists the valid ones |
| Modelled error inside the agentic stream | generic error | the service's own message |

## Testing

83 unit tests pass; `pyright`, `ruff check` and `ruff format` clean.

Verified end to end over MCP JSON-RPC against live managed and vector knowledge bases — every
message quoted above is copied from a real client response, not composed by hand:

- Non-managed knowledge base rejection, empty knowledge base list, and the reranking region check
  (server in `us-east-1`, where `amazon.rerank-v1:0` is not offered).
- Unknown data source id rejected on **both** tools; **valid ids unaffected** on both (5 and 10
  results respectively); multi-knowledge-base call scoped to a data source in only one of them
  accepted (20 results); vector knowledge base also validated.

One note on the tests: the management-client stubs previously left `list_data_sources` to
`MagicMock`, which reports no data sources. That is not what any live knowledge base looks like and
would have made validation reject valid ids, so the stubs now return data sources.

## Checklist

* [x] Changes are documented

No API, parameter or return-shape change. `AgenticRetrievalError` keeps its name and is still
catchable as before, now also as a `ToolError`.

One behaviour change worth calling out explicitly: a request with an unknown `data_source_ids`
value now fails instead of returning an empty result set. That request could not have returned
results, so nothing that previously worked stops working — but a caller relying on the empty
response rather than an error would now see an error. That seemed clearly preferable to leaving it
silent, and it is skipped entirely when the ids cannot be verified.

**RFC issue number**: n/a

Checklist:

## Notes for reviewers

- `ToolError` is imported from `mcp.server.mcpserver.exceptions`, matching this package's
  dependency. Most other servers in the repo use `fastmcp.exceptions`, which is not importable here.
- Happy to extend the same treatment to the pre-existing error paths in this server if you would
  like it in one change; I kept this scoped to the checks whose purpose is the message.
- If you would rather the data source check warn instead of fail, that is a small change — but a
  warning is easy for a model to skip past, and the failure it prevents is one where the caller
  concludes a data source is empty.

## Acknowledgment


ankdav-amazon:
Superseded by #4638, which combines this with my other open change into a single pull request. Closing to keep one review thread.

Nothing is dropped — #4638 contains every commit from here, verified identical in content.

PR #4637, opened 2026-09-16
chore: fix formatting drift breaking the Precommit check on main

ankdav-amazon:
Fixes

## Summary

`Precommit .` currently fails on `main`, and therefore on every open pull request. This is a
two-line fix to unblock it.

`src/timestream-for-influxdb-mcp-server/tests/test_server.py` is not formatted to the repo's own
ruff configuration. The pre-commit `ruff-format` hook rewrites it and the run fails with
`files were modified by this hook` — `1 file reformatted, 871 files left unchanged`. `detect-secrets`
then fails as a knock-on effect, because reformatting shifts the line number of an existing baseline
entry and the baseline no longer matches.

This is not a ruff version disagreement. Checked against `main`'s own copy of the file, independent
of any branch: both the ruff pinned in `.pre-commit-config.yaml` (`v0.14.10`) and the older one the
package itself pins (`0.11.2`) agree it needs reformatting.

### Changes

- `ruff format` on that one file. **Formatting only** — the changes collapse multi-line `assert`
  statements that now fit on a single line. No test logic is altered, and the package's **169 tests
  still pass**.
- Refreshed `.secrets.baseline` for the shifted line number (`621` → `612`). Only that line number
  and the `generated_at` timestamp change; the entry count is unchanged at **65 files / 126 entries**.

### User experience

No functional change. `Precommit .` goes green again, unblocking open pull requests.

## Testing

- `uvx ruff@0.14.10 format --check` clean on the file afterwards.
- Full pre-commit suite passes on the affected files.
- `uv run --frozen pytest tests/` in `timestream-for-influxdb-mcp-server`: 169 passed.

## Checklist

* [ ] Changes are documented

**RFC issue number**: n/a

Checklist:

## Notes for reviewers

I don't own this package — raising it because it blocks all pull requests, including my own
(#4636). Happy to close if a package owner would rather fix it themselves. Kept deliberately
separate from my feature work so it can be reviewed on its own and merge quickly.

Worth considering separately: the package pins `ruff>=0.11.2` while `.pre-commit-config.yaml` pins
`v0.14.10`, so a package-local `ruff format` can disagree with the repo-wide hook. That did not
cause this particular failure, but it makes the class of failure easy to reintroduce.

## Acknowledgment


ankdav-amazon:
Superseded by #4638, which combines this with my other open change into a single pull request. Closing to keep one review thread.

Nothing is dropped — #4638 contains every commit from here, verified identical in content.

PR #4639, opened 2026-09-16
fix(aws-healthomics-mcp-server): report truthful pagination state for date-filtered ListRuns

markjschreiber:
Fixes

## Summary

When created_after/created_before is supplied, list_runs filters runs
client-side after fetching raw batches from the HealthOmics API. If the
filtered set exceeded max_results, the response was truncated and
nextToken was always omitted, so callers had no signal that more
matching runs existed and would treat a partial page as complete.

### Changes

Re-emit the upstream nextToken on this path when one is available,
populating the same optional response key already used on the
unfiltered path with the same kind of value (the raw upstream
continuation token). This does not change what nextToken means for
existing callers.

### User experience

Known residual limitation, documented in code: resuming with this
token can skip matches that were already fetched into the truncated
batch but excluded by the max_results slice, since continuation only
resumes from unfetched upstream pages. This is a strict improvement
over the prior behavior (which always silently claimed completeness)
and is called out for follow-up rather than solved here, since a fully
correct cursor would require changing nextToken from a verbatim
upstream token into a synthetic cursor -- a reserved MCP-contract
change.

## Checklist

* [x] Changes are documented

N

**RFC issue number**:

Checklist:

## Acknowledgment


a-li (approved):
LGTM

alxawan (approved):
LGTM

PR #3094, opened 2026-04-14
feat(aws-dataprocessing-mcp): add support for --allow-sensitive-data-access mcp flag

naikvaib:
### Changes

Implemented enforcement for the `--allow-sensitive-data-access` CLI flag that was previously parsed and stored but never checked in any conditional. This security vulnerability allowed unauthorized access to sensitive customer data including database passwords, query results, and job execution outputs regardless of the flag's value.

**Security Fixes**:
- Added 8 enforcement points across 5 handler files to protect sensitive operations
- Fixed critical password exposure by forcing `hide_password=True` when flag is disabled
- Blocked query result retrieval operations that return customer data
- Protected job run details containing sensitive arguments and error messages

**Operations Protected**:

*CRITICAL - Database Credentials*:
- `get-connection` and `list-connections`: Enforces `hide_password=True` to prevent plaintext password exposure

*HIGH - Customer Data*:
- `get-query-results` (Athena): Blocks customer query result data
- `get-statement` (Glue Interactive Sessions): Blocks execution outputs with customer data
- `get-entity-records` (Data Catalog): Blocks preview data from connected sources

*MEDIUM - Job Outputs*:
- `get-job-run` (Glue ETL & EMR Serverless): Blocks job details with sensitive arguments/errors
- `describe-step` (EMR EC2): Blocks step configurations with potential credentials

### User experience

**Before this change**:
- README documented: "Access to sensitive data is restricted by default"
- Reality: `--allow-sensitive-data-access` flag had zero effect
- Users could retrieve plaintext database passwords by passing `hide_password=False`
- Query results, session outputs, and job logs were always accessible
- False security promise created compliance risk

**After this change**:
- **Default behavior (secure)**: Without `--allow-sensitive-data-access` flag:
  - `get-connection()` automatically hides passwords even if user passes `hide_password=False`
  - `get-query-results()` returns error: "requires --allow-sensitive-data-access flag"
  - `get-statement()` returns error: "requires --allow-sensitive-data-access flag"
  - `get-entity-records()` returns error: "requires --allow-sensitive-data-access flag"
  - Job run and step detail operations return similar errors

- **With flag enabled**: All operations work as originally intended:
  ```json
  {
    "args": [
      "awslabs.aws-dataprocessing-mcp-server@latest",
      "--allow-sensitive-data-access"
    ]
  }
  ```

- **Metadata operations unaffected**: List operations (e.g., `list-query-executions`, `list-statements`) continue to work without the flag as they only return IDs/status, not actual data

[...]

PR #4640, opened 2026-09-17
fix(aws-dataprocessing-mcp-server): apply --allow-sensitive-data-access per field on read operations

DDD-Joo:
## Summary

`--allow-sensitive-data-access` was applied inconsistently across operations that return records mixing operational metadata with customer content. The singular form of an operation checked the flag and refused the call; the list form returned the same records unfiltered.

This makes both forms behave the same way, and stops refusing either of them. Without the flag the operation succeeds and the fields that can carry customer content are omitted from each record.

Thanks @LiyuanLD for the review — this is the field-level approach you suggested rather than the blocking approach the first revision took.

### Changes

Applied to both forms of each operation:

| Tool | Operations | Omitted without the flag |
| --- | --- | --- |
| `manage_aws_emr_ec2_steps` | `describe-step`, `list-steps` | `Config.Args`, `Config.Properties`, `Status.StateChangeReason.Message`, `Status.FailureDetails` |
| `manage_aws_glue_jobs` | `get-job-run`, `get-job-runs` | `Arguments`, `ErrorMessage`, `StateDetail` |
| `manage_aws_glue_statements` | `get-statement`, `list-statements` | `Code`, `Output.Data`, `Output.ErrorValue`, `Output.Traceback` |
| `manage_aws_emr_serverless_job_runs` | `get-job-run`, `list-job-runs` | `jobDriver`, `configurationOverrides`, `tags`, `stateDetails` |

Filtering is allowlist based rather than a list of fields to strip. Each record type declares the fields it returns and anything not named is dropped, so a field added by a future API version is omitted by default rather than returned. The handlers currently pass raw AWS records straight through as `Dict[str, Any]`, so there is no schema that would otherwise flag a new field appearing.

Athena `get-query-results` and Glue `get-entity-records` still refuse without the flag. Their entire response is customer content, so filtering would return a successful but empty result, which is less useful than a message naming the flag. That leaves the flag with two behaviours, which is now stated explicitly in the README rather than left implicit.

New `utils/sensitive_data_filter.py` holds the allowlists and the filtering helpers. Handler docstrings, the `operation` field descriptions and the README are updated, since none of them previously described what the flag does for these tools.

### User experience

Before, with the server started without `--allow-sensitive-data-access`: `describe-step` was refused, while `list-steps` returned full step records including `Config.Args`.

After, for the same setup, both return:

```json
{ "Id": "s-3KEZ7RJVBHQ2M", "Name": "nightly-customer-etl",
  "Config": { "Jar": "s3://bucket/etl.jar", "MainClass": "com.acme.etl.NightlyLoad" },
  "ActionOnFailure": "CONTINUE",
  "Status": { "State": "FAILED", "StateChangeReason": { "Code": "STEP_FAILED" },
              "Timeline": { "StartDateTime": "...", "EndDateTime": "..." } } }
```

along with a message naming what was left out:

> Note: describe-step omitted the following fields because --allow-sensitive-data-access is not enabled: Config.Args, Config.Properties, Status.StateChangeReason.Message, Status.FailureDetails. Enable the flag to receive them.

So the flag now controls how much of each record is returned rather than whether the operation may be called. Callers can still enumerate steps and see which ones failed; the failure detail requires the flag. With the flag enabled, responses are unchanged.

Naming the omitted fields is a small addition beyond the review comment. Without it a caller cannot distinguish a field that was empty from one that was left out, and there is no hint that a flag governs it.

## Checklist

* [x] Changes are documented

Y, and more so than the first revision, so worth calling out. Four operations that previously returned an error without the flag (`describe-step`, Glue `get-job-run`, `get-statement`, EMR Serverless `get-job-run`) now return a filtered result instead, which changes behaviour shipped in #3094. Callers relying on the error are affected, and callers that were reading the omitted fields with the flag off will no longer see them. Nothing needs migrating beyond enabling the flag where those fields are wanted, but I would rather defer to you on whether this belongs in a release note, and on the field lists themselves — those are a judgement call and easy to adjust.

**RFC issue number**:

Checklist:

## Testing

* `tests/test_sensitive_data_access.py` covers all eight operations, asserting the omitted fields are absent, the retained fields are present, and the response names what was left out. EMR steps share one fixture between `describe-step` and `list-steps` so both forms are asserted against identical input.
* `uv run pytest`: 1151 passed, no other tests affected.
* `pre-commit run` passes over the full change, and `pyright` reports no errors on the changed modules.
* Verified out of band against the AWS APIs that each operation returns the filtered record with the flag off and the full record with it on.

## Notes for reviewers

Two judgement calls worth your attention:

`get-statement` and `list-statements` are included, but filtering leaves state and progress only, which answers whether a statement finished and not what it returned. A refusal may serve callers better there. Happy to drop that pair from this change.

`Name` and `Jar` are retained but are customer-controlled strings, so an allowlist is a reviewed decision about which fields are worth returning rather than a guarantee about their contents. This is noted in the module docstring rather than left implicit.

## Acknowledgment


src/aws-dataprocessing-mcp-server/awslabs/aws_dataprocessing_mcp_server/handlers/emr/emr_ec2_steps_handler.py
```diff
@@ -59,7 +59,7 @@ async def manage_aws_emr_ec2_steps(
         operation: Annotated[
             str,
             Field(
-                description='Operation to perform: add-steps, cancel-steps, describe-step, list-steps. Choose read-only operations when write access is disabled.',
+                description='Operation to perform: add-steps, cancel-steps, describe-step, list-steps. Choose read-only operations when write access is disabled. "describe-step" and "list-steps" return step arguments and error messages, so they require --allow-sensitive-data-access.',
```
LiyuanLD: Curious why step arguments and error message are categorized as sensitive data?

src/aws-dataprocessing-mcp-server/README.md
```diff
@@ -290,12 +290,15 @@ Enables access to operations that expose sensitive user data. When disabled (def
 
 **HIGH - User Data:**
 * `get-query-results` (Athena): Blocks retrieval of actual query result data
-* `get-statement` (Glue Interactive Sessions): Blocks retrieval of statement execution outputs
+* `get-statement` and `list-statements` (Glue Interactive Sessions): Blocks retrieval of statement execution outputs
```
LiyuanLD: I feel like the effective fix should be introducing  Amazon Comprehend for job arguments mask, if we're looking for preventing sensitive data from job arguments or error messages

src/aws-dataprocessing-mcp-server/awslabs/aws_dataprocessing_mcp_server/handlers/emr/emr_ec2_steps_handler.py
DDD-Joo: it is the existing behavior of get-job-run, describe-step, get-statement, just extend it to the siblings (get-job-runs, list-steps, list-statements).

src/aws-dataprocessing-mcp-server/README.md
DDD-Joo: It's a reasonable idea and it will address weakness of the hard gating. but it is it's orthogonal to the bug. we may need a separate design for it

src/aws-dataprocessing-mcp-server/awslabs/aws_dataprocessing_mcp_server/handlers/emr/emr_ec2_steps_handler.py
naikvaib: Instead please remove necessary arguments if the flag is not set from the response

src/aws-dataprocessing-mcp-server/awslabs/aws_dataprocessing_mcp_server/handlers/emr/emr_ec2_steps_handler.py
```diff
@@ -113,13 +113,14 @@ async def manage_aws_emr_ec2_steps(
...
         - The server must be run with the `--allow-write` flag for add-steps and cancel-steps operations
+        - The server must be run with the `--allow-sensitive-data-access` flag for describe-step and list-steps operations, which return step arguments and error messages
         - Appropriate AWS permissions for EMR step operations
 
         ## Operations
         - **add-steps**: Add new steps to a running EMR cluster (max 256 steps per job flow)
         - **cancel-steps**: Cancel pending or running steps on an EMR cluster (EMR 4.8.0+ except 5.0.0)
-        - **describe-step**: Get detailed information about a specific step's configuration and status
-        - **list-steps**: List and filter steps for an EMR cluster with pagination support
+        - **describe-step**: Get detailed information about a specific step's configuration and status (requires --allow-sensitive-data-access)
```
naikvaib: Instead please remove necessary arguments if the flag is not set from the response

LiyuanLD (dismissed):

src/aws-dataprocessing-mcp-server/awslabs/aws_dataprocessing_mcp_server/handlers/emr/emr_ec2_steps_handler.py
```diff
@@ -59,7 +65,7 @@ async def manage_aws_emr_ec2_steps(
         operation: Annotated[
             str,
             Field(
-                description='Operation to perform: add-steps, cancel-steps, describe-step, list-steps. Choose read-only operations when write access is disabled.',
+                description='Operation to perform: add-steps, cancel-steps, describe-step, list-steps. Choose read-only operations when write access is disabled. "describe-step" and "list-steps" omit step arguments and failure text unless --allow-sensitive-data-access is enabled; the operations themselves remain available either way.',
```
naikvaib: Can we remove this - the operations themselves remain available either way.'. instead we can use (step arguments and failure text require --allow-sensitive-data-access)

src/aws-dataprocessing-mcp-server/awslabs/aws_dataprocessing_mcp_server/handlers/emr/emr_ec2_steps_handler.py
```diff
@@ -300,28 +307,31 @@ async def manage_aws_emr_ec2_steps(
...
 
+                step = response.get('Step', {})
                 success_message = (
                     f'Successfully described step {step_id} on EMR cluster {cluster_id}'
                 )
+
+                # Step records mix operational metadata with fields that can carry customer
+                # values (Config.Args, Config.Properties, and the Status failure text). Without
+                # --allow-sensitive-data-access those fields are omitted rather than the whole
+                # operation refused, so the caller still sees step identity, state and timeline.
```
naikvaib: lets remove this

src/aws-dataprocessing-mcp-server/awslabs/aws_dataprocessing_mcp_server/handlers/emr/emr_serverless_job_run_handler.py
```diff
@@ -69,7 +75,7 @@ async def manage_aws_emr_serverless_job_runs(
         operation: Annotated[
             str,
             Field(
-                description='Operation to perform: start-job-run, get-job-run, cancel-job-run, list-job-runs, get-dashboard-for-job-run. Choose read-only operations when write access is disabled.',
+                description='Operation to perform: start-job-run, get-job-run, cancel-job-run, list-job-runs, get-dashboard-for-job-run. Choose read-only operations when write access is disabled. "get-job-run" and "list-job-runs" omit the job driver, configuration overrides, tags and failure details unless --allow-sensitive-data-access is enabled; the operations themselves remain available either way.',
```
naikvaib: same message as above to keep it concised

naikvaib (approved):

LiyuanLD (approved):

PR #4645, opened 2026-09-17
fix(aws-documentation-mcp-server)!: correct read-path output and make failures raise

alexisareyn:
Fixes N/A

## Summary

Correctness fixes in the AWS Documentation MCP Server's read path, plus one deliberate change to
how the read tools signal failure. All of it was found by reading real `docs.aws.amazon.com` pages
through the server and comparing the output to the rendered page.

**This is a breaking change.** `read_documentation`, `read_sections` and `search_table` now raise on
failure where they previously returned an error string, `extract_content_from_html` raises instead
of returning a tagged marker, and `SearchTableResponse.error` is removed. Details and rationale are
in "Breaking changes" below.

### Changes

**1. Text was fused together where markup separated it.** Cell and heading text was extracted with
`get_text(strip=True)`, which strips each text node individually and then joins them with nothing. Two
kinds of separation were lost.

Between sibling values: a cell listing two endpoints came back as
`sts.ap-southeast-4.amazonaws.comsts.ap-southeast-4.api.aws` and the matching protocol cell as
`HTTPSHTTPS`. On the pages checked the values are `<p>` siblings rather than `<br />`, so both are
handled, along with the other block-level tags that appear inside cells. Values join with `; `, which
keeps the Nth value in one column aligned with the Nth in the next — that alignment is what makes an
endpoint table usable.

Around inline tags: `use the <code>Switch Role</code> feature` came back as
`use theSwitch Rolefeature`. Cells containing a link took a separate extraction path that spaced
correctly, so identical prose rendered two different ways depending on whether an unrelated anchor
happened to be present. Unstripped text now goes to a joiner that already normalizes whitespace, at
all five sites: cell text, heading text, link text, and the section-title match in each of
`search_table` and `read_sections`.

The two title matches are not cosmetic. `<h2>Using the <code>Switch Role</code> API</h2>` normalized
to `UsingtheSwitchRoleAPI`, so a caller passing the heading as rendered got no match at all — and the
error then listed the fused string as the title to retry with, which is not a title any page displays.
Both sides of each comparison now go through the same normalizer.

Three boundaries need special handling:

- **A note or important callout inside a cell qualifies the value it follows rather than adding
  one.** On a quotas page, delimiting it produced `6,144 characters.; Note; IAM doesn't count white
  space…` — a two-item list where there is one value, with the callout's own label as a phantom
  entry. Callout labels are dropped and callout prose joins with a space, so that cell reads
  `The size of each managed policy can't exceed 6,144 characters. IAM doesn't count white space…`.
  The boundary *after* a callout is left intact, so a cell of `value / callout / value / value`
  yields three values, not two.
- **A callout with no value before it qualifies the value after it.** Absorbing only the preceding
  boundary would leave a cell that opens with a note carrying the note's prose as a value of its own.
  A leading callout attaches forwards instead, so `callout / a / b` is two values.
- **Newlines inside `<pre>` are value boundaries.** In `<pre>` a newline is significant markup, so
  `<pre>Key1 = a\nKey2 = b\nKey3 = c</pre>` would otherwise fuse into one string. It delimits.
  Newlines outside `<pre>` stay whitespace, which is what HTML says they are.

**2. `table_heading` was taken from a callout title.** The heading for a table was resolved with
`find_previous` over `h1`–`h6`. AWS docs pages give note, tip and important callouts their own `<h6>`
title, so a callout sitting between a section heading and its table won the search and the table was
labelled with the callout's title instead of its section. Callout titles are skipped when walking
backwards. This fixes `detected_section` the same way.

**3. Links whose target has no filename rendered as working links.** An unresolved cross-reference in
source content emits an href with a path but no filename (`./.html#some-anchor`). Those were emitted
as ordinary markdown links, so an agent following one gets a 404 while the output gave no hint the
link was broken. They are unwrapped to their own text, in the document read path and inside table
cells.

**4. A page that redirected elsewhere was reported as missing content.** Redirects were already
followed and every hop already re-validated against the domain allowlist, so this is not a
"redirects are not followed" bug. The defect was in reporting.

A request for a path that is not a page in a guide is answered with a 302 to that guide's index,
not a 404 — and that index is a script shell with no extractable prose. So:

```
GET /service-authorization/latest/reference/list_not-a-real-page.html
  -> 302 /service-authorization/latest/reference/     (script shell, no tables)
```

`search_table` on that URL reported `"No tables found on this page."` That statement is true of the
page it was served and false of the page it names, and the caller had no way to tell which it meant.
The stakes are set by what a correct request returns: the EC2 entry in that guide is `list_ec2.html`
and carries five tables. A caller that gets a page name slightly wrong was told the tables do not
exist.

Every message that can be affected by a substitution now leads with it:

```
Requested <REQUESTED_URL>; served <SERVED_URL>. <what happened>
```

and content responses head with the page that actually answered:

```
<e>Requested <REQUESTED_URL>; served <SERVED_URL>.</e>

AWS Documentation from <SERVED_URL>:
```

Legitimate renames are reported too, deliberately.
`/AmazonCloudWatch/latest/monitoring/CloudWatch_Alarms.html` is the correct successor to
`AlarmThatSendsEmail.html`, but an agent that requested the latter must not cite it as the former.
Multi-hop chains are covered, because the comparison is between the requested URL and the final one
rather than any single hop.

Cosmetic differences are not treated as substitutions: scheme, host case, fragment, the session query
string the server appends, and path spelling — including a doubled slash anywhere in the path and an
explicitly named directory index. The header still shows the canonical URL in those cases, so a
request for `…/dg//welcome.html` is answered under `…/dg/welcome.html` rather than echoing a
malformed URL back to the caller.

The structured fields agree with the prose. `SearchTableResponse.url` reports the page that was read,
matching `results`, `tables_searched` and `section_title`, which all describe the served page. A field
that named the requested page would be the one field on the response that can disagree with its
neighbours, and it is the field an agent is most likely to cite as its source. The two truncation
hints name the served page for the same reason: they emit an executable
`search_table(url=…)` suggestion, and that call should land on the page holding the tables rather than
redirect again.

**5. Every link repeated its own URL as a title.** `markdownify` was called with
`default_title=True`, which renders `[text](url "url")` for every anchor with no authored title. On
the pages measured that duplication was 4–10% of total read output — about 43 KB of 457 KB on the
Bedrock user guide page. Since `read_documentation` is bounded by `max_length`, that is 4–10% of the
caller's budget spent restating hrefs. Set to `False`; authored titles are still emitted. `autolinks`
is pinned `False` at the same time: `markdownify` gates its autolink shortcut on `not default_title`,
so that option had been dead code, and enabling it with this change would have silently switched bare
links from `[url](url)` to `<url>`. Both options now have tests that fail if either is flipped back.

**6. A page carrying only markup must not read as content.** Deciding that a page cannot be read is
what turns a substitution into an error rather than a misleading answer, so the test for readable text
has to be exact about what counts as text.

- Prose is excluded when any ancestor is `script`, `style` or `noscript`, not only the immediate
  parent. A JS-disabled banner sits several levels inside its `<noscript>`, and checking one level
  would count it as content.
- `find_all(string=True)` returns `Comment` nodes, which are markup. They are excluded, so a page
  whose body holds only a build comment is not readable.
- `script` and `style` elements are removed from the tree before conversion. `markdownify`'s `strip=`
  option drops a tag while still converting its children, so naming them there suppresses no text —
  a page whose body carries an inline bootstrap script would otherwise return that JavaScript as
  documentation. Removing the elements leaves nothing to convert, so the existing emptiness check
  reports the page as unreadable. `noscript` is removed by the navigation-element pass already.
- The emptiness check strips before testing. A body of `<br>` elements converts to spaces and
  newlines, which is a non-empty string and was returned as a successful read.

These are tested against the shell shapes directly rather than through a mocked extractor, so the
tests fail if any of the four conditions is relaxed.

**7. Every table returned a first row of its own column names.** AWS documentation tables carry a
`<thead>` and no `<tbody>`. With no `<tbody>` the row source falls back to the whole table, which
sweeps in the header row, so the header parsed as data. The STS endpoint table returned 37 rows for 36
regions, and `search_table(query="Region Name Endpoint Protocol")` returned that phantom row as its
only match — every word in the query is a column name, so nothing else could match it. Rows whose
parent is `<thead>` are excluded from the fallback; an explicit `<tbody>` is still honoured where a
page supplies one.

This is also why the table fixtures changed. Wrapping rows in `<tbody>` — which browsers inject but
the published HTML does not — means a fixture cannot reproduce the shape the server actually receives.
Both shapes are covered now, and the endpoint tables are asserted live.

**8. A cell's icon carried meaning that was dropped.** Support matrices mark a cell with an ``
and no text, so cell extraction saw nothing and *supported* and *not supported* rendered identically.
On a model card the input-modality column returned every modality the same way, so an agent would
report an unsupported input as available. Images contribute their `alt` text, which is where the
meaning is written down and what `read_documentation` already returns — so the two tools now agree
rather than `search_table` being the lossier of the pair. An image with no `alt` is decorative and
contributes nothing.

**9. `read_sections` heads its output the same way every time.** The
`AWS Documentation from <url>:` header was added only when a substitution had occurred, so the same
successful call produced two different shapes and a caller's parsing depended on whether a redirect
happened. `read_documentation` always heads its output; `read_sections` now does too, with the
substitution note above the header rather than replacing it.

**10. `<e>` marks notes, not failures.** The tag carries out-of-band notices — content truncated, no
further content, a substituted page. Every failure is an exception, so nothing needs a failure marker,
and the one path that still emitted `<e>` for a failure — the China partition's catalogue, which keeps
its payload when the prose will not simplify — would have made the same marker mean both. That note is
now labelled in words.

## Breaking changes

**The read tools raise instead of returning an error string.** Previously a failed fetch, a page
with no extractable content, or a non-HTML response came back as a normal successful tool result
whose payload happened to contain an error message. MCP reported success. A caller had to
string-match to notice anything had gone wrong, and the most common failure — a redirect onto a
script shell — produced a message (`"No tables found on this page."`) that reads as a valid answer.

Now the rule is: **a tool returns when it can give the caller something usable, and raises when it
cannot.** `read_documentation`, `read_sections` and `search_table` raise `ValueError` on transport
failure, on a `>= 400` status, on a page with no extractable content, and on a response that is not
HTML when the tool needs HTML.

Consequences for callers:

- `extract_content_from_html` raises `UnreadablePageError` (a `ValueError` subclass) instead of
  returning `<e>Empty HTML content</e>`, `<e>Page failed to be simplified from HTML</e>`, or
  `<e>Error converting HTML to Markdown: …</e>`. Anything matching on those strings must catch the
  exception instead.
- `parse_html_tables` raises `UnreadablePageError` for a page with no readable body text, where it
  previously returned `None` (which was indistinguishable from "this page has no tables").
- `SearchTableResponse.error` is **removed**. Every failure it carried is now an exception, so the
  field was unreachable. `hint` is unchanged and still carries non-failure guidance — no tables on
  the page, section not found, no rows matched.
- `read_documentation` for the China partition (`docs.amazonaws.cn`) shares the same implementation
  and so raises on the same conditions.

`search_table` still *returns* for genuine empty results — no tables on the page, named section not
found, no rows matched the query. Those are successful reads that found nothing, not failures.

One place deliberately does not raise. `get_available_services` (China partition) builds a service
catalogue from a table-of-contents fetch and then reads a page for prose. If that page will not
simplify, the catalogue is the payload worth keeping, so the read failure is reported alongside it
rather than discarding both.

## Behaviour on every redirect and failure case

Measured against live `docs.aws.amazon.com` through a real MCP tool call, so every row is what an
agent receives rather than what the impl returns. That distinction matters: the tool layer validates
the domain and requires a `.html` suffix before it delegates, which makes some impl states
unreachable. `result` = a normal tool result. `error` = an MCP tool error, which the agent sees as a
failed call.

| # | Case | `read_documentation` | `read_sections` | `search_table` |
|---|---|---|---|---|
| A | readable page, no redirect | result | result | result |
| B | 301 to a real page (rename) | result | result | result |
| C | 302 to a guide-index shell | **error** | **error** | **error** |
| D | cosmetic `//`, lands on the same page | result | result | result |
| E | redirect whose target 404s | **error** | **error** | **error** |
| F | plain 404 | **error** | **error** | **error** |
| G | bare directory URL | **error** (guard) | **error** (guard) | **error** (guard) |
| H | `.md` URL | **error** (guard) | **error** (guard) | **error** (guard) |

Two things an agent can rely on: a substitution is always named before anything else, and a page the
server cannot read is an error rather than a result that reads like an answer.

C is the case this PR exists for. Before, a request for a path that is not a page was answered with
a 302 to the guide index — a script shell with no prose — and `search_table` reported "No tables
found on this page." That is true of the page served and false of the page named.

G and H are rejected by the `.html` guard, so they never reach the fetch. Worth stating because it
retires two ideas: that a bare directory needs impl-side handling, and that an agent appending `.md`
is an MCP-side concern. Neither request can be made through these tools.

## User experience

`search_table` on the AWS General Reference STS endpoints table, before:

```json
{"Region": "ap-southeast-4", "Endpoint": "sts.ap-southeast-4.amazonaws.comsts.ap-southeast-4.api.aws", "Protocol": "HTTPSHTTPS"}
```

after:

```json
{"Region": "ap-southeast-4", "Endpoint": "sts.ap-southeast-4.amazonaws.com; sts.ap-southeast-4.api.aws", "Protocol": "HTTPS; HTTPS"}
```

`table_heading` for a table preceded by a note callout, before `"Note"`, after the section heading
the table actually belongs to.

An href with no filename, before `[Cross-region inference](./.html#some-anchor)` (a link that 404s),
after `Cross-region inference` as plain text.

A cell holding a preformatted block, before `Key1 = a Key2 = b Key3 = c`, after
`Key1 = a; Key2 = b; Key3 = c`.

`search_table` on a URL that is not a page, before:

```json
{"tables_searched": 0, "results": [], "hint": "No tables found on this page."}
```

after — an exception, whose message is:

```
Requested https://docs.aws.amazon.com/service-authorization/latest/reference/list_not-a-real-page.html; served https://docs.aws.amazon.com/service-authorization/latest/reference/. The page carries no readable content.
```

`read_documentation` on a page that was renamed, after:

```
<e>Requested https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/AlarmThatSendsEmail.html; served https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch_Alarms.html.</e>

AWS Documentation from https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch_Alarms.html:

# Using Amazon CloudWatch alarms
…
```

Link output, before
`[Amazon Bedrock](https://docs.aws.amazon.com/bedrock/ "https://docs.aws.amazon.com/bedrock/")`,
after `[Amazon Bedrock](https://docs.aws.amazon.com/bedrock/)`.

## Verification

- **428 passed, 25 skipped** (the skips are `@pytest.mark.live`, gated behind `--run-live`).
- With `--run-live` against the real site: **453 passed** on the commercial partition. The China
  partition's live tests require network reachability to `docs.amazonaws.cn`.
- Coverage 98% branch overall, with `server_utils.py` and `models.py` at 100%, `util.py` at 99% and
  `table_utils.py` at 97%.
- `pyright`: 0 errors, 0 warnings. `ruff check` and `ruff format --check`: clean.
- **Mutation testing as the acceptance gate.** Every fix was reverted in place and the covering
  tests were required to fail. Each row below is a revert and the number of tests that caught it:

| mutation | tests failed |
|---|---|
| substitution detection disabled entirely | 18 |
| `<thead>` rows parsed as data | 7 |
| callout titles kept when resolving a heading | 10 |
| cell text loses inline spacing | 3 |
| callout absorbs the boundary after it too | 3 |
| `default_title` flipped back to `True` | 3 |
| `<pre>` newline boundaries removed | 2 |
| broken links left as links | 2 |
| truncation hint names the requested page | 2 |
| heading text loses inline spacing | 1 |
| section-title match loses inline spacing (`search_table`) | 1 |
| section-title match loses inline spacing (`read_sections`) | 2 |
| image `alt` text dropped | 1 |
| image `alt` dropped on the link path | 1 |
| whitespace-only body passes as content | 1 |
| China note tagged as an error | 1 |
| `read_sections` header made conditional | 1 |
| link text loses inline spacing | 1 |
| leading callout becomes its own value | 1 |
| a comment counts as readable text | 1 |
| only the immediate parent checked for `noscript` | 1 |
| `script`/`style` left for `markdownify` to strip | 1 |
| `autolinks` flipped back to `True` | 1 |
| `url` reports the requested page | 1 |
| transport cause not chained | 1 |
| leading `//` collapse reverted | 1 |
| named directory index not normalized | 1 |
| China-partition catalogue guard removed | 1 |

## Checklist

* [ ] Changes are documented

Is this a breaking change? (Y/N) **Y**

Three things change for callers, all covered in "Breaking changes" above: the three read tools raise
instead of returning an error string; `extract_content_from_html` and `parse_html_tables` raise
instead of returning a tagged string or `None`; and `SearchTableResponse.error` is removed because
every failure it carried became an exception. Output formatting also changes for callers that were
receiving fused or mislabelled table values — a caller parsing a multi-value cell by string equality
against the fused form would need to split on `; ` instead, though that form was not a documented
contract and was not usable data.

**RFC issue number**: N/A

Checklist:

Callers that previously string-matched an error payload will now see a real MCP error, which is the
intended migration: the failure becomes visible instead of silent. Anything matching on
`<e>Page failed to be simplified from HTML</e>` or on `SearchTableResponse.error` needs updating.

## Acknowledgment


alexisareyn:
pre commit is failing due to issue in separate MCP, issue is being tracked https://github.com/awslabs/mcp/pull/4647

PR #4646, opened 2026-09-17
chore(timestream-for-influxdb-mcp-server): apply ruff format to test_server.py

alexisareyn:
Fixes N/A

## Summary

The `Precommit .` job runs `pre-commit --all-files` and currently fails on `main`, so every open
pull request reports the same failure regardless of what it changes.

One root cause, two reported hooks:

- `ruff format` rewrites `src/timestream-for-influxdb-mcp-server/tests/test_server.py`, collapsing
  multi-line `assert (...)` calls that fit on one line.
- That shifts line numbers, so `detect-secrets` then updates the recorded line for the known
  non-secret in that file (`621` → `612`) and refreshes `generated_at`.

Both hooks report modified files and the job exits 1.

## Changes

Applied the formatter output and the matching baseline line number. The diff is whitespace and line
joining only — no behaviour change.

## Verification

- `ruff format --check` on the file: clean.
- `pre-commit run ruff-format --all-files` and `pre-commit run detect-secrets --all-files`: both
  pass, with no further drift on a repeat run.
- `ruff check`, `check-license-header`, `trailing-whitespace`, `end-of-file-fixer`, `check-ast`,
  `pretty-format-json` across all files: pass.

One note, out of scope here. `test_get_timestream_influxdb_client_happy_path` asserts
`Session(region_name='us-east-1')` but does not patch the environment, and `server.py` reads
`os.environ.get('AWS_REGION', 'us-east-1')`. It therefore passes in CI, where `AWS_REGION` is unset,
and fails for any contributor who has a different region exported. The sibling test at line 157
already uses `@patch.dict('os.environ', ...)`; applying the same decorator would make it hermetic.
Unrelated to this change and no effect on `Precommit .`, which does not run pytest.

## Acknowledgment


grayhemp:
The fix was merged yesterday https://github.com/awslabs/mcp/pull/4647.

